Issue #33, opened 2019-02-23
Description/main account


azbulutlu:
As these users are essentially one way feeds, it would be great if the feeds when viewed from fediverse could have 

1) an icon
2) a description (e.g. this would be especially useful to point out that this is a wordpress install that cannot follow back or boost, and a brief description of the blog)
3) information slots (similar to up to 4 slots mastodon has.  this would be very useful to add custom information mastodon account associated with this blog)

Unless the plan is to have these eventually work as full fledged actors that can follow, boost and comment on other accounts? although these features would be useful even in that case.

Thanks for this great plugin!

pfefferle:
I am not sure I get you right, but all the points you mentioned are already implemented.

![bildschirmfoto 2019-02-24 um 09 31 27](https://user-images.githubusercontent.com/2373/53296998-fd342600-3816-11e9-92e5-5782bba50555.png)

The plugin uses the avatar of the user as icon, the description of the user as description and the links for the slots...

azbulutlu:
My mistake for missing this. :) I didn't think to try user settings of wordpress. Thank you for quick comment.

pfefferle:
no problem :)

let me know if you still miss something.

azbulutlu:
ok here is one issue I am noticing and one enhancement that would be useful if feasible. (https://www.azbulutlu.org/author/azbulutlu_tgtrrc/ being the guinea pig for this experiment) 

1) the avatar (from local media file) doesn't show when using a plugin (I use https://www.azbulutlu.org/wp-admin/plugin-install.php?tab=plugin-information&plugin=simple-local-avatars not sure if there is one that happens to be compatible with this addon) to replace gravatar which is a privacy issue.

2) This is a fairly minor issue but it would be nice to be able to customize which of these links. Right now there is 3 static categories. With the exception of website these are predetermined domain that repeats itself. it would be nice if
a) i was possible to hide any existing information (e.g. only have "blog" in my case)
b) being able to have custom slots as "key":"descriptor" pairs (similar to the way it is on mastodon. This would allow customizing these slots to have information that matches use case of the person using the site. (e.g. I'd like to use a link to my main mastodon, and link to my main website as "portfolio". Another person might want to link to their rss feed, and put a pronoun and email information in publicly, etc) 

Again thank you for the great work on this addon.

lapineige:
Could we configure the avatar manually ?
[...]

PR #54, opened 2019-07-21
Allow attachments to be filtered more easily by using array keys.

noplanman:
This allows easier filtering of attached `PropertyValue` fields.
A step towards allowing a plugin to be developed that makes #33 super easy for users to configure themselves.

(An alternative would be to add individual filters for all parts of the author array, instead of just the main one)

So using a filter like this:
```php
add_filter( 'activitypub_json_author_array', function($json) {

        // Don't show the Blog URL.
        unset($json->attachment['blog']);

        // Add a custom field.
        $json->attachment[] = array(
                'type'  => 'PropertyValue',
                'name'  => 'Coding...',
                'value' => '...for the World!',
        );

	return $json;
} );
```

Gives something like this:
mastodon-profile

### DIFF
--- a/templates/json-author.php
+++ b/templates/json-author.php
@@ -43,7 +43,7 @@ $json->publicKey = array(
 $json->tag = array();
 $json->attachment = array();
 
-$json->attachment[] = array(
+$json->attachment['blog'] = array(
 	'type' => 'PropertyValue',
 	'name' => __( 'Blog', 'activitypub' ),
 	'value' => html_entity_decode(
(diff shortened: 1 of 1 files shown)


noplanman:
@pfefferle I'd appreciate your feedback on this :innocent:

pfefferle:
hmmmm... I like the general idea to make it easier to replace the link-items... but the risk of your suggestion is inconsistency... In your example you remove the `blog` and replace it with an entry on a different key... so if a plugin, that calls the filter later in the chain, would like to drop the `Coding...` entry, it cannot find it or gets a warning, because it tries to `unset` a non existing array item (like you did in the example)...

noplanman:
Right, but that's the case with all filters. That's why it's important to treat them kindly. Arguably one shouldn't have multiple plugins playing with the same filters anyway, especially if they bite each other.

To solve what I'm trying now, would be to iterate through the array and check for "Blog" as the attribute value and then remove it like that, which is just much more cumbersome but returns the same array in the end :confused:

pfefferle:
close in favor of #132

thanks for the proposal and sorry for the delay :(

PR #66, opened 2019-09-08
WIP: Don't mark fediverse comments as spam, fixes #62

### LINKED ISSUE #62: Akismet treats federated comments as spam
On blog.ninapaley.org, I've finally managed to get the posts available in the fediverse (yay!) but federated comments aren't coming through into WordPress.

Possibly related, but closed without resolve: #55

alexgleason:
Fediverse comments will often be captured as spam, even though they will rarely or never be spam. This PR aims to ensure that fediverse comments are never marked as spam.

For now it makes them all be held. However, I think there should be a conditional which checks the site settings, and chooses "hold" or "approve" depending on whether comments must be approved manually. I don't know how to do that.


pfefferle:
Is this still WIP?

alexgleason:
I think it just needs a conditional around it but I don't know how to pull the site settings about comments. Maybe you or someone else can finish it?

mediaformat:
I have been testing this, and it seems the only conditional would be to add something like 
`if ( is_plugin_active( 'akismet/akismet.php' ) )`. 
`wp_new_comment` uses the `check_comment` function, which goes through the relevant site discussion settings.

PR #81, opened 2020-01-18
Federate image descriptions if present

BenLubar:
Add image alt text to the ActivityStreams attachment property in a format that Mastodon can read.


pfefferle:
Thanks!

PR #84, opened 2020-01-18
Use the comment URL rather than the comment author URL so the comment can be displayed in context.

BenLubar:


pfefferle:
Should we replace the author url or add a comment meta with the if to show the context? What do you expect when clicking on the author name?

BenLubar:
I'm coming from the fediverse side, so my expectations might not line up with what people who use WordPress more frequently expect, but I'd expect there to at least be *some* way to go directly to the remote post, and going to the site of the person who posted it would be lower in priority.

PR #92, opened 2020-05-18
Only return nodeinfo data if site is public

mediaformat:
Give nodeinfo services the same indexing privileges as web crawlers based on the wordpress setting `blog_public` :
> Discourage search engines from indexing this site

If the option is enabled, the nodeinfo path returns a 404 error as per the [protocol](https://github.com/jhass/nodeinfo/blob/master/PROTOCOL.md)

### DIFF
--- a/includes/rest/class-nodeinfo.php
+++ b/includes/rest/class-nodeinfo.php
@@ -13,7 +13,7 @@ class Nodeinfo {
 	 * Initialize the class, registering WordPress hooks
 	 */
 	public static function init() {
-		if ( 1 === get_option('blog_public') ) {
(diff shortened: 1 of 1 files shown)


mediaformat:
Background: My staging site somehow ended being heavily crawled by friendica servers.

@pfefferle Any objection to me merging this this?

pfefferle:
I like the change! Will merge that later today and try to also release a new version!

Thanks a lot!

pfefferle:
But I am not sure, this will solve the crawler problem.

mediaformat:
True, I will monitor my logs to see if it is effective. 
Thanks!

PR #103, opened 2020-10-22
Code Style: Filter option in favor of updating the database

akirk:
Just a quick improvement to avoid actually updating the database option. I didn't test this, though, please give it a quick test.


pfefferle:
nice idea, thanks!

Issue #104, opened 2020-12-09
Add support for latest HTTP Signatures spec draft


mediaformat:
Since Mastodon 3.2.1 servers require a Digest header in their POST requests.
https://github.com/tootsuite/mastodon/pull/14556

PR Forthcoming

pfefferle:
@mediaformat fixed it with #105 

Thanks a lot!

mediaformat:
Cheers

PR #105, opened 2020-12-10
add digest header

mediaformat:
adds a digest header
Mastodon now returns 202, instead of Digest missing error. 
Not sure if there is another issue, or something wrong with my setup...
#104


mediaformat:
Just tested in production.

This fixes #106

jgoerzen:
Works for me also

Issue #63, opened 2019-09-08
Timezones


alexgleason:
A post is made from a WordPress site where the system time is set to UTC-5: Chicago.

It reaches the Fediverse server, which stores time in UTC-0.

The timestamp is 5 hours off when displayed in the user's local time.

pfefferle:
But isn't that a bug of the Fediverse Server then?

ActivityPub specifies `published` here: https://www.w3.org/TR/activitystreams-vocabulary/#dfn-published

It has to be `xsd:dateTime` which defines:

> The time zone may be specified as Z (UTC) or (+|-)hh:mm. Time zones that aren't specified are considered undetermined.

http://books.xmlschemata.org/relaxng/ch19-77049.html

silksow:
I can confirm this issue with posts from my WP to mastodon.social

jgoerzen:
Observed as well here

pfefferle:
Fixed by #107 

Thanks @mediaformat !!

PR #107, opened 2020-12-17
Fix timezone

mediaformat:
In my own testing, I have found that the post timezone was skewed by its distance from GMT.
WordPress post object includes a timezone neutral `post_date_gmt`, using this allows peers to display the correct post date-time. 
#63


pfefferle:
Thanks! 👍

PR #111, opened 2021-01-12
Link to a GNU Social fork that is under active development

mkljczk:
What you can find on https://gnu.io/social doesn't even support ActivityPub.


pfefferle (approved):
thanks!

Issue #114, opened 2021-01-29
Don't set `to` field to receiving user id for each user on federation


lambadalambda:
https://github.com/pfefferle/wordpress-activitypub/blob/b49cc5333cce42638d4e1b8f9aae3a16c5bd8b6f/includes/class-activity-dispatcher.php#L34

The `to` field should contain the wordpress user's follower collection, as putting a user into the `to` field is treated like a direct mention, so it will generate notifications.

See https://git.pleroma.social/pleroma/pleroma/-/issues/2480#note_81046

pfefferle:
OK, maybe I misread:

> Origin servers sending publicly addressed activities to sharedInbox endpoints MUST still deliver to actors and collections otherwise addressed (through to, bto, cc, bcc, and audience) which do not have a sharedInbox and would not otherwise receive the activity through the sharedInbox mechanism.

pfefferle:
@lambadalambda ah, I see the issue! I have to use a collection instead of a single user and the actor url instead of the inbox, right?

pfefferle:
OK, I checked again, and this is always a collection. It might be that there is only one in the collection, but it is always a collection: https://github.com/pfefferle/wordpress-activitypub/blob/b49cc5333cce42638d4e1b8f9aae3a16c5bd8b6f/includes/functions.php#L202

pfefferle:
...or should the list empty and the server is handling all the inbox stuff?

mediaformat:
@lambadalambda you are correct, I had corrected this in an upstream feature branch, but I will submit a PR for this shortly.

mediaformat:
@lambadalambda I wonder if there is some MRF on the receiving instance causing this issue. I am currently testing with a vanilla pleroma and it is not triggering notifications

rinpatch:
Maybe you are testing with old Pleroma. We used to not generate notifications if there is no explicit mention in the `tag` field, but this behavior was lost to some refactorings unfortunately

lambadalambda:
Actually I think the error is on our side because we add a user to the `to` collection if the message arrives in their own inbox and not the shared inbox. Will have to check that.

PR #115, opened 2021-02-01
Fix public addressing

mediaformat:
Public activities are addressed to the special Public collection in the `to` field, individual recipients are added to the `cc` field

A reference to the Followers collection could be added to the `to` field, but I'm not sure it's needed at the moment.

 #114

### DIFF
--- a/includes/class-activity-dispatcher.php
+++ b/includes/class-activity-dispatcher.php
@@ -31,7 +31,8 @@ class Activity_Dispatcher {
 		$activitypub_activity->from_post( $activitypub_post->to_array() );
 
 		foreach ( \Activitypub\get_follower_inboxes( $user_id ) as $inbox => $to ) {
-			$activitypub_activity->set_to( $to );
+			$activitypub_activity->set_to( 'https://www.w3.org/ns/activitystreams#Public' );
+			$activitypub_activity->set_cc( $to );
 			$activity = $activitypub_activity->to_json(); // phpcs:ignore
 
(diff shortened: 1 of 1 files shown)


includes/class-activity-dispatcher.php
```diff
@@ -31,7 +31,8 @@ public static function send_post_activity( $activitypub_post ) {
 		$activitypub_activity->from_post( $activitypub_post->to_array() );
 
 		foreach ( \Activitypub\get_follower_inboxes( $user_id ) as $inbox => $to ) {
-			$activitypub_activity->set_to( $to );
+			$activitypub_activity->set_to( 'https://www.w3.org/ns/activitystreams#Public' );
```
pfefferle, line 34: Does it have to be a string? if we only set the `cc`, `to` falls back to the default and that is https://github.com/pfefferle/wordpress-activitypub/blob/master/includes/model/class-activity.php#L17

includes/class-activity-dispatcher.php
```diff
@@ -51,7 +52,8 @@ public static function send_update_activity( $activitypub_post ) {
...
+			$activitypub_activity->set_cc( $to );
```
pfefferle, line 56: `$to` includes only the followers that are on the same server as `$inbox`. Is that correct or should we add the complete followers list?

mediaformat: I think this is correct, the for loop will take care so that we get to all followers per server. 

We could add the user followers collection endpoint in `to` making it an array with the special Public address. But imho it is not necessary for now

PR #117, opened 2021-02-09
Prevent comments being flagged as spam by akismet

jejb:
It's very difficult to get akismet to approve of a mechanically posted
comment because all the various capchas and java script tests have to
pass.  So, instead of trying to make everything right, simply turn off
akismet for Activitypub comments (this means you should set your
default policy to moderate all comments which don't have one approved
just in case some bot works out how to spam via activitypub).

Signed-off-by: James Bottomley <James.Bottomley@HansenPartnership.com>

### DIFF
--- a/includes/rest/class-inbox.php
+++ b/includes/rest/class-inbox.php
@@ -383,7 +383,12 @@ class Inbox {
 		// do not require email for AP entries
 		\add_filter( 'pre_option_require_name_email', '__return_false' );
 
+		// disable akismet if necessary
+		\add_filter( 'akismet_get_api_key', '__return_null');
+
 		$state = \wp_new_comment( $commentdata, true );
+
+		\remove_filter( 'akismet_get_api_key', '__return_null');
 
 		\remove_filter( 'pre_option_require_name_email', '__return_false' );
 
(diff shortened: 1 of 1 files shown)


pfefferle:
Hey @jejb ! Thanks for your MR!

We discussed that here #62 and here #66 !

I would prefer using a more generic method like the `wp_set_comment_status` introduced in #66, because Akismet is not the only Spam Protection plugin (Antispam Bee is very popular in Europe, because of GDPR reasons) but this disables the spam protection completely.

I would love to have both of both worlds: Antispam Plugins filtering spam, but not ActivityPub completely.

jejb:
On Wed, 2021-02-10 at 09:53 -0800, Matthias Pfefferle wrote:
> Hey @jejb ! Thanks for your MR!
> 
> We discussed that here #62 and here #66 !
> 
> I would prefer using a more generic method like the
> `wp_set_comment_status` introduced in #66, because Akismet is not the
> only Spam Protection plugin (Antispam Bee is very popular in Europe,
> because of GDPR reasons) but this disables the spam protection
> completely

That doesn't work for me because akismet still marks the comment as
spam, which excites my fail2ban filter.  Also once the comment gets
marked as spam, you don't get a notification email that it's pending.

> I would love to have both of both worlds: Antispam Plugins filtering
> spam, but not ActivityPub completely.

All spam plugins go to great lengths to detect that a human actually
sent in the comment via a web browser, so they add javascript puzzles
and nonces.  Activitypub comments, by their nature, aren't generated by
humans, they're emitted by rest endpoints after the human submitted the
comment, so none of those tricks are ever going to pass.

One possible way out is to mimic standard mechanically generated
content, which would be pingbacks, but you'd basically have to submit
the comment as a pingback and then flip the type to comment after it
was added to the database, which sounds like a nasty hack.

James

Cambridgeport90:
In light of all of the newer protocols, such as activityPub, I am quite surprised that more changes haven't been made to spam blocking facilities to accommodate the changing landscape, so it seems that it will take both sides of this working together to fix the issue completely.

PR #123, opened 2021-06-08
Add missing `author_id` to actions and filters

pfefferle:


pfefferle:
@skysarwer I added the `author_id` to the actions and filters, so that you have the info for your buddypress needs.

PR #132, opened 2021-09-15
Add basic BuddyPress support

pfefferle:
fix #122

thanks and props @skysarwer

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -106,3 +106,12 @@ function flush_rewrite_rules() {
 }
 \register_activation_hook( __FILE__, '\Activitypub\flush_rewrite_rules' );
 \register_deactivation_hook( __FILE__, '\flush_rewrite_rules' );
+
+/**
+ * Only load code that needs BuddyPress to run once BP is loaded and initialized.
+ */
+function enable_buddypress_features() {
+	require_once \dirname( __FILE__ ) . '/integration/class-buddypress.php';
+	\Activitypub\Integration\Buddypress::init();
+}
+add_action( 'bp_include', '\Activitypub\enable_buddypress_features' );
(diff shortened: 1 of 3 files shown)


pfefferle:
@skysarwer @Cambridgeport90 can you verify/test the changes?

pfefferle:
props also to @noplanman for the idea with the named attachments!

Cambridgeport90:
Can when I’m not at work, and can actually set things up.

skysarwer:
This is amazing, thank you so much @pfefferle  ! 

I will try to test this soon. One thing I noticed right away, is that I don't see any code that effects the json['url]' data -- just the json['attachment']['blog_url'] data. 

I think it would make sense to overwrite the url data with the BP Member Profile link too, no?

pfefferle:
> I think it would make sense to overwrite the url data with the BP Member Profile link too, no?

Will do! thanks for the feedback!

@skysarwer can you focus on testing https://github.com/pfefferle/wordpress-activitypub/pull/132/files#diff-95fc90cd109ecd774f4949796c5f642eedee889e6e65773c6a943134fb333de1R32 ?

Cambridgeport90 (commented):
This looks really good. Will be testing when I get a chance.

pfefferle:
@skysarwer re-added `json->url`.

pfefferle:
@skysarwer could you verify the changes yet?

Cambridgeport90:
Unfortunately I don't have environment set up, but I can definitely do that.￼

skysarwer:
@pfefferle it's working! I disabled my custom filter and this is the results. I am going to detach the profile metadata since I find it redundant, and add back the custom XPROFILE field for the Bio (which is specific to my use case). 

Screenshot 2021-09-28 004145

skysarwer (approved):

pfefferle:
@skysarwer thanks :)

PR #139, opened 2022-02-28
Signature Verification with phpseclib3

mediaformat:
Starting this PR to get Signature Verification working.

It's based on [github.com/landrok/activitypub/](https://github.com/landrok/activitypub/)

It's not working for me locally, perhaps it's my setup, or  maybe you will understand where it's failing.

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -15,6 +15,8 @@
 
 namespace Activitypub;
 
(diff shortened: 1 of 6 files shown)


mediaformat:
Signature verification is failing with both methods. The following log is output:
```
openssl_error_string(): error:0909006C:PEM routines:get_name:no start line
```

PR #95, opened 2020-06-03
[WIP] Mentions

mediaformat:
Work in progress for constructing a per user inbox, federating replies.

- [x] Basic Reply to Comments (no mentions or audience yet) #42 
- [x] Threaded Replies
- [x] [Inbox forwarding](https://www.w3.org/TR/activitypub/#inbox-forwarding)

The wordpress comment system is fine for allowing Fediverse users to interact with a post, however currently Private Mentions (non-public, such as Followers Only, and Direct/Private Messages) end up as WP Comments.

The following types of mentions go to a custom Mentions Post Type  
- [x] Private mentions should bypass comments and go to inbox as per #72 
- [x] Mentions unrelated to a Post should bypass comment system and go to mentions post type as per #50
- [x] Reply to received Mention

PR #142, opened 2022-04-15
Federated Comments

mediaformat:
Restarting a fresh PR from #95 

Depends on #299

Comments:
- [x] Comments support threading
- [x] Local / Remote comments can be deleted
- [x] Local / Remote comments can be updated
- [x] Replies Collection
- [x] Harden ActivityPub compatibility using ugly permalinks
- [x] Support replying to legacy comments using `source_url` (refetch to get activity `id`)


johnspurlock:
👀

mediaformat:
@pfefferle just a note that SonarCloud code smell uses rules that conflict with `phpcs`.

pfefferle:
Yea, ignore sonar cube!

pfefferle:
@mediaformat but there seems to be a conflict in `activitypub.php`.

mediaformat:
> @mediaformat but there seems to be a conflict in `activitypub.php`.

@pfefferle I'm not seeing anything out of ordinary, any chance you could pinpoint the error?

pfefferle:
Nice, thanks. Is it still [WIP]?

mediaformat:
There is one last piece to this, a tools page for manually federating updates (deletes, or delete and announce with new ID).

mediaformat:
Ready to go!

mediaformat:
One more fix coming ;-)

includes/class-admin.php
```diff
@@ -121,6 +134,33 @@ public static function register_settings() {
...
+
+		// Needs update
+		if ( empty( $activitypub_db_version ) || \version_compare( $plugin_data['Version'], $activitypub_db_version, '>' ) ) {
+			// Check for specific migrations
+
+			if ( version_compare( '0.13.5', $activitypub_db_version, '>' ) ) {
+				// This updates post_meta with _activitypub_permalink_compat.
+				// Posts that have this meta will be backwards compatible with their permalink based ActivityPub ID (URI)
+
+				// This may create false positives, where the permalink has changed (slug, permalink structure) since federation,
```
mediaformat, line 155: @pfefferle Do you think this is a sufficient upgrade/migration plan? 

1. On update, Mark all existing activitypub post types with a post meta: `_activitypub_permalink_compat` of the existing permalink.

2. In the tools page, a List_Table displays all these marked posts, a column shows how many comments are associated with the post. The following text would alert the user:

> Posts with ActivityPub Comments, should be treated with care, existing interactions in the fediverse but not on site will be lost. 

2. a) I think I should add some more alerting style to the rows with comments. 
2. b) Row actions allow Deleting the post on the fediverse, or Deleting and Re-sharing (Announce/Boost) the original post with the new ID (this removes existing interactions from the fediverse)

Any thoughts, on this flow, or the language?

includes/table/migrate-list.php
```diff
@@ -0,0 +1,127 @@
...
+				'total_pages' => ceil( $total_items / $per_page ),
+			)
+		);
+
+		// actions
+		if ( isset( $_REQUEST['_wpnonce'] ) ) {
+			$nonce = \esc_attr( $_REQUEST['_wpnonce'] );
+		}
+		// delete
+		if ( isset( $_REQUEST['action'] ) && 'activitypub_tools' === $_REQUEST['page'] && 'delete' === $_REQUEST['action'] ) {
```
mediaformat, line 63: Thinking a bit further, I might remove the simple row_action delete, since there is also a Field to delete a federated URL

mediaformat:
Pagination of `migrate-list.php` / `wp-admin/tools.php?page=activitypub_tools` is not working, will fix, but shouldn't block review.

includes/model/class-comment.php
```diff
@@ -0,0 +1,282 @@
...
+
+/**
+ * ActivityPub Comment Class
+ *
+ * @author Django Doucet
+ */
+class Comment {
+	private $comment;
+	private $updated;
+	private $deleted;
```
pfefferle, line 12: Add missing class attributes

includes/model/class-post.php
```diff
@@ -15,6 +15,9 @@ class Post {
 	private $attachments;
 	private $tags;
 	private $object_type;
+	private $deleted;
+	private $updated;
+	private $permalink;
```
pfefferle, line 20: add `replies` to class attributes

includes/class-activitypub.php
```diff
@@ -85,6 +97,12 @@ public static function render_json_template( $template ) {
 	 */
 	public static function add_query_vars( $vars ) {
 		$vars[] = 'activitypub';
+		$vars[] = 'ap_comment_id';//comment_id doesn't work, 'c' is probably too short and prone to collisions
```
pfefferle, line 100: is there really no way to reuse a core param like `p` for posts?

includes/class-activitypub.php
```diff
@@ -24,6 +24,14 @@ public static function init() {
 		}
 
 		\add_action( 'transition_post_status', array( '\Activitypub\Activitypub', 'schedule_post_activity' ), 10, 3 );
+
+		\add_filter( 'preprocess_comment', array( '\Activitypub\Activitypub', 'preprocess_comment' ) );
+		\add_filter( 'comment_post', array( '\Activitypub\Activitypub', 'postprocess_comment' ), 10, 3 );
+		\add_filter( 'wp_update_comment_data', array( '\Activitypub\Activitypub', 'comment_updated_published' ), 20, 3 );
+		\add_action( 'transition_comment_status', array( '\Activitypub\Activitypub', 'schedule_comment_activity' ), 20, 3 );
+		\add_action( 'edit_comment', array( '\Activitypub\Activitypub', 'edit_comment' ), 20, 2 );//schedule_admin_comment_activity
+		\add_filter( 'get_comment_text', array( '\Activitypub\Activitypub', 'comment_append_edit_datetime' ), 10, 3 );
```
pfefferle, line 33: maybe we should add a `class-comment.php` file?

includes/activitypub.js
```diff
@@ -0,0 +1,52 @@
+(function($) {
```
pfefferle, line 1: I would not mix up PHP and JS code. Maybe we should add an assets folder to the root, including JS/CSS/... files:

`/assets/js/activitypub.js`

pfefferle:
We should also add some tests.

includes/class-activitypub.php
pfefferle: The Webmention plugin uses the `replytocom` query var, does that maybe work?

includes/assets/js/activitypub.js
```diff
@@ -0,0 +1,52 @@
+(function($) {
```
pfefferle, line 1: thanks for moving, but I would place the `assets` folder in the root folder, not inside the `includes` folder.

mediaformat:
@pfefferle I added a view **Posts with (activitypub) Comments** to the Migration page. 

There is a column which displays number of (published) comments. However some posts in this view may display `0` comment count (this number is only for published comments). Do you find that confusing? or have any suggestion? 

Screen Shot 2022-11-19 at 5 15 32 PM

includes/class-comments.php
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	/**
+	 * preprocess_comment()
+	 * preprocess local comments for federated replies
+	 */
+	public static function preprocess_comment( $commentdata ) {
+		// only process replies from local actors
+		if ( ! empty( $commentdata['user_id'] ) ) {
+			$commentdata['comment_type'] = 'activitypub';
```
pfefferle, line 31: shouldn't this be `comment_type` be `comment` and the `protocol` comment-meta be `activitypub`? Otherwise we can not differ between boosts, shares and comments.

mediaformat: I wrote that part of the code years ago 😅

It doesn't seem to interfere with likes and shares in `handle_reaction`, 
```
'comment_type' => \esc_attr( \strtolower( $object['type'] ) ),
```

I think the original concern was how to only process/federate replies from authors. I will review and refactor this. This also gives me an idea for a Test Case

mediaformat: @pfefferle For defining whose reply can federate, do you have a preference between `$user->roles`(ex. Author) & `$user->has_cap`(ex. `publish_posts`)? Capabilities are more flexible where there are custom roles, but could also have unintended outcomes.

templates/tools-page.php
```diff
@@ -0,0 +1,59 @@
+<?php 
+
+$action_nonce = wp_create_nonce( 'activitypub_action' );
+if ( isset( $_REQUEST['post_url'] ) && $_REQUEST['page'] == "activitypub_tools" ) {
```
valerio-bozzolan, line 4: (I suggest `===` instead of `==` for string comparison)

includes/model/class-comment.php
```diff
@@ -0,0 +1,291 @@
...
+	private $update;
+	private $deleted;
+	private $replies;
+
+	/**
+	 * Initialize the class
+	 */
+	public function __construct( $comment = null ) {
+		$this->comment 			  = $comment;
+		$this->id 				  = $this->generate_comment_id();
```
valerio-bozzolan, line 29: Sorry for my very unuseful comment: but probably here it should be aligned with spaces instead of tabs

(I love indenting with tabs and aligning with spaces ihih)

templates/replies-json.php
```diff
@@ -0,0 +1,84 @@
...
+$post = \get_post();
+
+$page =  \get_query_var( 'page' );
+$args = array(
+	'status'  => 'approve',
+	'number'  => '10',
+    'offset'  => $page,
+    'type'    => 'activitypub',
+	'post_id' => $post->ID,
+    'order'   => 'ASC',
```
valerio-bozzolan, line 12: Again sorry for my unuseful comment: probably here the spaces should be replaced with tabs, in order to indent with tabs

mediaformat: running phpcs/phpcbf now ;)

pfefferle:
@mediaformat I will fokus on this PR during the holidays and had a look at the migration part... I am not sure if this is a good idea, because it could be very distructive! I think we should stick with backwards compatibility in the first place.

mediaformat:
> @mediaformat had a look at the migration part... I am not sure if this is a good idea, because it could be very distructive! I think we should stick with backwards compatibility in the first place.

Ok, I'll remove all that code. Backwards Compat then will only be checking first for permalink...

pfefferle:
What do you think about this way (at least as quick fix) https://github.com/pfefferle/wordpress-activitypub/pull/231 ?

mediaformat:
@pfefferle the `\add_post_meta( $post_id, 'activitypub_canonical_url', \get_permalink( $post_id ), true );` looks good

johnspurlock:
Is there a public WP instance where this is available to read from?  Curious in how the AP payloads look for comments and posts.

sunjam:
It is a pull request and not actually in stable afaik

johnspurlock:
I know, I'm just wondering if there is a staging server or something

mediaformat:
There isn't a staging server, I'm testing this locally with ngrok, and mastodon. The AP payloads are pretty typical/similar to other implementations

johnspurlock:
Congrats on the [acquisition](https://wptavern.com/automattic-acquires-activitypub-plugin-for-wordpress) - can't wait to unlock ActivityPub comments/replies for WP posts - hopefully this is still a priority!

mediaformat:
Ok, a few things broke and are left to fix

- processing remote comments (update, delete)

Also, I'll probably extract the post/comments replies collection into a separate PR, to reduce the scope of this one.

includes/class-activitypub.php
```diff
@@ -91,7 +91,11 @@ public static function render_json_template( $template ) {
 		}
 
 		if ( \is_author() ) {
-			$json_template = ACTIVITYPUB_PLUGIN_DIR . '/templates/author-json.php';
+			$json_template = \dirname( __FILE__ ) . '/../templates/author-json.php';
```
pfefferle, line 94: please use `ACTIVITYPUB_PLUGIN_DIR`

includes/class-activitypub.php
```diff
@@ -91,7 +91,11 @@ public static function render_json_template( $template ) {
...
+		} elseif ( \Activitypub\is_ap_replies() ) {
+			$json_template = \dirname( __FILE__ ) . '/../templates/replies-json.php';
```
pfefferle, line 96: please use `ACTIVITYPUB_PLUGIN_DIR`

includes/functions.php
```diff
@@ -173,6 +173,327 @@ function url_to_authorid( $url ) {
...
+	return null;
+}
+
+/**
+ * get_parent_uri (in reply to)
+ * takes a comment and returns
+ * @param WP_Comment activitypub object id URI
+ * @return int comment_id
+ */
+function get_in_reply_to( $comment ) {
```
pfefferle, line 307: shouldn't this be part of the Comment Class?

includes/functions.php
```diff
@@ -173,6 +173,327 @@ function url_to_authorid( $url ) {
...
+			$in_reply_to = \get_permalink( $comment->comment_post_ID );
+		}
+	}
+	return $in_reply_to;
+}
+
+/**
+ * Checks a comment ID for a source_id, or source_url
+ */
+function get_source_id( $comment ) {
```
pfefferle, line 331: shouldn't this be part of the Comment Class?

includes/functions.php
```diff
@@ -173,6 +173,327 @@ function url_to_authorid( $url ) {
...
+	}
+	return null;
+}
+
+/**
+ * Verify if url is a wp_ap_comment,
+ * Or if it is a previously received remote comment
+ * @return int comment_id
+ */
+function is_ap_comment() {
```
pfefferle, line 363: I would prefer to remove the `ap`. It is namespaced, so should be no problem.

includes/functions.php
```diff
@@ -173,6 +173,327 @@ function url_to_authorid( $url ) {
...
+		}
+	}
+	return null;
+}
+
+/**
+ * Verify if url has a replies query,
+ * @return bool
+ */
+function is_ap_replies() {
```
pfefferle, line 379: I would prefer to remove the `ap`. It is namespaced, so should be no problem.

includes/functions.php
```diff
@@ -173,6 +173,327 @@ function url_to_authorid( $url ) {
...
+	return null;
+}
+
+/**
+ * Get recipients to insert/tag in reply tag from received AP object meta
+ * @param string $object_id a comment_id to search
+ * @param boolean $post defaults to searching a comment_id
+ * @return string space separated webfinger of tagged users
+ */
+function reply_recipients( $object_id, $post = null ) {
```
pfefferle, line 393: shouldn't this be part of the Comment Class?

includes/functions.php
```diff
@@ -173,6 +173,327 @@ function url_to_authorid( $url ) {
...
+		$comment = \get_comment( $object_id );
+		preg_match_all( '/@' . ACTIVITYPUB_USERNAME_REGEXP . '/', $comment->comment_content, $recipients );
+		return implode( ' ', $recipients[0] );
+	}
+}
+
+/**
+ * Add summary to reply
+ */
+function get_summary( $comment_id ) {
```
pfefferle, line 430: this is a very generic name for a Comment only thing, maybe also move to Comment Class?

includes/functions.php
```diff
@@ -173,6 +173,327 @@ function url_to_authorid( $url ) {
...
+			return \esc_attr( $ap_object['object']['summary'] );
+		}
+	}
+}
+
+/**
+ * @param string $user_url
+ * @return string $webfinger
+ */
+function url_to_webfinger( $user_url ) {
```
pfefferle, line 443: Maybe move to WebFinger Class?

templates/replies-json.php
```diff
@@ -0,0 +1,83 @@
...
+	$collection_page->next  = \add_query_arg( 'page', $page + 1, $collection_page->partOf ); // phpcs:ignore
+	$collection_page->last  = \add_query_arg( 'page', \ceil ( $collection_page->totalItems / 10 ), $collection_page->partOf ); // phpcs:ignore
+}
+
+foreach ( $comments as $comment ) {
+	$remote_url = \get_comment_meta( $comment->comment_ID, 'source_url', true );
+	if ( $remote_url ) { //
+		$collection_page->items[] = $remote_url;
+	} else {
+		$activitypub_comment = new \Activitypub\Model\Comment( $comment );
```
pfefferle, line 38: there is no model any more. (I think of re-adding it, also for posts).

templates/replies-json.php
```diff
@@ -0,0 +1,83 @@
...
+		$activitypub_activity = new \Activitypub\Model\Activity( 'Create', \Activitypub\Model\Activity::TYPE_NONE );
```
pfefferle, line 39: This is also the old Class model.

pfefferle:
@mediaformat are you fine with me adding some changes directly to your PR?

includes/class-mention.php
```diff
@@ -146,23 +145,4 @@ public static function get_inbox_by_mentioned_actor( $actor ) {
...
-	public static function extract_mentions( $mentions, $post_content ) {
-		\preg_match_all( '/@' . ACTIVITYPUB_USERNAME_REGEXP . '/i', $post_content, $matches );
-		foreach ( $matches[0] as $match ) {
-			$link = Webfinger::resolve( $match );
-			if ( ! is_wp_error( $link ) ) {
-				$mentions[ $match ] = $link;
-			}
-		}
-		return $mentions;
-	}
```
mediaformat, line 168: @pfefferle Was deleting this an error?

includes/transformer/class-comment.php
```diff
@@ -185,50 +121,15 @@ protected function get_tags() {
...
 	}
 
-	/**
-	 * Helper function to get the @-Mentions from the comment content.
-	 *
-	 * @return array The list of @-Mentions.
-	 */
-	protected function get_mentions() {
-		return apply_filters( 'activitypub_extract_mentions', array(), $this->wp_comment->comment_content, $this->wp_comment );
-	}
```
mediaformat: @pfefferle idem

mediaformat (commented):
wondering the extract mentions filter was removed

includes/class-mention.php
pfefferle: No, it was by intention, but not finished yet ;)

The removed version implements the same functionality as the Mention class already solves, so I planned to rewrite the Mention class a bit, to also support comments or finding mentions in texts in general.

pfefferle:
@mediaformat I planned to also work a bit more on the topic but then came the https://github.com/Automattic/wordpress-activitypub/issues/523.

I would love to:
* [ ] Simplify the code a bit (move global functions to transformer and similar stuff)
* [ ] use the Mention class also for comments
* [ ] Start with inbox-handlers to have the code a bit more structured. If we have a good idea how to do it, we can adapt it for `follow` and `Delete` afterwards. Something like
  * `/handler/class-create.php`
  * `/handler/class-like.php`
  * `/handler/class-update.php` 

What do you think?

includes/class-mention.php
mediaformat: @pfefferle ok, probably an error then because this ^ code IS the from the Mentions class ;-)

mediaformat:
> I would love to:
>     * simplify the code a bit (move global functions to transformer and similar stuff)
>     * Start with inbox-handlers to have the code a bit more structured. 
> What do you think?

I agree with the simplification, and inbox handlers and all.

I ended up re-added some of the code that was removed, because I wanted to fix some bugs from where I left off last time (federation wasn't working well for deletes, edits). But it took a while to figure out why things weren't working. 😅

Anyhow everything is working again: threads, mentions, replies, edits, deletes. 🎉

includes/transformer/class-comment.php
```diff
@@ -63,8 +63,9 @@ public function to_object() {
 		$comment = $this->wp_comment;
 		$object = new Base_Object();
 
-		$object->set_id( $this->get_id() );
+		$object->set_id( $this->get_id( $comment ) );
```
pfefferle, line 66: why this? we already have the comment as object var. no need to pass it to `get_id()`?!?

includes/transformer/class-comment.php
```diff
@@ -63,8 +63,9 @@ public function to_object() {
...
 		$object->set_url( \get_comment_link( $comment->ID ) );
+		$object->set_context( $this->get_context() );
```
pfefferle, line 68: same as above... we have the comment as object attribute... it feels a bit weird to use the getter to fill the setter...

pfefferle:
I would love to keep in in draft state for the next week, that it will not be merged by accident.

pfefferle:
`?replytocom=1` seems to redirect even with ActivityPub Accept header.

mediaformat:
> `?replytocom=1` seems to redirect even with ActivityPub Accept header.

I believe there are some checks, comment 1 is not an author

includes/transformer/class-comment.php
mediaformat: we need somethings more flexible, so that if a reply is to a local post we can pass in the parent_comment

includes/class-activity-dispatcher.php
```diff
@@ -30,48 +33,48 @@ public static function init() {
...
 
-		$object = Post::transform( $wp_post )->to_object();
+		$object = Factory::get_transformer( $wp_object )->to_object();
 
 		$activity = new Activity();
 		$activity->set_type( $type );
 		$activity->set_object( $object );
 
-		$follower_inboxes  = Followers::get_inboxes( $wp_post->post_author );
+		$follower_inboxes  = Followers::get_inboxes( $wp_object->post_author );
```
mediaformat, line 77: We will need a better check here since the comment->post_author isn't guaranteed to be the same user as the comment author

mediaformat (commented):
See comment

includes/transformer/class-comment.php
pfefferle: but then I would use a helper function to be a bit more consistent

includes/class-activity-dispatcher.php
pfefferle: good point! what do you think about transforming the object before scheduling it? so that the dispatcher already get the Activity Object?

pfefferle: but this seems to be a bigger change... let's try to fix this first...

PR #169, opened 2022-11-08
Unmark comments as spam (maybe needs an option?)

colinleroy:
Hi,
I've noticed comments from the Fediverse have a tendancy to get flagged as spam by Akismet. 
I think that's unfortunate as there - for now ? - a lot less chance that this is the case.

This PR unflags spam on these comments and mark them for moderation instead.

Hope this helps, and thanks for the nice plugin!

### DIFF
--- a/includes/rest/class-inbox.php
+++ b/includes/rest/class-inbox.php
@@ -390,7 +390,12 @@ class Inbox {
 		// do not require email for AP entries
 		\add_filter( 'pre_option_require_name_email', '__return_false' );
 
-		$state = \wp_new_comment( $commentdata, true );
+		$commentid = \wp_new_comment( $commentdata, true );
+
+		if ('spam' === \wp_get_comment_status($commentid)) {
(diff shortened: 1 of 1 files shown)


pfefferle:
Hey @colinleroy thanks for your pull request! What does `comment_status = 0` represent? Shouldn't it be hold, so that the site admin can review the comment?

colinleroy:
Hi! Yes, "0" is hold for moderation ("1" is published, and "spam" is spam)
Hope this helps :)

colinleroy:
I should try to see if Akismet has some sort of log to explain its decision, I'm unsure it does. I'll tell you :)

pfefferle:
I do not think akismet is moving the comment, because I do not use it and have also federated comments in spam. I think this is a core glitch.

colinleroy:
Interesting! I'll try to do more checks if it can help

pfefferle:
We are working with Akismet and AntispamBee to fix this issue, so I will close this for now.

PR #172, opened 2022-11-09
Add a parser to the Friends Plugin

akirk:
With this patch your WordPress can not only be followed on ActivityPub platforms (like Mastodon) but you can also follow people on those platforms, with the plan eliminating the need for having another account on an instance.

If the Friends plugin is activated on your WordPress then it will add an ActivityPub "parser" (that's the concept of getting content into the Friends plugin although here we're obviously subscribing to the ActivityPub activities and will receive them in real time).

Here are some screenshots of the patch in action:
Screenshot 2022-11-09 at 07 11 04

Screenshot 2022-11-09 at 07 11 23

Screenshot 2022-11-09 at 07 11 57

Screenshot 2022-11-09 at 07 20 13

While I have marked this PR as draft, it is functional but probably needs some refinement.

I also took the liberty to refactor the Health Check for the Webfinger URL into a `Webfinger::resolve` which is needed for determining the right URL from a username in the `@username@domain` format.


pfefferle:
This is awesome and exactly what I had in mind as we first talked about a cooperation!

akirk:
Ok, I released [2.1.0 of the Friends plugin](https://github.com/akirk/friends/releases/tag/2.1.0) with updated filters for this, so this PR can now move forward.

pfefferle:
Do you have an idea how we can guide this feature?

akirk:
What I do for plugins like [friends-parser-fraidyscrape](https://github.com/akirk/friends-parser-fraidyscrape) when they are installed on their own is to put the Friends icon there and an About menu entry like this:

Screenshot 2022-11-13 at 12 37 27

But I don't think it's such a good idea in this case, so I made this little mockup. We could add a "About" or "Welcome" screen in addition to the settings like this:

Screenshot 2022-11-13 at 13 10 27

The current settings would be on their own tab:
Screenshot 2022-11-13 at 12 43 22

Here is the text I whipped up for the about section, it would also allow to explain things better than today. I initially had a hard time understanding what service the plugin actually provided and which username to follow.
```
<p>ActivityPub turns your blog into a federated social network. This means you can share and talk to everyone using the ActivityPub protocol, including users of Friendica, Pleroma and Mastodon.</p>
<p>After you used <a href="/wp-admin/site-health.php">Site Health</a> to ensure that your site is compatible, people can follow you by using the username <tt>alex@akirk.blog</tt></p>
<p>To follow people on Mastodon or similar platforms using your own WordPress, you can use the <a href="https://wordpress.org/plugins/friends">Friends Plugin for WordPress</a> which uses this plugin to receive posts and display them on your own WordPress, thus making your own WordPress a Mastodon instance of its own.</p>
```

timnolte:
@akirk I'd like to ask a clarifying question. Does the Friends Plugin create posts from those you follow on the frontend of the WordPress site or just in the backend?

timnolte:
@akirk another question is whether the Friends Plugin works for non-Admin accounts. I explicitly do not use an Admin account even on my own personal sites. If the Friends Plugin has only been tested with an Admin account, and assumes all functionality within the Admin account context then that really needs to be changed. From a security standpoint people shouldn't be using a full admin account on their sites as their main account.

akirk:
There terms frontend and backend are probably not precise enough. There is a "friends frontend" at `/friends/` and the remote posts are cached in [a custom post type](https://github.com/akirk/friends/blob/79c9dec0d57a9837e6ddcfa8a4c7653c1e7ef17e/includes/class-friends.php#L163-L201) and those are displayed on that page. They are not exposed elsewhere to the public.

To your point about administrator priviledges, I have opened an issue here: https://github.com/akirk/friends/issues/120

It's a

timnolte:
@akirk how is this handling the visibility of posts? I ask because on Mastodon people can set a Toot to only be visible to followers, those posts shouldn't be publicly displayed on a WordPress instance.

akirk:
As I said, the cached post type is meant for your eyes and your friends page, plus external posts (at least from other Friends plugin users) that are posted as private, will also have the post_status private.

integration/class-friends-feed-parser-activitypub.php
```diff
@@ -0,0 +1,198 @@
...
+	const NAME = 'ActivityPub';
+	const URL = 'https://www.w3.org/TR/activitypub/';
+	private $friends_feed;
+
+	public function __construct( \Friends\Feed $friends_feed ) {
+		$this->friends_feed = $friends_feed;
+
+		\add_action( 'activitypub_inbox_create', array( $this, 'handle_received_activity' ), 10, 2 );
+		\add_action( 'activitypub_inbox_accept', array( $this, 'handle_received_activity' ), 10, 2 );
+		\add_filter( 'friends_user_feed_activated', array( $this, 'follow_user' ), 10 );
```
pfefferle, line 23: I used the cron to send the activities asynchronous, to not slow down activities on the WordPress plattform.

* https://github.com/pfefferle/wordpress-activitypub/blob/master/includes/class-activitypub.php#L121
* https://github.com/pfefferle/wordpress-activitypub/blob/master/includes/class-activity-dispatcher.php

akirk: Good point, I added that.

pfefferle (approved):
LGTM;

@akirk have you verified that it is functional?

akirk:
It didn't work. fixed it with another commit.

akirk:
@pfefferle, I have now added support for the `Announce` type = Mastodon boosts. I have also added unit tests, they only kick in [if you have the Friends plugin in the same directory as ActivityPub](https://github.com/pfefferle/wordpress-activitypub/pull/172/files#diff-96a82592013e5f4b91e682332583b7a4a6ca1ff0431d267c61179b1a21b167daR22-R27). I have reused some unit test code from the Friends plugin that invokes the REST API, so this also tests the [`user_inbox_post` function](https://github.com/pfefferle/wordpress-activitypub/blob/010ccc7e228618f10f6473c7a1f35a3b4fbb9c92/includes/rest/class-inbox.php#L141).

pfefferle:
@akirk ready to merge?

akirk:
Yes, I just also quickly added attachment support, the unit tests help ensure that things still work :) 

There will be follow-up PRs to improve it further but for now I am happy!

pfefferle (approved):
LGTM

PR #189, opened 2022-11-20
Add Custom Post Type support to outbox API

blueset:
This add contents from all enabled post types into the outbox API output, instead of the previously hardcoded `"post"` type.


pfefferle (approved):
LGTM

pfefferle:
Thanks @blueset 🙂

PR #198, opened 2022-11-23
fix #191 implement generate_the_title

### LINKED ISSUE #191: generate_the_title function doesn't appear to exist
this line
https://github.com/pfefferle/wordpress-activitypub/blob/master/includes/model/class-post.php#L24
sets the summary field of the post (which ends up as a content warning in Mastodon)
however I am getting null in the output, which appears to be because there is no generate_the_title function

AlanBell:
fix #191 by implementing a function to get the title, which becomes the ActivityPub summary/subject/content warning field.

### DIFF
--- a/includes/model/class-post.php
+++ b/includes/model/class-post.php
@@ -75,9 +75,14 @@ class Post {
 		return \str_replace( '__trashed', '', $permalink );
 	}
 	
-	public function generate_the_title(){
-		$post = $this->post;
-		return \get_the_title( $post->ID );
+	public function generate_the_title() {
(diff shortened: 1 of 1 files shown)


pfefferle:
Why have you changed to "note"? I liked the limitation to "Article", because "note" does not have a title and is more a microblogging style thing.

AlanBell:
Because my test blog posts were not working, because they were Note rather than Article. Does Wordpress makes things Note if they are short?

pfefferle:
there is an option in the settings of the plugin where you can choose to use the cutom formats of wordpress as AP formats, that means if you publish a Note, it will be a note.

Issue #192, opened 2022-11-20
Object of class WP_Error could not be converted to string


YiffyToys:
Version: 0.13.4
Wordpress 6.1.1
PHP 7.4-fpm
Apache

```
Got error 'PHP message: PHP Fatal error:  Uncaught Error: Object of class WP_Error could not be converted to string in ***/wp-includes/http.php:687\n
Stack trace:\n
#0 ***/wp-content/plugins/activitypub/includes/class-signature.php(76): wp_parse_url()\n
#1 ***/wp-content/plugins/activitypub/includes/functions.php(38): Activitypub\\Signature::generate_signature()\n
#2 ***/wp-content/plugins/activitypub/includes/rest/class-inbox.php(341): Activitypub\\safe_remote_post()\n
#3 ***/wp-includes/class-wp-hook.php(308): Activitypub\\Rest\\Inbox::handle_follow()\n
#4 ***/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters()\n
#5 ***/wp-includes/plugin.php(517): WP_Hook->do_action()\n
#6 ***/wp-content/plugins/activitypub/includes/rest/class-inbox.php(149)...'
```

I am experiencing the "waiting for approval" issue mentioned in
https://wordpress.org/support/topic/approve-follow-request-redux/

pfefferle:
Can you tell me your blog URL and the Username, so that I can have a look?

YiffyToys:
@suran@blog.yiffytoys.de

YiffyToys:
I have moved to a regular Mastodon server and another WP plugin.

thomas-pike:
Also experiencing this issue, same symptoms and same stacktrace. I added a debug line in line 75 of `includes/class-signature.php`: `error_log(print_r($url, true));` to find out what the `WP_Error` object being passed to `wp_parse_url` was. 

Result:
```
WP_Error Object
(
    [errors] => Array
        (
            [activitypub_no_valid_actor_url] => Array
                (
                    [0] => The "actor" is no valid URL
                )

        )

    [error_data] => Array
        (
            [activitypub_no_valid_actor_url] => https://social.vivaldi.net/users/example
        )

    [additional_data:protected] => Array
        (
        )

)
```

[...]

PR #206, opened 2022-12-04
fix for #192

craigconstantine:
Object of class WP_Error could not be converted to string #192

https://github.com/pfefferle/wordpress-activitypub/issues/192


craigconstantine:
upon further consideration, this is not correct.

PR #210, opened 2022-12-06
fix user count

### LINKED ISSUE #209: Adjust nodeinfo user count
Nodeinfo reports all WP users accounts, even though only those with the role author and above can technically post.
 
`$users = \count_users();`

Original report @heleucht https://pirati.ca/display/ec054ce7-1363-8f94-6546-944144035152

pfefferle:
props @mediaformat

fix #209

### DIFF
--- a/includes/rest/class-nodeinfo.php
+++ b/includes/rest/class-nodeinfo.php
@@ -95,10 +95,6 @@ class Nodeinfo {
 			'outbound' => array(),
 		);
 
-		$nodeinfo['metadata'] = array(
-			'email' => \get_option( 'admin_email' ),
-		);
-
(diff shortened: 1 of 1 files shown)


mediaformat:
LGTM

includes/rest/class-nodeinfo.php
```diff
@@ -95,10 +95,6 @@ public static function nodeinfo( $request ) {
 			'outbound' => array(),
 		);
 
-		$nodeinfo['metadata'] = array(
-			'email' => \get_option( 'admin_email' ),
```
mediaformat, line 99: ✊🏼

PR #213, opened 2022-12-09
Outgoing Mentions

akirk:
With this Pull Request, when you mention another ActivityPub account in the form `username@example.org` (or `@username@example.org`), it will add a "tag" and mention the user so that they'll get a notification. It will also send it to their inbox, even if they are not following you.

In the Friends Plugin, you can use the Gutenberg autocomplete to just mention a user and it will rewrite the mention to the correct ActivityPub username:

Screenshot 2022-12-09 at 12 03 31

### Other

This adds a number of unit tests to ensure that this doesn't break anything.


akirk:
By default, in the Friends plugin, @-mentions of others will be hidden (see https://github.com/akirk/friends/pull/124). This can be changed with a setting on the Edit Feeds page:

Screenshot 2022-12-10 at 09 33 23

activitypub.php
```diff
@@ -144,3 +148,49 @@ function( \Friends\Feed $friends_feed ) {
 		$friends_feed->register_parser( Friends_Feed_Parser_ActivityPub::SLUG, new Friends_Feed_Parser_ActivityPub( $friends_feed ) );
 	}
 );
+
+/**
+ * Disable webfinger for known example domains.
+*/
+add_filter(
```
pfefferle, line 155: shouldn't this be better part of the test and/or of the `debug.php` file?

akirk: Hm, the problem that I've been running into is that the `the_content` hook is called for any posts. So if you write a post that talks about Mastodon and uses example-mentions, the filter would kick in and try to look it up.

akirk: Ok, I have decided to move the whole parser into the Friends plugin https://github.com/akirk/friends/pull/137 so that I can add a filter there and I also feel it's better for me to be able to iterate more quickly on it for everything that does not need any ActivityPub protocol work.

HammyHavoc:
This looks great! Once available, I'll tag the authors of ActivityPub and Friends in on a post mentioning my use of it to give it a test.

pfefferle:
@akirk how difficult is it to port the Gutenberg autocomplete feature for the classic block editor. I fear that no one understands the feature and/or does not know how to use it, if there is no GUI?!?

akirk:
The autocomplete is actually not part of this PR since it autocompletes local users which are only present if you follow someone with the Friends plugin. You need to manually type the user you'd like to mention and it will be extracted here: https://github.com/pfefferle/wordpress-activitypub/pull/213/files#diff-0c17a58839e890fd735824ceba1483d38496d10ef779994bc90ab9d53843e763R76

akirk:
The idea of this PR is that if you do mention someone via their ActivityPub address, they also get [cc-ed](https://github.com/pfefferle/wordpress-activitypub/pull/213/files#diff-a5454f0022839a124be6a9068d22bd76fb3ec1f68aa5025c987139b10f4bcc12R62) and [tagged](https://github.com/pfefferle/wordpress-activitypub/pull/213/files#diff-44de75651674f44bc586c0f66e8dddf70ec9c6a4a9eb72418be87aa6c12a6be7R162).

pfefferle:
I understood that, I just feared that noone will understand that this is possible. Maybe we can add a block (that might check the webfinger address) in the future.

pfefferle (approved):
Thanks a lot!

remouk:
Hello, I'm not really sure how this is working, but the autocomplete is messing up my content. I have a page where I give my contacts: where to find me / follow me. For Mastodon I wrote "@myusername@mywebsite.com" in plain text.

Since the update, it's automatically replaced with "@myusername" which is a link to my "https://mywebsite.com/author/myusername"

I cannot bypass the change, I cannot simply share my Mastodon account anymore.

Am I missing something?

PR #215, opened 2022-12-10
Issue 214

### LINKED ISSUE #152: Cannot follow user with email address as name
In a WordPress environment with ActivityPub installed, I can successfully follow @Downes@leftish.media from Mastodon, but cannot follow @stephen@downes.ca@leftish.media

The problem may be related to the behaviour of the Web Finger plugin, as 'Downes' resolves nicely: https://leftish.media/.well-known/webfinger?resource=acct:Downes@leftish.media 
while 'stephen@downes.ca' does not: https://leftish.media/.well-known/webfinger?resource=acct:stephen@downes.ca@leftish.media

The error appears to be caused by the parsing of username and website URL from the identifier. It appears to be splitting at the first '@' which would result in a username of 'stephen' and a website of 'downes.ca@leftish.media' whereas it should be split at the last '@', which would result in a username of 'stephen@downes.ca' and 'leftish.media'.

This issue may exist in several plugins. It may also exist in ActivityPub clients as well, such as Mastodon.

### LINKED ISSUE #214: Webfinger not compatible with PHP < 8.0
I don't know if this is a valid bug because I'm not sure what the minimum requirements of this plugin are really supposed to be.  The readme specifies PHP 5.6, and class-webfinger.php is the only place I can find a reference to str_contains.  So my best theory is that this was an unintended side-effect of the fix to #152.  I'm either saying that this code should be made backwards-compatible, or that the readme or other documentation should be clarified.  I just don't know which.

Anyway.  I access this URL on an internal test machine:

    https://test1-wordpress.fritz.box/index.php?rest_route=/activitypub/1.0/webfinger&resource=acct:test@test1-wordpress.fritz.box

And I get:

    Fatal error: Uncaught Error: Call to undefined function str_contains() in /srv/wordpress/test_wordpress/wp-content/plugins/activitypub/includes/rest/class-webfinger.php:47 Stack trace: #0 /usr/share/wordpress/wp-includes/rest-api/class-wp-rest-server.php(1139): Activitypub\Rest\Webfinger::webfinger() #1 /usr/share/wordpress/wp-includes/rest-api/class-wp-rest-server.php(985): WP_REST_Server->respond_to_request() #2 /srv/wordpress/test_wordpress/wp-content/plugins/activitypub/includes/rest/class-server.php(22): WP_REST_Server->dispatch() #3 /usr/share/wordpress/wp-includes/rest-api/class-wp-rest-server.php(412): Activitypub\Rest\Server->dispatch() #4 /usr/share/wordpress/wp-includes/rest-api.php(354): WP_REST_Server->serve_request() #5 /usr/share/wordpress/wp-includes/class-wp-hook.php(292): rest_api_loaded() #6 /usr/share/wordpress/wp-includes/class-wp-hook.php(316): WP_Hook->apply_filters() #7 /usr/share/wordpress/wp-includes/plugin.php(551): WP_Hook->do_action() #8 /usr/share/wordpress/wp-includes/class-wp.php(388): do_a in /srv/wordpress/test_wordpress/wp-content/plugins/activitypub/includes/rest/class-webfinger.php on line 47`

This is on Raspbian 11.  PHP 7.4. Apache 2.4.54.  Wordpress 5.7.5.  If any of those things are not compatible, it probably should be made more clear in the readme.

mexon:
This reverts the previous fix to #152, and replaces it with a new implementation that is compatible with PHP 7.4 and therefore fixes #214.  Or at least, I think this fixes #152.  This has not been extensively tested.  All I did was change the regex `([^@]+)@(.+)` to `(.+)@([^@]+)`, so that the last at-sign will be the separator instead of the first.  So perhaps treat it as a suggestion.

### DIFF
--- a/includes/rest/class-webfinger.php
+++ b/includes/rest/class-webfinger.php
@@ -44,16 +44,15 @@ class Webfinger {
 	public static function webfinger( $request ) {
 		$resource = $request->get_param( 'resource' );
 
-		$matched = \str_contains( $resource, '@' );
+		$matches = array();
+		$matched = \preg_match( '/^acct:([^@]+)@(.+)$/', $resource, $matches );
 
 		if ( ! $matched ) {
 			return new \WP_Error( 'activitypub_unsupported_resource', \__( 'Resource is invalid', 'activitypub' ), array( 'status' => 400 ) );
 		}
 
-		$resource = \str_replace( 'acct:', '', $resource );
-
-		$resource_identifier = \substr( $resource, 0, \strrpos( $resource, '@' ) );
-		$resource_host = \substr( \strrchr( $resource, '@' ), 1 );
+		$resource_identifier = $matches[1];
+		$resource_host = $matches[2];
 
 		if ( \wp_parse_url( \home_url( '/' ), \PHP_URL_HOST ) !== $resource_host ) {
 			return new \WP_Error( 'activitypub_wrong_host', \__( 'Resource host does not match blog host', 'activitypub' ), array( 'status' => 404 ) );
@@ -98,7 +97,7 @@ class Webfinger {
 		$params['resource'] = array(
 			'required' => true,
 			'type' => 'string',
-			'pattern' => '^acct:(.+)@(.+)$',
+			'pattern' => '^acct:([^@]+)@(.+)$',
 		);
 
 		return $params;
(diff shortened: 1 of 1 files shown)


pfefferle:
I've made a quick fix, to release a new version ASAP! But I will also have a look at your proposal! Maybe we have to add some tests.

mexon:
Oh nice, thank you for the quick work!  Confirmed that your change fixes the problem for me.  I didn't expect such a quick response, otherwise I wouldn't have dived in 🤣

I'm pretty helpless with running PHP unit tests locally, I don't submit changes like that often enough for it to be second nature, and I don't have much spare time.  So I think I'll close this PR for now as being unnecessary.

PR #231, opened 2022-12-27
Security & privacy related fixes

pfefferle:


pfefferle:
just tested it on a live system and it seem to work as expected

Issue #233, opened 2022-12-28
Replies to comments from Friendica not surfacing on blog


mexon:
I'm testing the plugin against Friendica.  Suppose I post a WP blog entry A, and then a Friendica user posts a reply B.  This goes through to WP fine, and the reply B shows up as a comment and can be approved.  Suppose another Friendica user replies to B with their own reply C.  This reply does not show up in WP at all.

This is not what I would expect.  In general I would want all the discussion threads from around the Fediverse to arrive as comments, with the option of approving them and displaying them on the blog.

(Edit: I just realised "threaded comments support" is in the README as to implement.  So I guess this is a feature.)

mexon:
Digging in the code, I can see in `handle_create` that it uses the URL from the `inReplyTo` field to look up a post with `url_to_postid`.  But when I look at the data sent by Friendica, the original post is supplied as the `context`, while `inReplyTo` is the URL of the comment it replies to.  So it seems to me that:

1. `handle_create` should look up the `context` as well as / instead of `inReplyTo`
2. `handle_create` should also look up a comment using `inReplyTo`, and if present use that as the `comment_parent` field

However, I haven't checked what the ActivityPub spec says should happen here, and I don't know that much about the comment fields offhand to be sure I'm interpreting this right.

If my understanding is right, however, I could probably try implementing this myself.

pfefferle:
Keep

PR #241, opened 2023-01-03
support threaded comments from ActivityPub

mexon:
This change supports replies to comments as threaded comments in wordpress.  It implements #233.

Originally I planned to get the post ID from the `context`.  However, it turns out only older versions of Friendica populate that field.  So instead I have to get it by looking up the comment with `inReplyTo`.  I left the support for `context` in there in case other ActivityPub actors use it that way.  Otherwise it's not always possible to correctly thread together replies that are out of order.


mexon:
I'm not happy with the code as it is here, I think it needs refactoring into separate functions.  But I'll make the PR anyway.

PR #242, opened 2023-01-03
support threaded comments from ActivityPub

mexon:
This change supports replies to comments as threaded comments in wordpress. It implements https://github.com/pfefferle/wordpress-activitypub/issues/233.

Originally I planned to get the post ID from the context. However, it turns out only older versions of Friendica populate that field. So instead I have to get it by looking up the comment with inReplyTo. I left the support for context in there in case other ActivityPub actors use it that way. Otherwise it's not always possible to correctly thread together replies that are out of order.

The other fun part is that the source ID and URL are different.  Until now it's only been storing the URL, but that's not sufficient for looking up inReplyTo.  So now I have to store the source ID as well when comments are received.


mexon:
I'm not happy with the code as it is here, I think it needs refactoring into separate functions. But I'll make the PR anyway.

mexon:
OK, I've refactored the code and it seems to work.  This is ready for review now.

pfefferle:
Oh, this is awesome, can you maybe add one or two unittests? let me know if I could/should help!

mexon:
Yeah, I should totally do unit tests.  That will involve figuring out how to run them locally.  Don't expect a result for a while - I'm travelling this weekend and will be really busy at the other end.  I'll send you some questions if I get stuck.

mexon:
@pfefferle I've just had a go at adding the first unit tests (as warned, I don't have much time to work on this, so I'm slow).  Could you comment on the style or structure?  I decided to make `convert_object_to_comment_data` public and unit test it separately.  I also added my own filter as a kind of quasi dependency injection.

Do you have any thoughts about dependency injection?  My bias is towards very fine-grained unit tests with mock objects isolating them from the rest of the system, and I often test the implementation rather than limit myself to the API.  These are controversial positions at work 😉 so I'll follow your lead here.

If you give me some feedback, I'll fill in a bunch more tests.

mexon:
Also, I can't figure out how to solve the backwards-compatibility issues around the `setUp` function.  Any suggestions?

mediaformat (commented):
Hi, Thanks for this PR! 

While there is quite a bit of overlap, I like how you have `convert_object_to_comment_data` I'll merge as much of this into #142 where it makes sense.

mexon:
Alright, after a million years I finally found time to look at this again.  I solved the `setUp` problem (in WP you have to use `set_up`, and I'm happy with the test coverage now.  @pfefferle could you have another look?

PR #245, opened 2023-01-06
Settings page cleanups

toolstack:
- Add in some details to what each of the predefined post content actaully are in a hidden area, including a preview/sample of them.
- Remove the dashes in the "Post Content" and "Activity Object Type" to make them consistent with the other titles.
- Standardize the case of the "Supported Post Types" title to make it consistent with the other titles.

### DIFF
--- a/assets/css/activitypub-admin.css
+++ b/assets/css/activitypub-admin.css
@@ -55,6 +55,26 @@ summary {
 	color: #2271b1;
 }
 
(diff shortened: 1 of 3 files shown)


mattwiebe:
Sorry for getting to this so late and thanks for your PR @toolstack ! We've made some big changes in the admin page(s) and we hope you'll find them to your liking. `1.0.0` should be released very soon. :)

PR #246, opened 2023-01-06
Add new template patterns

### LINKED ISSUE #243: Template Pattern: Featured Image
Apologies if I missed it! I tried searching and didn't see it, so I might have searched for the wrong terms. But it'd be nice if the template patterns also included featured image somehow?

(Thanks! 💕)

toolstack:
This PR adds the following new template patterns:

- %hashcats% - The post's categories as hashtags
- %image% - The URL for the post's featured image, full size
- %thumbnail% - The URL for the post's featured image thumbnail size
- %author% - The author's name
- %authorurl% - The URL to the author's profile page
- %date% - The post's date
- %time% - The post's time
- %datetime% - The post's date/time formated as "date @ time"
- %blogurl% - The URL to the site
- %blogname% - The name of the site
- %blogdesc% - The description of the site

In additon it cleans up a few gramatical and param definitions associated with the origial template patterns.

Resolves #243

### DIFF
--- a/includes/model/class-post.php
+++ b/includes/model/class-post.php
@@ -228,6 +228,18 @@ class Post {
 		$content = \str_replace( '%permalink%', $this->get_the_post_link( 'permalink' ), $content );
 		$content = \str_replace( '%shortlink%', $this->get_the_post_link( 'shortlink' ), $content );
 		$content = \str_replace( '%hashtags%', $this->get_the_post_hashtags(), $content );
+		$content = \str_replace( '%thumbnail%', $this->get_the_post_image( 'thumbnail' ), $content );
+		$content = \str_replace( '%image%', $this->get_the_post_image(), $content );
+		$content = \str_replace( '%hashcats%', $this->get_the_post_categories(), $content );
+		$content = \str_replace( '%author%', $this->get_the_post_author(), $content );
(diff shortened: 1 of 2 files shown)


pfefferle:
Thanks a lot for your work, but I would like to replace the current pattern with shortcodes, to be more flexible and to be able to add some basic settings (like word-count/character-count).

pfefferle:
> The other issue is that the user would then be able to use any shortcode and that might cause unexpected problems.

But isn't that a good thing? It is an admin backend, so the possibility that something get's in the field that is not expected is very low and maybe there are nice shortcodes that might help/fit... maybe in the podcasting area or so...???

I liked the idea, because it is very similar and I do not have to implement the complete parsing thing. I haven't had a look at the shortcodes code yet, maybe there is a way to reuse the code, maybe with a different structure.

toolstack:
It's not neccessarily bad or good, just a conern.  There are lots of shortcodes kicking around that do partial html and other things that make a lot of sense in the content, but might not in an activitypub item.  Just something to keep in mind for this kind of thing.

If the user really wants to included shortcodes in their activitypub data, they can always just use the_content and that will render any shortcodes they want.

Obviously if the templates get more complex, then yes, using an existing parser makes a lot of sense.

PR #247, opened 2023-01-07
Make the excerpt code actually crop the excerpt at 400 characters.

toolstack:
The existing implementation crops at words and may return very short strings based upon filters, or very long strings based upon user inputted excerpts.

For example:
- Any user created excerpt isn't cropped at any length as it's just passed straight back.
- The excerpt length is filtered with `excerpt_length`, but that returns a number of words, not characters.
- User excerpts are not converted to plain text before being passed into the final filter which could cause unexpected outcomes.

This PR does the following:
- Treats manual and content based excerpts similarly, stripping tags, and making sure they are the correct length.
- Returns an excerpt based on character count, not word count.
- Does *not* use the `excerpt_length` filter.
- Tries to ensure that the excerpt is never more than the target length including all markup done by `the_excerpt` filter.  There could be a corner case where more than the target length is returned if `the_excerpt` filter adds more html than the allowed target size.

### DIFF
--- a/includes/model/class-post.php
+++ b/includes/model/class-post.php
@@ -309,24 +309,24 @@ class Post {
 		$current_excerpt_length = strlen( $excerpt );
 
 		// Setup a variable to keep track of our target length.
-		$target_excerpt_length = $current_excerpt_length - $excerpt_more_len;
+		$target_excerpt_length = $excerpt_length - $excerpt_more_len;
 
 		// Setup a variable to keep track of the current max length.
-		$current_expcerpt_max = $target_excerpt_length;
+		$current_excerpt_max = $target_excerpt_length;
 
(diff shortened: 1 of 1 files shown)


pfefferle:
There seem to be a calculation error. In my test, I get more than 4000 characters as response. You can check the output by adding `/activitypub` to a post URL.

See: https://notiz.blog/2022/11/14/the-at-protocol/activitypub/

toolstack:
I'll take a look and do some more testing.

pfefferle:
Thanks for the PR btw. ☺️

toolstack:
Hmmm... when I add ```/activitypub``` to a post url I just get a 404 error... anything I have to enable?

toolstack:
Ok, found the issue with the incorrct length, used a wrong variable to start the except length with.

pfefferle (approved):
LGTM;

PR #248, opened 2023-01-07
configuration item for number of images to attach

mexon:
Currently the plugin sends up to three images when publishing via ActivityPub.  This is probably more than Mastodon users, especially, are used to.  This is particularly true if you're only publishing an excerpt of the full article.  Alternatively, for very image-heavy posts, it might not be enough.  It should be configurable.

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -19,7 +19,7 @@ namespace Activitypub;
  * Initialize plugin
  */
 function init() {
-	\defined( 'ACTIVITYPUB_NUMBER_IMAGES' ) || \define( 'ACTIVITYPUB_NUMBER_IMAGES', 3 );
+	\defined( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS' ) || \define( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS', 3 );
 	\defined( 'ACTIVITYPUB_HASHTAGS_REGEXP' ) || \define( 'ACTIVITYPUB_HASHTAGS_REGEXP', '(?:(?<=\s)|(?<=<p>)|(?<=<br>)|^)#([A-Za-z0-9_]+)(?:(?=\s|[[:punct:]]|$))' );
 	\defined( 'ACTIVITYPUB_ALLOWED_HTML' ) || \define( 'ACTIVITYPUB_ALLOWED_HTML', '<strong><a><p><ul><ol><li><code><blockquote><pre><img>' );
 	\defined( 'ACTIVITYPUB_CUSTOM_POST_CONTENT' ) || \define( 'ACTIVITYPUB_CUSTOM_POST_CONTENT', "<p><strong>%title%</strong></p>\n\n%content%\n\n<p>%hashtags%</p>\n\n<p>%shortlink%</p>" );
--- a/activitypub.php
+++ b/activitypub.php
@@ -19,6 +19,7 @@ namespace Activitypub;
  * Initialize plugin
  */
 function init() {
+	\defined( 'ACTIVITYPUB_NUMBER_IMAGES' ) || \define( 'ACTIVITYPUB_NUMBER_IMAGES', 3 );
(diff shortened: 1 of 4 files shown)


includes/model/class-post.php
```diff
@@ -94,20 +94,22 @@ public function generate_attachments() {
...
-				'post_mime_type' => 'image',
-				'order' => 'ASC',
-				'orderby' => 'menu_order ID',
-				'posts_per_page' => $max_images,
-			)
-		);
-		foreach ( $query->get_posts() as $attachment ) {
-			if ( ! \in_array( $attachment->ID, $image_ids, true ) ) {
-				$image_ids[] = $attachment->ID;
+		if ( $max_images > 0 ) {
```
pfefferle, line 97: Max images will currently always be bigger than 0, see line 84:

```PHP
		// max images can't be negative or zero
		if ( $max_images <= 0 ) {
			$max_images = 1;
		}
```

pfefferle: I would also prefer to fast fail like:

```PHP
		// max images can't be negative or zero
		if ( $max_images <= 0 ) {
			return $images;
		}

File: activitypub.php [hunk 2]
```diff
@@ -19,6 +19,7 @@
  * Initialize plugin
  */
 function init() {
+	\defined( 'ACTIVITYPUB_NUMBER_IMAGES' ) || \define( 'ACTIVITYPUB_NUMBER_IMAGES', 3 );
```
pfefferle, line 22: what about `ACTIVITYPUB_NUMBER_POST_IMAGES` or something similar (I am no native speaker) to be more specific about the usecase?

includes/model/class-post.php
mexon: Except that `$max_images` now _can_ be zero after it's decremented on line 94 (if the user configures 1 image).  It can't return `$images` unless it gets to the metadata-building part in the loop on line 119.

However, I just realised this code is doing the wrong thing.  If the thumbnail is also one of the first attachments, which it normally would be, it will return one too few attachments.  I'll think about this some more.

activitypub.php
mexon: Yeah, that name is not clear.  I changed it to `ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS`.  I'm not sure whether attachments can also show up in comments, or if they can only be in posts?

pfefferle: shouldn't this simply work? https://github.com/pfefferle/wordpress-activitypub/pull/248/commits/6992fbbe226eb1753d34637cb50c10b9009d778c

includes/model/class-post.php
pfefferle: shouldn't this simply work: https://github.com/pfefferle/wordpress-activitypub/pull/248/commits/6992fbbe226eb1753d34637cb50c10b9009d778c ???

pfefferle: If `$max_images` is smaller than 1, it returns an empty array, otherwise the code collects the number of images needed.

pfefferle: I think this shows it a bit better: https://github.com/pfefferle/wordpress-activitypub/pull/248/files#diff-44de75651674f44bc586c0f66e8dddf70ec9c6a4a9eb72418be87aa6c12a6be7

pfefferle: you were right, I think we need both... the `if` check you added and the early return... what do you think @mexon ??

pfefferle (approved):
LGTM

pfefferle:
Thanks @mexon !!!

includes/model/class-post.php
mexon: Hey, sorry I missed this conversation, but FWIW I just tested it and it behaves correctly.  The interesting case is if it delivers the right number both when the featured image is in the first N and when it is not in the first N attachments.  And that's what it does.  Honestly, I'm not sure how that works, but it does 😳

Thanks so much for carrying the patch forward!

PR #250, opened 2023-01-13
Switch to shortcodes

toolstack:
As discussed in #246, this PR converts the existing template codes to shortcodes.  As well as add new shortcodes for:

- `[ap_hashcats]` - The post's categories as hashtags
- `[ap_image]` - The URL for the post's featured image, full size [ap_thumbnail] - The URL for the post's featured image thumbnail size 
- `[ap_author]` - The author's name
- `[ap_authorurl]` - The URL to the author's profile page [ap_date] - The post's date
- `[ap_time]` - The post's time
- `[ap_datetime]` - The post's date/time formated as "date @ time" [ap_blogurl] - The URL to the site
- `[ap_blogname]` - The name of the site
- `[ap_blogdesc]` - The description of the site

Finally, it adds in a `length` parameter to the `[ap_excerpt]` shortcode.

A few items of note/discussion:

- I chose the `ap_` prefix for the shortcodes as a compromise between length and uniqueness.  Two character prefixes aren't ideal, you can run into conflicts with other plugins, but likewise, the shortcodes should be easily typeable by the users.
- The shortcodes are only active during the post processing, so they are not available during the loop or in the WordPress editor.
- The current code 'updates' the custom template any time it is actually used, otherwise it remains as it was set by the user.  This adds a bit of extra processing per item (assuming the custom template is in use), but seems like a reasonable compromise to make sure things don't break.
- The new `length` parameter currently doesn't really work, it needs #247 to be in place first, as the exisiting code will apply the `excerpt_length` filter which is usually set by the theme.

Closes #246


pfefferle:
The shortcode does not seem to have the post context, if doing it that way. You can test it, by adding `activitypub` to the end of a post permalink. The shortcode will not be replaced.

I hacked together a simple (quick and dirty) version of how I thought it might work: https://github.com/pfefferle/wordpress-activitypub/pull/255

toolstack:
Seems to have been two problems:

- For some reason the constructor for the class was being called *after* the content was being genreated, so the shortcodes weren't registered yet.  Easy fix, just moved the `add_shortcode()` calls to the content function to make sure they exist.
- The second was a brain fart by me, I was echoing out the shortcodes instead of returning them as a string.

Both issues fixed in latest commit.

I did consider creating a single funciton per shortcode, but that seemed like overkill for such simple shortcodes, as does creating a whole new class for them as they are only ever used in the post.

toolstack:
Oh, and there is something broken on the setup for adding `activitypub` to the end of a post permalink, as it doesn't work unless you update your permalink settings (even just saving them as they are).

toolstack:
@pfefferle I've integrated the ideas from #255 into this PR, creating a seperate class for the shortcodes.

I added the class to the root plugin file, but didn't call an init on it, instead leaving it as part of the post class code.

I also added a size attribute to the image shortcode.

activitypub.php
```diff
@@ -19,14 +19,16 @@
  * Initialize plugin
  */
 function init() {
+	\defined( 'ACTIVITYPUB_EXCERPT_LENGTH' ) || \define( 'ACTIVITYPUB_EXCERPT_LENGTH', 400 );
 	\defined( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS' ) || \define( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS', 3 );
 	\defined( 'ACTIVITYPUB_HASHTAGS_REGEXP' ) || \define( 'ACTIVITYPUB_HASHTAGS_REGEXP', '(?:(?<=\s)|(?<=<p>)|(?<=<br>)|^)#([A-Za-z0-9_]+)(?:(?=\s|[[:punct:]]|$))' );
 	\defined( 'ACTIVITYPUB_ALLOWED_HTML' ) || \define( 'ACTIVITYPUB_ALLOWED_HTML', '<strong><a><p><ul><ol><li><code><blockquote><pre><img>' );
-	\defined( 'ACTIVITYPUB_CUSTOM_POST_CONTENT' ) || \define( 'ACTIVITYPUB_CUSTOM_POST_CONTENT', "<p><strong>%title%</strong></p>\n\n%content%\n\n<p>%hashtags%</p>\n\n<p>%shortlink%</p>" );
+	\defined( 'ACTIVITYPUB_CUSTOM_POST_CONTENT' ) || \define( 'ACTIVITYPUB_CUSTOM_POST_CONTENT', "<p><strong>[ap_title]</strong></p>\n\n[ap_content]\n\n<p>[ap_hashtags]</p>\n\n<p>[ap_shortlink]</p>" );
```
pfefferle, line 26: `%shortlink%` and `permalink` was HTML code, so maybe adding an attribute like `[ap_shortlink type=raw|html]`.

includes/model/class-post.php
```diff
@@ -226,18 +229,20 @@ public function generate_object_type() {
...
-		$content = \str_replace( '%title%', \get_the_title( $post->ID ), $content );
-		$content = \str_replace( '%excerpt%', $this->get_the_post_excerpt(), $content );
-		$content = \str_replace( '%content%', $this->get_the_post_content(), $content );
-		$content = \str_replace( '%permalink%', $this->get_the_post_link( 'permalink' ), $content );
-		$content = \str_replace( '%shortlink%', $this->get_the_post_link( 'shortlink' ), $content );
-		$content = \str_replace( '%hashtags%', $this->get_the_post_hashtags(), $content );
-		// backwards compatibility
-		$content = \str_replace( '%tags%', $this->get_the_post_hashtags(), $content );
+		// Register the shortcodes.
+		$shortcodes = new \Activitypub\Shortcodes( $post );
```
pfefferle, line 242: I prefer using:

```
setup_postdata( $post );
$content = do_shortcode( $content );
wp_reset_postdata();
```

It makes the handling of the shortcodes easier and more generic (and more compatible with the loop).

includes/class-shortcodes.php
```diff
@@ -0,0 +1,486 @@
...
+	 * @param string $tag 		shortcode tag name
+	 *
+	 * @return string
+	 */
+	public function excerpt( $atts, $content, $tag ) {
+		if( $this->post === false ) {
+			return '';
+		}
+
+		$length = ACTIVITYPUB_EXCERPT_LENGTH;
```
pfefferle, line 95: use `shortcode_atts` to set default params: https://developer.wordpress.org/reference/functions/shortcode_atts/

includes/class-shortcodes.php
```diff
@@ -0,0 +1,486 @@
...
+	 * @return string
+	 */
+	public function image( $atts, $content, $tag ) {
+		if( $this->post === false ) {
+			return '';
+		}
+
+		$size = 'full';
+
+		if( is_array( $atts ) && array_key_exists( 'size', $atts ) ) {
```
pfefferle, line 272: use `shortcode_atts` to set default params: https://developer.wordpress.org/reference/functions/shortcode_atts/

includes/class-shortcodes.php
```diff
@@ -0,0 +1,486 @@
...
+			$post = \get_post();
+		}
+
+		if( \is_object( $post ) && ! $post instanceof WP_POST ) {
+			$this->post = $post;
+		} else {
+			$this->post = false;
+		}
+
+		foreach( get_class_methods( $this ) as $shortcode ) {
```
pfefferle, line 28: I like the idea, but I would still stick to a static class with an init function, because (if we remove the `post` var, see other comment) there is no need for an object context.

pfefferle (commented):
Thanks for all your work so far!

I have some feedback inline.

activitypub.php
toolstack: Done, added options for:

- raw - just the raw url
- esc - url with esc_html()
- html - the traditional a tag
- blank - the a tag with a target of blank added

On a side note, should the esc_html() be replaced with esc_attr() in these functions?

includes/model/class-post.php
toolstack: I'm not a fan of this technique, but I've coverted the code over to.  Adding the shortcodes even when not used isn't that big of a hit for overhead 🤷

includes/class-shortcodes.php
toolstack: Done.

includes/class-shortcodes.php
toolstack: Done.

includes/class-shortcodes.php
toolstack: I modified the code to just ignore the `init` function instead of looking for the double underscore, so the fundamental technique still works.

adegans:
Aren't shortcodes sort of reserved for use in widgets, post and page content?

pfefferle:
@adegans but why not re-use them here? otherwise we had to implement everything shortcodes already do.

adegans:
You can indeed 're-use' them. But think of the dumb users. While it may look familiar it may also prompt them to use [permalink] or whatever in the wrong places.

If you have a distinctly different tag like %permalink% which can work just as easy with a str_replace() or preg_replace() people won't try funny stuff.

As an example (or comparison), I've been using tags like that in my AdRotate plugin for over 10 years and it never had any confusion as it's a distinctly different tag from what people use in actual content fields.

Just something to consider.

templates/settings.php
```diff
@@ -81,8 +69,7 @@
...
 										<li><code>[ap_hashtags]</code> - <?php \esc_html_e( 'The post\'s tags as hashtags.', 'activitypub' ); ?></li>
 										<li><code>[ap_hashcats]</code> - <?php \esc_html_e( 'The post\'s categories as hashtags.', 'activitypub' ); ?></li>
-										<li><code>[ap_image <i>size=xxx</i>]</code> - <?php \esc_html_e( 'The URL for the post\'s featured image, defaults to full size. The size attribute can be any of the following:', 'activitypub' ); echo esc_html( $registered_sizes ); ?></li>
-										<li><code>[ap_thumbnail]</code> - <?php echo \wp_kses( sprintf( __( 'The URL for the post\'s featured image thumbnail size (%s).', 'activitypub'), $thumbnail_size ), 'default' ); ?></li>
+										<li><code>[ap_image <i>type=full</i>]</code> - <?php \esc_html_e( 'The URL for the post\'s featured image, defaults to full size. The type attribute can be any of the following: thumbnail, medium, large, full', 'activitypub' ); ?></li>
```
toolstack, line 72: I'd suggest keeping the custom list, many sites add addtional sizes or may even remove standard ones.

pfefferle:
@adegans the problem is, that there are a lot requests from users that want to customize the output more than just some placeholders, like the size of an image, the length of an excerpt, ... and to implement that, I have to implement everything a shortcode does.

pfefferle (approved):
LGTM

PR #251, opened 2023-01-14
Give the notice boxes some margin so they have some space.

toolstack:
Currently, if you save settings, the two notice boxes sit righ on top of each other like so:
Screenshot from 2023-01-13 20-15-59

Which is a little ugly, so give them some space by adding a top margin:
Screenshot from 2023-01-13 20-15-39


pfefferle (approved):
Nice, thanks!

PR #256, opened 2023-01-16
Move the activitypub endpoint rule to the main rewrite addition function

toolstack:
This is for two reasons:
- No need to add the endpoint every time the plugin loads.
- The old code didn't flush the rewrite rules, making the endpoint non-functional until something did (like the user saving the permalink settings)


pfefferle:
Good point! Thanks for investigating!

pfefferle (approved):
LGTM!

PR #260, opened 2023-01-23
Update followers list to be present more data

toolstack:
This PR does several things, but all aimed at making the followers list more useful.  At the moment, it is very bare bones, showing only the folower id.

Included in the PR are:

- Use nodeinfo and the mastodon API (new classes for each) to get more infomration about the follower.
- Use three new database tables to store the information in.
- Update the followers class to use the database tables.
- Update the followers list to include additonal columns:  Avatar, Name, Description, From Service, Since, Is Bot.
- Create an upgrade class to create/manage the db tables going forward.
- Migraiton code to move the followers list from the user options to the database.
- Renamed the "Followers (Fediverse)" menu item to "ActivityPub" for two reasons; the first to be more consistent with the rest of the plugin, the second to expand its purpose.
- Moved the plugin info from the user profile page (where it gets lost in the giant list of info there) to the new ActivityPub menu item.
- By default, avatar usage is disabled (load/security concerns), but there is an option to enable them.
- Enabled pagnation of the followers list (TODO: add search and I think sorting is broken at the moment).

Things that probably need some thought include how to update the data about users/services over the long term (currenly only done when a new follower joins from that service/user).

Screenshot from 2023-01-23 14-59-48


pfefferle:
Close in favor of #304

PR #261, opened 2023-01-27
Hashtags: Protect HTML tags from being broken

akirk:
If your post has HTML that contains something that resembles a hashtag, the ActivityPub plugin currently breaks that tag by replacing new HTML tags into the other tag.

Specifically, this HTML:
```
<span style="color: #ccc">Only friends</span>
```

will be transformed to this:
```
<span style="color: <a rel="tag" class="u-tag u-category" href="https://alex.kirk.at/tag/ccc/">#ccc</a>;&#8221;>Only friends</span>
```

which visually appears like this:

Screenshot 2023-01-27 at 12 07 34

This PR adds protection of the tag before the replacing operation happens and then replaces them back in, resulting in unmodified tags inside HTML:

Screenshot 2023-01-27 at 12 09 23

### DIFF
--- a/includes/class-hashtag.php
+++ b/includes/class-hashtag.php
@@ -43,7 +43,21 @@ class Hashtag {
 	 * @return string the filtered post-content
 	 */
 	public static function the_content( $the_content ) {
+		$protected_tags = array();
+		$the_content = preg_replace_callback(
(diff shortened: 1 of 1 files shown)


pfefferle (approved):
Awesome! Thanks a lot!

PR #262, opened 2023-01-27
Fix accessing post properties

akirk:
In dbaddd91890f4926ce3f8055e1c35947119dbd2f accessing the properties of the inner WP_Post was broken. I discovered this while rebasing #213 which has tests for this.


pfefferle (approved):
LGTM

PR #263, opened 2023-01-27
Protect HTML between certain tags

akirk:
We also need to prevent the code between HTML elements in some cases, for example inside `<code>` or `<pre>`. This adds some more unit tests as well.

```
<pre>
Please don't mention @username@example.org
  here.
</pre>
```

should not become
```
<pre>
Please don't mention <a rel="mention" class="u-url mention" href="https://example.org/users/username">@<span>username</span></a>
   here.
 </pre>
```

See https://wordpress.org/support/topic/block-shows-raw-html-on-frontend/


pfefferle (approved):
LGTM

PR #265, opened 2023-02-02
optimize publishing

pfefferle:


mediaformat:
I think tagged users should go into the `to` field.

Followers collection goes in the `cc` field. 

And depending where you put the special `public` uri:
- `to`: makes an status Public
- `cc`: makes it Unlisted

pfefferle:
@mediaformat my thoughts: Having the default things in `to` (public and collection) and the changing things in `cc` for an easier handling. Mastodon seems to see no difference if mentioned ids are in `cc` or `to`.

Btw. here the documentation: https://docs.joinmastodon.org/spec/activitypub/

akirk (commented):
Seems reasonable to me. I'd like to suggest to update the `__call()` function with something like:
```
list( $action, $var ) = explode( '_', strtolower( $method ), 2 );
```
instead of the `strncasecmp` calls, see https://3v4l.org/8NUrC about compatibility.

pfefferle:
@akirk can you give some context why to use:

```
list( $action, $var ) = explode( '_', strtolower( $method ), 2 );
```

instead of `strncasecmp`???

akirk:
Of course it's a matter of taste but already at the beginning of the function you'd set `$action` and `$var` and can compare it later more explicitly.

Issue #165, opened 2022-11-06
Request: Custom text for profile biography


jaschaurbach:
I was rewriting my biography on my blog to for both, the website as well as the Fediverse account description. While doing so I recognized that it is not that easy.

I would love to have a second field on my author page: Fediverse Biography. If set, use it for fediverse related activities, if empty, take default biography from author. 

What do you think?

danimrich:
I think it would be nice to have this feature. For a start, it would help alleviate confusion between the authors' Mastodon profiles and the profiles created by the plugin.

eficklin:
Looking for opportunities to contribute and happy to take a crack at this. PR forthcoming next week, hopefully.

pfefferle:
Keep!

pfefferle:
done, will be part of the next release.

PR #294, opened 2023-03-14
Add ActivityPub specific user description

eficklin:
Address issue #165, a feature request for a user bio specific to the fediverse, separate from the WP standard user description field.

The PR provides the following

- A const for the user meta key for storing/retrieving the AP bio
- Profile fields and save handling for the AP bio on the user edit screen; as implemented here the form field only displays when a user is editing their own profile; if the ability to see and/or edit another user's AP bio is desired, e.g. an Administrator editing a Contributor's AP bio, I can add that to this PR; it's two additional action hooks and some extra capability checks at most
- Template helper to display the AP specific bio first, if set, or defer to the default; an empty string is returned if both are unset
- Updates the author-json.php template to use the new template function to display the correct description in the summary field.
ap-user-desc

Verification:

- edit and save the Fediverse Biography field on a user edit screen to confirm form data handling is correct
- using the dev console or command line, request an author profile (e.g. http://activitypub.local/author/admin) with an `Accept: application/activity+json` header to verify the `summary` field in the returned JSON has the correct value: fediverse bio if set, standard author description if no fediverse bio, empty string if neither is set


pfefferle:
Thanks a lot @eficklin

PR #299, opened 2023-04-02
Signature verification

mediaformat:
> federated servers ... SHOULD NOT TRUST content received from a server other than the content's origin without some form of verification. [SocialCG](https://www.w3.org/wiki/SocialCG/ActivityPub/Authentication_Authorization)

This PR implements http signatures and digests verification thereby preventing impersonation and other MITM attacks.

Features:
- Verify signatures of all incoming POST requests (comments)
- Provides an Application actor (sometimes known as a Service actor) to sign all outgoing GET requests (instead of leaking the identity of users making the requests)
- Optionally verify incoming GET requests. This is also known as [Secure Mode](https://docs.joinmastodon.org/admin/config/#authorized_fetch). And is a useful feature in deterring harassment. (Blocked users are prevented from accessing profiles, and posts, and therefore cannot redistribute or interact with them). 

Tested:
- [x] Mastodon
- [x] Akkoma
- [x] GoToSocial
- [x] Friendica
- [x] Calckey
- [ ] <del>Lemmy</del> seems to not work at all. lemmy seem to have a allow list, that blocks any other instance.
- [x] Pixelfed


mediaformat:
I'll make a separate PR for LD Signatures

pfefferle:
Is the LD signature thing something we have to support?

mediaformat:
I'm not even sure. Maybe for deleted users but I'm not sure how it's supposed to be processed (are we meant to delete (or anonymize) their content. These questions are why I wanted to separate them from this PR!

mediaformat:
> the signature is not part of the spec

Not officially, no. But this seems to be because the Editors had pressure to finalize the spec... and the HTTP Signatures spec itself was a [working draft](https://tools.ietf.org/html/draft-cavage-http-signatures-08).

> B. Security Considerations
> Unfortunately at the time of standardization, there are no strongly agreed upon mechanisms for authentication. Some possible directions for authentication are laid out [in the Social Web Community Group Authentication and Authorization best practices report](https://www.w3.org/wiki/SocialCG/ActivityPub/Authentication_Authorization).
>
> B.2 [Verification](https://www.w3.org/TR/activitypub/#security-verification)
> ... federated servers also should not trust content received from a server other than the content's origin without some form of verification.

What do you think?

mediaformat:
> make the verification configurable

We could do this for the GET requests, similar to Mastodon's [Authorized Fetch](https://docs.joinmastodon.org/admin/config/#authorized_fetch)

> As a result, through the authentication mechanism and avoiding re-distribution mechanisms that do not have your server in the loop, it becomes possible to enforce who can and cannot retrieve even public content from your server, e.g. servers whose domains you have blocked.

includes/rest/class-inbox.php
```diff
@@ -241,6 +233,16 @@ public static function user_inbox_post_parameters() {
 			'sanitize_callback' => 'esc_url_raw',
 		);
 
+		$params['signature'] = array(
```
pfefferle, line 236: this will check if the JSON blob will have an attribute `signature` and should therefore always fail!?!

I would assume that this is the wrong place to verify HTTP headers.

pfefferle (changes requested):
Can you add some PHPDoc?

includes/class-signature.php
```diff
@@ -107,8 +110,169 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
...
 	}
 
-	public static function verify_signature( $headers, $signature ) {
+	public static function verify_http_signature( $request ) {
+		$headers = $request->get_headers();
+		$actor = isset( json_decode( $request->get_body() )->actor ) ? json_decode( $request->get_body() )->actor : '';
+		$headers['(request-target)'][0] = strtolower( $request->get_method() ) . ' /wp-json' . $request->get_route();
+
+		if ( ! $headers ) {
+			$headers = self::default_server_headers();
```
pfefferle, line 119: Why do you add defaults? Shouldn't the code break if there are no headers?

includes/class-signature.php
```diff
@@ -107,8 +110,169 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
 		}
 	}
 
-	public static function verify_signature( $headers, $signature ) {
+	public static function verify_http_signature( $request ) {
+		$headers = $request->get_headers();
+		$actor = isset( json_decode( $request->get_body() )->actor ) ? json_decode( $request->get_body() )->actor : '';
+		$headers['(request-target)'][0] = strtolower( $request->get_method() ) . ' /wp-json' . $request->get_route();
```
pfefferle, line 116: this should be after the `if ( ! $headers )` check!

pfefferle: You can rewrite the API endpoint, so it is not always `/wp-json` (see my blog for example).

includes/class-signature.php
```diff
@@ -107,8 +110,169 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
...
+	}
+
+	public static function get_signature_algorithm( $signature_block ) {
+		switch ( $signature_block['algorithm'] ) {
+			case 'rsa-sha-512':
+				return 'sha512';
+			default:
+				return 'sha256';
+		}
+		return false;
```
pfefferle, line 191: shouldn't the default always be `sha256`?

includes/class-signature.php
```diff
@@ -107,8 +110,169 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
...
+			if ( \strpos( $k, 'HTTP_' ) === 0 ) {
+				$field = \str_replace( '_', '-', \strtolower( \substr( $k, 5 ) ) );
+				$headers[ $field ] = $v;
+			}
+		}
+		return $headers;
+	}
+
+	public static function get_signature_algorithm( $signature_block ) {
+		switch ( $signature_block['algorithm'] ) {
```
pfefferle, line 185: we should check if there is an "algorithm" key.

includes/rest/class-inbox.php
mediaformat: I have updated the code locally, will push today

mediaformat:
I've run into the need to implement a "Service" actor to signing GET requests. For example when a post is queried in secure mode. 

The way I've implemented it so far is to use `0` on the function to get_*_key(), if 0, set/get_option.

It occurred to me that this might also be useful for issue #1

What do you think?

pfefferle:
The users start with id `1`? Then `0` for blog is totally fine, otherwise we may have to use `-1` or something.

includes/class-activitypub.php
```diff
@@ -66,23 +68,24 @@ public static function render_json_template( $template ) {
...
-
 		if (
+			\stristr( $accept_header, 'application/activity+json' ) ||
+			\stristr( $accept_header, 'application/ld+json' ) ||
 			\in_array( 'application/ld+json; profile="https://www.w3.org/ns/activitystreams"', $accept, true ) ||
 			\in_array( 'application/activity+json', $accept, true ) ||
 			\in_array( 'application/ld+json', $accept, true ) ||
 			\in_array( 'application/json', $accept, true )
 		) {
+			$secure_mode = \get_option( 'activitypub_use_secure_mode', '0' );
```
pfefferle, line 81: We have so much issues with caching and content negotiations, that I would not add signature verification on get requests! Besides of that, these requests are synchron and generate a lot of latency for the requester.

pfefferle (dismissed):

includes/class-http.php
```diff
@@ -60,9 +60,9 @@ public static function post( $url, $body, $user_id ) {
 	 *
 	 * @return array|WP_Error The GET Response or an WP_ERROR
 	 */
-	public static function get( $url, $user_id ) {
+	public static function get( $url ) {
 		$date = \gmdate( 'D, d M Y H:i:s T' );
-		$signature = Signature::generate_signature( $user_id, 'get', $url, $date );
+		$signature = Signature::generate_signature( -1, 'get', $url, $date );
```
pfefferle, line 65: I would not send every request signed from the blog-author! Is this even accepted by Mastodon and others?

pfefferle: It is a nice idea to maybe use it for getting remote data, but not for sending `create` activities I think.

includes/rest/class-server.php
```diff
@@ -0,0 +1,103 @@
...
+		\add_filter( 'rest_request_before_callbacks', array( self::class, 'authorize_activitypub_requests' ), 10, 3 );
+		\add_action( 'rest_api_init', array( self::class, 'register_routes' ) );
+	}
+
+	/**
+	 * Register routes
+	 */
+	public static function register_routes() {
+		\register_rest_route(
+			'activitypub/1.0',
```
pfefferle, line 29: this is the URL that will be shown on the other platforms if we implement the catch all account, so I would create a user friendly URL and not an API endpoint?

Should we mix this up with the signature use-case?

includes/class-http.php
mediaformat: This is just for get requests, mastodon does the same, providing a minimum user anonymity

mediaformat: I'm using `-1` for the application actor, leaving `0` for the blog-author

includes/class-activitypub.php
mediaformat: This is a user setting which could be appreciated by some privacy minded folks. It is off by default.

includes/rest/class-server.php
mediaformat: This is strictly for the Application acct, and is only used for Signature verification. I think I might need to give it a webfinger...

includes/class-activitypub.php
pfefferle: but it uses the same setting that enables/disables the incoming requests. I really would remove this for now and think about that for a furure release!

includes/rest/class-server.php
pfefferle: oh ok!

includes/class-http.php
pfefferle: if the number of non-user rise, should we use constants instead? to not get confused?

mediaformat: Good idea!

includes/class-activitypub.php
mediaformat: In both cases it only applies to get requests.

jeherve (changes requested):
Coming in late to this PR, looks like you've done a lot already @mediaformat 🙇‍♂️ 

Like @pfefferle, I think it would be important to add tests for all of this before we merge it:

https://github.com/Automattic/wordpress-activitypub/pull/299#issuecomment-1517372534

This would help ensuring we're not missing any potential regressions that would be added with this change.

Do you think you could also extend the Pull Request description? I'm especially curious to know how you went about testing this with the different platforms you mentioned.

Thank you!

includes/class-activitypub.php
pfefferle: But following the WordPress philosopgy: Decisions, not options, does it provide any real value, that we need to add it?

(I do not want to convince you both, I want to realy understand why to add it!)

includes/rest/class-server.php
```diff
@@ -0,0 +1,103 @@
...
+	 *
+	 * @param                  $response
+	 * @param                  $handler
+	 * @param \WP_REST_Request $request
+	 *
+	 * @return mixed|\WP_Error
+	 */
+	public static function authorize_activitypub_requests( $response, $handler, $request ) {
+		$route = $request->get_route();
+		if ( str_starts_with( $route, '/activitypub' ) ) {
```
pfefferle, line 83: @mediaformat what do you think about an `verify` filter, where we hook the signature mechansim in, but that could also be implemented by others? Like WordPress is doing with the login for example: `_wp_get_current_user` https://github.com/WordPress/WordPress/blob/f28299d50422286d8b2c021e5405a3b2fec44a22/wp-includes/user.php#L3581

pfefferle: That would allow us, to add other verification methods like a simple Webmention/Pingback check if signature is not available: https://www.w3.org/TR/webmention/#request-verification

mediaformat:
@jeherve i have a yunohost server where I am testing these against WP

includes/class-activitypub.php
mediaformat: There is a clear enough reason not to make it a default: Performance.

But there is also a clear enough reason to keep the option, which fits in with many of WP's own moderation options: Abuse prevention.

> As a result, through the authentication mechanism ... it becomes possible to enforce who can and cannot retrieve even public content from your server, e.g. servers whose domains you have blocked.

Typically WP blogs have been more concerned with SPAM, but [many](https://www.pewresearch.org/internet/2021/01/13/the-state-of-online-harassment/) [studies](https://www.sciencedirect.com/science/article/pii/S8755461521000037) have shown that  harassment is increasingly prevalent in social media. ActivityPub removes even more barriers to interaction with blog posts, so I believe a corresponding option to limit is a fair compromise.

includes/class-signature.php
```diff
@@ -107,12 +124,244 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
...
+	}
+
+	/**
+	 * Get public key from key_id
+	 *
+	 * @param string $key_id
+	 * @return string $publicKeyPem
+	 * @author Django Doucet <django.doucet@webdevstudios.com>
+	 */
+	public static function get_remote_key( $key_id ) { // phpcs:ignore
```
pfefferle, line 208: don't we have an actor URL here? can we always be sure, that the key_id is the actur URL?

pfefferle: I think we have the request here, why not use the Actor-URL instead?

includes/class-signature.php
```diff
@@ -107,12 +124,244 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
...
+		$actor = \Activitypub\get_remote_metadata_by_actor( strtok( strip_fragment_from_url( $key_id ), '?' ) ); // phpcs:ignore
```
pfefferle, line 209: the meta is now also stored int the followers collection. you do not have to get the remote data any more.

includes/rest/class-server.php
```diff
@@ -0,0 +1,103 @@
...
+	public static function application_actor() {
+		$json = new \stdClass();
+
+		$json->{'@context'} = \Activitypub\get_context();
+		$json->id = \get_rest_url( null, 'activitypub/1.0/application' );
+		$json->type = 'Application';
+		$json->preferredUsername = wp_parse_url( get_site_url(), PHP_URL_HOST ); // phpcs:ignore WordPress.NamingConventions
+		$json->name = get_bloginfo( 'name' );
+		$json->summary = 'WordPress-ActivityPub application actor';
+		$json->manuallyApprovesFollowers = true; // phpcs:ignore WordPress.NamingConventions
```
pfefferle, line 55: is this required by mastodon or activitypub? if not, we should remove it, because you should not be able to follow the application user at all.

includes/rest/class-server.php
```diff
@@ -0,0 +1,103 @@
...
+	 *
+	 * @return mixed|\WP_Error
+	 */
+	public static function authorize_activitypub_requests( $response, $handler, $request ) {
+		$route = $request->get_route();
+		if ( str_starts_with( $route, '/activitypub' ) ) {
+			if ( 'POST' === $request->get_method() ) {
+				$verified_request = Signature::verify_http_signature( $request );
+				if ( \is_wp_error( $verified_request ) ) {
+					return $verified_request;
```
pfefferle, line 87: shouldn't this return a valid WP_REST_Response? currently it seem to respond with the result of `openssl_verify` which might be a boolean.

pfefferle: I see, it only returns the result if it is a WP_Error!

includes/class-signature.php
```diff
@@ -14,16 +18,17 @@ class Signature {
 	 * @return mixed
 	 */
 	public static function get_public_key( $user_id, $force = false ) {
-		$key = \get_user_meta( $user_id, 'magic_sig_public_key' );
```
pfefferle, line 17: this change prevents that a key will be generated at all!

pfefferle: so the key will ever be empty.

pfefferle: fixed

pfefferle:
@mediaformat I made some small changes and fixed the unit-tests. Now we should add some tests to check if verification works.

includes/class-activitypub.php
pfefferle: OK, agreed. But what about hiding it from the user by using a CONST instead of a setting, to play around a bit for some reseases? Like a feature switch?

And if it works fine, to add a setting afterwards. We could add the documentation about using the CONST in the FAQ?

includes/class-signature.php
mediaformat: Sometimes we need to get a key from a non-follower, ex: someone commenting on our post.

pfefferle: ok, true!

includes/class-signature.php
mediaformat: I think the content-negotiated get requests didn't include an actor.

I haven't yet seen an implementation whose keyId wasn't added to the actor url:
- #main-key
- /main-key

pfefferle: Can you make post an example how the request looks??

mediaformat: Here are some example requests from Mastodon

```
WP_REST_Request Object
(
    [method:protected] => POST
    [params:protected] => Array
        (

            [JSON] => Array
                (
                    [@context] => Array
                        (
                            [0] => https://www.w3.org/ns/activitystreams
                            [1] => Array
                                (
                                    [ostatus] => http://ostatus.org#
                                    [atomUri] => ostatus:atomUri
                                    [inReplyToAtomUri] => ostatus:inReplyToAtomUri
                                    [conversation] => ostatus:conversation
                                    [sensitive] => as:sensitive
                                    [toot] => http://joinmastodon.org/ns#
                                    [votersCount] => toot:votersCount
                                )

                        )

                    [id] => https://masto.example.app/users/mastodon/statuses/110082446069839861/activity
                    [type] => Create
                    [actor] => https://masto.example.app/users/mastodon
                    [published] => 2023-03-25T06:13:44Z
                    [to] => Array
                        (
                        )

                    [cc] => Array
                        (
                            [0] => https://masto.example.app/users/mastodon/followers
                            [1] => https://wp-example.ngrok.app/author/wordpress/
                        )

                    [object] => Array
                        (
                            [id] => https://masto.example.app/users/mastodon/statuses/110082446069839861
                            [type] => Note
                            [summary] => 
                            [inReplyTo] => https://wp-example.ngrok.app/sad-eros/
                            [published] => 2023-03-25T06:13:44Z
                            [url] => https://masto.example.app/@mastodon/110082446069839861
                            [attributedTo] => https://masto.example.app/users/mastodon
                            [to] => Array
                                (
                                )

                            [cc] => Array
                                (
                                    [0] => https://masto.example.app/users/mastodon/followers
                                    [1] => https://wp-example.ngrok.app/author/wordpress/
                                )

                            [sensitive] => 
                            [atomUri] => https://masto.example.app/users/mastodon/statuses/110082446069839861
                            [inReplyToAtomUri] => https://wp-example.ngrok.app/sad-eros/
                            [conversation] => tag:masto.example.app,2023-03-18:objectId=35:objectType=Conversation
                            [content] => <p><span class="h-card"><a href="https://wp-example.ngrok.app/author/wordpress/" class="u-url mention">@<span>wordpress</span></a></span> $request</p>
                            [contentMap] => Array
                                (
                                    [en] => <p><span class="h-card"><a href="https://wp-example.ngrok.app/author/wordpress/" class="u-url mention">@<span>wordpress</span></a></span> $request</p>
                                )

                            [attachment] => Array
                                (
                                )

                            [tag] => Array
                                (
                                    [0] => Array
                                        (
                                            [type] => Mention
                                            [href] => https://wp-example.ngrok.app/author/wordpress/
                                            [name] => @wordpress@wp-example.ngrok.app
                                        )

                                )

                            [replies] => Array
                                (
                                    [id] => https://masto.example.app/users/mastodon/statuses/110082446069839861/replies
                                    [type] => Collection
                                    [first] => Array
                                        (
                                            [type] => CollectionPage
                                            [next] => https://masto.example.app/users/mastodon/statuses/110082446069839861/replies?only_other_accounts=true&page=true
                                            [partOf] => https://masto.example.app/users/mastodon/statuses/110082446069839861/replies
                                            [items] => Array
                                                (
                                                )

                                        )

                                )

                        )

                    [signature] => Array
                        (
                            [type] => RsaSignature2017
                            [creator] => https://masto.example.app/users/mastodon#main-key
                            [created] => 2023-03-25T06:13:44Z
                            [signatureValue] => Ic9JGB+Q3DkYnx/LlGmOkW6eYJbfhqaQxaDEeL74urpOpKYwjFOfPSyHZkIiGW5kzqaAqC8OSp3Yzds441yoQZSLruWEpjL764o9J9FKkYM4cZr7lBasq8tdRtYLag0hEJtgt9yvXS7JiM3Tk4F/5XVrQ1B9D4OtQQwBtILB1zUifzocqi+x31VeV45fwkRI61yjFhhOE7NTHB0P9hfnrArP9OcCKYqISqOgH2qApGXVeZgs4hW5LMxjJyXG+KLBmyNFiaLBCyyAgiwKxgDzeOGMwbisEnSxCZwyLTuEeqAp7HUbj3FVQZUlbLVN/B678Q2jVakyN+mAe6sUHsKV1w==
                        )

                )

            [defaults] => Array
                (
                )

        )

    [headers:protected] => Array
        (
            [x_original_host] => Array
                (
                    [0] => wp-example.ngrok.app
                )

            [signature] => Array
                (
                    [0] => keyId="https://masto.example.app/users/mastodon#main-key",algorithm="rsa-sha256",headers="(request-target) host date digest content-type",signature="Cva/xbz0ntRh6qsrFANPkiXGMTdfWy03+uK7Hqb7QQW/QpgUZkK1A6iQLtcpiI1rTIStyhFc96ySdwtzVHjLRr40dRqQAaqyxJ8BkXh7ScNwS2lVv4j6LXQJSNUNkY/Vcs+na+4++Lc80Pl82AFQPODAxrOuwmbdmOkqCFNu/lQ2I1JvG3WgsTgQrntlruS7oYZQ/9UyUnG3nWi3mYrqnKtgmPc5MFLkDFSsreHFD9KCW932fvn5VCpmaFxKS+PVmXbRVgGBh7UZHS+hTRgxrmwAL8VXc5V0QqHEm5zb5JpjvcLJ+OjZ8l0T1J73uHvis4sZY9sExYPSHDSRKc27uQ=="
                )

            [digest] => Array
                (
                    [0] => SHA-256=hdi8H0pMg8q8KebwR8/HIEZFRBW5V6tNprKvxAFm8xE=
                )

            [date] => Array
                (
                    [0] => Sat, 25 Mar 2023 06:13:44 GMT
                )

            [content_type] => Array
                (
                    [0] => application/activity+json
                )

            [accept_encoding] => Array
                (
                    [0] => gzip
                )

            [user_agent] => Array
                (
                    [0] => http.rb/5.1.0 (Mastodon/4.0.2; +https://masto.example.app/)
                )

            [content_length] => Array
                (
                    [0] => 3035
                )

            [host] => Array
                (
                    [0] => wp-example.ngrok.app
                )

        )

    [body:protected] => {"@context":["https://www.w3.org/ns/activitystreams",{"ostatus":"http://ostatus.org#","atomUri":"ostatus:atomUri","inReplyToAtomUri":"ostatus:inReplyToAtomUri","conversation":"ostatus:conversation","sensitive":"as:sensitive","toot":"http://joinmastodon.org/ns#","votersCount":"toot:votersCount"}],"id":"https://masto.example.app/users/mastodon/statuses/110082446069839861/activity","type":"Create","actor":"https://masto.example.app/users/mastodon","published":"2023-03-25T06:13:44Z","to":["https://www.w3.org/ns/activitystreams#Public"],"cc":["https://masto.example.app/users/mastodon/followers","https://wp-example.ngrok.app/author/wordpress/"],"object":{"id":"https://masto.example.app/users/mastodon/statuses/110082446069839861","type":"Note","summary":null,"inReplyTo":"https://wp-example.ngrok.app/sad-eros/","published":"2023-03-25T06:13:44Z","url":"https://masto.example.app/@mastodon/110082446069839861","attributedTo":"https://masto.example.app/users/mastodon","to":["https://www.w3.org/ns/activitystreams#Public"],"cc":["https://masto.example.app/users/mastodon/followers","https://wp-example.ngrok.app/author/wordpress/"],"sensitive":false,"atomUri":"https://masto.example.app/users/mastodon/statuses/110082446069839861","inReplyToAtomUri":"https://wp-example.ngrok.app/sad-eros/","conversation":"tag:masto.example.app,2023-03-18:objectId=35:objectType=Conversation","content":"\u003cp\u003e\u003cspan class=\"h-card\"\u003e\u003ca href=\"https://wp-example.ngrok.app/author/wordpress/\" class=\"u-url mention\"\u003e@\u003cspan\u003ewordpress\u003c/span\u003e\u003c/a\u003e\u003c/span\u003e $request\u003c/p\u003e","contentMap":{"en":"\u003cp\u003e\u003cspan class=\"h-card\"\u003e\u003ca href=\"https://wp-example.ngrok.app/author/wordpress/\" class=\"u-url mention\"\u003e@\u003cspan\u003ewordpress\u003c/span\u003e\u003c/a\u003e\u003c/span\u003e $request\u003c/p\u003e"},"attachment":[],"tag":[{"type":"Mention","href":"https://wp-example.ngrok.app/author/wordpress/","name":"@wordpress@wp-example.ngrok.app"}],"replies":{"id":"https://masto.example.app/users/mastodon/statuses/110082446069839861/replies","type":"Collection","first":{"type":"CollectionPage","next":"https://masto.example.app/users/mastodon/statuses/110082446069839861/replies?only_other_accounts=true\u0026page=true","partOf":"https://masto.example.app/users/mastodon/statuses/110082446069839861/replies","items":[]}}},"signature":{"type":"RsaSignature2017","creator":"https://masto.example.app/users/mastodon#main-key","created":"2023-03-25T06:13:44Z","signatureValue":"Ic9JGB+Q3DkYnx/LlGmOkW6eYJbfhqaQxaDEeL74urpOpKYwjFOfPSyHZkIiGW5kzqaAqC8OSp3Yzds441yoQZSLruWEpjL764o9J9FKkYM4cZr7lBasq8tdRtYLag0hEJtgt9yvXS7JiM3Tk4F/5XVrQ1B9D4OtQQwBtILB1zUifzocqi+x31VeV45fwkRI61yjFhhOE7NTHB0P9hfnrArP9OcCKYqISqOgH2qApGXVeZgs4hW5LMxjJyXG+KLBmyNFiaLBCyyAgiwKxgDzeOGMwbisEnSxCZwyLTuEeqAp7HUbj3FVQZUlbLVN/B678Q2jVakyN+mAe6sUHsKV1w=="}}
    [route:protected] => /activitypub/1.0/users/2/inbox
)

```

GET request for a `/convenience/` page (filtered out the irrelevant headers)
```
$_SERVER array
(
    [SERVER_SOFTWARE] => nginx/1.16.0
    [REQUEST_URI] => /convenience/
    [HTTP_X_ORIGINAL_HOST] => wp-example.ngrok.app
    [HTTP_SIGNATURE] => keyId=\"https://masto.example.app/actor#main-key\",algorithm=\"rsa-sha256\",headers=\"(request-target) host date accept\",signature=\"i8vqEeFPld/evOgJeKVW8X0zbmeJ76Mcyg3EAknbNREX+CP6CffkZ+jyVzhOnXQfWVgis1r0mC6c6aShLoxAkKo/Np9SbzRZptlay85tbEJ4Tx4jvC/CVoFmJ35x6zngic8+pXTaM33OGQRk9VO778xdqXrV1FQ7qQmddPymsx0uT3nbVTwPCsqyXp0f9bxnR5J8dzwj6mNzWFGUelqlgpFHytSU10ebY9i46iJ9OOBHl7McYJknZrdB42eabPcohUfRjvZjKbMvgJX+Ze/LdLYiEd0el0+ijb5veYzLp6VbGql7a+9VT0NYRCXUzFJlvLeEVvQHHrmGhJ1zRB8piQ==\"
    [HTTP_DATE] => Tue, 18 Apr 2023 02:31:19 GMT
    [HTTP_ACCEPT_ENCODING] => gzip
    [HTTP_ACCEPT] => application/activity+json, application/ld+json; profile=\"https://www.w3.org/ns/activitystreams\", text/html;q=0.1
    [HTTP_USER_AGENT] => http.rb/5.1.0 (Mastodon/4.0.2; +https://masto.example.app/)
    [CONTENT_LENGTH] => 
    [CONTENT_TYPE] => 
    [REQUEST_METHOD] => GET
    [REQUEST_TIME_FLOAT] => 1681785079.3728
    [REQUEST_TIME] => 1681785079
)
```

pfefferle: Puh... I understand the idea behind that, but there are already three different versions of the `key` ID and (as far as I know) not every plattform does support it for `GET`.

I totally understand it for `POST` requests, and I am fine to sign also `GET` requests, but I am not really convinced by checking for signatures on receiving `GET` requests.

I fear that this might cause a lot of support questions!

mediaformat: If only some GET requests are verified, the setting would not provide the intended security.

On a practical level, despite not having the actor url for these Get requests, they come from instance actors, and would be cached in transients after the first request.

Also, I haven't yet seen an implementation that didn't sign requests. It would have difficulty federating with Mastodon.

pfefferle: But only for `POST` requests. The `GET` stuff is (as far as I can remember) an optional opt-in thing!?! At least I can query my Mastodon endpoints without signature headers.

And I am totally fine, sending the headers for all requests, I am only sceptical on checking signatures on incoming `GET` requests!

I am totally fine to add that feature with a feature flag, but at this state, I fear too much problems if it is a visible setting, everyone could opt-in.

pfefferle: I also checked a lot of other platforms (calckey, miskey, friendi.ca), that all have no signature verification on `GET` requests (checked to load the profile URL and the follower list).

pfefferle: I just checked the Mastodon Settings and couldn't find anything on `GET`-Request verification. Can you point me to any documentation about that?

I see that they add verifivation to all requests, but I could not find any documentation about verifying them.

pfefferle: Just checked the Code... Mastodon seem to always use the KeyID to get and verify the signature, so at least that should work... still sceptical about the `GET` verification thing...

mediaformat: > The GET stuff is (as far as I can remember) an optional opt-in thing

Exactly, this is the same model with ` \get_option( 'activitypub_use_secure_mode', '0' ) `

mediaformat: Sorry if this wasn't clear: Verifying GET requests would ONLY happen if a site has activated Secure Mode!

pfefferle: I understand that, but as I wrote earlier in the thread, I fear that a lot of people will activate it (because what could be wrong with being even more secure) and that we see a lot of issues on GitHub and in the WP.org forums, because AP is a complex topic and not every WP user understands the problems with shared hosting, content negotiation, signature verification and caching.

pfefferle: That's why I vote for a const instead of a setting, that only people who know what they are doing activate it in the first place. If this first try works good we could re-add the setting afterwards. I would see it as a soft launch. Mastodon also has no setting for that and gives it in the responsibility of the host of a mastodon instance.

pfefferle: And if not that way, than we need at least an opt-out... A const that disables the setting/feature when set.

includes/rest/class-server.php
```diff
@@ -0,0 +1,108 @@
...
+	 *
+	 * @return WP_REST_Response The JSON profile of the Application Actor.
+	 */
+	public static function application_actor() {
+		$json = new \stdClass();
+
+		$json->{'@context'} = \Activitypub\get_context();
+		$json->id = \get_rest_url( null, 'activitypub/1.0/application' );
+		$json->type = 'Application';
+		$json->preferredUsername = wp_parse_url( get_site_url(), PHP_URL_HOST ); // phpcs:ignore WordPress.NamingConventions
```
mediaformat, line 53: @pfefferle 

I haven't run into any issues with this application actor so far, it has worked fine without webfinger.

The preferredUsername for this actor would be `example.com@example.com` however it seems that this could conflict with brid.gy fed.

https://fed.brid.gy/docs#profile

pfefferle: Should only be an issue if you are also using brid.gy fed and I see no reason to use both plugins!?!

/cc @snarfed

mediaformat: This might be an edge case:
https://fed.brid.gy/r/https://gregsplace.net/2023/05/15/saying-hello/

pfefferle: What about using `example-com@example.com` or `example_com@example.com` instead?

includes/rest/class-server.php
```diff
@@ -0,0 +1,110 @@
...
+			return $response;
+		}
+
+		if ( 'POST' === $request->get_method() ) {
+			$verified_request = Signature::verify_http_signature( $request );
+			if ( \is_wp_error( $verified_request ) ) {
+				return $verified_request;
+			}
+		} else {
+			if ( get_rest_url_by_path( 'webfinger' ) !== $route ) {
```
pfefferle, line 97: it is not only WebFinger, but also NodeInfo. Should we have a allow-list?

pfefferle (changes requested):

includes/rest/class-server.php
mediaformat: I wonder if an site that chooses secure mode would even want to broadcast their stats.

pfefferle: What do you mean by that?

mediaformat: Sorry that was vague. 
My thinking is that nodeinfo is about exposing site metadata (number of users) to the public, while secure mode goes in the direction of privacy, requiring signed fetches even for public posts/author profiles.
Considering how those two features are opposed it might make sense to disable nodeinfo entirely when secure mode is activated.

includes/class-admin.php
```diff
@@ -144,15 +144,6 @@ public static function register_settings() {
 				'default'      => array( 'post', 'pages' ),
 			)
 		);
-		// \register_setting(
```
mediaformat: I was going to go with your suggestion, where a filter would be required to display the option

includes/rest/class-server.php
pfefferle: True, but I would not implicate it. If we want to disable it, we should it the same way as we've done it with NodeInfo and "hide from search engines".

pfefferle:
Looks good for me so far. I will test it on my live site for some time, to check if it really works as expected. thanks @mediaformat !

pfefferle:
@mediaformat Mastodon seem to work fine, but pixelfed does not come through.

mediaformat:
I will further test Lemmy & Pixelfed

pfefferle:
Lemmy does not seem to work at all:

* https://mastodon.social/@pfefferle/110412905895045931
* https://mastodon.social/@pfefferle/110413015534399373

pfefferle:
@mediaformat any findings so far? Otherwise I would try to dig into this issue today, because it has some nice changes, I would love to use in other PRs. ;)

pfefferle:
@mediaformat is it easier if the signature class would be an object instead of a static helper class?

pfefferle:
So then I will run it on my live server for some more days and then we can merge it 😍

pfefferle (approved):

pfefferle (dismissed):
No issues so far! I will merge it and if we find issues, we can improve on that.

pfefferle:
@mediaformat could you add some tests, then I would merge it?

mediaformat:
@pfefferle i have a test written but it won't seem to validate.
Maybe needs a fresh pair of eyes!

tests/test-class-activitypub-rest-post-signature-verification.php
```diff
@@ -0,0 +1,57 @@
...
+			'digest' => [ "SHA-256=$digest" ],
+			'signature' => [ $signature ],
+			'date' => [ $date ],
+			'host' => [ $host ],
+			'(request-target)' =>  [ 'POST ' . $route ]
+		);
+
+		// Start verification
+		// parse_signature_header, get_signed_data, get_public_key
+		$signature_block = Activitypub\Signature::parse_signature_header( $headers['signature'] );
```
pfefferle, line 44: how difficult is it to use `\Activitypub\Signature::verify_http_signature` here?

pfefferle:
Thanks @mediaformat :)

tests/test-class-activitypub-rest-post-signature-verification.php
mediaformat: I guess that is a fair test, I'll try to mock a `WP_REST_Request`

includes/class-signature.php
```diff
@@ -242,6 +242,11 @@ public static function verify_http_signature( $request ) {
 	 * @return string $publicKeyPem
 	 */
 	public static function get_remote_key( $key_id ) { // phpcs:ignore
+		$pre = apply_filters( 'pre_get_remote_key', false, $key_id );
```
mediaformat, line 245: Tried adding this filter to short-circuit the function during Tests, maybe someone else can get this to work

mediaformat (commented):
I won't have any time to dedicate for the coming weeks, hopefully someone can help fix the `test_rest_activity_signature` test case

pfefferle:
Awesome! Seems to work like a charm and tests are green!

Thanks @mediaformat !

pfefferle (approved):

PR #302, opened 2023-04-10
Webfinger info: avoid PHP warning when user isn't defined

jeherve:
This should avoid warnings like this one:

```
PHP Warning:  Attempt to read property "user_login" on bool in /var/www/html/wp-content/plugins/activitypub/includes/class-webfinger.php on line 27
```

### DIFF
--- a/includes/class-webfinger.php
+++ b/includes/class-webfinger.php
@@ -23,6 +23,9 @@ class Webfinger {
 		}
 
 		$user = \get_user_by( 'id', $user_id );
+		if ( ! $user ) {
+			return '';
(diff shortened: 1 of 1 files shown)


pfefferle:
Is it even possible to have no user? Shouldn't it fail way before?

jeherve:
I didn't get the full stacktrace so I cannot really say, but it did happen somehow (when logging in to write a new post).

pfefferle (approved):
LGTM

pfefferle:
Thanks @jeherve

PR #304, opened 2023-04-21
use a taxonomy to save the list of followers

pfefferle:


akirk:
@pfefferle there are conflicts. I think they are worth fixing before we review.

includes/class-migration.php
```diff
@@ -0,0 +1,72 @@
...
+	/**
+	 * The Migration for Plugin Version 1.0.0 and DB Version 1.0.0
+	 *
+	 * @since 5.0.0
+	 *
+	 * @return void
+	 */
+	public static function migrate_to_1_0_0() {
+		foreach ( get_users( array( 'fields' => 'ID' ) ) as $user_id ) {
+			$followes = get_user_meta( $user_id, 'activitypub_followers', true );
```
akirk, line 63:
```suggestion
			$followers = get_user_meta( $user_id, 'activitypub_followers', true );
```

includes/collection/class-followers.php
```diff
@@ -0,0 +1,383 @@
...
+	/**
+	 * Add a new Follower
+	 *
+	 * @param int    $user_id The ID of the WordPress User
+	 * @param string $actor   The Actor URL
+	 *
+	 * @return array|WP_Error The Follower (WP_Term array) or an WP_Error
+	 */
+	public static function add_follower( $user_id, $actor ) {
+		$meta = get_remote_metadata_by_actor( $actor );
```
akirk, line 198: It is important to note that because of this requirement, it could be that a follower might not be converted. For example if their Mastodon server is temporarily down.

I'd suggest to add a `$force` parameter and skip that check when converting, and then add a maintenance tool that can clean out stale followers.

akirk (commented):
I think it is a great effort but it's worth discussing how the migration should take place. In the current implementation you might lose followers.

includes/class-migration.php
```diff
@@ -0,0 +1,72 @@
...
+
+			if ( $followes ) {
+				foreach ( $followes as $follower ) {
+					Collection\Followers::add_follower( $user_id, $follower );
+				}
```
akirk, line 68:
```suggestion
			if ( $followers ) {
				foreach ( $followers as $follower ) {
					Collection\Followers::add_follower( $user_id, $follower );
				}
```

includes/collection/class-followers.php
pfefferle: I have to add a task that updates the user-profiles anyways! So I can combine both!

includes/class-migration.php
```diff
@@ -0,0 +1,72 @@
...
+	 * Updates the database structure if necessary.
+	 */
+	public static function maybe_migrate() {
+		if ( self::is_latest_version() ) {
+			return;
+		}
+
+		$version_from_db = self::get_version();
+
+		if ( version_compare( $version_from_db, '1.0.0', '<' ) ) {
```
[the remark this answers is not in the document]
pfefferle: you mean that I use the version of the previous release and if the migration is done I add this to the db?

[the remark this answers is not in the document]
akirk: Why not use the plugin version number? Why introduce a separate one?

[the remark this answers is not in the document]
pfefferle: so simple? :)

includes/model/class-follower.php
```diff
@@ -0,0 +1,258 @@
...
+		if ( \strncasecmp( $method, 'set', 3 ) === 0 ) {
+			$this->$var = $params[0];
+		}
+	}
+
+	public function from_meta( $meta ) {
+		$this->meta = $meta;
+
+		foreach ( $this->map_meta as $remote => $internal ) {
+			if ( ! empty( $meta[ $remote ] ) ) {
```
akirk, line 150: This (and all `$meta['...']` notations below) causes a fatal when `$meta` is a `WP_Error`.

includes/class-migration.php
```diff
@@ -0,0 +1,101 @@
...
+			self::migrate_from_0_17();
+			self::migrate_from_0_16();
+		}
```
akirk, line 43: Ideally, you can just extend a construct like this:
```suggestion
		if ( version_compare( $version_from_db, '0.16.0', '<' ) ) {
			self::migrate_from_0_16();
		}
		if ( version_compare( $version_from_db, '0.17.0', '<' ) ) {
			self::migrate_from_0_17();
		}
```

includes/collection/class-followers.php
pfefferle: I updated the `add_follower` function to be more error-tolerant. I just ignore HTTP responses with code 404 and 410 because they seem to be deleted.

Still missing: The update schedule, to finally remove profiles with errors (`errors >= 5` for example) and/or to updated profiles.

pfefferle:
@akirk can you check if this PR breaks something on the Friends plugin?

akirk (approved):
I've done a number of tests with the Friends plugin and this PR and everything seems fine. Also my followers for which the metadata was not available were successfully converted.

pfefferle:
@akirk I am currently working on two schedulers, to check the followers with errors and to remove them after x checks and on a second to update the meta-data in a regular period.

Thanks for having a look!

PR #305, opened 2023-04-21
Fix Edge Caching Problems

akirk:
I saw that many people have problems (including with WordPress caching plugins) but also with Cloudflare, that JSON is being delivered instead of HTML. This is because we're not announcing that multiple versions exist depending on an incoming `Accept:` header. This can be fixed through a `Vary` HTTP header, see https://www.fastly.com/blog/best-practices-using-vary-header


pfefferle (approved):
LGTM!

PR #307, opened 2023-04-24
Compatibility: do not serve images with Jetpack CDN when active

jeherve:
When Jetpack's image CDN is active, core calls to retrieve images return an image served by the CDN.

Since Fediverse instances usually fetch and cache the data themselves, we do not need to use the CDN for those images when returned by the ActivityPub plugin. In fact, we really do not want that to happen, as Fediverse instances may get errors when trying to fetch images from the CDN (they may get blocked / rate-limited / ...).

Let's hook into Jetpack's CDN to avoid that.

> **Warning**
> Jetpack's Image CDN is currently changing, and will soon rely on a different class for those things:
> https://github.com/Automattic/jetpack/pull/30050/
> It may be interesting to support both the old and the new class in ActivityPub for now.
> I will update this PR once progress is made on the Jetpack side of things.


donnchawp (approved):
It works. Image was served by my blog rather than Photon.

pfefferle (approved):

pfefferle:
thanks @jeherve and @donnchawp

timnolte:
I see that this was already merged but i don't agree with this change being global without it being an option. The whole purpose of a CDN is to leverage serving assets closer to the source. This has nothing to do with Fediverse instances doing their own caching. Every single Fediverse instance is going to hit your blog for their own copy of that image. Yes, a single instance will cache the image but not all instances. I don't think this should be the default but be an option if someone has some problem.

timnolte:
BTW, I've been using the JetPack CDN with the ActivityPub plugin for years and have never had a problem. I'm chiming in now because this was apparently open/reviewed/merged while I was asleep and couldn't even chime in before it was just approved to be merged in. This also directly affects my site since I use both.

pfefferle:
I would avoid third plugin specific settings to be honest. But maybe we add an action/filter and provide a demo snippets for disabling jetpack photon for AP?!?

timnolte:
Yes, I would agree that there should be minimal third-party specific functionality embedded in the plugin.

pfefferle:
@timnolte And sorry for merging it so fast, but I am so excited about the progress ☺️

I practice patience! 🙄

jeherve:
> maybe we add an action/filter and provide a demo snippet for disabling jetpack photon for AP?!?

I opened #309 for this.

PR #309, opened 2023-04-25
Compat: only disable Jetpack's image CDN via filter

jeherve:
This follows the discussion in #307.

1. Do not disable Jetpack's image CDN in ActivityPub requests by default.
2. ~Add a new filter, activitypub_images_use_jetpack_image_cdn, that site owners can use to disable Jetpack's Image CDN if they'd like to.~
3. ~Extract image getting into its own method for improved readability.~

~One can use the filter like so:~

### Testing instructions

1. Start with a site where you have installed and activated the Jetpack plugin.
2. Go to Jetpack > Settings > Performance and enable the site accelerator feature.
3. Publish a post that includes an image.
4. Make a GET request for that post, with the `Accept: application/activity+json` header.
5. You should see the image using a `i0.wp.com` prefix in the response
6. ~Now add the above snippet in a functionality plugin, or in your theme's `functions.php` file.~
7. ~Make the same request again; the `i0.wp.com` should be gone.~

cc @timnolte

### DIFF
--- a/includes/model/class-post.php
+++ b/includes/model/class-post.php
@@ -345,9 +345,6 @@ class Post {
 	/**
 	 * Return details about an image attachment.
 	 *
-	 * Can return a CDNized URL if Jetpack's image CDN is active.
-	 * This can be disabled with a filter.
-	 *
 	 * @param int    $id         The attachment ID.
 	 * @param string $image_size The image size to retrieve. Set to 'full' by default.
 	 *


includes/model/class-post.php
```diff
@@ -305,41 +305,16 @@ public function get_attachments() {
...
-
-			$mimetype = \get_post_mime_type( $id );
-
+			$thumbnail = $this->get_image( $id );
 			if ( $thumbnail ) {
-				$image = array(
-					'type' => 'Image',
-					'url' => $thumbnail[0],
+				$mimetype = \get_post_mime_type( $id );
+				$alt      = \get_post_meta( $id, '_wp_attachment_image_alt', true );
```
jeherve, line 311: Noting that this change moves those 2 functions inside the `$thumbnail` check, so they only run when necessary. Before this PR they ran regardless of the response from `wp_get_attachment_image_src`.

pfefferle:
@jeherve This is very specific to JetPack, isn't it possible to have general `pre_` and `post_` hooks, where we can hook into, to add the `disable` code?

Kind of

`do_action( 'activitypub_pre_get_image' );`

instead of line: https://github.com/pfefferle/wordpress-activitypub/pull/309/files#diff-44de75651674f44bc586c0f66e8dddf70ec9c6a4a9eb72418be87aa6c12a6be7R360

So we can do something like:

```PHP
add_action( 'activitypub_pre_get_image', 'activitypub_pre_get_image' );
function activitypub_pre_get_image() {
		/*
		 * Disable Jetpacks image CDN image processing for this request.
		 *
		 * Note: old versions of Jetpack used the Jetpack_Photon class to do this.
		 * New versions use the Image_CDN class.
		 * Let's handle both.
		 */
		if ( \class_exists( '\Automattic\Jetpack\Image_CDN\Image_CDN' ) ) {
			\remove_filter( 'image_downsize', array( \Automattic\Jetpack\Image_CDN\Image_CDN::instance(), 'filter_image_downsize' ) );
		} elseif ( \class_exists( 'Jetpack_Photon' ) ) {
			\remove_filter( 'image_downsize', array( \Jetpack_Photon::instance(), 'filter_image_downsize' ) );
		}
}

```

So we wouldn't have any JetPack specific code in the plugin and could all be done with a simple action?!?

jeherve:
> isn't it possible to have general pre_ and post_ hooks, where we can hook into, to add the disable code?

We could definitely go that route. I've updated this PR accordingly.

pfefferle (approved):
I think this is fine and can be merged independent from the JetPack PR.

PR #318, opened 2023-05-04
update scheduler for followers

pfefferle:


includes/collection/class-followers.php
```diff
@@ -404,4 +404,94 @@ public static function get_inboxes( $user_id ) {
...
+		$args = array(
+			'taxonomy'   => self::TAXONOMY,
+			'number'     => $number,
+			'meta_key'   => 'updated_at',
+			'orderby'    => 'meta_value_num',
+			'order'      => 'DESC',
+			'meta_query' => array(
+				array(
+					'key'        => 'updated_at',
+					'value'      => strtotime( 'now' ) - $older_than,
```
akirk, line 427:
```suggestion
					'value'      => time() - $older_than,
```

includes/collection/class-followers.php
```diff
@@ -404,4 +404,94 @@ public static function get_inboxes( $user_id ) {
...
+	/**
+	 * Get all Followers that have not been updated for a given time
+	 *
+	 * @param enum $output     The output format, supported ARRAY_N, OBJECT and ACTIVITYPUB_OBJECT.
+	 * @param int  $number     Limits the result.
+	 * @param int  $older_than The time in seconds.
+	 *
+	 * @return mixed The Term list of Followers, the format depends on $output.
+	 */
+	public static function get_outdated_followers( $output = ARRAY_N, $number = 50, $older_than = 604800 ) {
```
akirk, line 417: I am not a big fan of the variable output dependent on the parameter (same for the next function). What's the benefit? Is it something you couldn't do differently? For example, I see that there is a normal array and an `ACTIVITYPUB_OBJECT` which is actually not an object but an array as well. If you added a `__toString()` function to `Follower` that returns the name, it would do the same as `ARRAY_N`.

For the other use case you could just by convention use something like this (you still need to expose the term inside Follower)

```
$outdated_followers = Followers::get_outdated_followers();
$outdated_terms = array_map( $outdated_followers, 'get_term' );
```

If you don't want to do the `__toString()`, you could also use this construct:
```
$outdated_followers = Followers::get_outdated_followers();
$outdated_terms = array_map( $outdated_followers, 'get_name' );
```

activitypub.php
```diff
@@ -31,74 +31,66 @@ function init() {
...
+
+		$class = strtolower( $class );
 
-	if ( \WP_DEBUG ) {
-		require_once \dirname( __FILE__ ) . '/includes/debug.php';
+		if ( strncmp( $class, $base, strlen( $base ) ) === 0 ) {
+			$class = str_replace( 'activitypub\\', '', $class );
+
+			if ( strpos( $class, '\\' ) ) {
+				list( $sub_dir, $class ) = explode( '\\', $class );
```
akirk, line 70: You should handle the case if there are more than one `\` in the class

activitypub.php
```diff
@@ -31,74 +31,66 @@ function init() {
...
+		$base     = 'activitypub';
+
+		$class = strtolower( $class );
 
-	if ( \WP_DEBUG ) {
-		require_once \dirname( __FILE__ ) . '/includes/debug.php';
+		if ( strncmp( $class, $base, strlen( $base ) ) === 0 ) {
+			$class = str_replace( 'activitypub\\', '', $class );
+
+			if ( strpos( $class, '\\' ) ) {
```
akirk, line 69: This would fail if the `\` is the first character, thus:
```suggestion
			if ( false !== strpos( $class, '\\' ) ) {
```

akirk, line 69: This would fail if the `\` is the first character, thus:
```suggestion
			if ( false !== strpos( $class, '\\' ) ) {
```

activitypub.php
```diff
@@ -31,74 +31,66 @@ function init() {
...
+	Scheduler::init();
+}
+\add_action( 'plugins_loaded', '\Activitypub\init' );
+
+/**
+ * Class Autoloader
+ */
+spl_autoload_register(
+	function ( $class ) {
+		$base_dir = trailingslashit( __DIR__ ) . 'includes' . DIRECTORY_SEPARATOR;
```
akirk, line 61: Just catches my eye that you are using `trailingslashit` which adds slash but then `DIRECTORY_SEPARATOR` which could be a backslash on Windows.

activitypub.php
```diff
@@ -31,74 +31,66 @@ function init() {
...
-		require_once \dirname( __FILE__ ) . '/includes/debug.php';
+		if ( strncmp( $class, $base, strlen( $base ) ) === 0 ) {
+			$class = str_replace( 'activitypub\\', '', $class );
+
+			if ( strpos( $class, '\\' ) ) {
+				list( $sub_dir, $class ) = explode( '\\', $class );
+				$base_dir = $base_dir . $sub_dir . DIRECTORY_SEPARATOR;
+			}
+
+			$filename = 'class-' . str_replace( '_', '-', $class );
```
akirk, line 74:
```suggestion
			$filename = 'class-' . strtr( $class, '_', '-' );
```

activitypub.php
```diff
@@ -31,74 +31,66 @@ function init() {
...
+			$file     = $base_dir . $filename . '.php';
+
+			if ( file_exists( $file ) ) {
+				require_once $file;
+			}
```
akirk, line 79: Worth throwing some kind of warning here?

activitypub.php
```diff
@@ -31,74 +31,66 @@ function init() {
...
+
+			if ( strpos( $class, '\\' ) ) {
+				list( $sub_dir, $class ) = explode( '\\', $class );
+				$base_dir = $base_dir . $sub_dir . DIRECTORY_SEPARATOR;
+			}
+
+			$filename = 'class-' . str_replace( '_', '-', $class );
+			$file     = $base_dir . $filename . '.php';
+
+			if ( file_exists( $file ) ) {
```
akirk, line 77: You should probably also test if `is_readable()`

activitypub.php
```diff
@@ -31,74 +31,66 @@ function init() {
...
+				require_once $file;
+			}
+		}
 	}
+);
+
+require_once \dirname( __FILE__ ) . '/includes/functions.php';
+
+// load NodeInfo endpoints only if blog is public
+if ( true === (bool) \get_option( 'blog_public', 1 ) ) {
```
akirk, line 87:
```suggestion
if ( \get_option( 'blog_public', 1 ) ) {
```

includes/class-scheduler.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @return void
+	 */
+	public static function cleanup_followers() {
+		$followers = Followers::get_faulty_followers( ACTIVITYPUB_OBJECT );
+
+		foreach ( $followers as $follower ) {
+			$meta = get_remote_metadata_by_actor( $follower->get_actor() );
+
+			if ( empty( $meta ) || ! is_array( $meta ) || is_wp_error( $meta ) ) {
+				if ( 5 >= $follower->count_errors() ) {
```
akirk, line 127: I think you got this flipped. This would delete the follower if there have been fewer than 5 errors.
```suggestion
				if ( 5 >= $follower->count_errors() ) {
```

akirk, line 127: I think you got this flipped. This would delete the follower if there have been fewer than 5 errors.
```suggestion
				if ( 5 >= $follower->count_errors() ) {
```

includes/class-scheduler.php
```diff
@@ -0,0 +1,136 @@
...
+					$follower->delete();
+				} else {
+					$follower->set_error( $meta );
```
akirk, line 130: Maybe I am missing something in the Follower class but I dont see how `set_error` increases the `count_errors`?

akirk (commented):
I haven't tested the code but noticed a few things.

includes/class-scheduler.php
pfefferle: not directly, but after the next call.

activitypub.php
pfefferle: is WordPress casting this correctly if '0'? I may remember having issues in the past.

activitypub.php
pfefferle: like: file not accessible/readable?

activitypub.php
pfefferle: @akirk see https://core.trac.wordpress.org/ticket/44579

akirk: This is a PHP question and `if` does the `true === (bool) expr` implicitly, so you can just drop it. See this example that I crafted: https://3v4l.org/9YSjN

pfefferle: wasn't aware of that, thanks!

includes/class-scheduler.php
pfefferle: I reset the errors now, so that they will be requested again.

PR #319, opened 2023-05-05
Profile settings: use string to fetch user description instead of constant

jeherve:
Follow-up from #304

Since we do not use a constant anywhere else just yet, let's keep using a string in the settings page.


pfefferle:
Oh, have I missed on 🙄

Thanks!

pfefferle (approved):
LGTM!

PR #322, opened 2023-05-05
fix migrate_from_0_17() error

mediaformat:
I think this should fix the problem of [$actor](https://github.com/Automattic/wordpress-activitypub/blob/master/includes/class-migration.php#L63) being undefined.

Untested.


mediaformat:
My fault for running code from Master ;-)

Maybe we could use feature branches...

pfefferle:
no, this was already merged and is a real issue! thanks for fixing!

pfefferle (approved):
LGTM

PR #324, opened 2023-05-09
Introduce Caching

pfefferle:
Adds Inbox-Caching

## Proposed changes:

* Cache Follower-Inbox-List

### Other information:

### DIFF
--- a/includes/collection/class-followers.php
+++ b/includes/collection/class-followers.php
@@ -19,6 +19,7 @@ use function Activitypub\get_remote_metadata_by_actor;
  */
 class Followers {
 	const TAXONOMY = 'activitypub-followers';
+	const CACHE_KEY_INBOXES = 'activitypub_follower_inboxes_for_%s';
 
 	/**
 	 * Register WordPress hooks/actions and register Taxonomy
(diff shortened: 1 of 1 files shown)


pfefferle:
@mediaformat @akirk @mattwiebe @jeherve any idea what should also be cached?

includes/collection/class-followers.php
```diff
@@ -369,6 +374,13 @@ public static function count_followers( $user_id ) {
 	 * @return array The list of Inboxes
 	 */
 	public static function get_inboxes( $user_id ) {
+		$cache_key = sprintf( self::CACHE_KEY_INBOXES, $user_id );
+		$inboxes = wp_cache_get( $cache_key );
```
akirk, line 378: Instead of using a very verbose cache key, you'd rather want to use the `$group` parameter of the `wp_cache_*` calls.

pfefferle: ...and the user_id as key?

pfefferle:
@mattwiebe @akirk @jeherve @mediaformat any idea what else to cache? otherwise I would love to at least merge this one.

PR #327, opened 2023-05-11
sanitize output

pfefferle:
Fixes sanitisation. thanks @xknown for your feedback!

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/help.php
+++ b/includes/help.php
@@ -9,8 +9,8 @@
 			'<dl>' .
 				'<dt><code>[ap_title]</code></dt>' .
 				'<dd>' . \wp_kses( __( 'The post\'s title.', 'activitypub' ), 'default' ) . '</dd>' .
-				'<dt><code>[ap_content apply_filters="yes"]</code></dt>' .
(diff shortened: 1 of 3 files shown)


includes/help.php
```diff
@@ -9,8 +9,8 @@
 			'<dl>' .
 				'<dt><code>[ap_title]</code></dt>' .
 				'<dd>' . \wp_kses( __( 'The post\'s title.', 'activitypub' ), 'default' ) . '</dd>' .
-				'<dt><code>[ap_content apply_filters="yes"]</code></dt>' .
-				'<dd>' . \wp_kses( __( 'The post\'s content. With <code>apply_filters</code> you can decide if filters should be applied or not (default is <code>yes</code>). The values can be <code>yes</code> or <code>no</code>. <code>apply_filters</code> attribute is optional.', 'activitypub' ), 'default' ) . '</dd>' .
+				'<dt><code>[ap_content]</code></dt>' .
+				'<dd>' . \wp_kses( __( 'The post\'s content.', 'activitypub' ), 'default' ) . '</dd>' .
```
mattwiebe, line 13: We can migrate this from `wp_kses` to simply `esc_html` now since there are no HTML components left in the string to be translated.

Furthermore, when there was still `a string with <code>some code</code>`, we can be more explicit to `wp_kses` about what we'll allow through, like:

```php
wp_kses(__(  'a string with <code>some code</code>' ), [ 'code' ] );
```

PR #328, opened 2023-05-11
Compat: introduce a conditional to detect ActivityPub requests

jeherve:
## Proposed changes:

This conditional could be used within the plugin, but also by third-party plugins, to detect whether a request is an ActivityPub request, without having to manually check for query vars and headers every time.

### Other information:

## Testing instructions:

This is better tested with a code snippet, just like a third-party plugin would use on their end. Something like this:

```php
add_action(
	'wp',
	function () {
		if ( function_exists( '\Activitypub\is_activitypub_request' ) ) {
			error_log( 'the function exists' );
			if ( \Activitypub\is_activitypub_request() ) {
				error_log( 'This is an ActivityPub request' );
			}
		}
	}
);
```

Then make requests and see the output of the function:

* Single posts
* Single posts with a custom `Accept: application/activity+json` header.
* Archive pages.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -228,3 +228,47 @@ function is_tombstone( $wp_error ) {
 
 	return false;
 }
+
+/**
+ * Check if a request is for an ActivityPub request.
+ *
+ * @return bool False by default.
+ */
+function is_activitypub_request() {
+	global $wp_query;
+
+	/*
+	 * ActivityPub requests are currently only made for
+	 * author archives, singular posts, and the homepage.
+	 */
+	if ( ! \is_author() && ! \is_singular() && ! \is_home() ) {
+		return false;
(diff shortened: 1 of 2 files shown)


pfefferle (approved):
LGTM

pfefferle:
Maybe we could/should also use this function for the internal detection, to be consistant.

See: https://github.com/Automattic/wordpress-activitypub/blob/master/includes/class-activitypub.php#L37

jeherve:
> Maybe we could/should also use this function for the internal detection, to be consistant.

Yeah, that seems like a good idea. I think we should do it in a follow-up PR though, since it will require refactoring that method. What do you think?

pfefferle:
@jeherve fine!

PR #329, opened 2023-05-11
Allow setting the REST namespace with `ACTIVITYPUB_REST_NAMESPACE`

mattwiebe:
As discussed elsewhere, the REST API will need to land within a `wpcom/*` namespace to work with our very particular requirements on wordpress.com. This change should allow it to be configurable by us or anyone who would want to use a different namespace in their own environment.

## Proposed changes:

* Introduce a `ACTIVITYPUB_REST_NAMESPACE` constant that can be overridden from the default `activitypub/1.0` namespace.

### Other information:

## Testing instructions:
I don't know how, yet. :)

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -32,8 +32,8 @@ function init() {
 	\define( 'ACTIVITYPUB_PLUGIN_FILE', plugin_dir_path( __FILE__ ) . '/' . basename( __FILE__ ) );
 
 	Migration::init();
-	Activity_Dispatcher::init();
 	Activitypub::init();
+	Activity_Dispatcher::init();
 	Collection\Followers::init();
 
 	// Configure the REST API route
@@ -41,13 +41,13 @@ function init() {
 	Rest\Inbox::init();
 	Rest\Followers::init();
 	Rest\Following::init();
+	Rest\Nodeinfo::init();
 	Rest\Webfinger::init();
 
 	Admin::init();
 	Hashtag::init();
 	Shortcodes::init();
 	Mention::init();
-	Debug::init();
 	Health_Check::init();
 	Scheduler::init();
 }
@@ -96,6 +96,7 @@ if ( \get_option( 'blog_public', 1 ) ) {
 $debug_file = \dirname( __FILE__ ) . '/includes/debug.php';
 if ( \WP_DEBUG && file_exists( $debug_file ) && is_readable( $debug_file ) ) {
 	require_once $debug_file;
+	Debug::init();
 }
 
 /**
--- a/activitypub.php
+++ b/activitypub.php
@@ -25,6 +25,7 @@ function init() {
 	\defined( 'ACTIVITYPUB_HASHTAGS_REGEXP' ) || \define( 'ACTIVITYPUB_HASHTAGS_REGEXP', '(?:(?<=\s)|(?<=<p>)|(?<=<br>)|^)#([A-Za-z0-9_]+)(?:(?=\s|[[:punct:]]|$))' );
 	\defined( 'ACTIVITYPUB_USERNAME_REGEXP' ) || \define( 'ACTIVITYPUB_USERNAME_REGEXP', '(?:([A-Za-z0-9_-]+)@((?:[A-Za-z0-9_-]+\.)+[A-Za-z]+))' );
 	\defined( 'ACTIVITYPUB_CUSTOM_POST_CONTENT' ) || \define( 'ACTIVITYPUB_CUSTOM_POST_CONTENT', "<strong>[ap_title]</strong>\n\n[ap_content]\n\n[ap_hashtags]\n\n[ap_shortlink]" );
+	defined( 'ACTIVITYPUB_REST_NAMESPACE' ) || define( 'ACTIVITYPUB_REST_NAMESPACE', 'activitypub/1.0' );
 
 	\define( 'ACTIVITYPUB_PLUGIN_DIR', plugin_dir_path( __FILE__ ) );
 	\define( 'ACTIVITYPUB_PLUGIN_BASENAME', plugin_basename( __FILE__ ) );
@@ -120,12 +121,12 @@ function plugin_settings_link( $actions ) {
  */
 function add_rewrite_rules() {
 	if ( ! \class_exists( 'Webfinger' ) ) {
-		\add_rewrite_rule( '^.well-known/webfinger', 'index.php?rest_route=/activitypub/1.0/webfinger', 'top' );
+		\add_rewrite_rule( '^.well-known/webfinger', 'index.php?rest_route=/' . ACTIVITYPUB_REST_NAMESPACE . '/webfinger', 'top' );
 	}
 
 	if ( ! \class_exists( 'Nodeinfo' ) || ! (bool) \get_option( 'blog_public', 1 ) ) {
-		\add_rewrite_rule( '^.well-known/nodeinfo', 'index.php?rest_route=/activitypub/1.0/nodeinfo/discovery', 'top' );
-		\add_rewrite_rule( '^.well-known/x-nodeinfo2', 'index.php?rest_route=/activitypub/1.0/nodeinfo2', 'top' );
+		\add_rewrite_rule( '^.well-known/nodeinfo', 'index.php?rest_route=/' . ACTIVITYPUB_REST_NAMESPACE . '/nodeinfo/discovery', 'top' );
+		\add_rewrite_rule( '^.well-known/x-nodeinfo2', 'index.php?rest_route=/' . ACTIVITYPUB_REST_NAMESPACE . '/nodeinfo2', 'top' );
 	}
 
 	\add_rewrite_endpoint( 'activitypub', EP_AUTHORS | EP_PERMALINK | EP_PAGES );
--- a/activitypub.php
+++ b/activitypub.php
@@ -25,7 +25,7 @@ function init() {
 	\defined( 'ACTIVITYPUB_HASHTAGS_REGEXP' ) || \define( 'ACTIVITYPUB_HASHTAGS_REGEXP', '(?:(?<=\s)|(?<=<p>)|(?<=<br>)|^)#([A-Za-z0-9_]+)(?:(?=\s|[[:punct:]]|$))' );
 	\defined( 'ACTIVITYPUB_USERNAME_REGEXP' ) || \define( 'ACTIVITYPUB_USERNAME_REGEXP', '(?:([A-Za-z0-9_-]+)@((?:[A-Za-z0-9_-]+\.)+[A-Za-z]+))' );
 	\defined( 'ACTIVITYPUB_CUSTOM_POST_CONTENT' ) || \define( 'ACTIVITYPUB_CUSTOM_POST_CONTENT', "<strong>[ap_title]</strong>\n\n[ap_content]\n\n[ap_hashtags]\n\n[ap_shortlink]" );
-	defined( 'ACTIVITYPUB_REST_NAMESPACE' ) || define( 'ACTIVITYPUB_REST_NAMESPACE', 'activitypub/1.0' );
+	\defined( 'ACTIVITYPUB_REST_NAMESPACE' ) || \define( 'ACTIVITYPUB_REST_NAMESPACE', 'activitypub/1.0' );
 
 	\define( 'ACTIVITYPUB_PLUGIN_DIR', plugin_dir_path( __FILE__ ) );
 	\define( 'ACTIVITYPUB_PLUGIN_BASENAME', plugin_basename( __FILE__ ) );
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -233,6 +233,7 @@ function is_tombstone( $wp_error ) {
  * Get the REST URL relative to this plugin's namespace.
  *
  * @param string $path Optional. REST route path. Otherwise this plugin's namespaced root.
+ *
  * @return string REST URL relative to this plugin's namespace.
  */
 function get_rest_url_by_path( $path = '' ) {
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -228,3 +228,16 @@ function is_tombstone( $wp_error ) {
 
 	return false;
 }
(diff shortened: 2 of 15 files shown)


includes/model/class-post.php
```diff
@@ -142,7 +142,8 @@ class Post {
 	 */
 	public function __construct( $post ) {
 		$this->post = \get_post( $post );
-		$this->add_to( \get_rest_url( null, '/activitypub/1.0/users/' . intval( $this->get_post_author() ) . '/followers' ) );
+		$url = sprintf( '/%/users/%d/followers', ACTIVITYPUB_REST_NAMESPACE, intval( $post->get_post_author() ) );
```
pfefferle, line 145: @mattwiebe Shouldn't this be a different CONST, because it has to use the full URL (public-api.wordpress.com/...) instead of only the different namespace?

pfefferle:
I think we need different configuration possibilities for registering API endpoints (because we maybe want to simply change the namespace) and the URLs that are shown to the user (which should be the complete URL with the public-api.wordpress.com host).

activitypub.php
```diff
@@ -120,12 +121,12 @@ function plugin_settings_link( $actions ) {
  */
 function add_rewrite_rules() {
 	if ( ! \class_exists( 'Webfinger' ) ) {
-		\add_rewrite_rule( '^.well-known/webfinger', 'index.php?rest_route=/activitypub/1.0/webfinger', 'top' );
+		\add_rewrite_rule( '^.well-known/webfinger', 'index.php?rest_route=/' . ACTIVITYPUB_REST_NAMESPACE . '/webfinger', 'top' );
```
pfefferle, line 124: Does this even work if .com blocks the local API endpoint or do we have to rethink this completely?

includes/rest/class-nodeinfo.php
```diff
@@ -173,7 +173,7 @@ public static function discovery( $request ) {
 		$discovery['links'] = array(
 			array(
 				'rel' => 'http://nodeinfo.diaspora.software/ns/schema/2.0',
-				'href' => \get_rest_url( null, 'activitypub/1.0/nodeinfo' ),
+				'href' => \get_rest_url( null, ACTIVITYPUB_REST_NAMESPACE . '/nodeinfo' ),
```
pfefferle, line 176: I would also use `sprintf` here to be consistent.

tests/test-class-activitypub-activity.php
```diff
@@ -22,7 +22,7 @@ function( $mentions ) {
 		$activitypub_activity = new \Activitypub\Model\Activity( 'Create' );
 		$activitypub_activity->from_post( $activitypub_post );
 
-		$this->assertContains( \get_rest_url( null, '/activitypub/1.0/users/1/followers' ), $activitypub_activity->get_to() );
+		$this->assertContains( \get_rest_url( null, '/' . ACTIVITYPUB_REST_NAMESPACE . '/users/1/followers' ), $activitypub_activity->get_to() );
```
pfefferle, line 25: I would also use `sprintf` here to be consistent.

includes/model/class-post.php
pfefferle, line 145: there is a typo, it has to be `$this->get_post_author()` instead of `$post->get_post_author()`! (that will fix the unit-tests)

activitypub.php
mattwiebe: So it's the `/wp-json/` route that's blocked but the underlying `index.php?rest_route=*` probably still works fine. (Checked now and nope, not fine in initial testing)

But the `add_rewrite_call` itself gets handled uniquely on wpcom so more investigation will be required, and if we need to get help from systems to allow any calls within `.well-known`.

pfefferle: Maybe we have to simply use a different route, not under the API endpoint. Something like: 

`\add_rewrite_rule( '^.well-known/webfinger', 'index.php?webfinger=endpoint', 'top' );`

includes/rest/class-nodeinfo.php
mattwiebe: I think that, instead, I'm going to make an internal helper for this.

activitypub.php
mattwiebe: I investigated a bit more and we're almost certainly going to need help from the systems folks to get anything handled under the `.well-known` namespace. I can't even put a PHP file at `.well-known/webfinger/index.php` and reach it. It's being directed elsewhere at the nginx level, so we'll have to ask. I assume it has to be exactly of the pattern `/.well-known/webfinger` for spec conformity?

I also see `nodeinfo` and `x-nodeinfo2`, anything else that needs to be in `.well-known`? And my impression was that only `webfinger` was _required_ and can't be changed to another URL pattern right?

includes/rest/class-webfinger.php
```diff
@@ -24,6 +24,7 @@ public static function init() {
 	 * Register routes
 	 */
 	public static function register_routes() {
+		\l( 'register webfinger' );
```
pfefferle, line 27: remove debug function

includes/model/class-activity.php
```diff
@@ -148,8 +148,8 @@ public function from_post( Post $post ) {
 			$this->published = $object['published'];
 		}
 
-		$url = sprintf( '/%/users/%d/followers', ACTIVITYPUB_REST_NAMESPACE, intval( $post->get_post_author() ) );
-		$this->add_to( \get_rest_url( null, $url ) );
+		$path = sprintf( 'users/%d/followers', intval( $post->get_post_author() ) );
+		$this->add_to( get_rest_url_by_path( $path ) );
```
pfefferle, line 152: `get_rest_url_by_path` is in the `Activitypub` namespace, so you have to add `use function Activitypub\get_rest_url_by_path` as the top of the class.

includes/model/class-post.php
```diff
@@ -142,8 +142,8 @@ class Post {
 	 */
 	public function __construct( $post ) {
 		$this->post = \get_post( $post );
-		$url = sprintf( '/%/users/%d/followers', ACTIVITYPUB_REST_NAMESPACE, intval( $post->get_post_author() ) );
-		$this->add_to( \get_rest_url( null, $url ) );
+		$path = sprintf( 'users/%d/followers', intval( $this->get_post_author() ) );
+		$this->add_to( get_rest_url_by_path( $path ) );
```
pfefferle, line 146: get_rest_url_by_path is in the Activitypub namespace, so you have to add use function Activitypub\get_rest_url_by_path as the top of the class.

includes/rest/class-followers.php
```diff
@@ -78,7 +78,7 @@ public static function get( $request ) {
 		$json->actor = \get_author_posts_url( $user_id );
 		$json->type = 'OrderedCollectionPage';
 
-		$json->partOf = \get_rest_url( null, sprintf( '/%s/users/%d/followers', ACTIVITYPUB_REST_NAMESPACE, $user_id ) ); // phpcs:ignore
+		$json->partOf = get_rest_url_by_path( sprintf( 'users/%d/followers', $user_id ) ); // phpcs:ignore
```
pfefferle, line 81: get_rest_url_by_path is in the Activitypub namespace, so you have to add use function Activitypub\get_rest_url_by_path as the top of the class.

activitypub.php
pfefferle: only WebFinger and NodeInfo for the moment.

includes/rest/class-following.php
```diff
@@ -72,7 +72,7 @@ public static function get( $request ) {
 		$json->actor = \get_author_posts_url( $user_id );
 		$json->type = 'OrderedCollectionPage';
 
-		$json->partOf = \get_rest_url( null, "/activitypub/1.0/users/$user_id/following" ); // phpcs:ignore
+		$json->partOf = get_rest_url_by_path( sprintf( 'users/%d/following', $user_id ) ); // phpcs:ignore
```
pfefferle, line 75: `get_rest_url_by_path` is in the `Activitypub` namespace, so you have to add `use function Activitypub\get_rest_url_by_path` as the top of the class.

includes/rest/class-inbox.php
```diff
@@ -108,7 +108,7 @@ public static function user_inbox_get( $request ) {
 		$json->id = \home_url( \add_query_arg( null, null ) );
 		$json->generator = 'http://wordpress.org/?v=' . \get_bloginfo_rss( 'version' );
 		$json->type = 'OrderedCollectionPage';
-		$json->partOf = \get_rest_url( null, "/activitypub/1.0/users/$user_id/inbox" ); // phpcs:ignore
+		$json->partOf = get_rest_url_by_path( sprintf( 'users/%d/inbox', $user_id ) ); // phpcs:ignore
```
pfefferle, line 111: `get_rest_url_by_path` is in the `Activitypub` namespace, so you have to add `use function Activitypub\get_rest_url_by_path` as the top of the class.

includes/rest/class-nodeinfo.php
```diff
@@ -173,7 +173,7 @@ public static function discovery( $request ) {
 		$discovery['links'] = array(
 			array(
 				'rel' => 'http://nodeinfo.diaspora.software/ns/schema/2.0',
-				'href' => \get_rest_url( null, 'activitypub/1.0/nodeinfo' ),
+				'href' => get_rest_url_by_path( 'nodeinfo' ),
```
pfefferle, line 176: `get_rest_url_by_path` is in the `Activitypub` namespace, so you have to add `use function Activitypub\get_rest_url_by_path` as the top of the class.

includes/model/class-activity.php
mattwiebe: Aha! I am not accustomed to thinking in namespaces :)

includes/rest/class-outbox.php
```diff
@@ -72,7 +72,7 @@ public static function user_outbox_get( $request ) {
 		$json->generator = 'http://wordpress.org/?v=' . \get_bloginfo_rss( 'version' );
 		$json->actor = \get_author_posts_url( $user_id );
 		$json->type = 'OrderedCollectionPage';
-		$json->partOf = \get_rest_url( null, "/activitypub/1.0/users/$user_id/outbox" ); // phpcs:ignore
+		$json->partOf = get_rest_url_by_path( sprintf( 'users/%d/outbox', $user_id ) ); // phpcs:ignore
```
pfefferle, line 75: `get_rest_url_by_path` is in the `Activitypub` namespace, so you have to add `use function Activitypub\get_rest_url_by_path` as the top of the class.

tests/test-class-activitypub-activity.php
```diff
@@ -22,7 +22,7 @@ function( $mentions ) {
 		$activitypub_activity = new \Activitypub\Model\Activity( 'Create' );
 		$activitypub_activity->from_post( $activitypub_post );
 
-		$this->assertContains( \get_rest_url( null, '/activitypub/1.0/users/1/followers' ), $activitypub_activity->get_to() );
+		$this->assertContains(  get_rest_url_by_path( 'users/1/followers' ), $activitypub_activity->get_to() );
```
pfefferle, line 25: use `\Activitypub\get_rest_url_by_path`

templates/blog-json.php
```diff
@@ -27,10 +27,10 @@
 	);
 }
 
-$json->inbox = \get_rest_url( null, '/activitypub/1.0/blog/inbox' );
-$json->outbox = \get_rest_url( null, '/activitypub/1.0/blog/outbox' );
-$json->followers = \get_rest_url( null, '/activitypub/1.0/blog/followers' );
-$json->following = \get_rest_url( null, '/activitypub/1.0/blog/following' );
+$json->inbox = get_rest_url_by_path( 'blog/inbox' );
```
pfefferle, line 30: use `\Activitypub\get_rest_url_by_path`

templates/author-json.php
```diff
@@ -28,10 +28,10 @@
 	);
 }
 
-$json->inbox = \get_rest_url( null, "/activitypub/1.0/users/$author_id/inbox" );
-$json->outbox = \get_rest_url( null, "/activitypub/1.0/users/$author_id/outbox" );
-$json->followers = \get_rest_url( null, "/activitypub/1.0/users/$author_id/followers" );
-$json->following = \get_rest_url( null, "/activitypub/1.0/users/$author_id/following" );
+$json->inbox = get_rest_url_by_path( sprintf( 'users/%d/inbox', $author_id ) );
```
pfefferle, line 31: use `\Activitypub\get_rest_url_by_path`

includes/rest/class-webfinger.php
mattwiebe: Oh haha good thing it wasn't a spicier one! ;)

mattwiebe: I'll have to figure out a way to run these tests rather than crossing my fingers!

includes/functions.php
```diff
@@ -238,8 +238,11 @@ function is_tombstone( $wp_error ) {
 function get_rest_url_by_path( $path = '' ) {
 	// we'll handle the leading slash.
 	$path = ltrim( $path, '/' );
-	$url = sprintf( '/%s/%s', ACTIVITYPUB_REST_NAMESPACE, $path );
-	return \get_rest_url( null, $url );
+	$namespaced_path = sprintf( '/%s/%s', ACTIVITYPUB_REST_NAMESPACE, $path );
+	$rest_url = \get_rest_url( null, $namespaced_path );
+	// Just in case there are non-default ways of handling REST URLs.
+	$rest_url = \apply_filters( 'activitypub_rest_url', $rest_url, $path, ACTIVITYPUB_REST_NAMESPACE );
```
pfefferle, line 244: do we need this filter? `get_rest_url` already has a filter that might work already: https://developer.wordpress.org/reference/hooks/rest_url/

mattwiebe: We may not. I was mostly expecting it to already be transformed and it wasn't so I reached for this first. Further investigation needed

mattwiebe:
@pfefferle do you think this is ready to merge?

pfefferle (approved):
LGTM

PR #336, opened 2023-05-17
support query handling

pfefferle:
support simple query handling as alternative to the rest-endpoint, to support WP.com

Fixes issues with API-Endpoints on WP.com

## Proposed changes:

*

### Other information:

## Testing instructions:

* Use `\define( 'ACTIVITYPUB_WELL_KNOWN_HANDLER', 'query' );` in the loader script or wp-config.php to test the changes


mattwiebe (commented):
Left one comment on approach. I'm also considering that we leave the plugin alone and just manage this complexity on the wpcom side

mattwiebe:
> @mattwiebe We can also install the WebFinger plugin https://github.com/pfefferle/wordpress-webfinger

I am open to this when and if we have a use case outside of ActivityPub itself, as I noted when I pinged you about Gravatar. For now I've got stuff working for wpcom without needing this.

pfefferle:
Perfect, then I will close this for now!

PR #337, opened 2023-05-19
Prevent activation error: Set `ACTIVITYPUB_REST_NAMESPACE` early

mattwiebe:
Needed to prevent activation errors.

Fixes bug introduced in #329 

## Proposed changes:

* defines ACTIVITYPUB_REST_NAMESPACE early

### Other information:

## Testing instructions:
Activate plugin. On `master` it will produce a fatal error since `\Activitypub\init()` isn't called first to define it. Will activate fine with this.

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -15,6 +15,8 @@
 
 namespace Activitypub;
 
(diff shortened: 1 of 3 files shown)


pfefferle (approved):
LGTM

PR #342, opened 2023-05-24
prepare pseudo users like a blog wide user. (Catch-All)

### LINKED ISSUE #1: Catchall-Account
Hej,
I write my comment here because I am not sure if your OStatus-suite has a future with ActivityPub at the horizon.

It would be cool if it was possible to follow the whole blog and not just accounts. At the moment you can just follow @user1@example.com, @user2@example.com, @user3@example.com - it would be nice if I could just follow *@example.com. I don't care what * is or if I am able to customize it...

pfefferle:
this allows also other constructs like tag oder category users

fix #1

## Proposed changes:

* Introduces @-mention URLs like used by Mastodon and others `https://example.org/@username` to have a general URL scheme that works for all users including pseudo users. The URL will forward to the canonical URL if not called with the ActivityPub Accept header, this is the author page for Users and the root URL for the blog user.
* Generates a default Username for the blog user, based on the following cascade
  1. host, if no subdomain and smaller than x (for example `notiz.blog@notiz.blog`)
  2. the title (for example `notizblog@notiz.blog`)
  3. a random string from the following list: 'feed', 'all', 'everyone', 'authors', 'follow', 'posts'.

### Other information:

## Testing instructions:

* Go to '..'
*


includes/class-activitypub.php
```diff
@@ -241,6 +214,11 @@ public static function add_rewrite_rules() {
 				'index.php?rest_route=/' . ACTIVITYPUB_REST_NAMESPACE . '/nodeinfo2',
 				'top'
 			);
+			\add_rewrite_rule(
+				'^@([\w]+)',
+				'index.php?rest_route=/' . ACTIVITYPUB_REST_NAMESPACE . '/users/$matches[1]',
+				'top'
+			);
```
mattwiebe, line 221: This is going to allow us to do something like this?

```
https://example.com/@user@example.com
```

includes/model/class-blog-user.php
```diff
@@ -0,0 +1,95 @@
...
+	 * Get the User-Url.
+	 *
+	 * @return string The User-Url.
+	 */
+	public function get_url() {
+		return \esc_url( \trailingslashit( get_home_url() ) . '@' . $this->get_username() );
+	}
+
+	public function get_username() {
+		return \esc_html( \get_bloginfo( 'activitypub_blog_identifier', 'feed' ) );
```
mattwiebe, line 59: Curious about the use of `get_bloginfo` here? (And the other one for `activitypub_application_identifier`)

Do we want `get_option()`? `get_bloginfo` is just a proxy for a variety of standard site options.

includes/class-activitypub.php
pfefferle: The second one. To have URLs for pseudo users (like the blog user) that has no author page. It's kind of the same as on Mastodon, where the @username URLs redirect to the classic URLs.

includes/model/class-blog-user.php
pfefferle: Oops! That was a copy and paste issue! I wanted to use get_option! Good catch!

mattwiebe:
An idea I'm playing with, what do you think?

Use `@example.com@example.com` as the catchall account. Perhaps allow overriding via constant or filter (and I would even see if we can get away without that), but `URL@URL` is a consistent, predictable, workable default that requires no setting and won't collide with any given WP user name. It's truly an _identity_ for the whole site. A good place for decisions not options!

And if we can't collide on usernames, there's a bunch of chances for code simplification in here. And I think we can then support both catchall user and single users at the same time with no trouble? Or is there a problem with content duplication that way?

I also remember back when I knew I could try `/feed` to any given site and find out if it was WP that way. I like predictable, hackable URLs and this provides that for our whole-site identifier.

mediaformat:
> Use `@example.com@example.com` as the catchall account.

We [considered this](https://github.com/Automattic/wordpress-activitypub/pull/299#discussion_r1195936738) for the "application" actor, but found a risk for WP users who use / have previously used brid.gy fed.

mattwiebe:
> We [considered this](https://github.com/Automattic/wordpress-activitypub/pull/299#discussion_r1195936738) for the "application" actor, but found a risk for WP users who use / have previously used brid.gy fed.

Hi @mediaformat , thanks a bunch for the context!

My perspective is that I would hate to make our product worse simply because some users have hooked their site up to Bridgy at some point. Requiring everyone to default to the weird-looking `example-com@example.com` just because it won't conflict with an edge case seems backwards.

But this is good to understand so that we can ensure there's a sensible path to allowing those users to work around their issue by setting a custom identifier as a fallback. Can we outright detect Bridgy usage?

It might also mean that we don't want to enable the catchall "account" for existing sites that upgrade to the new version, but we do opt in fresh plugin activations. The catchall account seems to be what most people would like. It's certainly what I was expecting when I came to the plugin.

timnolte:
I really don't think the site-wide user should ever be on by default. I actually wouldn't expect that to be the default. Honestly, if someone only has 1 account on their site why would they not just be using their single account as the author anyways. As I see it this sort of "user" is really just for sites with multiple authors that want content to appear from only a single account regardless of the author, or as an aggregate type of account when there are multiple authors.

mattwiebe:
> I really don't think the site-wide user should ever be on by default. I actually wouldn't expect that to be the default. Honestly, if someone only has 1 account on their site why would they not just be using their single account as the author anyways. As I see it this sort of "user" is really just for sites with multiple authors that want content to appear from only a single account regardless of the author, or as an aggregate type of account when there are multiple authors.

Thanks for the input! I am agnostic at this point about if both user- and site-centric modes should or can be enabled simultaneously. 

But I do think that onboarding new users and setting expectations so that they can decide which works best for them is the way forward. The user-based model is agreeable to you and that's great, but that's far from universal: #1's popularity attests to that.

timnolte:
Is this feature going to be implementing the ActivityPub Groups feature?

https://socialhub.activitypub.rocks/t/groups-implementation/591

It seems like for WordPress this might be classified as a "Working Group".

FoxLee:
> I really don't think the site-wide user should ever be on by default. I actually wouldn't expect that to be the default. Honestly, if someone only has 1 account on their site why would they not just be using their single account as the author anyways.

Since you asked—because this isn't all about blogs. Sometimes I'm publishing content that I want divorced from authorial identity; for example, I have a site which is a vaguely wiki-esque rules compendium for a game. On the site there is a clear delineation between blog articles and rules content, and I display author info only on blog posts, because that's the only time I'm writing conversationally. I'm still the only author on the site, but only one type of content acknowledges the presence of an author. 

With both types of content in mind, it's definitely more appropriate for the feed to be <site> as an identity, not <me> as an identity; at the _very_ least, it should have the site's icon for a profile pic, not my personal user icon. But it's silly to ask a user to change their name and icon for the sake of the feed's presentation (plus I _do_ want to look like a person when my info is displayed on the blog articles). I could make two authors, but I don't want two feeds; this is not a high volume site, any anybody following it almost certainly wants updates of both types.

In short, a unified feed would give me exactly the end result I want without any silly hacks.

timnolte:
@FoxLee ultimately my point was about defaults. While your use case warrants using the full site pseudo identity I still don't think it should be the default but easy enough for folks like yourself that do need this feature.

pfefferle:
Does it hurt if the blog-wide user is active by default? I think you can force visitors to follow either of the identifiers, by only providing the identifiers of the users you want them to follow.

I am also thinking about: maybe the blog-wide user will "only" boost the users/authors post (at least at default setting).

But I hear you and will think about a way to enable/disable users.

timnolte:
@pfefferle so as a plugin developer myself I'm always looking at how I am handling changes for existing users when my plugins are upgraded. I never want my plugins to unexpectedly do something, or change a users site, in ways they were never expecting. I never want to force something on my users that they didn't ask for, so I work to always give them options.

Additionally, I feel like WordPress sites aren't necessarily multiuser by majority. I'd be very curious to look at the FediDb tracker to see how many WordPress sites it finds are actually multi-author/user sites. My take on this feature is that a large portion of the user base has no need for it. That doesn't mean I'm saying that it's not needed but for those that do need it let them turn it on when they upgrade or install it.

With any WordPress plugin it is generally better to have less on by default and let people turn on the things they need. This also helps to reduce the load on a site and potentially reduce the attack surface a plugin adds to a site.

pfefferle:
Hey @timnolte 

> Additionally, I feel like WordPress sites aren't necessarily multiuser by majority. I'd be very curious to look at the FediDb tracker to see how many WordPress sites it finds are actually multi-author/user sites. My take on this feature is that a large portion of the user base has no need for it. That doesn't mean I'm saying that it's not needed but for those that do need it let them turn it on when they upgrade or install it.

That is exactly the point, why a lot of users requested a blog wide account. To not use a WordPress user, but to have the blog itself as a user. If you refer to the "boost" feature, you are right!

What I want:

* Be able to disable/enable users and/or the blog wide user
* Decide if the blog-wide user should post real Activities or if he should "only" boost the user posts (maybe this could be implicit: If user accounts are active -> boost, if not -> create full activity)

I would love to have some kind of onboarding process, to check what setting is the best for every user.

timnolte:
@pfefferle yeah, I think an onboarding process would be very helpful.

mediaformat:
Since the [2nd commenter](https://github.com/Automattic/wordpress-activitypub/issues/1#issuecomment-416265531) on #1 there were actually 2 models proposed:
1. An Group type actor that will Announce all published posts.
2. A single user mode, where all Posts regardless of the Author post as the Single Actor. 

I think it makes sense for both to be options (off by default).

mediaformat:
Single Actor mode brings up another issue of what could/should be done about existing Accounts / Posts

pfefferle:
@mattwiebe this is ready to review now

includes/collection/class-followers.php
```diff
@@ -482,9 +477,28 @@ public static function get_faulty_followers( $number = 10 ) {
...
 		foreach ( $posts->get_posts() as $follower ) {
-			$items[] = Follower::from_custom_post_type( $follower ); // phpcs:ignore
+			$items[] = Follower::init_from_cpt( $follower ); // phpcs:ignore
 		}
 
 		return $items;
 	}
+
+	/**
+	 * Undocumented function
```
mattwiebe, line 487: We could surely document it a bit more than this?

includes/model/class-blog-user.php
```diff
@@ -0,0 +1,212 @@
...
+	public static function get_default_username() {
+		$username = \get_option( 'activitypub_blog_user_identifier' );
+
+		if ( $username ) {
+			return $username;
+		}
+
+		// check if domain host has a subdomain
+		$host       = \wp_parse_url( \get_home_url(), \PHP_URL_HOST );
+		$host       = \str_replace( 'www.', '', $host );
```
mattwiebe, line 86: I think it would be better to explicitly use `preg_replace( '/^www\.', '', $host )` here? People put `www` in the strangest-possible places :)

includes/model/class-blog-user.php
```diff
@@ -0,0 +1,212 @@
...
+		$host_parts = \explode( '.', $host );
+
+		if ( \count( $host_parts ) <= 2 && strlen( $host ) <= 15 ) {
+			\update_option( 'activitypub_blog_user_identifier', $host );
```
mattwiebe, line 90: I don't really love setting from a getter - what problem is this solving that just allow the default to be calculated doesn't?

includes/model/class-user.php
```diff
@@ -1,23 +1,261 @@
...
 	 *
-	 * @var int
+	 * @var string
+	 */
+	protected $type = 'Person';
+
+	public static function from_wp_user( $user_id ) {
+		if (
+			is_user_disabled( $user_id ) ||
+			! get_user_by( 'id', $user_id )
```
mattwiebe, line 31: We can probably just leave the `get_user_by` logic to `is_user_disabled`?

includes/rest/class-followers.php
```diff
@@ -5,6 +5,7 @@
 use stdClass;
 use WP_REST_Server;
 use WP_REST_Response;
+use Activitypub\Collection\Users as User_Collection;
```
mattwiebe, line 8: Small nitpick but we should probabl normalize to `UserCollection` or change the below to `Follower_Collection`?

mattwiebe (commented):
Left a few comments, not explicitly approving this since I haven't even tested it yet, only managed to do a first pass on code review so far.

But nothing that screams "blocker" and I'm inclined towards "merge now, fix bugs later"

includes/functions.php
```diff
@@ -302,3 +272,68 @@ function is_activitypub_request() {
...
+			}
+
+			return false;
+		// if the user is any other user, it's enabled if it can publish posts.
+		default:
+			if ( ! \user_can( $user_id, 'publish_posts' ) ) {
+				return true;
+			}
+
+			if ( defined( 'ACTIVITYPUB_DISABLE_USER' ) ) {
```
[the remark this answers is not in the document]
pfefferle: I thought about using the consts instead?!?

includes/model/class-blog-user.php
pfefferle: Oops!

mattwiebe:
I tried running this branch on a live dev server, `wiebeatomic.wpcomstaging.com`. By default it generated `wiebeatomic@wiebeatomic.wpcomstaging.com` which seemed fine but I couldn't follow it as a profile from Mastodon. Then I used the settings to change over to `wiebeatomic.wpcomstaging.com@wiebeatomic.wpcomstaging.com` and the setting appears to have taken but alas I still cannot follow it properly from Mastodon.

`mattwiebe@wiebeatomic.wpcomstaging.com` works fine.

Do I need to set any constants?

mattwiebe (dismissed):
Whatever specific problems arise, including my inability to get it working, should be addressed in follow-up PRs. Let's get this merged and go from there.

PR #344, opened 2023-06-02
Followers Block

mattwiebe:
This adds a Follower block so that sites can display their fediverse followers on the frontend!

Needs testing instructions, but this is mostly done.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


mattwiebe:
I don't love adding `/build` but it's useful for now. (

I also don't love plugins that need to be built before they can be used and I greatly appreciated that this was not one of them and now here I am potentially ruining it!)

mediaformat:
> I don't love adding /build but it's useful for now. 
@mattwiebe I might need to use it for another thing, perhaps this block code could go under `src/followers` or something

src/index.js
```diff
@@ -0,0 +1,10 @@
+import { registerBlockType } from '@wordpress/blocks';
+import edit from './edit';
+const save = () => {
+  return (
+    <div className='fediverse-follows'>
+      <h2>Fediverse Follow Save</h2>
+    </div>
+  );
+};
+registerBlockType( 'automattic/fediverse-follows', { edit, save } );
```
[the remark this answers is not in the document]
mattwiebe: The namespace is encouraged to be of the form `namespace/blockname` so I think that `activitypub/follows` should do it, just muscle memory from doing most things as `a8c/block` for most things.

includes/model/class-follower.php
```diff
@@ -205,25 +215,38 @@ public function from_meta( $meta ) {
...
+	 */
+	public function get_url() {
+		if ( $this->get( 'url' ) ) {
+			return $this->get( 'url' );
+		}
+		$actor = $this->get_actor();
+		// normalize
+		$actor = ltrim( $actor, '@' );
+		$parts = explode( '@', $actor );
+		return sprintf( 'https://%s/@%s', $parts[1], $parts[0] );
```
pfefferle, line 247: This will not work for every platform! If you have a WebFinger-ID and you want to have a URL, you have to use a WebFinger Call to get the Profile-URL.

includes/model/class-follower.php
```diff
@@ -306,11 +329,6 @@ public function get_meta_by( $attribute ) {
 			}
 		}
 
-		// try ActivityPub attribtes
```
pfefferle, line 309: This helps querying meta attributed directly from the meta JSON.

includes/model/class-follower.php
```diff
@@ -335,10 +353,7 @@ public function get_meta() {
 	public function update() {
 		$term = wp_update_term(
 			$this->id,
-			Followers::TAXONOMY,
-			array(
```
pfefferle, line 339: I think this is needed for the process that updates the follower data.

includes/model/class-follower.php
```diff
@@ -355,8 +370,7 @@ public function save() {
 			$this->actor,
 			Followers::TAXONOMY,
 			array(
-				'slug'        => sanitize_title( $this->get_actor() ),
-				'description' => wp_json_encode( $this->get_meta() ),
```
pfefferle, line 359: The description is used to store the complete meta blob. Why should we remove it?

includes/model/class-follower.php
pfefferle: The actor is a URL in most of the cases btw. I am not aware of a platform that sends WebFinger IDs as Actor. So we should check if the Actor is already a URL and if not, get a URL using a WebFinger call.

pfefferle: I just checked the followers of my blog and all 500++ are having URLs as Actors.

pfefferle: I would simply change the code to something like

```php
if ( filter_var( $this->actor, FILTER_VALIDATE_URL ) )
    $this->url = $this->actor;
} else {
    $this->url = Webfinger::resolve( $this->actor );

```

mattwiebe: Thanks! This is what comes from doing a lot of manual testing locally.

includes/model/class-follower.php
mattwiebe: It was creating many problems for me, I don't remember any more what they were. Probably another aspect of testing locally.

includes/rest/class-followers.php
```diff
@@ -97,6 +88,25 @@ function( $item ) {
 		return $response;
 	}
 
+	private static function get_followers( $request ) {
+		$user_id = $request->get_param( 'user_id' );
+		$order = $request->get_param( 'order' );
+		$per_page = $request->get_param( 'per_page' );
+		$page = $request->get_param( 'page' );
+		$offset = ( $page - 1 ) * $per_page;
+		$query = FollowerCollection::get_followers_query( $user_id, $per_page, $offset, array( 'order' => ucwords( $order ) ) );
```
pfefferle, line 97: why not using `get_followers` if you make the transformation to a `Follower` anyways?

mattwiebe: I'm saving a separate call to `Followers::count_followers` by doing it this way, and guaranteeing that the identical query is the basis for the count.

pfefferle: 🤔 hmmm... good point...

pfefferle: 🤔 hmmm... good point...

mattwiebe: We could also possibly make some kind of internal API like

```php
$query = new WP_Query( $args );
$followers = transform_posts_to_followers( $query->posts );
$total = $query->found_posts;
return compact( 'followers', 'total' );
```

to just parse out what we consistently need from the queries

includes/collection/class-followers.php
```diff
@@ -292,14 +292,30 @@ public static function send_follow_response( $actor, $object, $user_id, $followe
...
+	 * @param int    $user_id The ID of the WordPress User.
+	 * @param int    $number  Maximum number of results to return.
+	 * @param int    $page    Page number.
+	 * @param array  $args    The WP_Query arguments.
+	 *
+	 * @return array
+	 *               followers List of `Follower` objects.
+	 *               total     Total number of followers.
+	 */
+	public static function get_followers_raw( $user_id, $number = -1, $page = null, $args = array() ) {
```
pfefferle, line 318: I think `raw` is a bit misleading, I would use a different function name something like `_with_count` or similar.

includes/rest/class-followers.php
```diff
@@ -51,6 +51,10 @@ public static function register_routes() {
 	 * @return WP_REST_Response
 	 */
 	public static function get( $request ) {
+		$context = $request->get_param( 'context' );
+		if ( 'view' === $context ) {
```
pfefferle, line 55: I would change the output of the `orderedItems` to have the ActivityPub headers and to not have too much duplicated content:

```PHP
$json->orderedItems = array_map(
	function( $item ) {
		if ( 'view' === $context ) {
			return $item->to_array();
		} else {
			return $item->get_url();
		}
	},
	Follower_Collection::get_followers( $user->get__id(), 20, $page )
);
```

includes/rest/class-followers.php
```diff
@@ -102,6 +106,23 @@ function( $item ) {
 		return $response;
 	}
 
+	private static function get_followers( $request ) {
```
pfefferle, line 109: see comment above

includes/rest/class-followers.php
```diff
@@ -115,9 +136,33 @@ public static function request_parameters() {
...
+			'validate_callback' => function( $param, $request, $key ) {
+				// despite being an integer, user_id is passed as string.
+				$param = (int) $param;
+				return 0 === $param || user_can( $param, 'publish_posts' );
+			},
+		);
+
+		$params['context'] = array(
+			'type' => 'string',
+			'default' => 'outbox',
```
pfefferle, line 162: `outbox`? What about `simple` and `full`? Both views (the link and the full object) are ActivityPub compatible, I will have a look at the spec maybe there is something like an output filter defined.

includes/rest/class-followers.php
```diff
@@ -115,9 +136,33 @@ public static function request_parameters() {
...
+			'validate_callback' => function( $param, $request, $key ) {
+				return in_array( $param, array( 'outbox', 'view' ), true );
```
pfefferle, line 164: shouldn't we use an enum here to, instead of the validator callback?

includes/rest/class-followers.php
```diff
@@ -115,9 +136,33 @@ public static function request_parameters() {
...
+		$params['order'] = array(
+			'type'    => 'string',
+			'default' => 'desc',
+			'enum'    => array( 'asc', 'desc' ),
+		);
+
 		$params['user_id'] = array(
 			'required' => true,
 			'type' => 'string',
+			'validate_callback' => function( $param, $request, $key ) {
```
pfefferle, line 153: this will be implicitly checked with

```PHP
$user    = User_Collection::get_by_various( $user_id );

if ( is_wp_error( $user ) ) {
	return $user;
}
```

I would remove the check, because otherwise we might ran into inconsistencies.

includes/rest/class-followers.php
```diff
@@ -115,9 +136,33 @@ public static function request_parameters() {
 			'default' => 1,
 		);
 
+		$params['per_page'] = array(
+			'type' => 'integer',
+			'default' => 10,
```
pfefferle, line 141: can we use "20" as default, to not break backwards compatibility?

includes/rest/class-followers.php
mattwiebe: Hmm let's see if I can make that work, I like the idea of it

includes/rest/class-followers.php
mattwiebe: Yup good catch

includes/rest/class-followers.php
mattwiebe: Yup seems fine, was just aligning it with the default of 10 from the widget

src/followers/followers.js
```diff
@@ -0,0 +1,112 @@
...
+	const setPage = passedSetPage || setLocalPage;
+	const prevLabel = (
+		<>
+			<span class="wp-block-query-pagination-previous-arrow is-arrow-arrow" aria-hidden="true">
+				{
+					/* translators: arrow for previous followers link */
+					__( '←', 'activitypub' )
+				}
+			</span>
+			{ __( 'Less', 'activitypub' ) }
```
jeherve, line 50: Maybe this (and the below) could be combined into a single translatable string, with `createInterpolateElement` from `@wordpress/element`. It would make things easier in RTL languages, for example.

Something like this (not tested):

```diff
diff --git a/src/followers/followers.js b/src/followers/followers.js
index aee1c3c..b756f8f 100644
--- a/src/followers/followers.js
+++ b/src/followers/followers.js
@@ -1,6 +1,7 @@
 import { useState, useEffect } from 'react';
 import apiFetch from '@wordpress/api-fetch';
 import { addQueryArgs } from '@wordpress/url';
+import { createInterpolateElement } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { Pagination } from './pagination';
 import { ExternalLink } from '@wordpress/components';
@@ -39,27 +40,19 @@ export function Followers( {
 	const [ localPage, setLocalPage ] = usePage();
 	const page = passedPage || localPage;
 	const setPage = passedSetPage || setLocalPage;
-	const prevLabel = (
-		<>
-			<span class="wp-block-query-pagination-previous-arrow is-arrow-arrow" aria-hidden="true">
-				{
-					/* translators: arrow for previous followers link */
-					__( '←', 'activitypub' )
-				}
-			</span>
-			{ __( 'Less', 'activitypub' ) }
-		</>
+	const prevLabel = createInterpolateElement(
+		/* translators: arrow for previous followers link */
+		__( '<span>←</span> Less', 'activitypub' ),
+		{
+			span: <span class="wp-block-query-pagination-previous-arrow is-arrow-arrow" aria-hidden="true" />,
+		}
 	);
-	const nextLabel = (
-		<>
-			{ __( 'More', 'activitypub' ) }
-			<span class="wp-block-query-pagination-next-arrow is-arrow-arrow" aria-hidden="true">
-				{
-					/* translators: arrow for next followers link */
-					__( '→', 'activitypub' )
-				}
-			</span>
-		</>
+	const nextLabel = createInterpolateElement(
+		/* translators: arrow for next followers link */
+		__( 'More <span>→</span>', 'activitypub' ),
+		{
+			span: <span class="wp-block-query-pagination-next-arrow is-arrow-arrow" aria-hidden="true" />,
+		}
 	);
 
 	useEffect( () => {
@@ -109,4 +102,4 @@ function Follower( { name, icon, url, preferredUsername } ) {
 			</span>
 		</ExternalLink>
 	)
-}
\ No newline at end of file
+}

```

mattwiebe: TIL about `createInterpolateElement` - thanks so much for this, great point about RTL and just keeping the context together.

pfefferle (dismissed):
oh, you are right!

pfefferle (approved):
😍

PR #365, opened 2023-07-20
fix some issues and re-add "ACTIVITYPUB_SINGLE_USER_MODE" const

pfefferle:
Fixes some small issues and glitches

## Proposed changes:

* re-add "ACTIVITYPUB_SINGLE_USER_MODE" const
* fix some issues that happens when single-user-mode is enabled


tests/test-class-activitypub-activity-dispatcher.php
```diff
@@ -95,6 +95,13 @@ function( $mentions ) {
 	}
 
 	public function test_dispatch_announce() {
+		add_filter(
+			'activitypub_is_user_type_disabled',
+			function( $value ) {
```
mattwiebe, line 100: For this you can use one of the handy internal `__return` aliases:

```
add_filter( 'activitypub_is_user_type_disabled', '__return_false' );
```

mattwiebe (approved):
🚢

PR #369, opened 2023-07-24
Add default values for name, preferredUsername and icon

pfefferle:
Better default values, that will be used by all representations (Objects, Arrays and JSON (API)).

Adding the defaults in the `to_array` method would produce different outputs when using JSON/arrays versus using the Follower-Object and it's getters.

Because the `to_array` method uses the Objects getters, this change will affect all different usage possibilities.


mattwiebe (approved):
Looks great to me! Appreciate how this keeps the compat layer in the getters as well.

PR #370, opened 2023-07-26
Compat: avoid PHP notices when using PHP 8.2

jeherve:
## Proposed changes:

This PR should just avoid a PHP notice when using PHP 8.2:

```
PHP Deprecated:  Creation of dynamic property Activitypub\Transformer\Post::$attachments is deprecated in /var/www/html/wp-content/plugins/activitypub/includes/transformer/class-post.php on line 249
```

### Other information:

## Testing instructions:

Not much to test here; if you have a local installation of WordPress using PHP 8.2, you should no longer see the deprecation notice when trying to access a post in an ActivityPub request.

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -29,13 +29,6 @@ class Post {
 	protected $wp_post;
 
 	/**
-	 * Attachment images, used in the content.
-	 *
-	 * @var array
-	 */
-	protected $attachments = array();
-
-	/**
 	 * The Allowed Tags, used in the content.
 	 *
 	 * @var array
@@ -252,8 +245,6 @@ class Post {
 				$images[] = $image;
 			}
 		}
-
-		$this->attachments = $images;
 
 		return $images;
 	}
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -29,6 +29,13 @@ class Post {
 	protected $wp_post;
 
 	/**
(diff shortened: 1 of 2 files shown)


pfefferle:
Good catch! But I would prefer to simply remove `$this->attachments = $images;` in the `get_attachment` function. This seems to be a Copy&Paste issue that comes from the old Post-Model. In the actual case, there seems to be no need to have it as a Object-Attribute.

jeherve:
I'm not at all familiar with the codebase; do you mean I should get rid of it entirely?

```diff
diff --git a/includes/transformer/class-post.php b/includes/transformer/class-post.php
index 6f1fed3..e9453dc 100644
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -28,13 +28,6 @@ class Post {
 	 */
 	protected $wp_post;
 
-	/**
-	 * Attachment images, used in the content.
-	 *
-	 * @var array
-	 */
-	protected $attachments = array();
-
 	/**
 	 * The Allowed Tags, used in the content.
 	 *
@@ -253,8 +246,6 @@ class Post {
 			}
 		}
 
-		$this->attachments = $images;
-
 		return $images;
 	}
 

```

pfefferle:
Yes, not needed anymore.

jeherve:
👍 Done.

pfefferle (approved):
thanks @jeherve :)

PR #384, opened 2023-07-31
Add/small improvements

pfefferle:
Small improvements.

To discuss: Can we equate `@notiz.blog@notiz.blog` with `@notizblog@notiz.blog`?

### DIFF
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -28,7 +28,7 @@ class Activitypub {
 		\add_action( 'wp_trash_post', array( self::class, 'trash_post' ), 1 );
 		\add_action( 'untrash_post', array( self::class, 'untrash_post' ), 1 );
 
-		self::add_rewrite_rules();
+		\add_action( 'init', array( self::class, 'add_rewrite_rules' ), 11 );
 
 		\add_action( 'after_setup_theme', array( self::class, 'theme_compat' ), 99 );
 	}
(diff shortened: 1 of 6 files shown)


jeherve:
> To discuss: Can we equate `@notiz.blog@notiz.blog` with `@notizblog@notiz.blog`?

What are the requirements for ActivityPub / Activity Streams? Is there any precedent conversation about this, by any chance?

mattwiebe:
What problem is this fixing?

pfefferle:
GoToSocial seems to not support dots in pathes

pfefferle:
@mattwiebe @jeherve if this is too tricky, I will remove it, to discuss it separately.

But there are some other improvements that should be merged!

includes/class-activitypub.php
```diff
@@ -28,7 +28,7 @@ public static function init() {
 		\add_action( 'wp_trash_post', array( self::class, 'trash_post' ), 1 );
 		\add_action( 'untrash_post', array( self::class, 'untrash_post' ), 1 );
 
-		\add_action( 'init', array( self::class, 'add_rewrite_rules' ), 11 );
+		self::add_rewrite_rules();
```
mattwiebe, line 31: I would prefer to keep this as an action because I need to unhook it on dotcom due to our different rewrite requirements.

Does it cause problems elsewhere?

mattwiebe (dismissed):
Approved, with a question

includes/class-activitypub.php
pfefferle: done

mattwiebe (approved):
Thanks for the change!

PR #386, opened 2023-08-01
re-add post model

pfefferle:
for a better backwards compatibility

@akirk can you check the code with the friends plugin?


akirk (approved):
I can confirm that this no longer throws an error with Friends version before [2.7.2](https://github.com/akirk/friends/releases/tag/2.7.2) (which contained the fix to not blindly require this class).

PR #389, opened 2023-08-04
Update comment type

mattwiebe:
It's useful to have more than just metadata to identify our comments, and a useful cue for spam systems to validate replies.

This introduces a new helper function, `get_reply_comment_type` that can also be filtered. A bonus was that it simplified our logic in the avatar filtering.

Will we need to introduce some form of upgrade functionality so that old comments get the new comment_type ?


mattwiebe:
We decided against this.

PR #391, opened 2023-08-09
Disable the akismet nonce when processing our comments

mattwiebe:
Akismet tries to add a nonce to harden spam submissions. Since these comments aren't submitted through the comment box, we should disable that.


pfefferle:
I am fine to merge this for the moment, but I would suggest to have an `integration` class for akismet at one day, because not every ap-plugin user is also an akismet user.

Here is an example for a buddypress "integration": https://github.com/Automattic/wordpress-activitypub/tree/master/integration

PR #356, opened 2023-07-04
started remote-follow endpoint

pfefferle:
Provides an Endpoint for starting the remote-follow process.

/cc @nuriapenya 

## Proposed changes:

* This is an endpoint to get the URL to remote-follow someone.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Go to '..'
*

PR #392, opened 2023-08-10
remote-follow endpoint

pfefferle:
Added in favour of #356 

The API works like: `https://example.org/wp-json/activitypub/1.0/users/1/remote-follow?resource=pfefferle@mastodon.social` and returns `{"url":"https:\/\/mastodon.social\/authorize_interaction?uri=test@localhost"}`

## Proposed changes:

* Added functions to get Remote-Follow endpoint from Remote-Server
* Added API endpoint to query that endpoint

### DIFF
--- a/includes/class-webfinger.php
+++ b/includes/class-webfinger.php
@@ -150,7 +150,7 @@ class Webfinger {
 	public static function get_data( $identifier, $host ) {
 		$webfinger_url = 'https://' . $host . '/.well-known/webfinger?resource=' . rawurlencode( $identifier );
 
-		$response = wp_remote_get(
+		$response = wp_safe_remote_get(
 			$webfinger_url,
 			array(
 				'headers' => array( 'Accept' => 'application/jrd+json' ),
--- a/includes/class-webfinger.php
+++ b/includes/class-webfinger.php
@@ -62,7 +62,7 @@ class Webfinger {
 		$response = \wp_remote_get(
 			$url,
 			array(
-				'headers' => array( 'Accept' => 'application/activity+json' ),
+				'headers' => array( 'Accept' => 'application/jrd+json' ),
 				'redirection' => 0,
 				'timeout' => 2,
 			)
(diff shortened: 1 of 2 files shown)


includes/class-webfinger.php
```diff
@@ -94,4 +94,110 @@ public static function resolve( $resource ) {
...
+	 * @param string $identifier The Identifier: <identifier>@<host>
+	 * @param string $host       The Host: <identifier>@<host>
+	 *
+	 * @return WP_Error|array Error reaction or array with
+	 *                        identifier and host as values
+	 */
+	public static function get_data( $identifier, $host ) {
+		$webfinger_url = 'https://' . $host . '/.well-known/webfinger?resource=' . rawurlencode( $identifier );
+
+		$response = wp_remote_get(
```
mattwiebe, line 153: Especially because this is pinging remote servers based on user-provided information, let's use `wp_safe_remote_get` here

includes/rest/class-users.php
```diff
@@ -40,6 +42,25 @@ public static function register_routes() {
 				),
 			)
 		);
+
+		\register_rest_route(
+			ACTIVITYPUB_REST_NAMESPACE,
+			'/users/(?P<user_id>[\w\-\.]+)/remote-follow',
```
mattwiebe, line 48: Do we really need a `user_id` in this URL? What is it doing for us?

(looks at the code below)

Oh. :) That. Carry on!

includes/rest/class-users.php
```diff
@@ -80,6 +101,33 @@ public static function get( $request ) {
...
+		$resource = $request->get_param( 'resource' );
+		$user_id  = $request->get_param( 'user_id' );
+
+		$template = Webfinger::get_remote_follow_endpoint( $resource );
+
+		if ( is_wp_error( $template ) ) {
+			return $template;
+		}
+
+		$resource = Webfinger::get_user_resource( $user_id );
```
mattwiebe, line 122: can we carry on with this resource? what about calls to a nonexistent/disabled `user_id` ?

mattwiebe (commented):
This will be really good to have for #395 

I can address the issues I raised here if you're already AFK!

includes/rest/class-users.php
pfefferle: we need the ID, because we need the identifier of that user to be passed as param for the remote-follow endpoint.

Example: `https://mastodon.social/authorize_interaction?uri=username.wordpress.com@username.wordpress.com`

includes/rest/class-users.php
pfefferle: good point! missed to add the check :(

pfefferle:
now I am AFK :)

mattwiebe (approved):
looks great! I'll merge it.

PR #395, opened 2023-08-11
Add "Follow Me" block

mattwiebe:
This will add a "Follow Me" block that will produce a dialog for educating and enabling fab fediverse following!

Screenshot 2023-08-11 at 12 32 23
Screenshot 2023-08-11 at 12 33 44

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


mattwiebe:
I'd like to figure out what's going on with those phpcs CI builds, I can't replicate those errors locally at all with `composer run lint` which is theoretically running the same command

mattwiebe:
> I can't replicate those errors locally at all 

I just needed to `rm -rf vendor && composer install` - something had gone wrong in my local install. Now added a `composer run lint:fix` command to call `phpcbf`

src/follow-me/follow-me.js
```diff
@@ -0,0 +1,153 @@
...
+					<Button ref={ ref }>
+						<Icon icon={ buttonIcon } />
+						{ __( 'Copy', 'activitypub' ) }
+					</Button>
+				</div>
+			</div>
+			<div className="apmfd__section">
+				<h4>{ __( 'Different Server', 'activitypub' ) }</h4>
+				<div className="apfmd-description">
+					{ __( 'Give us your Username/URL and we will start the process for you. (E.g. https://example.com/username or username@example.com)', 'activitypub' ) }
```
pfefferle, line 119: Use URL/Username instead of Username/URL, to match the order of the examples (E.g ...)

```suggestion
					{ __( 'Give us your URL or Username and we will start the process for you. (E.g. https://example.com/username or username@example.com)', 'activitypub' ) }
```

mattwiebe: I actually decided not to name them at all in the end, but otherwise this was a good catch :)

mattwiebe:
This is pretty close to ready now?

We are abusing the Link color to set the button color. Otherwise there's no really good way to set it.

pfefferle (dismissed):
:)

pfefferle:
Can you also update the changelog?!?

mattwiebe:
https://github.com/Automattic/wordpress-activitypub/pull/395/commits/dce6daa30cb826b4c4c52be184e9c0df78833f75 - it looks a bit busier because I picked a consistent "no periods" style. :)

PR #405, opened 2023-08-26
restore compatibility with php7

mexon:
## Proposed changes:
No functional changes, just replace PHP8-only `str_starts_with` and `str_contains` calls with PHP7-compatible versions.

### DIFF
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -156,7 +156,8 @@ class Signature {
 	public static function verify_http_signature( $request ) {
 		if ( is_object( $request ) ) { // REST Request object
 			// check if route starts with "index.php"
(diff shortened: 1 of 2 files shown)


mexon:
I don't really get what's up with the testing here.  I don't know how to run unit tests locally with a specific version.  But generally - how can unit tests on PHP 7.X be passing if the code uses `str_starts_with`?  Certainly I get errors about these function calls when I try to run on a Debian 11 image with PHP 7.4.  Something is fishy here - are the tests really running what they claim to run?

pfefferle:
WordPress provides backwards compatibility for that 😊

mexon:
@pfefferle Ah, I see.  Those compatibility functions were introduced in WP 5.9.0.  So that would mean that instead of this change, the "Requires at least" should be bumped, right?

PR #410, opened 2023-08-31
Fix some signature and application user issues

pfefferle:
Fixes some signature issues and the functionality of the application user.


includes/model/class-blog-user.php
```diff
@@ -136,6 +136,11 @@ public function get_icon() {
 		}
 
 		$image = wp_get_attachment_image_src( $icon_id, 'full' );
+
+		if ( ! $image ) {
```
mattwiebe, line 140: I'm going to push a change to remove this as I'll be dealing with the same thing while providing a default icon in #412

PR #411, opened 2023-09-01
fix Secops issues

pfefferle:
Fix issues detected by Secops


includes/rest/class-inbox.php
```diff
@@ -331,61 +331,6 @@ public static function shared_inbox_post_parameters() {
 		return $params;
 	}
 
-	/**
-	 * Handles "Reaction" requests
-	 *
-	 * @param  array $object  The activity-object
-	 * @param  int   $user_id The id of the local blog-user
-	 */
-	public static function handle_reaction( $object, $user_id ) {
```
mattwiebe, line 340: Are we deleting this because it wasn't working anyway? Or I guess we'd already removed the hook? I think we mention "reactions" in the readme too, we should probably remove those references as well.

tests/test-class-db-activitypub-followers.php
```diff
@@ -268,6 +294,30 @@ public function test_add_duplicate_follower() {
 		$this->assertCount( 1, $meta );
 	}
 
+	public function test_migration() {
```
mattwiebe, line 297: Thanks for testing!

mattwiebe (approved):
This looks good to me!

includes/rest/class-inbox.php
pfefferle: It was not in use and I do not know if it really works. I think we should give the comments feature a bit more love in one of the next major versions.

PR #412, opened 2023-09-01
Add default blog user icon

mattwiebe:
This will add a default WP icon for the Blog User.


pfefferle:
Isn't there a core file we could use?

mattwiebe:
> Isn't there a core file we could use or what about the MP.jpg? https://github.com/Automattic/wordpress-activitypub/blob/master/assets/img/mp.jpg

My brain said "nothing in Core" but that isn't quite true, there are a couple of WP icons [here](https://github.com/WordPress/WordPress/tree/master/wp-includes/images) but they are very small.

As for re-using mystery-person, I guess I thought that a Blog User isn't a person? Plus I want to market WP in the fediverse a bit.

pfefferle:
What about https://github.com/WordPress/WordPress/blob/master/wp-includes/images/w-logo-blue-white-bg.png ?? Do we use it bigger than 80x80?

If we use our own, it should have a white background, otherwise it might look crappy on other networks. Had this problem already.

mattwiebe:
80x80 is very small and will look bad. but sure we can add a white bg and make it a bit smaller

pfefferle:
No, it's not about making it smaller, I was only thinking about re-using the core logos. If we use our own, we can keep the size, but should make the background white.

mattwiebe:
Ah right gotcha. I've downsized the WP logo and added a white bg. Also upsized the Mystery Person to be a bit less blurry

pfefferle (dismissed):
👍

pfefferle:
Nice!

pfefferle (approved):

PR #416, opened 2023-09-07
Store keypairs as options keyed to user IDs.

mattwiebe:
A problem we've run into for WP Multisite is that there is theoretical key leakage for a given user who is a member of more than one network blog.

Imagine `example.com` as a WP Multisite install. Two users, `foo` (ID 1) and `bar` (ID 2), have a shared site called `foobar.example.com`, while `foo` has a solo site, `foo.example.com`. Since usermeta is shared across the whole network, `bar` could be sneaky and retrieve `foo`'s private key from their shared site. `bar` could then use that private key to sign private messages on the fediverse _as though it was coming from `@foo@foo.example.com`_ rather than `foo@foobar.example.com`, or any other site on that network on which `foo` is a member.

What this patch does is to migrate keypair storage to be based on blog options instead, which will keep each keypair functioning only within a single blog context per user in the network. `foo` will have blog options of `activitypub_keypair_for_1` on both sites but it will be unique to each site.

There was no reason to not just standardize on this so I've made the Blog and Application Users use this style as well for consistency. They will receive `activitypub_keypair_for_0` and `activitypub_keypair_for_-1`, respectively.

This is a speculative PR, put up before I went to bed while the idea was still in my head. I have not tested it other than in my head, and that it doesn't throw fatals. I'm sure it needs unit tests.

### DIFF
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -4,7 +4,6 @@ namespace Activitypub;
 use WP_Error;
 use DateTime;
 use DateTimeZone;
-use Activitypub\Model\User;
 use Activitypub\Collection\Users;
 
 /**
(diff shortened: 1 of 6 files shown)


pfefferle:
@mattwiebe thanks for preparing the PR!

I thought it might be a good idea to remove all signature relevant informations/functions from the User objects (except the public-key output for the ActivityPub-Profile) to be sure that the private key will never unintentionally displayed by a `to_json` or `to_array` function.

I also added the legacy check as one option to the `generate_key_pair_for` method, because it is kind of a generation of the keys using the old DB persistence.

Let me know what you think.

/cc @ebinnion

mattwiebe:
Nice refactor, gonna test this out on wpcom a bit

includes/class-signature.php
```diff
@@ -49,32 +40,51 @@ public static function get_public_key( $user_id, $force = false ) {
 	 *
 	 * @return mixed The private key.
 	 */
-	public static function get_private_key( $user_id, $force = false ) {
+	public static function get_private_key_for( $user_id, $force = false ) {
 		if ( $force ) {
```
mattwiebe, line 44: I think we can get rid of the `$force` argument here, since `get_keypair_for` is already generating the key if one isn't found.

mattwiebe (commented):
This tested fine on dotcom, the same key continued to be used with or without this PR

(I can't approve because I started the PR but consider it so)

pfefferle (approved):
Thanks @mattwiebe

PR #423, opened 2023-09-13
Follow redirects in class-webfinger.php

### LINKED ISSUE #422: Incorrect health message if author URl redirected
I'm receiving this error because I use a subdirectory:

Screenshot of Site health error.

I'm using `.htaccess` to redirect to `/blog/.well-known/...`

This can be validated by running:

`curl -L "https://shkspr.mobi/.well-known/webfinger?resource=acct:blog@shkspr.mobi" -H 'Accept: application/jrd+json'`

After following redirects, it works.

I believe this is caused by:

https://github.com/Automattic/wordpress-activitypub/blob/42d525c904e2fcf2e275614f81cdde9526f6873c/includes/class-webfinger.php#L66

The [default number of allowed redirects is 5](https://developer.wordpress.org/reference/functions/wp_remote_get/#comment-1906)

I'm going to propose a PR for:

```php
$response = \wp_remote_get(
	$url,
	array(
		'headers' => array( 'Accept' => 'application/jrd+json' ),
		'redirection' => 2,
		'timeout' => 2,
	)
);
```

That fixes the issue for me.

edent:
Fixes #422

## Proposed changes:

* Authors who use a subdirectory have to redirect `/.well-known` - this allows the health check to pass.

### Other information:

## Testing instructions:

* Create a blog at example.com/blog/
* Install the plugin
* Configure .htaccess as described to redirect `/.well-known/` to `/blog/.well-known/`
* Run the WordPress health check
* Note that it fails.
* Apply this patch.
* Success!

### DIFF
--- a/includes/class-webfinger.php
+++ b/includes/class-webfinger.php
@@ -63,7 +63,7 @@ class Webfinger {
 			$url,
 			array(
 				'headers' => array( 'Accept' => 'application/jrd+json' ),
-				'redirection' => 0,
+				'redirection' => 2,
 				'timeout' => 2,
 			)
 		);


pfefferle (approved):
Thanks @edent

PR #438, opened 2023-09-15
add path to route

### LINKED ISSUE #421: Follower requests never complete - ModSecurity off
I've successfully set up `@blog@shkspr.mobi` - it appears on the Fediverse. But people can't follow it.

ModSecurity is set to "off" in my cPanel. As per #341 and https://blog.rac.me.uk/2022/11/10/activitypub-for-wordpress-how-to-fix-modsecurity-to-make-it-work/#adding-the-new-mod-security-rule that should fix it. But it doesn't.

Screenshot showing ModSecurity off.

Are there any tests I can run to see what the problem is?

Site Health shows:
>  Author URL is not accessible ActivityPub
> Your author URL https://shkspr.mobi/blog/author/admin/ does not return valid JSON for application/activity+json. Please check if your hosting supports alternate Accept headers.

However, I am not using `admin` as the author; I've disabled the author in the settings and have a generic blog profile ID.

pfefferle:
fix #421

## Proposed changes:

* The current code ignores the 'path' when generating the signature string. This PR should fix the issue.

### Other information:

## Testing instructions:

* Install WordPress in a sub directory
* Install ActivityPub
* Try to follow your blog from Mastodon or a similar platform

### DIFF
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -238,6 +238,15 @@ class Signature {
 			} else {
 				$route = '/' . rest_get_url_prefix() . '/' . ltrim( $request->get_route(), '/' );
 			}
+
+			// fix route for subdirectory installs
+			$path = wp_parse_url( get_home_url(), PHP_URL_PATH );
+			$path = trim( $path, '/' );
+
+			if ( $path ) {
+				$route = '/' . $path . $route;
+			}
+
 			$headers = $request->get_headers();
 			$headers['(request-target)'][0] = strtolower( $request->get_method() ) . ' ' . $route;
 		} else {
--- a/README.md
+++ b/README.md
@@ -107,9 +107,10 @@ Project maintained on GitHub at [automattic/wordpress-activitypub](https://githu
 
 ### Next ###
 
-* Compatibility: allow deactivating blocks registered by the plugin.
-* Compatibility: avoid Fatal Errors when using ClassicPress.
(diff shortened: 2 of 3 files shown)


pfefferle:
@edent can you verify that this PR fixes your issue?

edent:
🙌🙌🙌🙌🙌

I can now follow `@blog@shkspr.mobi`!

Will check in an hour or so if it pushes out the scheduled post to subscriber.

pfefferle:
@mattwiebe might this break something on WP.com?

includes/class-signature.php
```diff
@@ -238,6 +238,15 @@ public static function verify_http_signature( $request ) {
 			} else {
 				$route = '/' . rest_get_url_prefix() . '/' . ltrim( $request->get_route(), '/' );
 			}
+
+			// fix route for subdirectory installs
+			$path = wp_parse_url( get_home_url(), PHP_URL_PATH );
+			$path = trim( $path, '/' );
+
+			if ( $path ) {
```
mattwiebe, line 246: I'd maybe check instead for `! empty( $path )` here instead, although this would probably be ok.

mattwiebe (approved):
> wp_parse_url( get_home_url(), PHP_URL_PATH ); should be empty on public-api.wordpress.com/!?!

This shouldn't cause any trouble on dotcom or any non-subdirectory-based install. Looks good.

PR #441, opened 2023-09-16
More Group meta-data to play nicely with existing platforms

pfefferle:
I had a long discussion with some friendi.ca users on Mastodon about Groups and to not break compatibility of existing platforms.

This PR adds some more meta data, to be more compatible with Kbin, Lemmy and Friendi.ca.

## Proposed changes:

* Adds `postingRestrictedToMods = true` for Group users
* Adds `moderators` endpoint

### Other information:


includes/rest/class-collection.php
```diff
@@ -56,6 +57,19 @@ public static function register_routes() {
 				),
 			)
 		);
+
+		\register_rest_route(
+			ACTIVITYPUB_REST_NAMESPACE,
+			'/users/(?P<user_id>[\w\-\.]+)/collections/moderators',
```
mattwiebe, line 63: Do we really even need to bother with the user ID here? Since it's only ever for the Blog User, we could just use `/collections/moderators` or even just `/moderators` ?

mattwiebe (dismissed):
The endpoint works and it seems like this will help compat - did have one question but it looks good to me.

includes/rest/class-collection.php
pfefferle: I will think about that... good point!

PR #444, opened 2023-09-18
better handling when data is missing

pfefferle:
Fixes https://wordpress.org/support/topic/php-error-when-viewing-ap-followers-via-json/

## Proposed changes:

* Better error handling when data is missing

### Other information:


mattwiebe (approved):
let's keep getting more reliable!

PR #448, opened 2023-09-19
WP_Error: add status message if it might be returned by API

pfefferle:
It seems that version 1.0.0 produces a higher amount of http 500 errors. It seems that this is the case if you return an WP_Error without defining the status code.

## Proposed changes:

* Add status code to most of the WP_Error messages

### Other information:

## Testing instructions:

* Go to '..'
*


mattwiebe (approved):
Looks good to me!

PR #453, opened 2023-09-20
Follow Me: truncate long blog titles and handles

### LINKED ISSUE #427: Follow Me Block: too big for sidebar
First up thanks a lot for working on this plugin. this is a massive help in bringing wordpress to fediverse. 

I am using it for past 2-3 months and its been a smooth ride. New features are awesome and kudoes for the massive work on the upgrade.

Now for my greedy demands.

For http://blog.anantshri.info/ i have a big title and the userid also becomes big. follow block right now adds full title in large font and then adds full id below it. this squeezes the follow button and makes the whole widget ugly.

image

The same scenario plays out in mobile view also. 

I have tried reducing font size assuming that was the culprit even then the widget didn't worked out. 

Couple of suggestion.

1. A re-designed widget may be.
2. A cutomizable block which allows me to change the text could also work out. 
3. A simple Follow me on Fediverse with a logo of fediverse or simmilar might also be a good option.

mattwiebe:
Also add typography control

Fixes #427

## Proposed changes:
This will truncate the text of the blog title and fediverse handle:
Screenshot 2023-09-20 at 12 21 08

### Other information:

## Testing instructions:
Try using the widget in a narrow column, and/or using a long blog title.


pfefferle:
We should also apply this to the username in lightbox to prevent overflow when opened on a mobile for example.

mattwiebe:
> We should also apply this to the username in lightbox to prevent overflow when opened on a mobile for example.

I don't think it's needed, there. It can just flow onto multiple lines, like on my local site.

Screenshot 2023-09-21 at 12 28 33

mattwiebe:
> Do we have to re-generate the JS files, because of these changes

Nope, `block.json` is meant to be totally static, it doesn't change anything about the JS, which we didn't change here.

PR #454, opened 2023-09-21
Remove html_entity_decode to prevent encoding issues

### LINKED ISSUE #445: Incorrectly escaped content when using Markdown?
### Quick summary

My [blog post](https://shkspr.mobi/blog/2023/09/false-friends-html-elements-which-are-also-top-level-domains/) contains the following Markdown:

```
For example `<footer>` to display a footer, or `<ol>` for an ordered list.
```

On the HTML page, those are rendered as `<code>&lt;ol&gt;</code>`

Looking at [the outbox](https://shkspr.mobi/blog/wp-json/activitypub/1.0/users/0/outbox), I see:
```
For example <code><footer><\/code> to display a footer, or <code><ol><\/code> for an ordered list.<\/p>
```

But when I view it on Mastodon, it is rendered as:
```
<p>For example <code></code></p>  to display a footer, or <code></code><ol><code></code> for an ordered list.<p></p>
```

### Steps to reproduce

1. Use Markdown with a single \` surrounding HTML elements
2. Publish post

Elements wrapped in `<code>` blocks should be properly escaped.

### What actually happened

The HTML was garbled when viewed on Mastodon.

### Impact

One

### Available workarounds?

### Logs or notes

I use a custom "post content" of:

```
<strong>[ap_title]</strong>
<a href="[ap_permalink]">[ap_permalink]</a>

[ap_content]

<a href="[ap_permalink]">[ap_permalink]</a>

[ap_hashtags]
```

edent:
Fixes #445

## Proposed changes:

Previously, content like `&lt;` was incorrectly converted to `<` - this meant that HTML elements would be passed which wouldn't render correctly.

By removing this, I'm able to post content with encoded entities.

I've tested this on content which includes MarkDown, HTML, encoded entities, unencoded entities, etc.

### Other information:

## Testing instructions:

* Write a post which includes the literal text `The new &lt;search&gt; element is <em>&quot;gr&euml;át&quot;</em> - It makes me &#x1F970; and &#128573; - you should spend your &pound;s on it.`
* Under the old version, the `<search>` element would be inserted and the rendering of the post would break.
* After this patch, the entities are unchanged and the post displays as 

> The new &lt;search&gt; element is *"grëát"* - It makes me 🥰 and 😽 - you should spend your £s on it.


pfefferle:
I thinks we should add some tests here, to not break something else! I will also check why this was added in the first place.

PR #458, opened 2023-09-21
Blocks: ensure that only a valid user can be selected

### LINKED ISSUE #440: Follow me widget does not update to selected author
### Quick summary

If the blog has only one author and if the blog author is disabled, the widget will only show the default "hello dolly" dummy.

### Steps to reproduce

1. Disable Blog-User and enable Authors
2. Use widget.

Use the only available author as default.

### What actually happened

Dummy user "hello dolly" shows

### Impact

All

### Available workarounds?

Yes, easy to implement

### Logs or notes

Screenshot 2023-09-15 at 20 47 50

mattwiebe:
Fixes #440

## Proposed changes:

The default user is the Blog User, but when that is disabled, we are unable to switch over to an Author User when there is only one option.

It would be nice to just adjust the default value but I don't know how to, or if it's even possible, when that's set statically in `block.json`.

So instead we simply set `selectedUser` to the first available user in the list, when it's set to a value that doesn't exist in `usersOptions`.

### Other information:

## Testing instructions:

- have only one user on the blog
- remove all "follow me" widgets from the test site
- disable blog-user and enable authors
- add "follow me" or "followers" block

Before this PR, you would not be able to select the single user from the list and it would default to the disabled Blog User.

With this PR, the single author (or, even with multiple authors, the first in the list) will be selected by default when inserting the new widget.


pfefferle (approved):
this fixes the issue! thanks!

PR #460, opened 2023-09-22
Activity: try to parse image IDs using blocks

### LINKED ISSUE #447: Image removed from post is still kept in json attachment list
### Quick summary

I was working on a blogpost. added an image and removed it to replace it with new version. When the post was published the old image is still referenced as attachment.

### Steps to reproduce

1. Create a blogpost
2. add an image
3. remove that image
4. add another image
publish the page

json contains both the old image and new image reference.

only new image as attachment

### What actually happened

old image as well as new image in attachment

image

**Github-Actions-1.png**

Only when i deleted the image from media library did that image went away from the attachment

image

Double checked both html of blogpost and feed xml they were not referencing to **Github-Actions-1.png** only to **Github-Actions-1-1.png**

### Impact

One

### Available workarounds?

Yes, easy to implement

### Logs or notes

_No response_

mattwiebe:
This will prevent the issue of attaching images that don't were uploaded to the post but not used in the post

The post needs to be using blocks to get the introspection required.

Fixes #447

## Proposed changes:
Use the blocks parser to fetch actually-used images from the post.

### Other information:

## Testing instructions:
With this PR active, upload two images to a post. Only use one of them. Before this, both images would be attached to the Activity as attachments. After this, only the image that gets inserted in the post will be used.

This has the added benefit of being able to use images in a post that weren't uploaded against it, eg reusing previously-used images.

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -9,6 +9,7 @@ use Activitypub\Activity\Base_Object;
 use function Activitypub\esc_hashtag;
 use function Activitypub\is_single_user;
 use function Activitypub\get_rest_url_by_path;
+use function Activitypub\site_supports_blocks;
 
 /**
  * WordPress Post Transformer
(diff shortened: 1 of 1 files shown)


pfefferle (approved):
thanks a lot! this is a so simple and perfect solution (if you know that something like that is even possible), I love it!

PR #471, opened 2023-09-25
check if blog-user collides with a username

### LINKED ISSUE #470: New posts are not being published on Mastodon after upgrade to 1.0.1
### Quick summary

I upgraded to 1.0.1 earlier today and since then the posts on the blog are not being published to Mastodon. 
Blog: https://www.suramya.com/blog/
Mastodon : https://mastodon.social/@suramya@www.suramya.com

As you can see the last post pushed to Mastodon was 5 days ago, even though I posted 2 posts today. I didn't see any error messages and the only change made was to upgrade the plugin version.

### Steps to reproduce

1. Log into suramya.com/blog
2. Make new post 
3. Submit

The post should be available at https://mastodon.social/@suramya@www.suramya.com 

### What actually happened

The post is not available at https://mastodon.social/@suramya@www.suramya.com 

### Impact

All

### Available workarounds?

### Logs or notes

_No response_

pfefferle:
Fixes #470

## Proposed changes:

* Add a check that validates that the blog-users name is not colliding with the login name of an Author (WordPres user).

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -1,6 +1,9 @@
 <?php
 namespace Activitypub;
 
+use WP_User_Query;
+use Activitypub\Model\Blog_User;
+
 /**
  * ActivityPub Admin Class
  *
@@ -168,7 +171,7 @@ class Admin {
 				'type'              => 'string',
 				'description'       => \esc_html__( 'The Identifier of the Blog-User', 'activitypub' ),
 				'show_in_rest'      => true,
-				'default'           => \Activitypub\Model\Blog_User::get_default_username(),
+				'default'           => Blog_User::get_default_username(),
 				'sanitize_callback' => function( $value ) {
 					// hack to allow dots in the username
 					$parts     = explode( '.', $value );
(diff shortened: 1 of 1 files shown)


includes/class-admin.php
```diff
@@ -178,7 +181,31 @@ public static function register_settings() {
...
+							'hide_empty'     => true,
+							'fields'         => 'ID',
+						)
+					);
+
+					if ( $user->results ) {
+						add_settings_error(
+							'activitypub_blog_user_identifier',
+							'activitypub_blog_user_identifier',
+							\esc_html__( 'Please avoid using an existing author’s name as the blog profile ID. Fediverse platforms might use caching and this could break the functionality completely.', 'activitypub' ),
```
mattwiebe, line 201: I think I'd remove the extra detail about caching here, something a little more directive like:

"You cannot use an existing author's name for the blog profile ID."

includes/class-admin.php
```diff
@@ -178,7 +181,31 @@ public static function register_settings() {
 						$sanitized[] = \sanitize_title( $part );
 					}
 
-					return implode( '.', $sanitized );
+					$sanatized = implode( '.', $sanitized );
```
mattwiebe, line 184: Spelling error here: should be `sanitized` not `sanatized`

mattwiebe (approved):
This will be really good to have!

mediaformat:
Probably a separate PR, but it would be worth extending this idea to hooking into validate_username filter, to prevent registering with the chosen blog name, and also perhaps usernames that wouldn't federate. (such as those with a space or an `@`)

PR #487, opened 2023-10-04
Posts: add pre-fetch hook to allow plugins to hang filters on

kraftbj:
Some plugins, like Jetpack, add additional content to the post via filters, which is then included in the ActivityPub dispatch. I would not expect AP to track 
all that, but by adding a filter, plugins can add third-party integration to unhook filters in the right context.

Example usage:
```
add_action( 'activitypub_before_get_content', function( $post ) {
	remove_action( 'the_content', 'sharing_display', 19 );
});
```

With the Jetpack example, while there is already a check for `is_activitypub_request`, I don't believe this returns true when a new post is published and pushed 
to subscribers, e.g.:
* https://religious.social/@kraft@kraft.blog/110946655386025049
* https://religious.social/@kraft@photo.kraft.blog/111171770931037815

## Proposed changes:

* Adds a filter before post content retrival to allow for extensibility.

### Other information:

## Testing instructions:

* I need to actually put this on my site to confirm it works, so feel free to ignore until I do that and update the issue. :)
* Setup WordPress, Jetpack with Sharing buttons, ActivityPub, then follow your blog with another AP client (Mastodon).
* Do something like a photoblog entry with only a picture. No text for the sake of the experiment.
* See the post on Mastodon, etc.
* See the post including sharing items.
* Assuming this patch is applied, add a filter like the example above to remove the sharing_display.
* New post.
* See post without sharing stuff.


pfefferle (approved):
LGTM

kraftbj:
Thanks! I'll get the compact hooks in Jetpack (and wp.com if applicable/different)

PR #497, opened 2023-10-07
Fix #493

### LINKED ISSUE #493: `p` elements stripped from comments?
### Quick summary

I had [this here comment](https://indieweb.social/@janboddez/111150270077953809) appear on my blog as "[@jan](https://jan.boddez.net/author/jan) And that’s what “long-form” posts currently look like.I mean, I’m still tricking Mastodon [...]"

There should be a line break (or two) between "currently look like." and "I mean"; I think what's happening is the same as https://github.com/pfefferle/wordpress-webmention/issues/399 (`wp_kses()` being a bit too greedy).

How I've personally fixed this elsewhere (i.e., in my own Webmention plugin) is not by not filtering at all (even though WP core will (!) still sanitize comments inserted using `wp_new_comment()`, _I think_) but by adding the allowed tags in a `wp_kses_allowed_html` callback and making sure to use `pre_comment_content` for the "current filter" argument. That's what core will use. And then when I myself call `wp_kses()`, I make sure to call it with the same second argument.

### Steps to reproduce

Use Mastodon or something to comment on a federated post, and have at least two newlines in there.

For the `p` and `br`s to remain intact or get replaced by the correct number of newline characters.

### What actually happened

Elements got stripped, causing sentences and paragraphs to get concatenated wrongly.

### Impact

Most (> 50%)

### Available workarounds?

Yes, difficult to implement

### Logs or notes

_No response_

janboddez:
Fixes #493

## Proposed changes:

* Replaces call to `wp_filter_kses( $data )` with `addslashes( wp_kses( $data, 'pre_comment_content' ) )` when creating a new comment. This also fixes disappearing slashes (the AP JSON content is not pre-slashed).
* Additionaly, this PR adds a `wp_kses_allowed_html` filter that allows comments to contain `p` and `br` tags (only for AP comments, as the filter is removed immediately after the content is inserted).

### Other information:

- I have not changed (or run) test cases.
- I believe WP core will _again_ run new comments created thru `wp_new_comment()` through a kses filter. It normally uses the `pre_comment_content` for this, so the tags will be allowed there, too (the filter is only removed after this second call).

## Testing instructions:

* Reply to a federated post (e.g., on Mastodon), and use newline characters. The resulting WP comment should now include the `p` or `br` elements that were stripped before.
* Do the same thing, but use backslashes (e.g., respond using `¯\_(ツ)_/¯`). The backslash that would previously be stripped should appear in the comment.


janboddez:
Note that re: the filter being applied site-wide: should be able to remove it immediately after the call to `wp_new_comment()`. Didn't yet do so because I accidentally had left a (now fixed) bug in my code :-S

includes/rest/class-inbox.php
```diff
@@ -361,7 +361,7 @@ public static function handle_create( $object, $user_id ) {
 			'comment_post_ID' => $comment_post_id,
 			'comment_author' => \esc_attr( $meta['name'] ),
 			'comment_author_url' => \esc_url_raw( $object['actor'] ),
-			'comment_content' => \wp_filter_kses( $object['object']['content'] ),
+			'comment_content' => addslashes( \wp_kses( $object['object']['content'], 'pre_comment_content' ) ),
```
pfefferle, line 364: isn't it easier to directly add the array of allowed HTML as second param instead of using the filter?

janboddez: Alternatively, could temporarily unhook the (second) core call to `wp_kses`

pfefferle (approved):

PR #503, opened 2023-10-12
fix function call

pfefferle:
Fixes `[12-Oct-2023 10:15:38 UTC] PHP Fatal error:  Uncaught Error: Call to undefined method Activitypub\Integration\Webfinger::get_profile() in /.../wp-content/plugins/activitypub/integration/class-webfinger.php:55`

thanks for reporting @akirk


akirk (approved):
Thank you!

PR #518, opened 2023-10-16
increase timeout

### LINKED ISSUE #514: POST in inbox gives Error 500
### Quick summary

[ cut & paste from wordpress support site ]

I now (re-)installed the Activity-Pub Plugin. But it just randomly works. Most of the time the following request just produces error 500.)

From the Log of the Mastodon Instance (the access log from the wordpress system shows the same).
Mastodon::UnexpectedResponseError: https://le-chat-a-velo.at/wp-json/activitypub/1.0/users/0/inbox returned code 500

Some requests eventually get through. But

I tried disabling various plugins without success. The Error-Log of wordpress or PHP.

After a bit of playing around, we observed the following:
Pixelfed <-> WordPress works reliable.
Mastodon <-> WordPress does not work – gives the 500 on the POST reqest to the inbox.

We cleared various caches on both sides, enabled and disabled caches etc. with no difference in the outcome. We also tried 2 different Mastodon Instances with the same result. (Mastodon versions 4.2.0 and 4.3.0 alpha).

What I observed is, that sometimes the followers even end up in the database of wordpress. But for the failed requests the “activitypub_user_id” meta attribute to the follower is missing. However manually adding it does not change the behavior (it just shows up in the list in the admin-panel afterwards).

key problem for me is how to find out what causes the 500 – since the log is empty.

Cheers, ulrich

### Steps to reproduce

See above

202 instead of 500 ;-)

### What actually happened

Error 500
le-chat-a-velo.at "POST /wp-json/activitypub/1.0/users/0/inbox HTTP/1.1" 500 680 "-" "http.rb/5.1.1 (Mastodon/4.2.0; +https://miau.le-chat-a-velo.at/)"

### Impact

Some (< 50%)

### Available workarounds?

### Logs or notes

blog running at le-chat-a-velo.at
User: lechat@le-chat-a-velo.at

pfefferle:
because of several issues

fix #514

Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -79,7 +79,7 @@ function get_remote_metadata_by_actor( $actor, $cached = true ) {
 	}
 
 	$short_timeout = function() {
-		return 3;
+		return 10;
 	};
 	add_filter( 'activitypub_remote_get_timeout', $short_timeout );
 	$response = Http::get( $actor );


mattwiebe (approved):
Let's try it :)

Issue #523, opened 2023-10-18
Stuck in "Follow Requested" state


pfefferle:
### Quick summary

We experienced issues with some Mastodon instances, where the `Follow` request will not be properly accepted by WordPress. This is no general issue, but seems to be depending on a special kind of setup/setting/hosting. We found out, that it is an issue with quite a lot of instances hosted on Masto.host.

### Steps to reproduce

1. Go to Mastodon.art
2. Follow your blog
3. Stuck in pending state

### What you expected to happen

Follow should be auto-accepted

### What actually happened

Stuck in pending state

### Impact

Some (< 50%)

### Available workarounds?

No but the platform is still usable

### Logs or notes

We assume that this is a race condition, where the `Accept` of a `Follow` request will be sent too early.

pfefferle:
If you experience similar issues, please add the Mastodon Instance in the comments.

pfefferle:
After sending a manually `Accept` it seem to also work, what hardens the assumption that it is a race condition. I will add a cron to back the `Accept` response and I also think a CLI extension and an Action at the Followers-Table to re-send the `Accept` would be also nice and helps with debugging.

GoneCaving:
Seeing similar issues trying to follow my own blog from mastodon.ie

pfefferle:
OK, I tried mastodon.ie and it is even worse, it seems to be no race condition, there is no initial Follow request to the inbox at all!

I will file an issue on the Mastodon repo!

pfefferle:
Also https://mastodon.social/@pfefferle/111255975069274376

uk3:
@pfefferle A note: the symptoms of https://github.com/Automattic/wordpress-activitypub/issues/514 were the same. Mastodon shows the request stuck in 'Follow requested' 

518 fixed that for me.

Timothyjchambers:
Experiencing this from Indieweb.social that is hosted on Mastohost.

supernovia:
We seem to have another case here: https://wordpress.com/forums/topic/fediverse-followers-not-auto-approved-how-to-approve/

ssawczyn:
I'm having a similar problem when trying to follow from mspsocial.net . I'm also hearing that there are problems when folks try to follow my blog from mastodon.social.  In my case and at least one other that I know of, the follow request seems stuck in some sort of pending state.  

[...]

PR #524, opened 2023-10-18
Improve Accept Activities on Follow Requests

pfefferle:
This is mostly for debugging purposes

Improves #523 

## Proposed changes:

* A CLI Command to re-send Accept Activities when Follow request is stuck
* Schedule `Accept` Activity to prevent race conditions

### DIFF
--- a/includes/class-cli.php
+++ b/includes/class-cli.php
@@ -1,0 +1,74 @@
+<?php
+namespace Activitypub;
+
+use WP_CLI;
+use WP_CLI_Command;
+use Activitypub\Collection\Users;
+use Activitypub\Activity\Activity;
+use Activitypub\Collection\Followers;
+
+/**
+ * The ActivityPub CLI
+ */
+class Cli extends WP_CLI_Command {
+
+	/**
+	 * Re-Send Accept Activity as response to a Follow
+	 *
+	 * ## OPTIONS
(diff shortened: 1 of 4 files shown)


includes/class-cli.php
```diff
@@ -0,0 +1,74 @@
...
+
+		$follower = esc_url( $assoc_args['follower'] );
+		$follower = Followers::get_follower( $user->get__id(), $follower );
+
+		if ( ! $follower ) {
+			WP_CLI::error( __( 'Unknown Follower', 'activitypub' ) );
+		}
+
+		// get inbox
+		$inbox = $follower->get_shared_inbox();
```
mattwiebe, line 47: Only shared inbox? Or does the function fall back?

mattwiebe (commented):
This looks nice and well worth a try.

I'll have to change the command on dotcom from `activitypub` to something else, but that's totally fine.

pfefferle:
But does not solve this issue 😞 I have opened an issue on the Mastodon repo https://github.com/mastodon/mastodon/issues/27458

includes/class-cli.php
pfefferle: it falls back to the inbox. Maybe we should rename it to something like `preferred_inbox`?!?

PR #526, opened 2023-10-18
Shortcodes: only register when needed

mattwiebe:
I was poking around with post rendering and I thought it would be better to make sure that our shortcodes are never available outside of our own rendering context.

And of course I pushed these changes to `master` accidentally. 🤦 

I'll have to see if this breaks tests.


mattwiebe:
I did break tests but fixed them. I tried to use `setUpBeforeClass`/`tearDownAfterClass` instead but adding the `:void` return type to the methods caused the PHP 5.6 tests to fail.

PR #527, opened 2023-10-19
Various improvements

pfefferle:
Fixes various errors

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -214,7 +214,7 @@ class Signature {
 		\openssl_sign( $signed_string, $signature, $key, \OPENSSL_ALGO_SHA256 );
 		$signature = \base64_encode( $signature ); // phpcs:ignore
 
-		$key_id = $user->get_url();
+		$key_id = $user->get_url() . '#main-key';
 
 		if ( ! empty( $digest ) ) {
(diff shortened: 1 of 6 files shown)


includes/class-signature.php
```diff
@@ -214,7 +214,7 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
 		\openssl_sign( $signed_string, $signature, $key, \OPENSSL_ALGO_SHA256 );
 		$signature = \base64_encode( $signature ); // phpcs:ignore
 
-		$key_id = $user->get_url() . '#main-key';
```
pfefferle, line 217: @jeherve @mediaformat @mattwiebe this seem to fix the pending follow request on GoToSocial. But I have no idea why. Does that makes sense?

jeherve: I am able to follow a site running this branch from my GoToSocial instance, and new posts are also pushed, so this seems to work. I do not know if this would break something else, though. :|

mattwiebe (approved):
I don't know how to test, but it should be worth a try :)

akirk (approved):
:+1:

PR #530, opened 2023-10-24
Add monthly active users

pfefferle:
for better stats on FediDB

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -479,40 +479,24 @@ function is_blog_public() {
 /**
  * Get active users based on a given duration
  *
- * @param string $duration The duration to check
+ * @param int $duration The duration to check in month(s)
  *
  * @return int The number of active users
  */
-function get_active_users( $duration = '1 month ago' ) {
-	// get all distinct authors that have published a post in the last 30 days
-	$posts = get_posts(
-		array(
-			'post_type'      => 'post',
-			'post_status'    => 'publish',
-			'orderby'        => 'post_count',
-			'order'          => 'DESC',
-			'posts_per_page' => 4,
-			'date_query'     => array(
-				array(
-					'after' => $duration,
-				),
-			),
-		)
-	);
+function get_active_users( $duration = 1 ) {
+	global $wpdb;
 
-	if ( ! $posts ) {
-		return 0;
+	$duration = intval( $duration );
+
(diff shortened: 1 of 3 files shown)


includes/functions.php
```diff
@@ -474,3 +475,91 @@ function is_json( $data ) {
...
+ */
+function get_active_users( $duration = '1 month ago' ) {
+	// get all distinct authors that have published a post in the last 30 days
+	$posts = get_posts(
+		array(
+			'post_type'      => 'post',
+			'post_status'    => 'publish',
+			'orderby'        => 'post_count',
+			'order'          => 'DESC',
+			'posts_per_page' => 4,
```
mattwiebe, line 494: Having this in the query will cause problems, I think? 

The query will be limited to 4 posts, which means that only a maximum of 4 users would be able to be active on a given site, and also if, say, one prolific author has published the 4 most recent posts, only that author will be counted?

mattwiebe: I think we can probably do this better with a custom SQL query. We would probably then also want to cache the query. I asked ChatGPT for help and it looks pretty promising:

```php
global $wpdb;
$query = "
  SELECT COUNT(DISTINCT post_author) 
  FROM {$wpdb->posts} 
  WHERE post_type = 'post' 
  AND post_status = 'publish'
  AND post_date >= DATE_SUB(NOW(), INTERVAL 1 MONTH)
";
$num_authors = $wpdb->get_var($query);
```

We could wrap it all in a transient with a `DAY_IN_SECONDS` expiry?

mattwiebe (commented):
Left some ideas

includes/functions.php
pfefferle: Why have I added a posts per page? 😳🙄

pfefferle:
@mattwiebe can you re-check?

mattwiebe (approved):
This looks good to me now :)

pfefferle:
👍

But Why do you prefer transients?

mattwiebe:
> But Why do you prefer transients?

WP mostly runs on shared hosts, which do not have an object cache, so the `wp_cache_*` functions in those cases do not persist beyond a single request. Transients were designed to overcome this limitation, by instead leveraging the Options API to store the results of slow things like network calls or expensive computations, like our custom query.

For hosts with an object cache (like WordPress.com with `memcached`, the transients API will use that instead of Options.

https://developer.wordpress.org/apis/transients/

PR #536, opened 2023-10-25
Attachments: add support for audio 🔈 and video 📼

mattwiebe:
## Proposed changes:
* support attaching both audio and video uploads!

Note that audio and video are only supported in the block editor. 

Audio and video files are also currently constrained by our max_images number. That may lead to confusion?

### Other information:

## Testing instructions:
* upload a video and/or audio file to a post, they should show up in the activitypub version of the post. You can append `/activitypub/` to any given post URL and inspect the `attachment` value.

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -187,7 +187,7 @@ class Post {
 		}
 		$media_ids = \array_unique( $media_ids );
 
-		return array_map( array( self::class, 'wp_attachment_to_activity_attachment' ), $media_ids );
+		return \array_filter( \array_map( array( self::class, 'wp_attachment_to_activity_attachment' ), $media_ids ) );
 	}
 
 	/**
@@ -234,7 +234,7 @@ class Post {
 		}
 		$image_ids = \array_unique( $image_ids );
 
-		return array_map( array( self::class, 'wp_attachment_to_activity_attachment' ), $image_ids );
+		return \array_filter( \array_map( array( self::class, 'wp_attachment_to_activity_attachment' ), $image_ids ) );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/transformer/class-post.php
```diff
@@ -143,15 +143,96 @@ protected function get_attributed_to() {
...
+			return array();
+		}
+
+		$id = $this->wp_post->ID;
+
+		$image_ids = array();
+
+		// list post thumbnail first if this post has one
+		if ( \function_exists( 'has_post_thumbnail' ) && \has_post_thumbnail( $id ) ) {
+			$image_ids[] = \get_post_thumbnail_id( $id );
```
mattwiebe, line 179: Note that I fixed the previous bug in the parser that led to #500 - it was because I did `$max_images--` here when that was incorrect for the parser, which needed the global value, including the thumbnail

pfefferle: shouldn't we use `media_ids` or `attachment_ids` where we can?

I thinks we have to keep the `activitypub_max_image_attachments` for now, but that should be the only thing!?!

mattwiebe: Yup I think so, this was just a lot of copy-paste from the previous code as I restructured it.

I'll go and change the copy in the settings page while I'm at it to clarify that it will constrain audio/video as well

pfefferle:
@mattwiebe I added some small changes/tweaks. In my docker-env, there is a high possibility that uploads are incomplete. They will show up in the post, but are not indexed correctly as an attachment and some of the resizes are missing. This results in empty attachment array entries. The changes will prevent that problem.

pfefferle (approved):

PR #542, opened 2023-10-31
Profiles: `Update` followers when profile fields change

mattwiebe:
Currently we are at the mercy of remote servers refreshing profile data on their schedule. This PR will instead send an `Update` Activity to all Follower inboxes whenever an update to a relevant option or user meta is updated.

## Proposed changes:

* Schedule an update when relevant fields are changed

### Other information:

## Testing instructions:
This is completely untested so far.


includes/class-scheduler.php
```diff
@@ -22,6 +25,21 @@ public static function init() {
...
+			\add_action( 'update_option_blogdescription', array( self::class, 'blog_user_update' ) );
+			\add_action( 'update_option_blogname', array( self::class, 'blog_user_update' ) );
+			\add_filter( 'pre_set_theme_mod_custom_logo', array( self::class, 'blog_user_update' ) );
+			\add_filter( 'pre_set_theme_mod_header_image', array( self::class, 'blog_user_update' ) );
+		}
+
+		// profile updates for user options
+		if ( ! is_user_type_disabled( 'user' ) ) {
+			\add_action( 'updated_user_meta', array( self::class, 'user_update' ), 10, 3 );
+			// @todo figure out a feasible way of updating the header image since it's not unique to any user.
```
mediaformat, line 41: @mattwiebe @pfefferle I have a Profile Header code stub, which could live inside this plugin, but also makes sense as an independent block for Block Theme use. Any thoughts?

mediaformat (commented):
Should we include user Profile Banner code within this plugin?

includes/class-scheduler.php
pfefferle: @mediaformat sounds interesting! Do you have a public repo with the code?

pfefferle (dismissed):
LGTM

mattwiebe:
When I tested this on dotcom it did not necessarily seem like it worked, but that may have been a result of our other troubles there. I will try again now that we fixed our problems with the application user there before merging this.

pfefferle (approved):

PR #546, opened 2023-11-01
More reliable way to get author and autorurl

### LINKED ISSUE #520: sometimes the content of ap_author is missing from the post
### Quick summary

sometimes the content of ap_author is missing from the post.
When i update a post, sometimes the author (```[ap_author]```) disappears from the post.
Updating the post again (without actually changeing anythin) usually helps.

No Idea when this happens.

### Steps to reproduce

see above.

the author is displayed

### What actually happened

Nothing is displayed

### Impact

One

### Available workarounds?

### Logs or notes

_No response_

uk3:
Fixes #520

## Proposed changes:

* Change the way how to get the author and the authorurl of a post. See #520 for details

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-shortcodes.php
+++ b/includes/class-shortcodes.php
@@ -390,7 +390,8 @@ class Shortcodes {
 			return '';
 		}
 
-		$name = \get_the_author_meta( 'display_name', $item->post_author );
+		$author_id = \get_post_field( 'post_author', $item->ID );
+		$name = \get_the_author_meta( 'display_name', $author_id );
 
 		if ( ! $name ) {
 			return '';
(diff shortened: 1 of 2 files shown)


mattwiebe:
Thanks for the contribution! I have no idea why the previous code wouldn't have worked, as `get_post_field( 'post_author', $item->ID )` and `$item->post_author` should be effectively identical. Very odd, but I'm also used to that in WP. :)

uk3:
Funny thing, the previous code worked sometimes. Mostly the original version did not work on the initial posting, but cured itself on an update. 
In our blog the Author and the one who hits publish are 2 different users - maybe its related to that.
or it is some kind of type mixup - the author-id is a string as far as I understood.
or a race-contition somewhere (since ```$item->post_author``` returned 0 in this cases

I did not dig into too much detail so the above are my best guesses 🤷‍♀️😉

mattwiebe (approved):
It won't harm, and seems to help.

PR #552, opened 2023-11-08
Add/User Delete activities

### LINKED ISSUE #566: How to remove the fediverse-account for my closed weblog?
### Quick summary

I closed my WordPress blog that had the Activitypub plugin installed and activated. I deactivated the plugin, but the account is still discoverable. See https://micro.blog/diggingthedigital.com
How can I remove these accounts permanently from the fediverse? 

### Steps to reproduce

1. Search for domain @diggingthedigital or @admin@diggingthedigital.com in Mastodon or other Fediverse places
2. See the accounts. 

Since the Activitypub is no longer active in my WordPress installation, I'd expect these accounts to be gone as well. 

### What actually happened

I can still find the accounts. 

### Impact

Some (< 50%)

### Available workarounds?

### Logs or notes

_No response_

mediaformat:
Adds a Server class for processing server activities, such as Federating a `delete_user` action 

Fixes #566
https://gdpr.eu/right-to-be-forgotten/

## Proposed changes:

* Adds a Server class for processing server related activities
* Adds an activitypub_send_server_activity method for Server initiated activities
* Adds a `delete_user` action to Federate an actor Delete activity
 
## Testing instructions:

* Go to '..'
*


includes/class-scheduler.php
```diff
@@ -22,6 +26,23 @@ public static function init() {
...
+		if ( ! is_user_type_disabled( 'blog' ) ) {
+			\add_action( 'update_option_site_icon', array( self::class, 'blog_user_update' ) );
+			\add_action( 'update_option_blogdescription', array( self::class, 'blog_user_update' ) );
+			\add_action( 'update_option_blogname', array( self::class, 'blog_user_update' ) );
+			\add_filter( 'pre_set_theme_mod_custom_logo', array( self::class, 'blog_user_update' ) );
+			\add_filter( 'pre_set_theme_mod_header_image', array( self::class, 'blog_user_update' ) );
+		}
+
+		// profile updates for user options
+		if ( ! is_user_type_disabled( 'user' ) ) {
```
mediaformat, line 40: This is the main reason I started from https://github.com/Automattic/wordpress-activitypub/pull/542

If needed I can rebase from master adding this myself

mediaformat (commented):
See if rebase is needed for removing dependency on #542

pfefferle:
looks interesting

 I think sooner or later we need an outbox when the number of outgoing activities steadily grows...

Cambridgeport90:
Looks like this plugin’s really improving markedly since the last time I used it.

mediaformat:
For now we're just deriving known Servers based on current users followers, I'll add the blog followers here as well.

But (in a subsequent PR) we'll have to think about tracking known users:

Receiving an [actor Delete](https://github.com/Automattic/wordpress-activitypub/issues/554) means we should delete any Comments they've sent. And possibly Inbox Forward the Delete Activity.

mediaformat:
@pfefferle after testing, the activity must be signed by the actor (mastodon ignores the actor `Delete` signed by the application actor).

Since the activity is scheduled, it creates a race condition where even if we pass the user_id to `safe_remote_post` the user is likely to be deleted by the time the signature is generated.

So a workaround would be to store key pair in an option during `delete_user` action and delete the option during `deleted_user`.

Code update forthcoming

includes/class-scheduler.php
```diff
@@ -330,4 +338,86 @@ public static function schedule_profile_update( $user_id ) {
...
+			add_option( 'activitypub_temp_sig_' . $user->ID, $temp_private_key );
+
+			$author_url = \get_author_posts_url( $user->ID );
+
+			$activity = new Activity();
+			$activity->set_id( $author_url . '#delete' );
+			$activity->set_type( 'Delete' );
+			$activity->set_actor( $author_url );
+			$activity->set_object( $author_url );
+			$activity->set_to( [ 'https://www.w3.org/ns/activitystreams#Public' ] );
```
mattwiebe, line 406: Total aside, but seeing the creation of an `Activity` object here, I think it might be nice to have a constructor shorthand, like:
```php
$activity = new Activity([
    'id' => $author_url . '#delete',
    'type' => 'Delete',
    'actor' => $author_url,
    'object' => $author_url,
    'to' => ['https://www.w3.org/ns/activitystreams#Public'],
]);
```

Just a little bit easier to mentally parse, for me anyway. Would need a `Base::__contstruct()` method added but the logic should be pretty simple. I'll maybe try it sometime. :)

mattwiebe (dismissed):
Thanks so much for this important feature @mediaformat !

My understanding is that this will work for deleted users but not yet for the Blog User? Should we merge this and a button to the Settings page that allows us to issue DELETE calls for the all Users (both Blog and User types) followed by deactivating the plugin? 

In any case that can be a followup PR, this does look ok to get in and iterate on.

pfefferle:
I would not merge it as is yet. I think we have to update some parts (like user caps) and discuss some missing parts, like: Should we send a delete if you remove the `ActivityPub` capability from a user, should we ask the admin if we should do it, should we schedule it some time in the future, so the user will be able to revert his decision, ...?

includes/class-application.php
```diff
@@ -0,0 +1,34 @@
...
+use function Activitypub\is_user_disabled;
+
+/**
+ * ActivityPub Application Class
+ *
+ * @author Django Doucet
+ */
+class Application {
+
+	public static function known_inboxes() {
```
pfefferle, line 17: I would not add it to the application class, but to the followers collection. We should also have a function to get all inboxes regardless of the user. Maybe we can use https://github.com/Automattic/wordpress-activitypub/blob/f6b4783dedbc5ad07c42a9c49a275ecaed05a227/includes/collection/class-followers.php#L200

includes/class-scheduler.php
```diff
@@ -75,6 +78,8 @@ function ( $comment_id ) {
 
 		// profile updates for user options
 		if ( ! is_user_type_disabled( 'user' ) ) {
+			\add_action( 'delete_user', array( self::class, 'schedule_profile_delete' ), 10, 3 );
+			\add_action( 'deleted_user', array( self::class, 'schedule_user_delete' ), 10, 3 );
```
pfefferle, line 82: As I mentioned in the comments: what about capability changes?

includes/class-signature.php
```diff
@@ -186,7 +186,11 @@ protected static function check_legacy_key_pair_for( $user_id ) {
 	 */
 	public static function generate_signature( $user_id, $http_method, $url, $date, $digest = null ) {
 		$user = Users::get_by_id( $user_id );
-		$key  = self::get_private_key_for( $user->get__id() );
+		if ( ! is_wp_error( $user ) ) {
+			$key = self::get_private_key_for( $user->get__id() );
+		} else {
+			$key = get_option( 'activitypub_temp_sig_' . $user_id, true );
```
pfefferle, line 192: this is not the user-id! see: https://github.com/mediaformat/wordpress-activitypub/blob/cf5e95eb07a4044e28481e7f15982c07e84b027b/includes/class-signature.php#L131C28-L131C57

pfefferle (changes requested):

pfefferle:
But I agree with @mattwiebe ! This is an important feature and we should work on getting it merged!

mediaformat:
hey @mattwiebe, that is correct there isn't any option to handle the Blog User. 
> Should we merge this and a button to the Settings page that allows us to issue DELETE calls for the all Users (both Blog and User types) followed by deactivating the plugin?

Yes, this is what I was envisioning, with warnings about the irreversible nature of the operation.

mattwiebe:
> Should we send a delete if you remove the ActivityPub capability from a user

Yes! Great chance to educate the user on what will happen.

> Should we ask the admin if we should do it, should we schedule it to some time in the future, so the user will be able to revert his decision

I think give fair warning, at least in the case of removing the capability, that it will be sent. Adding the UI to manage the delay and cancel the deletion seems like solving a hypothetical that we can circle back later if there's demand.

includes/class-application.php
mediaformat: yes, that looks like a great fit!

includes/class-scheduler.php
mediaformat: @pfefferle since capability changes don't have any associated action hooks, we might consider adding some sort of user management section to the plugin settings.

pfefferle: We could add some actions here https://github.com/Automattic/wordpress-activitypub/blob/f6b4783dedbc5ad07c42a9c49a275ecaed05a227/includes/class-admin.php#L450

includes/class-signature.php
mediaformat: l'd like to reuse this however there is a [race condition](https://github.com/Automattic/wordpress-activitypub/pull/552#issuecomment-1806592481) preventing the use of $user object at this stage.
https://github.com/Automattic/wordpress-activitypub/blob/f6b4783dedbc5ad07c42a9c49a275ecaed05a227/includes/class-signature.php#L131-L137

Hence the use of a temporary option to store the signature before user deletion.

includes/class-scheduler.php
mediaformat: @pfefferle Should we tackle this in a followup PR?

includes/class-signature.php
```diff
@@ -186,7 +186,14 @@ protected static function check_legacy_key_pair_for( $user_id ) {
 	 */
 	public static function generate_signature( $user_id, $http_method, $url, $date, $digest = null ) {
 		$user = Users::get_by_id( $user_id );
-		$key  = self::get_private_key_for( $user->get__id() );
+		if ( ! is_wp_error( $user ) ) {
+			$key = self::get_private_key_for( $user->get__id() );
+			$key_id = $user->get_url() . '#main-key';
+		} else {
+			$temp_sig_options = get_option( 'activitypub_temp_sig_' . $user_id );
```
pfefferle, line 193: this feels a bit hacky and might break things in the future, if we maybe introduce key rotation: https://swicg.github.io/activitypub-http-signature/#key-rotation

pfefferle: Do we need a sig for deletes at all? The remote server is not able to verify it anyways!?!

This is very confusing https://swicg.github.io/activitypub-http-signature/#handling-deletes-of-actors

mediaformat: Agreed it is very hacky! In my previous tests Mastodon ignored actor deletes signed by the instance actor, but I will do some more tests and report back.

pfefferle: or maybe we do it as you mentioned it here: https://github.com/Automattic/wordpress-activitypub/pull/552#issuecomment-1806592481

> So a workaround would be to store key pair in an option during delete_user action and delete the option during deleted_user.

pfefferle: or we store the complete delete object in the schedule on the delete?!?

mediaformat: > So a workaround would be to store key pair in an option during delete_user action and delete the option during deleted_user.

The first part is the hack I've implemented, the problem with the second part is that `deleted_user` will occur almost immediately, and the Delete activities will fail.

mediaformat: > or we store the complete delete object in the schedule on the delete?!?

Hmm, the scheduler runs before signature generation.

pfefferle:
This is tricky! I think the real "delete" from the DB should be easy to handle.

But deleting users from the fediverse, but keeping it in WordPress might be hard. With https://swicg.github.io/activitypub-http-signature/#handling-deletes-of-actors we have to either still provide the signature for all users that had the plugin enabled, or return a 410 when requested with `Accept: application/activity+json`.

mediaformat:
> we have to either still provide the signature for all users that had the plugin enabled, or return a 410 when requested with `Accept: application/activity+json`.

I think we might need both.
1. Keeping signatures to sign the user delete activities and notify known remote servers.
2. Returning 410 for Servers to discover that the user has been deleted.

pfefferle:
The foundational work was done with 7.3.0 https://activitypub.blog/2025/08/28/7-3-0-ctrl-fed-delete/

PR #551, opened 2023-11-08
this introduces activity-type-handlers

pfefferle:
Should be easier to handle different activities.

## Proposed changes:

*

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Go to '..'
*

PR #561, opened 2023-11-13
Add/incoming actor deletes

### LINKED ISSUE #554: Feature Request: Drop (certain) Delete requests?
### What

Seeing a lot of Delete activities in debug.log, followed by a 401 response ("Remote profile not accessible" or something).

But I'm (i.e., my WordPress user is) not following anyone, and less than 0.01% of these Deletes concern followers of mine.

Don't think it makes sense to look up remote profiles for all of these Delete requests. If they concern a follower, we should probably verify the delete actually took place. If not, can't we just drop the request and be done with it? (Or maybe that's already happening and the debug is simply ambiguous?)

### Why

Don't think it makes sense to look up remote profiles for Delete requests _if they don't concern a profile that we actually know_.

Hoping this frees up some server resources.

### How

_No response_

mediaformat:
Ignore Actor Deletes unless site has Comments from Actor

Fixes #554 

## Proposed changes:

* Checks for incoming Actor Delete 
* Compares incoming Actor Delete activity object to known commenters
* Changes `Signature::get_key` error codes to `404` for accuracy
* Schedules Deletion of actor's comments if verification fails with `404` 

### Other information:
* If the deleted actor's key is still in cache, the Deletion of actor's comments should be handled in appropriate activity-type-handler #551  

## Testing instructions:

* Go to '..'
*


mediaformat:
This method is a good starting point for the PR, it has a weakness though:

If the deleted actor's public key is not in cache, we are taking the servers word that the activity is authentic, without being able to prove it.

If a server is shut down and the dns changes control a malicious actor could send out fake activities. 

Mastodon uses LD-Signature for this, but it also caches all known actors keys basically indefinitely.

includes/rest/class-server.php
```diff
@@ -108,8 +108,28 @@ public static function authorize_activitypub_requests( $response, $handler, $req
 
 		// POST-Requets are always signed
 		if ( 'GET' !== $request->get_method() ) {
+
+			if ( Application::is_actor_delete_request( $request ) ) {
+				if ( ! Application::is_known_commenter( $request ) ) {
```
pfefferle, line 113: I would not handle that here, because this skips all the inbox actions, so no other implementation could interpret and react to it. lets skip signature verification for `Deletes` (even with the introduced filter or directly in the server class) and implement the code in the delete handler I am introducing here: https://github.com/Automattic/wordpress-activitypub/pull/551

I would also prefer adding the delete code to the matching collections, like followers for the actor deletes and https://github.com/Automattic/wordpress-activitypub/pull/551/files#diff-17683f8d8c7226d3e07a6c94ffb52b7bf270d9711a36bdb118b6b4c55fd32820 (we can discuss the name of it) for the comment/reaction deletes

pfefferle (changes requested):
Can you maybe point your PR to the #551 PR and match the new file structure?

mediaformat:
Closing in favor of #575

PR #565, opened 2023-11-17
Attachments: dedupe earlier to prevent incorrect max_media

### LINKED ISSUE #560: Number of attached images incorrect in case of duplicates
### Quick summary

If there are images in a post more than once, the number of attached images is incorrect.
Most common case is that the post thumbnail is also used in the Post.

### Steps to reproduce

1. Attach an image more than once to a post.
2. Viewer images are attached to the post

Configured number of attached images

### What actually happened

Viewer Images attached.

### Impact

All

### Available workarounds?

### Logs or notes

The issue is in the way the limit is imposed in https://github.com/Automattic/wordpress-activitypub/tree/master/includes/transformer/class-post.php line 184+

First the max_media number of images is fetched. After that a unique is done on the resulting array. This ends up in having viewer elements than max_media if there are duplicates. (this goes for both variants of how to get attached images)

The correct order should be: Fetch all -> unique -> cut out the first max_media.

mattwiebe:
Our dedupe logic ran too late, after already scoping to max_media, which would sometimes result in fewer attachments than warranted by `max_media`. 

This would happen especially in the case of a Featured Image also appearing in the `post_content`.

Fixes #560 

## Proposed changes:

* dedupe attachments before cutting to max_media

### Other information:

## Testing instructions:
Make a post with more images than your max media. Add a feature image and also add it as a standalone image in post_content. The ActivityPub representation will have one fewer attachment than it should have without this patch, the correct number with it.

Eg, with a local post and `max_images=3`:

```
curl -H 'Accept: application/activity+json' http://wp.test/2023/10/301/ | jq .attachment
## Before:
[
  {
    "type": "Image",
    "url": "http://wp.test/wp-content/uploads/2023/08/avatar-2017-large-upsidedown.jpeg",
    "mediaType": "image/jpeg"
  },
  {
    "type": "Image",
    "url": "http://wp.test/wp-content/uploads/2023/09/Screenshot-2023-09-20-at-15.33.15.png",
    "mediaType": "image/png"
  }
]

## After:
[
  {
    "type": "Image",
    "url": "http://wp.test/wp-content/uploads/2023/08/avatar-2017-large-upsidedown.jpeg",
    "mediaType": "image/jpeg"
  },
  {
    "type": "Image",
    "url": "http://wp.test/wp-content/uploads/2023/09/Screenshot-2023-09-20-at-15.33.15.png",
    "mediaType": "image/png"
  },
  {
    "type": "Image",
    "url": "http://wp.test/wp-content/uploads/2023/09/Screenshot-2023-09-20-at-15.40.13.png",
    "mediaType": "image/png"
  }
]
```

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -185,7 +185,6 @@ class Post {
 			$blocks = \parse_blocks( $this->wp_post->post_content );
 			$media_ids = self::get_media_ids_from_blocks( $blocks, $media_ids, $max_media );
 		}
-		$media_ids = \array_unique( $media_ids );
 
 		return \array_filter( \array_map( array( self::class, 'wp_attachment_to_activity_attachment' ), $media_ids ) );
 	}
@@ -278,6 +277,9 @@ class Post {
 					}
 					break;
 			}
+
+			// depupe
+			$media_ids = \array_unique( $media_ids );
 
 			// stop doing unneeded work
 			if ( count( $media_ids ) >= $max_media ) {


pfefferle (approved):
LGTM; Thanks @mattwiebe

PR #569, opened 2023-11-21
Add/extendable transformers

Menrath:
This draft pull request should for now make it easy for everyone to track the status of the proposal ["flexible management of how posts gets transformed to ActivityPub objects".](https://github.com/Automattic/wordpress-activitypub/discussions/541).

It is still under heavy, early development, but questions and discussion are also welcome here.

## Proposed changes:
Transformers are now extendable by other plugins via an API.
More details will follow.

### Other information:
Will follow.

## Testing instructions:
Will follow.

We develop this changes [here](https://code.event-federation.eu/Event-Federation/wordpress-activitypub/src/branch/add/extendable-transformers), if you are interested in following the internal discussions more closely.


includes/class-transformer-base.php
```diff
@@ -0,0 +1,597 @@
+<?php
```
pfefferle, line 1: I would use the `Transformer` Namespace and move them into the transformer folder.

pfefferle: same with the manager

includes/activity/class-event.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Event is an implementation of one of the
+ * Activity Streams Event object type
+ *
+ * The Object is the primary base type for the Activity Streams
+ * vocabulary.
+ *
+ * @see https://www.w3.org/TR/activitystreams-vocabulary/#dfn-event
+ */
+class Note extends Base_Object {
```
pfefferle, line 21:
```suggestion
class Event extends Base_Object {
```

includes/activity/class-event.php
```diff
@@ -0,0 +1,23 @@
+<?php
```
pfefferle, line 1: why these classes? I tried to avoid them, because otherwise we also have to map all other types https://github.com/Automattic/wordpress-activitypub/blob/master/includes/transformer/class-post.php#L398

includes/class-transformer-base.php
```diff
@@ -0,0 +1,597 @@
...
+		$class_reflection = new \ReflectionClass( $transformer );
+
+		$plugin_basename = plugin_basename( $class_reflection->getFileName() );
+
+		$plugin_directory = strtok( $plugin_basename, '/' );
+
+		$plugins_data = get_plugins( '/' . $plugin_directory );
+		$plugin_data = array_shift( $plugins_data );
+
+		return $plugin_data['Name'] ?? esc_html__( 'Unknown', 'activitypub' );
```
pfefferle, line 81: please be downwards compatible to at least PHP 7.0!!

includes/class-transformer-base.php
pfefferle: let me know if I can help!

includes/activity/class-event.php
Menrath: We are still evaluating all pros and cons. At the moment we have only tried to get an idea of what the consequences would be. As is often the case: the deeper I dig into this repository, the more I can understand certain design decisions.

includes/transformer/class-transformers-manager.php
```diff
@@ -0,0 +1,292 @@
...
+	/**
+	 * Require files.
+	 *
+	 * Require ActivityPub transformer base class.
+	 *
+	 * @since version_number_transformer_management_placeholder
+	 * @access private
+	 */
+	private function require_files() {
+		require ACTIVITYPUB_PLUGIN_DIR . 'includes/transformer/class-base.php';
```
pfefferle, line 115: you do not have to require classes, they are all auto-loaded.

includes/transformer/class-transformers-manager.php
```diff
@@ -0,0 +1,292 @@
...
+/**
+ * ActivityPub transformers manager.
+ *
+ * ActivityPub transformers manager handler class is responsible for registering and
+ * initializing all the supported WP-Pobject to ActivityPub transformers.
+ *
+ * @since version_number_transformer_management_placeholder
+ */
+class Transformers_Manager {
+	const DEFAULT_TRANSFORMER_MAPPING = array(
```
pfefferle, line 30: there are a lot of function/attributes that are already defined by the Base object.

includes/transformer/class-transformers-manager.php
```diff
@@ -0,0 +1,292 @@
...
+	 * @access public
+	 *
+	 * @param WP_Post|WP_Comment $object The WordPress Post/Comment.
+	 *
+	 * @return \ActivityPub\Transformer\Base|null Registered transformers.
+	 */
+	public function get_transformer( $object ) {
+		switch ( get_class( $object ) ) {
+			case 'WP_Post':
+				$post_type = get_post_type( $object );
```
pfefferle, line 279: this looks a bit complicated and redundant. the way you implemented it, does not require a switch/case, I think.

includes/rest/class-outbox.php
```diff
@@ -105,7 +105,8 @@ public static function user_outbox_get( $request ) {
 			);
 
 			foreach ( $posts as $post ) {
-				$post = Post::transform( $post )->to_object();
+				$transformer = \Activitypub\Transformer\Transformers_Manager::instance()->get_transformer( $post );
```
pfefferle, line 108:
```suggestion
				$transformer = Transformers_Manager::instance()->get_transformer( $post );
```

you do not have to use the full path (with namespaces) when using `use`.

includes/transformer/class-transformers-manager.php
```diff
@@ -0,0 +1,292 @@
+<?php
```
pfefferle, line 1: I would remove the `transformers` in the name, because this is already defined by the namespace and I would go with `factory` instead of `manager` because this is a known pattern.

pfefferle:
@Menrath thanks for your work, I really appreciate it!

I added some feedback, and just wanted to mention to please take it as "feedback" and not as "must be changed"! I simply would love to start a discussion about some things!

includes/activity/class-event.php
pfefferle: True, but there are things that are simply legacy decisions! Be aware that this project started as a spare time thing and still has some crappy code everywhere. So do not try to understand everything and feel free to challenge us, that's what I also try here ;)

I discuss to understand, not to win :)

includes/transformer/class-transformers-manager.php
Menrath: Thanks for those suggestions. Another naming I thought about is using `registry`, that is the naming the Gutenberg project uses for its blocks.

includes/activity/class-event.php
Menrath: Thank you, for offering help, even at our _Drafting Stage_. I gladly accept this invitation for discussions :)

One of my goals is,  that it is very easy an straight forward, for feature developers adding their own transformers to their plugins. Very likely they will have to extend the _Base_Object_ class anyway, especially if they want to add any custom fields that are not defined yet in our _Base_Objet_ class, which until now is consistent with the _Activitystreams Vocabulary Standard_. For example even for supported object types like 'Event' Mobilizon makes usage of PeerTubes `commentsEnabled` feature.

Other developers might have a look, how thinks are done in this project, and might get confused if it does things a bit differently than the future docs advice them to do.

Another related issue is, that when adding a custom field, one also has to add the appropriate json-ld context. 
Though the filter `activitypub_json_context` is available, editing the array directly seemed kind of rough to me and applying the filter within the transformer code leaded to a runtime issue, so I had to apply this filter in a different part of the code, which is not ideal in my opinion.

I will share our repository with some example transformers registering to the ActivityPub-plugin hopefully today, after doing some renaming and initial documentation.

pfefferle: > I will share our repository with some example transformers registering to the ActivityPub-plugin hopefully today, after doing some renaming and initial documentation.

That would be perfect!

I agree with your point, but I am not sure if the "extensions" (Event-Activity) should be part of this repo or part of yours. I agree, that it has to be as easy as possible and maybe we have to change the current context implementation for that.

You can maybe have a look what I did to extend the base object(s) for the Follower (https://github.com/Automattic/wordpress-activitypub/blob/master/includes/model/class-follower.php) or internal User mapping (https://github.com/Automattic/wordpress-activitypub/blob/master/includes/model/class-user.php).

Menrath:
As this Pull request is prepared by two persons at once and things at the moment change quite frequently, we thought it would be best to keep minor changes in a [seperated dev branch](https://code.event-federation.eu/Event-Federation/wordpress-activitypub/src/branch/dev/extendable-transformers). We will keep merging and squashing better understandable packages in the branch of this pull request soon.

Menrath:
@pfefferle I decided to close this PR and am currently preparing a new one after reviewing https://github.com/Automattic/wordpress-activitypub/commit/b744dc551dc9ca34e2ca805ca99755fcc86f4013.

I will keep the filter for applying the transformer you used as a fallback, but keep the transformer setting in the admin as proposed here, with a new UI though.

I you disagree I can arrange with that. However, what is most important to me is that the admin-users see in the settings page which transformer is used and which plugin provided it.

If you manage to give some short feedback before christmas break that would be awesome!

PR #570, opened 2023-11-23
Check if the current post type supports ActivityPub.

### LINKED ISSUE #521: Pages show in ActivityPub despite being disabled (in a weird way)
### Quick summary

I have checked ActivityPub only for Posts (not Pages or media).
When I do an activity-Pub search for a Page it still shows up, but with all the [ap_*] fields being empty. The other custom fields and the page thumbnail are still populated.

### Steps to reproduce

1. Search for https://le-chat-a-velo.at/reisen/karpatenbecken-2023/ in on a mastodon server.

i'd expect either the full content (since it is still publically avaliable) or a not found response (since pages are disabled)

### What actually happened

image

### Impact

One

### Available workarounds?

### Logs or notes

_No response_

pfefferle:
Fixes #521 

## Proposed changes:

* Check if post type is supported by ActivityPub

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -283,14 +283,7 @@ function is_activitypub_request() {
 	if ( \is_singular() ) {
 		$queried_object = \get_queried_object();
 		$post_type      = \get_post_type( $queried_object );
-		$post_types     = \get_post_types_by_support( 'activitypub' );
-
-		if (
-			false === $post_type ||
-			! \in_array( $post_type, $post_types, true )
-		) {
-			return false;
-		}
+		return ! empty( $post_type ) && \post_type_supports( $post_type, 'activitypub' );
 	}
 
 	// One can trigger an ActivityPub request by adding ?activitypub to the URL.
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -279,6 +279,20 @@ function is_activitypub_request() {
...
+		$queried_object = \get_queried_object();
+		$post_type      = \get_post_type( $queried_object );
+		$post_types     = \get_post_types_by_support( 'activitypub' );
+
+		if (
+			false === $post_type ||
+			! \in_array( $post_type, $post_types, true )
+		) {
+			return false;
+		}
```
jeherve, line 293: Once you have the post type, I think you may be able to rely on `post_type_supports`:

```suggestion
		return ! empty( $post_type ) && \post_type_supports( $post_type, 'activitypub' );
```

pfefferle: thanks for the feedback, but I still have to rewrite it a bit, to only return on `false`.

mattwiebe (approved):
Looks good to me

jeherve (approved):

PR #578, opened 2023-11-26
Clarifying what [ap_excerpt] does

### LINKED ISSUE #577: Clarify on settings page and in help what [ap_excerpt] does exactly
### Quick summary

The plugin settings are not explicit about what `[ap_excerpt]` does exactly. I was relieved that it supports `the_excerpt()` if that is set for a post in WordPress. So [ap_excerpt] will a) use `the_excerpt()` if set for a post, b) supply an excerpt itself by truncating at `length`, where length=400 by default. I propose to clarify this in the relevant places on the Settings page.

### Steps to reproduce

Select "excerpt" or use [ap_excerpt] in the Post content field in Settings. 

I expect the settings or help page to mention `the_excerpt()` explicitly.

### What actually happened

It is unclear what [ap_excerpt] does exactly until you test.

### Impact

One

### Available workarounds?

Yes, easy to implement

### Logs or notes

_No response_

mdingemanse:
Fixes #577 

## Proposed changes:

* Clarifies what [ap_excerpt] does exactly in the relevant places (main Settings page and help.php)
* No change in behaviour, just a change to clarify documentation. It should be clear to users that this respects `the_excerpt` if available, and otherwise truncates the post at `length`.

## Testing instructions:

* Go to ActivityPub > Settings
* The relevant descriptions should be updated to be more clear and less ambiguous about what it means to choose [ap_excerpt]. It will be clear to users that it respects `the_excerpt` if available, and otherwise truncates the post at `length`.

### DIFF
--- a/includes/help.php
+++ b/includes/help.php
@@ -12,7 +12,7 @@
 				'<dt><code>[ap_content apply_filters="yes"]</code></dt>' .
 				'<dd>' . \wp_kses( __( 'The post\'s content. With <code>apply_filters</code> you can decide if filters (<code>apply_filters( \'the_content\', $content )</code>) should be applied or not (default is <code>yes</code>). The values can be <code>yes</code> or <code>no</code>. <code>apply_filters</code> attribute is optional.', 'activitypub' ), array( 'code' => array() ) ) . '</dd>' .
 				'<dt><code>[ap_excerpt length="400"]</code></dt>' .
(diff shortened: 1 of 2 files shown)


mdingemanse:
I can't actually find where the "help" text is generated that I also would propose to change. This is what it says now:

> [ap_excerpt length="400"]
> The post's excerpt (default 400 chars). length attribute is optional.

This is what I think it should say:

> [ap_excerpt length="400"]
> The post's excerpt (uses `the_excerpt()` if available). If no excerpt is provided, this will present the start of the post truncated at  `length` characters. Length attribute is optional.

pfefferle (approved):
Thanks @mdingemanse

PR #588, opened 2023-12-02
Fix sending posts to at most ten instances

### LINKED ISSUE #587: Posts are sent to at most 10 instances
### Quick summary

`WP_Query` paginates by default. This is used to look up the list of inboxes to post to in `Followers::get_inboxes()`, which has the effect of returning at most ten users to post to, which are then deduped on the inbox URL.

### Steps to reproduce

1. Have a profile with more than 10 followers.
2. Make a post.
3. Check where it's delivered to.

All followers should get the post.

### What actually happened

Followers on at most ten instances get the post.

### Impact

All

### Available workarounds?

### Logs or notes

_No response_

Sentynel:
Fixes #587 

## Proposed changes:

Disable pagination on the query in `Followers::get_inboxes()` that looks up follower IDs. This causes it to return all followers, rather than a paged list of ten of them. This means that we end up delivering our post to all followers, and not to the instances belonging to a subset of ten of them.

### Other information:

## Testing instructions:

* You need an instance with more than ten followers from different instances for a profile (user or blog) to demonstrate the issue here.
* Log outgoing posts in some way.
* Prior to the fix, you'll see that when a post goes out, at most ten instances (and often fewer, if you have more than one follower per instance) actually get sent the message.
* With this fix, all should get them.
* Alternatively, use something like `wp shell` from `wp-cli` to run the queries in before and after fix forms directly and observe the difference.

### DIFF
--- a/includes/collection/class-followers.php
+++ b/includes/collection/class-followers.php
@@ -267,7 +267,7 @@ class Followers {
 		// get all Followers of a ID of the WordPress User
 		$posts = new WP_Query(
 			array(
-				'nopaging' => true,
+				'nopaging'   => true,
 				'post_type'  => self::POST_TYPE,
 				'fields'     => 'ids',
 				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
--- a/includes/collection/class-followers.php
+++ b/includes/collection/class-followers.php
@@ -267,6 +267,7 @@ class Followers {
 		// get all Followers of a ID of the WordPress User
 		$posts = new WP_Query(
 			array(
+				'pagination' => false,
 				'post_type'  => self::POST_TYPE,
 				'fields'     => 'ids',
 				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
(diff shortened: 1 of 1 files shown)


pfefferle:
😱 oh no! What a bad bug! thanks a lot for finding it! I definitely have to write some tests for that case!

But I couldn't find the arg `pagination` but `nopaging` instead!?!

Can you verify that?

Sentynel:
Whoops, yes, you're right! Tested it right in the console, immediately copied it down wrong. Don't submit PRs at midnight, folks. Fixed now.

includes/collection/class-followers.php
```diff
@@ -267,6 +267,7 @@ public static function get_inboxes( $user_id ) {
 		// get all Followers of a ID of the WordPress User
 		$posts = new WP_Query(
 			array(
+				'nopaging' => true,
```
pfefferle, line 270:
```suggestion
				'nopaging'   => true,
```

pfefferle:
Thanks a lot @Sentynel !!!

ThisIsMissEm:
Could similar exist elsewhere, e.g., `Followers::count_followers` or `Followers::get_outdated_followers` and `Followers::get_faulty_followers` (these seem to take per page, but no arguments to actually paginate)?

PR #593, opened 2023-12-08
Outbox Collection

### LINKED ISSUE #961: Deleted WP post remains undeleted on Mastodon
### Quick summary

There doesn’t appear to be any way to remove a post from Mastodon if the post has been deleted in Wordpress. Instead the Mastodon post remains live even after the post that created it in Wordpress has been deleted.

### Steps to reproduce

1. Create a post in WP.
2. it becomes a federated post on Mastodon thanks to AP plugin
3. i delete the post in WP.
4. it remains undeleted on Mastodon and there is no means to manually delete it.

Ideally when the WP post gets deleted this would also delete the Mastodon post it created. Or, if not, it would be good to be able to manage existing Mastodon posts created by the AP plugin and ve able to manually delete it.

### What actually happened

Post remained active in Fediverse and so links back to original, now deleted WP post go to a 404.

### Impact

One

### Available workarounds?

### Logs or notes

_No response_

pfefferle:
Working on an Outbox Collection, to better support Blogs with lots of Followers, Retries and Activities other than Post.

/cc @mediaformat @Menrath @ruru4143 @akirk @mattwiebe 

## Proposed changes:

* Add Outbox
* Scheduler, that handles unpublished Outbox items in chunks
* Implement Retry
* Actions to push data into the Outbox

### Other information:

## Testing instructions:

* Go to '..'
*

Fixes #961 #1203


includes/transformer/class-activity-object.php
```diff
@@ -0,0 +1,155 @@
...
+		$cc       = array();
+		$mentions = $this->get_mentions();
+
+		if ( $mentions ) {
+			foreach ( $mentions as $url ) {
+				$cc[] = $url;
+			}
+		}
+
+		return $cc;
```
obenland:
```suggestion
		return $this->get_mentions() ?: array();
```
Could this be simplified to ^?
If `get_mentions()` always returned an array, the fallback would not be needed.

pfefferle: We can remove the `if` then and directly iterate but we can't simply return `get_mention()` because it has some more data, not only the list of URLs.

pfefferle: And we could use an `array_filter` instead!

obenland: Hm, maybe `return array_values( $this->get_mentions() );`? 

Currently the function doesn't remove anything from `$mentions`, it just puts everything in `$cc`.

pfefferle: 🤔

pfefferle: You are right, array_values should do the trick! It thought it has some more informations, but it's simply the key is the WebFinger handle and the URL is the value!

includes/functions.php
```diff
@@ -1593,3 +1595,48 @@ function is_self_ping( $id ) {
...
+
+	$activity = $transformer->to_object();
+
+	if ( ! $activity || is_wp_error( $activity ) ) {
+		return false;
+	}
+
+	set_wp_object_state( $data, 'federate' );
+
+	$id = Outbox::add( $activity, $type, $user_id, $content_visibility );
```
mattwiebe: It seems to me that all of this functionality in `add_to_outbox` should just live inside of `Outbox::add`? Which would then either leave us with a thin wrapper in this function, or we can then get rid of the wrapper entirely?

includes/scheduler/class-actor.php
```diff
@@ -0,0 +1,102 @@
...
+
+	/**
+	 * Theme mods only have a dynamic filter so we fudge it like this.
+	 *
+	 * @param mixed $value Optional. The value to be updated. Default null.
+	 *
+	 * @return mixed
+	 */
+	public static function blog_user_update( $value = null ) {
+		self::schedule_profile_update( 0 );
```
mattwiebe:
```suggestion
		self::schedule_profile_update( Actors::BLOG_USER_ID );
```

As the person who wrote the original code and several times thought, "I should patch that"…

mattwiebe (commented):
Haven't tried running this yet, but read through the changes and left a couple of comments/questions.

Really like the scheduler separation. 👍 

What do we still need to do to ship this?

includes/functions.php
```diff
@@ -1547,3 +1549,48 @@ function is_self_ping( $id ) {
...
+	$args = array( $id );
+
+	if ( false === wp_next_scheduled( $hook, $args ) ) {
+		\wp_schedule_single_event(
+			\time() + 10,
+			$hook,
+			$args
+		);
+	}
+
```
pfefferle, line 1665:
```suggestion

	if ( 'Create' === $type && ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC && $content_visibility && ACTIVITYPUB_ACTOR_AND_BLOG_MODE === \get_option( 'activitypub_actor_mode', ACTIVITYPUB_ACTOR_MODE ) ) {
		// @todo transform $data to announce
		add_to_outbox( $data, 'Announce', Actors::BLOG_USER_ID );
	}
```

mattwiebe: Would edits also need to be supported, or do those merely reference the original post? If so might we need to send those direct post updates to followers of the Blog Actor?

pfefferle: you only share the id of a post via an `Announce`, so only `Creates` are important here... the updates will be done automatically afterwards... no need to re-announce them...

mattwiebe: > you only share the id of a post via an Announce, so only Creates are important

Perfect. But the further question:

Let's say I follow a WP `example.com` site from `mattwiebe@mastodon.social` with `ACTIVITYPUB_ACTOR_AND_BLOG_MODE === true`. I follow `example.com@example.com`, but I don't follow `user@example.com`. I receive the `Announce` of a post by `user@example.com`. But the user then edits the post. Will `mastodon.social` get notified of the edit with nobody on the instance following `user@example.com`?

pfefferle: 🤔 good point...

includes/class-activitypub.php
```diff
@@ -518,6 +519,102 @@ private static function register_post_types() {
...
+					return $value;
+				},
+			)
+		);
+
+		\register_post_meta(
+			Outbox::POST_TYPE,
+			'_activitypub_activity_actor',
+			array(
+				'type'              => 'integer',
```
Menrath: Shouldn't the type be string?

includes/functions.php
mattwiebe: So this means that we'll need to also POST to the inboxes of the Blog Actor for`Update`s. Are there any other actions you can think of that should be included in that behaviour?

I'll add a commit to #1186 to do that.

pfefferle: I would start with these two

pfefferle: (and I am still not sure that update needs to be federated)

pfefferle: the current code supports `Create`, `Update` and `Delete` https://github.com/Automattic/wordpress-activitypub/blob/1f7abb76ef4fa171ec7d120eaa03eac3913d3c78/includes/class-activity-dispatcher.php#L104

includes/collection/class-outbox.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @param \Activitypub\Activity\Base_Object $activity_object    The Activity-Object  to add as JSON.
+	 * @param string                            $activity_type      The activity type.
+	 * @param int                               $user_id            The user ID.
+	 * @param string                            $content_visibility Optional. The visibility of the content. Default 'public'.
+	 *
+	 * @return false|int|\WP_Error The added item or an error.
+	 */
+	public static function add( $activity_object, $activity_type, $user_id, $content_visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) { // phpcs:ignore
+		switch ( $user_id ) {
+			case -1:
```
Menrath: Shouldn't the constants `Activitypub\Collection\Actors::BLOG_USER_ID` and `Activitypub\Collection\Actors::APPLICATION_USER_ID` be used?

includes/collection/class-outbox.php
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Outbox collection file.
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\Collection;
+
+/**
+ * ActivityPub Outbox Collection
```
Menrath, line 11:
```suggestion
 * ActivityPub Outbox Collection
 *
 * @link https://www.w3.org/TR/activitypub/#outbox
```

includes/collection/class-outbox.php
```diff
@@ -0,0 +1,75 @@
...
+ */
+class Outbox {
+	const POST_TYPE = 'ap_outbox';
+
+	/**
+	 * Add an Item to the outbox.
+	 *
+	 * @param \Activitypub\Activity\Base_Object $activity_object    The Activity-Object  to add as JSON.
```
Menrath:
```suggestion
	 * @param \Activitypub\Activity\Base_Object $activity_object    The object of the activity that will be added to the outbox.
```

includes/collection/class-outbox.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @param string                            $activity_type      The activity type.
+	 * @param int                               $user_id            The user ID.
```
Menrath: The real or imaginary user ID of the actor that published the activity that will be added to the outbox.
```suggestion
	 * @param int                               $user_id            The real or imaginary user ID of the actor that published the activity that will be added to the outbox.
```

includes/collection/class-outbox.php
```diff
@@ -0,0 +1,75 @@
...
+				break;
+			default:
+				$actor = 'user';
+				break;
+		}
+
+		$outbox_item = array(
+			'post_type'    => self::POST_TYPE,
+			'post_title'   => $activity_object->get_id(),
+			'post_content' => $activity_object->to_json(),
```
Menrath: Maybe always using `post_content_filtered` to store the JSON could make the beneath removal of filter needless.

includes/collection/class-outbox.php
```diff
@@ -0,0 +1,75 @@
...
+			// ensure that user ID is not below 0.
```
Menrath, line 45: It would be helpful to know why this has to be ensured.

includes/functions.php
```diff
@@ -1547,3 +1549,48 @@ function is_self_ping( $id ) {
...
+		return false;
+	}
+
+	$activity = $transformer->to_object();
+
+	if ( ! $activity || is_wp_error( $activity ) ) {
+		return false;
+	}
+
+	set_wp_object_state( $data, 'federate' );
```
Menrath, line 1593: The namings within the function set_wp_object_state should probably be updated as well.

includes/scheduler/class-post.php
```diff
@@ -0,0 +1,113 @@
...
+			return;
+		}
+
+		// Do not send activities if post is password protected.
+		if ( \post_password_required( $post ) ) {
+			return;
+		}
+
+		// Check if post-type supports ActivityPub.
+		$post_types = \get_post_types_by_support( 'activitypub' );
```
Menrath: Wouldn't it be better if this check is in the `is_post_disabled` function? Because that one allows third parties to filter. If I want to disable a post type in general, but have one I want to still federate, then I also need to take care of my own scheduler.

Menrath (commented):
I have looked through that code, and am really looking forward to see this in production. I made some comments, but didn't find anything severe.

I will test my Event Bridge for ActivityPub plugin against this PR in the next days and maybe have some more feedback then. 

Keep on the amazing work!

includes/functions.php
pfefferle: this is still only for WP_Objects and might be deprecated over time

includes/collection/class-outbox.php
pfefferle: because WP_Post does not support negative ints

includes/collection/class-outbox.php
pfefferle: Addressed here: https://github.com/Automattic/wordpress-activitypub/pull/1200/commits/4d9f1a40f6401a16db2ccc19e5e7b4eeeb036860

includes/functions.php
mattwiebe: Perfect. I did `Update` but I added `Delete` as well - https://github.com/Automattic/wordpress-activitypub/pull/1186/commits/2caab9bdd6f7488f944fce5b6a132b0413f298d8

includes/functions.php
```diff
@@ -1547,3 +1551,121 @@ function is_self_ping( $id ) {
...
 	return false;
 }
+
+/**
+ * Add an object to the outbox.
+ *
+ * @param mixed   $data               The object to add to the outbox.
+ * @param string  $activity_type      The type of the Activity.
+ * @param integer $user_id            The User-ID.
+ * @param string  $content_visibility The visibility of the content.
```
akirk: Could you add the valid strings in the documentation here? Or where you can find them.

pfefferle: I added the information

obenland (approved):

obenland (approved):

Issue #211, opened 2022-12-07
BIO and Avatar


marc0janssen:
Hi There,

Question: Sinds this a beta software, are the BIO en AVATAR "sync" not just implemented?

I don't see a BIO and the avatar is empty when looking up the account on mastodon.

pfefferle:
The plugin uses the author description as bio and the gravatar as avatar.

marc0janssen:
Waited 6 hours or so... still no avatar and bio from Wordpress activatypub... any hints?

marc0janssen:
Thanks Django. I tested this. But no result yet. Also hashtags are no coming through.

marc0janssen:
The BIO text is updated. SO this seems to work.

Hashtages and avatar are not coming through.

Do you have a hint or so?

mediaformat:
I wonder if it's because of the ports, if you lookup your webfinger you will see `domain:443/`

marc0janssen:
This is the output.

`Request Log

21:09:38 Looking up WebFinger data for acct:marco@theattic.mjanssen.nl
21:09:38 GET https://theattic.mjanssen.nl/.well-known/webfinger?resource=acct%3Amarco%40theattic.mjanssen.nl
JSON Resource Descriptor (JRD)

{
  "subject": "acct:marco@theattic.mjanssen.nl",
  "aliases": [
    "acct:marco@theattic.mjanssen.nl",
    "https://theattic.mjanssen.nl:443/author/marco/",
    "mailto:marco@mjanssen.nl"
  ],
  "links": [
    {
      "rel": "http://webfinger.net/rel/profile-page",
      "type": "text/html",
      "href": "https://theattic.mjanssen.nl:443/author/marco/"
    },
    {
      "rel": "http://webfinger.net/rel/avatar",
      "href": "https://secure.gravatar.com/avatar/df726990448c149512cd5ced0f5718dd?s=96&d=mm&r=g"
    },
    {
      "rel": "http://webfinger.net/rel/profile-page",
      "type": "text/html",
      "href": "https://theattic.mjanssen.nl:443"
    },
    {
      "rel": "self",
      "type": "application/activity+json",
      "href": "https://theattic.mjanssen.nl:443/author/marco/"
    },
    {
      "rel": "http://nodeinfo.diaspora.software/ns/schema/2.0",
      "href": "https://theattic.mjanssen.nl:443/wp-json/nodeinfo/2.0"
    },
    {
[...]

PR #599, opened 2023-12-11
Add timestamp to icon url of class-user.php

andrigamerita:
As complained in #211, when a user updates their Gravatar profile picture, the change fails to be reflected on many ActivityPub instances that already know the user. This is usually because, when changing the Gravatar image, since the URL remains the same, instances won't try to redownload the data, and thus their media cache isn't updated. From my testing, simply adding a changing dummy parameter to the URL hash returned in the user's JSON is a change that is noticed by other instances, that will try to redownload the image from Gravatar (my experience is minutes for Misskey, and ~1 day for Mastodon).

## Proposed changes:

A decent solution to this problem is thus to add the current day string to the URL hash ($icon as returned by get_icon() as 'url'), in the format Ymd, which will result in the URL changing every day, which should be enough for any user, while at the same time avoiding overloading instances which try to do frequent refreshes. (See https://github.com/Automattic/wordpress-activitypub/issues/211#issuecomment-1848997738)

## Testing instructions:

* Change your profile picture on Gravatar without this change in the code
* And notice how even after days your image change shouldn't be reflected on any instance
* Now change the code as in this PR, and notice how in some time the change will get reflected


pfefferle:
Is this really working? I think the Avatar is cached and would not be requested at all, until there will be a profile update call.

PR #610, opened 2023-12-16
replace `excerpt_more` with custom filter `activitypub_excerpt_more`  for better consistency

### LINKED ISSUE #558: Excerpt way too short
### Quick summary

The Excerpt is way shorter than 400 Characters

(The Excerpt also shows a Link)

## Settings
grafik

## Post
grafik

### Steps to reproduce

Add Excerpt to the Posts
It's sometimes just two words

The Excerpt should be 400 Characters and without link

### What actually happened

It's just two words with a Link (the Link isn't always there)

### Impact

One

### Available workarounds?

### Logs or notes

Theme: Neve

kanru:
Fixes #558 

## Proposed changes:

* `excerpt_more` might already be customized by themes. Using `activitypub_excerpt_more` to allow more specific control.

### Other information:

## Testing instructions:

* Use a theme with custom `excerpt_more` filter, for example twentyfiteen.
* The `[ap_excerpt]` shortcode should not be affected by the `excerpt_more` filter.

### DIFF
--- a/includes/class-shortcodes.php
+++ b/includes/class-shortcodes.php
@@ -127,8 +127,7 @@ class Shortcodes {
 		// Strip out any remaining tags.
 		$excerpt = \wp_strip_all_tags( $excerpt );
 
-		/** This filter is documented in wp-includes/formatting.php */
-		$excerpt_more = \apply_filters( 'excerpt_more', ' [...]' );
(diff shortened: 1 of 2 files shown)


pfefferle (approved):
Thanks ☺️

PR #615, opened 2023-12-20
show admin notice when permalink structure is plain

### LINKED ISSUE #609: Feature Request: Flag when permalink structure is "plain"
### What

I want wordpress to flag an issue for the administrator if the plugin is installed on a system with a "plain" permalink structure.

### Why

When setting up the plugin, you can't use a permalink structure of "plain".  Anything else works.

I'm not sure exactly why that is, but I think it's because "plain" means wordpress doesn't add any rewrite rules, which means webfinger URLs don't work.  I guess that's a good reason.

So, the plugin should notice when it's installed with such a permalink structure and raise this as an issue for the administrator to fix.

### How

_No response_

mexon:
Fixes #609

## Proposed changes:

This displays an admin notice if the permalink structure is plain.  This is because the plugin cannot function correctly with such a permalink structure, it prevents webfinger from working.

### Other information:

Tests have been added.

### Implementation notes

I am not sure about the implementation.  The admin notice is pretty bold.  Perhaps a more subtle notice makes sense?

The implementation is designed to allow for many warnings.  At the moment there is only this one.  I imagine in future other issues could be detected here.  For example there is currently conflicts with the One User Avatar and Comment Like Dislike plugins.  This would be an appropriate mechanism to flag those issues.

## Testing instructions:

Testing:

* Create a fresh wordpress instance, but do not set a permalink structure.
* Install the activitypub plugin
* Observe administrator console.  An error about the permalink structure should be displayed.
* Change the permalink structure to anything else.
* Observe that the error has disappeared.

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -21,8 +21,8 @@ class Admin {
 		\add_action( 'personal_options_update', array( self::class, 'save_user_description' ) );
 		\add_action( 'admin_enqueue_scripts', array( self::class, 'enqueue_scripts' ) );
 
-		Admin::$admin_notices = Admin::get_admin_notices();
-		if ( !empty( Admin::$admin_notices ) ) {
+		self::$admin_notices = self::get_admin_notices();
+		if ( ! empty( self::$admin_notices ) ) {
 			\add_action( 'admin_notices', array( self::class, 'show_admin_notices' ) );
 		}
 
@@ -71,7 +71,7 @@ class Admin {
 	 * Display admin menu notices about configuration problems or conflicts
 	 */
 	public static function show_admin_notices() {
-		foreach ( Admin::$admin_notices as $admin_notice ) {
+		foreach ( self::$admin_notices as $admin_notice ) {
 			?>
 
 			<div class="notice notice-error">
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -10,8 +10,6 @@ use Activitypub\Model\Blog_User;
  * @author Matthias Pfefferle
  */
 class Admin {
-	private static $admin_notices = array();
-
 	/**
 	 * Initialize the class, registering WordPress hooks
 	 */
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -20,11 +20,7 @@ class Admin {
 		\add_action( 'admin_init', array( self::class, 'register_settings' ) );
 		\add_action( 'personal_options_update', array( self::class, 'save_user_description' ) );
 		\add_action( 'admin_enqueue_scripts', array( self::class, 'enqueue_scripts' ) );
-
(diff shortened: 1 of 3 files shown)


includes/class-admin.php
```diff
@@ -46,6 +53,35 @@ public static function admin_menu() {
...
 
+	/**
+	 * Collect admin menu notices about configuration problems or conflicts
+	 */
+	public static function get_admin_notices() {
+		$admin_notices = array();
+
+		$permalink_structure = \get_option( 'permalink_structure' );
+		if ( empty( $permalink_structure ) ) {
+			array_push( $admin_notices, 'You are using the ActivityPub plugin without setting a permalink structure.  This will prevent ActivityPub from working.  Please set a permalink structure.' );
```
pfefferle, line 64: why not adding the admin notice directly here?

mexon: Possibly over-engineering, but I think there are potentially many conflicts, environment problems, or misconfigurations that could be checked here.  They would all be displayed the same.  So it seems like repeating all of the display block for each one is a bit weird.  It's also an instinctive attempt to separate business logic from display logic.  If you prefer, I can certainly do it that way.

pfefferle: I like the `get_admin_notices` function, but I would avoid the static `$admin_notices` var. So why not call `show_admin_notice( $admin_notice, 'error' )` to build the HTML directly here instead of pushing it to the array? this way we could add others later on?!?

This way we might also remove the `! \WP_Block_Type_Registry::get_instance()->is_registered( 'activitypub/followers' )` check?!?

mexon: OK, sounds good.  I'll implement that when I get a chance.  Thanks for the review!

includes/class-admin.php
```diff
@@ -46,6 +49,30 @@ public static function admin_menu() {
...
+		}
+	}
+
+	/**
+	 * Display one admin menu notice about configuration problems or conflicts
+	 */
+	private static function show_admin_notice( $admin_notice, $level ) {
+		?>
+
+		<div class="notice notice-<?php echo wp_kses( $level, 'data' ); ?>">
```
pfefferle, line 69: I would use `esc_attr` instead of `wp_kses` here, because we do not want to allow html at all!

includes/class-admin.php
pfefferle: thanks a lot!

pfefferle (dismissed):

pfefferle (approved):

pfefferle:
Thanks @mexon 😊

mexon:
Oh merged already 😀  Thanks for the tidy-up!

pfefferle:
I hoped to release the new version before Xmas so I was in hurry, but I think I have to test a bit more!

PR #618, opened 2023-12-21
Add nodeinfo metadata

Menrath:
This is a very tiny pull request that adds some metadata fields for the nodeinfo representation.

Both `nodeName` and `nodeDescription` are already widely adopted.
More info to the current status can be found here: https://codeberg.org/thefederationinfo/nodeinfo_metadata_survey

`nodeIcon` is not used yet, I proposed it myself as it could yield to use cases later. Most WordPress sides are highly customized and styled. They have a favicon that people already identify the site with. Maybe other software will make use of this in the feature, like Mobilizon when following the application actor of a WordPress site (see https://framagit.org/framasoft/mobilizon/-/merge_requests/1513).

### DIFF
--- a/includes/rest/class-nodeinfo.php
+++ b/includes/rest/class-nodeinfo.php
@@ -106,8 +106,8 @@ class Nodeinfo {
 		);
 
 		$nodeinfo['metadata'] = array(
-			'nodeName' =>  \get_bloginfo( 'name' ),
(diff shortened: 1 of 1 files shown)


pfefferle:
Can you also add the changes to `integration/class-nodeinfo.php` if someone uses the NodeInfo plugin?

pfefferle (approved):

PR #624, opened 2023-12-26
Add improved classic editor image detection

### LINKED ISSUE #553: Images showing in ActivityPub feed even though they are not in the published post
### Quick summary

When authoring a post I changed my mind about which image to include a couple of times. Only the latest image is included in the published version but the fediverse feed still embeds all 3

### Steps to reproduce

1. In /wp-admin/edit.php click to Add New Post
2. Insert an image to the post
3. Remove the image from the post
4. Add a different image to the post
5. Publish the post

The ActivityPub version of the post to include only the image(s) that are in the published post
https://blog.mattedgar.com/2023/11/04/weeknote-30-october-to-3-november-2023/

### What actually happened

The ActivityPub version of the post includes images that were inserted then deleted from the draft before publication
https://mastodon.me.uk/@blog.mattedgar.com@blog.mattedgar.com/111353363264938089

### Impact

All

### Available workarounds?

### Logs or notes

I used Classic Editor on wordpress.com; not tested with Blocks editor or non-WP.com blogs
Privacy risk if users are unaware of the bug, and inadvertently leak private information that they have decided not to share

### LINKED ISSUE #612: Posts pictures added and subsequently removed from the post
### Quick summary

I uploaded images to the media library and decided I did not want to use one of them. I removed the image from the post. When previewing and publishing the post, the image was not publicly present on the blog. However, it *was* visible to subscribed mastodon users. Same issue exists when using the Share on Mastodon plugin https://wordpress.org/plugins/share-on-mastodon/ Had to entirely delete the image from the media library to resolve the issue.

### Steps to reproduce

1. Upload images to media library
2. Include images in the blog post
3. Preview post
4. Remove one of the images
5. Publish the post
6. Observe that the un-used image is present in my social media feed
7. Delete the image from the media library
9. Re-share

I expected the publish process to share only images I explicitly stipulate in the blog post.

### What actually happened

Images that I did not want to include in the post were shared.

### Impact

One

### Available workarounds?

Yes, easy to implement

### Logs or notes

_No response_

Sentynel:
Fixes #612 
Fixes #553 

I note that work on this issue hasn't been officially prioritised due to the deprecation of the classic editor, but we're still using it so here we are.

## Proposed changes:

* If we're using the classic editor on a new enough version of WordPress to have the `WP_HTML_Tag_Processor` available, use this to pull all `` tags out of the post. Grab their sources, and if they point at our uploads directory, use them in the federated representation of the post.
* There's a slightly hairy bit of code to handle different-sized versions of images - I played around with a few options here including trying to parse `srcset` but decided the simplest and most reliable option is to attempt to look the URL up directly, and _iff_ that fails use a regex to drop suffixes of the type `-500x500` or `-scaled` and look that up. If that also fails, give up. (This works fine if you're using a plugin for alternate `.webp` representations as well, as these show up in a different bit of the `<picture>` tag in this case.)
* If you don't have WP6.2+ for some reason, the old behaviour is retained.

### Other information:

## Testing instructions:
1. Use the classic editor.
2. Create a post.
3. Upload and add an image to the post.
4. Remove the image.
5. Add a different, previously uploaded image to the post.
6. Publish and look at the federated version.

Prior to this patch, you'll see the uploaded image and not the added image. With this patch, you correctly see the image(s) which are actually present in the post. (Note: We still preserve showing the featured/thumbnail image first, even if it's not actually in the post.)

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -188,7 +188,7 @@ class Post extends Base {
 		$image_ids = array();
 		$query = new \WP_Query(
 			array(
-				'post_parent' => $id,
+				'post_parent' => $this->object->ID,
 				'post_status' => 'inherit',
 				'post_type' => 'attachment',
 				'post_mime_type' => 'image',
@@ -221,6 +221,9 @@ class Post extends Base {
 		$base = \wp_get_upload_dir()['baseurl'];
 		$content = \get_post_field( 'post_content', $this->object );
 		$tags = new \WP_HTML_Tag_Processor( $content );
+		// This linter warning is a false positive - we have to
+		// re-count each time here as we modify $image_ids.
+		// phpcs:ignore Squiz.PHP.DisallowSizeFunctionsInLoops.Found
 		while ( $tags->next_tag( 'img' ) && ( \count( $image_ids ) < $max_images ) ) {
 			$src = $tags->get_attribute( 'src' );
 			// If the img source is in our uploads dir, get the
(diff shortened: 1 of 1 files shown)


pfefferle (approved):
Nice! Thanks a lot!

pfefferle:
> I note that work on this issue hasn't been officially prioritised due to the deprecation of the classic editor, but we're still using it so here we are.

We are always open to PRs even more if they have no high priority!

PR #625, opened 2023-12-27
Add readme section for "WordPress in sub directory" with different si…

### LINKED ISSUE #538: [Addition] ReadMe section on "wordpress in subdirectory" rewrites if wp_siteurl != wp_home
Hi,

I stumbled across the [readme](https://github.com/Automattic/wordpress-activitypub/blob/master/README.md#what-if-you-are-running-your-blog-in-a-subdirectory) section describing:

> What if you are running your blog in a subdirectory?
>
> In order for webfinger to work, it must be mapped to the root directory of the URL on which your blog resides.
>
> Apache
>
>Add the following to the .htaccess file in the root directory:
>
>RedirectMatch "^\/\.well-known/(webfinger|nodeinfo|x-nodeinfo2)(.*)$" /blog/.well-known/$1$2
>
>Where 'blog' is the path to the subdirectory at which your blog resides.
>
>(…)

In my WordPress install I have wordpress in its own directory (/blog/), but  it is answering directly under "/".

( the [method II described here](https://wordpress.org/documentation/article/giving-wordpress-its-own-directory/) )

In this case, the Redirect mentioned above section is not needed, since the index.php directly under "/" takes care.

I discovered only after `curl` ing the .well-known/webfinger url that the redirect in this cae will produce a 404 for /blog/.well-known/webfinger.

After deleting the RedirectMatch again in the top-level .htaccess, the fediverse profile works.

I'm not sure if this applies only to my setting or in general in such a scenario with the copied index.php and .htaccess and modifications under the root.

So maybe another section 

===

What if you are running your blog in a subdirectory, but have a different wp_siteurl?

In that that case you don't need the RedirectMatch, because the index.php will take care of that.

===

Could help clarify that.

Thanks for the great plugin and the work!

iamwebrocker:
Adds readme section for "WordPress in sub directory" rewirte when site_url differs from home url

Fixes #538

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -101,6 +101,10 @@ Add the following to the site.conf in sites-available:
 
 Where 'blog' is the path to the subdirectory at which your blog resides.
 
+= What if you are running your blog in a subdirectory, but have a different [wp_siteurl](https://wordpress.org/documentation/article/giving-wordpress-its-own-directory/)? =
+
+In that that case you don't need the redirect, because the index.php will take care of that.
+
 == Changelog ==
 
 Project maintained on GitHub at [automattic/wordpress-activitypub](https://github.com/automattic/wordpress-activitypub).
--- a/readme.txt
+++ b/readme.txt
@@ -103,7 +103,7 @@ Where 'blog' is the path to the subdirectory at which your blog resides.
 
 = What if you are running your blog in a subdirectory, but have a different [wp_siteurl](https://wordpress.org/documentation/article/giving-wordpress-its-own-directory/)? =
 
-In that that case you don't need the redirect, because the index.php will take care of that.
+In that case you don't need the redirect, because the index.php will take care of that.
 
 == Changelog ==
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -101,6 +101,10 @@ Add the following to the site.conf in sites-available:
 
 Where 'blog' is the path to the subdirectory at which your blog resides.
 
+= What if you are running your blog in a subdirectory, but have a different [wp_siteurl](https://wordpress.org/documentation/article/giving-wordpress-its-own-directory/)? =
+
+In that that case you don't need the redirect, because the index.php will take care of that.
```
pfefferle, line 106: Duplicate "that" ☺️

iamwebrocker: oopsie… sry

pfefferle (approved):

pfefferle:
thanks @iamwebrocker :)

PR #629, opened 2024-01-02
Add/event objects

Menrath:
I started this pull request as a minimal one, with only adding an Event class but I ended up also shaping the way the Json-LD context is build a bit. It is now tied to the ActivityPub objects.

Reason: Don't store and send context which is never referenced.


pfefferle:
The `context` handling is a bit complicated and confusing.

Can we maybe find a way to have only the context attribute and work with `set_context` to replace it and `add_context` to extend it instead of using `_context`?

templates/author-json.php
```diff
@@ -1,10 +1,6 @@
 <?php
 $user = \Activitypub\Collection\Users::get_by_id( \get_the_author_meta( 'ID' ) );
 
-$user->set_context(
```
pfefferle, line 4: I used that to set it only where needed. With your changes, we have the context also on the items in the collections, for example in the outbox and in sub-objects like the `Object` attribute. But we only want the context on the top level.

In the following example we have 3 different context definitions for example:

```JSON
{
    "@context": [
        "https:\\/\\/www.w3.org\\/ns\\/activitystreams"
    ],
    "id": "http:\\/\\/localhost:8076\\/wp-json\\/activitypub\\/1.0\\/users\\/1\\/outbox",
    "generator": "http:\\/\\/wordpress.org\\/?v=6.4.2",
    "actor": "http:\\/\\/localhost:8076\\/author\\/test\\/",
    "type": "OrderedCollectionPage",
    "partOf": "http:\\/\\/localhost:8076\\/wp-json\\/activitypub\\/1.0\\/users\\/1\\/outbox",
    "totalItems": 15,
    "first": "http:\\/\\/localhost:8076\\/wp-json\\/activitypub\\/1.0\\/users\\/1\\/outbox?page=1",
    "last": "http:\\/\\/localhost:8076\\/wp-json\\/activitypub\\/1.0\\/users\\/1\\/outbox?page=2",
    "next": "http:\\/\\/localhost:8076\\/wp-json\\/activitypub\\/1.0\\/users\\/1\\/outbox?page=2",
    "orderedItems": [
        {
            "@context": [],
            "type": "Create",
            "object": {
                "@context": [
                    "https:\\/\\/www.w3.org\\/ns\\/activitystreams",
                    {
                        "Hashtag": "as:Hashtag"
                    }
                ],
                "id": "http:\\/\\/localhost:8076\\/2023\\/12\\/26\\/classic-editor-images\\/",
                "type": "Note",
                "attachment": [
                    {
                        "type": "Image",
                        "url": "http:\\/\\/localhost:8076\\/wp-content\\/uploads\\/2023\\/12\\/337b3d7a89119c55a1a4c13bf5dd9332.jpeg",
                        "mediaType": "image\\/jpeg"
                    },
                    {
                        "type": "Image",
                        "url": "http:\\/\\/localhost:8076\\/wp-content\\/uploads\\/2023\\/10\\/273020110-810ab794-55f9-4a71-999f-0463921219d2.jpg",
                        "mediaType": "image\\/jpeg"
                    }
                ],
                "attributedTo": "http:\\/\\/localhost:8076\\/author\\/test\\/",
                "content": "<p>&nbsp;<br \\/>Hallo<br \\/>&nbsp;<\\/p><p><a href=\"http:\\/\\/localhost:8076\\/2023\\/12\\/26\\/classic-editor-images\\/\">http:\\/\\/localhost:8076\\/2023\\/12\\/26\\/classic-editor-images\\/<\\/a><\\/p>",
                "contentMap": {
                    "en": "<p>&nbsp;<br \\/>Hallo<br \\/>&nbsp;<\\/p><p><a href=\"http:\\/\\/localhost:8076\\/2023\\/12\\/26\\/classic-editor-images\\/\">http:\\/\\/localhost:8076\\/2023\\/12\\/26\\/classic-editor-images\\/<\\/a><\\/p>"
                },
                "published": "2023-12-26T20:04:02Z",
                "tag": [],
                "url": "http:\\/\\/localhost:8076\\/2023\\/12\\/26\\/classic-editor-images\\/",
                "to": [
                    "https:\\/\\/www.w3.org\\/ns\\/activitystreams#Public",
                    "http:\\/\\/localhost:8076\\/wp-json\\/activitypub\\/1.0\\/users\\/1\\/followers"
                ],
                "cc": []
            },
            "actor": "http:\\/\\/localhost:8076\\/author\\/test\\/",
            "id": "http:\\/\\/localhost:8076\\/2023\\/12\\/26\\/classic-editor-images\\/#activity",
            "published": "2023-12-26T20:04:02Z",
            "to": [
                "https:\\/\\/www.w3.org\\/ns\\/activitystreams#Public",
                "http:\\/\\/localhost:8076\\/wp-json\\/activitypub\\/1.0\\/users\\/1\\/followers"
            ],
            "cc": []
        }
    ]
}
```

Also: If you change it to have different and extendable contexts, we need a way to merge sub contexts (in collections or sub-objects) into the main.

Menrath:
Or maybe don't use a property for the JsonLD context at all, just use a getter function called "get_json_ld_context".
Then there should be no confusion between the jsonld-context ('@context') and the term context ('context') from the ActivityStreams.

pfefferle:
Ah, that is the difference! Makes sense!

templates/author-json.php
Menrath: I had a dream of really getting the a compacted JsonLD context from the objects and their attributes, but its anything but trivial. I gave this idea up, in the end nobody parses it yet.

So right now am fine with the most minimal hard-coded approach as long it does not require to use the context filter for objects like Event. 

And if at the moment we need to change the JsonLD context we can still filter it, or wait for a proper solution until it becomes relevant. If one wants to get things right this would be needed for the address of Events: some send a xsd:Text, other send a PostalAddress (schema.org).

Menrath: And thanks for pointing out those other use cases: I did not have these cases on my radar.

includes/activity/class-postal-address.php
```diff
@@ -0,0 +1,65 @@
+<?php
+/**
+ * PostalAddress is a custom ActivityPub object firstly used by Mobilizon
+ * derived from https://schema.org/PostalAddress.
+ *
+ * @license AGPL-3.0-or-later
+ */
+
+namespace Activitypub\Activity;
```
pfefferle, line 9: Should we use `namespace Activitypub\Activity\Object;` (and move them into a subfolder) instead, to differentiate between core concepts (Object, Actor, Activity) and sub-types?

includes/activity/class-base-object.php
```diff
@@ -622,9 +630,11 @@ public function from_array( $array ) {
 	 * It tries to get the object attributes if they exist
 	 * and falls back to the getters. Empty values are ignored.
 	 *
+	 * @param bool $include_json_ld_context
+	 *
 	 * @return array An array built from the Object.
 	 */
-	public function to_array() {
+	public function to_array( $include_json_ld_context = true ) {
```
pfefferle, line 637: I am not sure this is a good idea, because we have to add this var to every method that calls `to_array`, like `to_json` for example... Is there an easier, more elegant way?

Maybe re-add a class attribute like `private $json_ld_context;` that can be set to false?

I would love to see something like that, to have an easy way to extend it through code, like the `set_json_ld_context` (replaces the const) or `add_json_ld_context` (extends the const).

includes/activity/class-postal-address.php
pfefferle: > The first confusion for me was the namespace starting with Activitypub\Activity\ as 'Activity' is already a subtype.

I see it from that perspective: https://www.w3.org/TR/activitystreams-vocabulary/#types

There are core types, like Object, Link, Activity, IntransitiveActivity, Collection, ... and Extended Types: https://www.w3.org/TR/activitystreams-vocabulary/#extendedtypes

So that's why I thought it might be enough to stick with the core types, to keep it simple... but I agree that a sub-structure helps a lot with complicated formats like events, but let's keep them as sub-types and maybe put them in subfolders/namespaces then.

pfefferle: So you are right, but not with `Activity`, but with `Actor` beeing an extended type, but it makes still sense to have them on top level ideas (at least in my head 🫣 )

pfefferle: Kind of

```
activity/
├─ object/
│  ├─ class-event.php
│  ├─ class-place.php
│  ├─ ...
├─ class-base-object.php
├─ class-actor.php
├─ class-activity.php

```

...and if it makes sense to also have `Activity` subtypes, put them in the sub-directory and the same with `Actors`.

pfefferle: Or maybe that makes more sense: The objects at the main level are all the main "idea" of an Acticity/Object/Actor and you have to make it more concrete using the `type` attribute. For `Actor` these are `Person`, `Application`, ... and for `Object` it is `Event`, `Note`, ... and for `Activity` it is `Create`, `Update`, ...

Menrath:
Converted it to draft to signalize that there is no hurry and we might discover some little enhancements in the next few weeks anyway!

includes/activity/class-base-object.php
```diff
@@ -692,4 +700,13 @@ public function to_json() {
...
 		return \array_keys( \get_object_vars( $this ) );
 	}
+
+	/**
+	 * Returns the JSON-LD context of this object.
+	 *
+	 * @return array $context A compacted JSON-LD context for the ActivityPub object.
+	 */
+	public function get_json_ld_context() {
+		return $this::JSON_LD_CONTEXT;
```
pfefferle, line 710: shouldn't this be `self::JSON_LD_CONTEXT`?
```suggestion
		return self::JSON_LD_CONTEXT;
```

includes/model/class-application-user.php
```diff
@@ -52,7 +50,7 @@ public function get_name() {
 	}
 
 	public function get_preferred_username() {
-		return $this::get_name();
+		return self::get_name();
```
pfefferle, line 53: this is not a static function
```suggestion
		return $this->get_name();
```

includes/rest/class-collection.php
```diff
@@ -186,7 +186,7 @@ public static function featured_get( $request ) {
 	 */
 	public static function moderators_get( $request ) {
 		$response = array(
-			'@context'     => Activity::CONTEXT,
+			'@context'     => Activity::JSON_LD_CONTEXT,
```
pfefferle, line 189: With your changes, I think `Activity::JSON_LD_CONTEXT` is no longer enough, it has to be the Object context now and this can be a bit complicated?!?

includes/transformer/class-base.php
```diff
@@ -84,11 +96,12 @@ public function to_activity( $type ) {
 
 		$activity = new Activity();
 		$activity->set_type( $type );
-		$activity->set_object( $object );
```
pfefferle, line 87: this should be as is, because it pre-fills the Activity with data and it will no longer be done with your change.

pfefferle:
@Menrath I added some changes, can you verify them and I had one more inline question!

Menrath (commented):
I agree with that. Great that you saw that, it was probably a leftover before the function change_wp_user_id was available: Mobilizon currently requires the actor attribute also on the event object itself, although it is against the ActivityPub standard.

Menrath:
@pfefferle Thanks for your review again! Looks good to me!

A slightly off topic last questionthat comes to my mind: what about calling the `to_object` function `to_activitypub_object`?
For improving searchability of the where its used.

Menrath (commented):
LGTM

includes/activity/class-base-object.php
```diff
@@ -709,6 +709,6 @@ public function get_object_var_keys() {
 	 * @return array $context A compacted JSON-LD context for the ActivityPub object.
 	 */
 	public function get_json_ld_context() {
-		return self::JSON_LD_CONTEXT;
+		return static::JSON_LD_CONTEXT;
```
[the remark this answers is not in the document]
pfefferle: no, this is totally fine and thanks for the help (wasn't aware of it)!

pfefferle (commented):
oops 😳

thanks!

Menrath (commented):
Mixed licensing. I have no experience to be honest. I am fine with it. Our extensions plugin will be A-GPLv3 only though :)

pfefferle (approved):

carstingaxion:
What an enormous PR! Congrats and Thank you very much for your contribution to @Menrath and @pfefferle 

… and all this even without an issue ;)

PR #630, opened 2024-01-04
Improve the property names of the transformer classes

Menrath:
I tried to improve the readability and the efficiency of certain custom transformers via:

- Renaming the property 'object' to 'wp_object' of the _Transformer_ classes  to indicate that its the source
- Adding an optional property 'activitypub_object' which may make already transformed parts available to other "getter" functions.
- Enhancing some PHP-Docs

### DIFF
--- a/includes/transformer/class-base.php
+++ b/includes/transformer/class-base.php
@@ -1,6 +1,9 @@
 <?php
 namespace Activitypub\Transformer;
 
+use WP_Post;
+use WP_Comment;
+
 use Activitypub\Activity\Activity;
 use Activitypub\Activity\Base_Object;
 
@@ -12,18 +15,20 @@ use Activitypub\Activity\Base_Object;
  */
 abstract class Base {
 	/**
-	 * The WP_Post object.
+	 * The WP_Post or WP_Comment object.
 	 *
-	 * @var
+	 * This is the source object of the transformer.
+	 *
+	 * @var WP_Post|WP_Comment
 	 */
-	protected $object;
+	protected $wp_object;
 
 	/**
 	 * Static function to Transform a WordPress Object.
 	 *
 	 * This helps to chain the output of the Transformer.
 	 *
-	 * @param stdClass $object The WP_Post object
+	 * @param WP_Object|WP_Comment $object The WP_Post object
 	 *
 	 * @return void
 	 */
@@ -34,10 +39,10 @@ abstract class Base {
 	/**
 	 * Base constructor.
 	 *
-	 * @param stdClass $object
+	 * @param WP_Object|WP_Comment $wp_object
 	 */
-	public function __construct( $object ) {
-		$this->object = $object;
+	public function __construct( $wp_object ) {
+		$this->wp_object = $wp_object;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


includes/transformer/class-base.php
```diff
@@ -46,9 +61,9 @@ public function __construct( $object ) {
 	 * @return Activitypub\Activity\Base_Object
 	 */
 	public function to_object() {
-		$object = new Base_Object();
+		$this->activitypub_object = new Base_Object();
 
-		$vars = $object->get_object_var_keys();
+		$vars = $this->activitypub_object->get_object_var_keys();
```
pfefferle, line 66: why storing it in a local var?

pfefferle: it is not used outside of this function, so it might be more confusing than it would help?!?

Menrath: You are right. From this Pull Requests point of view custom transformers might override the to_object function anyway. And they could also define their own protected property for storing the already processed part of the object. 

I would suggest reverting this part of the code and deal with the performance optimizations by reusing already transformed parts later. One would also then have to take care of the race-conditions which properties gets transformed when.

includes/transformer/class-base.php
```diff
@@ -12,18 +15,20 @@
...
-	protected $object;
+	protected $wp_object;
 
 	/**
 	 * Static function to Transform a WordPress Object.
 	 *
 	 * This helps to chain the output of the Transformer.
 	 *
-	 * @param stdClass $object The WP_Post object
+	 * @param WP_Object|WP_Comment $object The WP_Post object
```
pfefferle, line 31:
```suggestion
	 * @param stdClass|WP_Post|WP_Comment $wp_object The WordPress object
```

includes/transformer/class-base.php
```diff
@@ -34,10 +39,10 @@ public static function transform( $object ) {
 	/**
 	 * Base constructor.
 	 *
-	 * @param stdClass $object
+	 * @param WP_Object|WP_Comment $wp_object
```
pfefferle, line 42:
```suggestion
	 * @param stdClass|WP_Post|WP_Comment $wp_object The WordPress object
```

includes/transformer/class-base.php
```diff
@@ -12,18 +15,20 @@
  */
 abstract class Base {
 	/**
-	 * The WP_Post object.
+	 * The WP_Post or WP_Comment object.
+	 *
+	 * This is the source object of the transformer.
 	 *
-	 * @var
+	 * @var WP_Post|WP_Comment
```
pfefferle, line 22:
```suggestion
	 * @var stdClass|WP_Post|WP_Comment
```

includes/transformer/class-base.php
Menrath: Do you already have use cases in mind where a non WP_Post or WP_Comment class is passed? Just curious! Of course I am fine with accepting that suggestion.

I know that for example 'The Events Calendar' offers it's own object with handy functions, but one could always create an instance of that within the transformer by the WP_Post object.

pfefferle: I fear that not all WordPress plugins work with custom post types, but I am not sure... You are the expert there... We can remove `stdClass`.

pfefferle (approved):
LGTM!

includes/transformer/class-base.php
Menrath: Some indeed use custom tables to handle complex functionality like recurring events, e.g. The Events Calendar. But that functionality is still built on top of a custom post type.

WordPress has been so clear for decades to use Custom Post Types wherever possible. I have not crossed any plugin that violated that suggestion.

pfefferle (approved):

Issue #633, opened 2024-01-05
Change state of comment back to moderation after comment got updated


jaschaurbach:
### Quick summary

I just found out that after an already published comment from activitybub is changed the changed comment just is there without any oversight or chance to check it out before it is published now. 

Anyone could put anything on my blog after a comment got approved and that is bad.

### Steps to reproduce

1. comment via Activitypub on an article
2. waqit till comment is approved
3. change comment
4. See changed comment without further check by the site owner

### What you expected to happen

I'd expect that the comment get back in moderations state.

### What actually happened

The comment just changed,

### Impact

All

### Available workarounds?

No but the platform is still usable

Schoeneh:
I consider this bug / missing feature a big moderation issue.
As long as it's not fixed, I'm not using this (otherwise great) plugin.

janboddez:
It looks like we _could_ leverage the following hook to temporarily work around this:

https://github.com/Automattic/wordpress-activitypub/blob/4c297acfd454a13260900ff0f97a9e3f425e960c/includes/handler/class-update.php#L76

Except I'm not sure what the 2nd parameter even is? Why is it `null`? This seems to be the only time this action is called?

So, untested, but you could try putting something like this in a mu-plugin or your theme's functions.php?

```
add_action( 'activitypub_handled_update', function( $activity, $second_param, $state, $reaction ) {
	if ( $reaction instanceof \WP_Comment ) {
		wp_set_comment_status( $reaction, 'hold' );
	}
}, 10, 4 );
```

Note that this wouldn't actually check if anything got changed (yet)!

janboddez:
So, that didn't work.

In fact, this whole bit of code is kinda strange:
```
		$state    = Interactions::update_comment( $activity );
		$reaction = null;

		if ( $state && ! \is_wp_error( $reaction ) ) {
			$reaction = \get_comment( $state );
		}

		\do_action( 'activitypub_handled_update', $activity, null, $state, $reaction );
```
`$reaction` is literally set equal to `null`, so `\get_comment( $state );` will never run.

But ... `$state` seems to actually be the comment! Let me edit my other response.

janboddez:
OK, forget it.

https://github.com/Automattic/wordpress-activitypub/blob/4c297acfd454a13260900ff0f97a9e3f425e960c/includes/collection/class-interactions.php#L138

Doesn't actually return a comment. It returns `1` if the update was successful.

[...]

PR #634, opened 2024-01-05
Address some of things found in #633

janboddez:
Fixes some of the things found in #633

## Proposed changes:

* So, `wp_update_comment()` doesn't actual return a comment ID.
* So I made it so that `Interactions::update_comment()` takes this into account and returns the `$commentdata` array, but only when the update was successful.
* If the update somehow failed, `Interactions::update_comment()` either returns `false` or `'invalid'` (just like before) or whatever is returned by `wp_update_comment()` (which can be an error, or `0`, or even `false`). Actually, in this case, the function behaves just like it did before.
* Subsequently, in the "update handler," I replaced `\get_comment( $state )` with something that I think makes a bit more sense (as mentioned, `$state` isn't actually a comment ID).

### Other information:

## Testing instructions:

The way I tested this, is I created a callback in a separate mu-plugin like so:
```
add_action( 'activitypub_handled_update', function( $activity, $second_param, $state, $reaction ) {
	if ( $reaction instanceof \WP_Comment ) {
		wp_set_comment_status( $reaction, 'hold' );
	}
}, 99, 4 );
```
Thing is, and this might be another bug but I left it untouched for now, I have no clue what the second param really is.

Prior to this PR, `$reaction` would always be `null` or the comment with ID `1` (but, as mentioned, the `1` returned by `wp_update_comment()` isn't actually a comment ID).

With this PR, `$reaction` is either a comment object or `null`. And `$state` basically the result of the `wp_update_comment()` call, kind of like before.

Either way, this short callback addresses #633. So if this PR were merged, site admins or developers would be able to much more easily force re-approval of updated comments.

### DIFF
--- a/includes/collection/class-interactions.php
+++ b/includes/collection/class-interactions.php
@@ -102,7 +102,7 @@ class Interactions {
 	 *
 	 * @param array $activity The activity-object
 	 *
-	 * @return array|string|false The commentdata or false on failure
+	 * @return array|string|int|\WP_Error|false The commentdata or false on failure
 	 */
 	public static function update_comment( $activity ) {
 		$meta = get_remote_metadata_by_actor( $activity['actor'] );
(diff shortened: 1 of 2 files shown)


janboddez:
So the important thing is this change attempts to take into account all the different return types of `Interactions::update_comment()`.

And then sets `$state` and `$reaction` (which in the original code would _always_ be `null`) based on that.

pfefferle (approved):
LGTM; Thanks :)

PR #643, opened 2024-01-10
Fix classic editor image finding for large images

Sentynel:
The filename with -scaled in it is used as the primary key in this case (for images wider than 2560px, by default). If a small version of one of these images, say `foo-500x500,jpg`, is embedded into a post, the correct primary URL to look up is `foo-scaled.jpg`. Sorry I didn't catch this before!

Fixes #

## Proposed changes:
* Fix lookups to try the -scaled URL correctly.

### Other information:

## Testing instructions:
* Upload an image wider than 2560 pixels.
* Insert this into a post (but don't make it the featured image).
* View the post via AP.
* Confirm that the image is now correctly attached.

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -242,16 +242,27 @@ class Post extends Base {
 			$src = $tags->get_attribute( 'src' );
 
 			// If the img source is in our uploads dir, get the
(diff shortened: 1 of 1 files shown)


pfefferle (approved):
LGTM

Issue #642, opened 2024-01-09
Federation of comments does not work as expected


jaschaurbach:
### Quick summary

I answered a comment from the fediverse on my blog. The comment is here: https://jascha.wtf/schoeffe-arbeitszeit-und-entschaedigung/#comment-2977

The comment is indeed published in the fediverse but it misses the reference:

<img width="335" alt="Bildschirmfoto 2024-01-09 um 20 55 17" src="https://github.com/Automattic/wordpress-activitypub/assets/2017105/cf5874f0-6db6-4c20-8ea0-8d5ca1928104">

### Steps to reproduce

Answer to a fediverse comment from your blog 

### What you expected to happen

the answer gets posted and has a reference to the post answering to.

### What actually happened

it has no reference nor mention.

### Impact

One

### Available workarounds?

No but the platform is still usable

pfefferle:
Was the comment you replied to before the update to 2.0.0?

pfefferle:
I added a comment to https://jascha.wtf/gedicht-slash-slash/ from Mastodon, can you reply to it?

jaschaurbach:
Ah yes, it worked! Thanks for testing with me!

This issue now can be closed!

![image](https://github.com/Automattic/wordpress-activitypub/assets/2017105/a1e2be58-4c17-454f-a9df-a1189c7e3afd)

PR #645, opened 2024-01-10
improve federated comments

pfefferle:
Fixes and optimisations around threaded, federated comments.

* Better handling/polifilling of #642 
* Better separation of local and federated comments #652


pfefferle:
@edent @janboddez I am now more restrictive with setting a `in_reply_to`, but I need a way to prevent federation. Any ideas?

I thought about returning a `WP_Error` on the `to_object` and `to_activity` functions instead of the object or activity. This way we could use the transformer classes to build up a basic verification system!?!

janboddez:
I think it would be very helpful if posts/comments could be prevented from being federated. I'd be super interested in a (filter?) hook to, to "force stop" "processing" of objects that are, e.g., in a certain category or something.

I would also argue that replies to comments without a `source_url` should simply not be federated. (Maybe let users override this option if they really want to?)

Ideally, you'd also bypass the JSON templates for GET requests (i.e., disable content negotation) for these objects. I think this _could_ be done by hooking into the `template_include` hook with a low enough prio and removing the plugin's callback if the post meets our criteria. But a custom hook (`activitypub_should_federate` or something) that's called in both places (the template callback _and_ the transformer class[es]) would be easier to use.

The plugin itself could then also use this hook. Or not.

Returning an error would work. A simple boolean might work too, but a `WP_Error` instance may be more future-proof. As long as the calling function knows how to handle it, i.e., skip federation (and/or return the default [HTML] template).

(Not entirely on-topic) I was also experimenting with a way to prevent "superfluous updates" (i.e., when nothing has actually been updated but the `transition_post_status` hook is run anyway) from being sent out. I currently compute and store a hash of the object just before activities are sent out, which I then compare with the next time around. Only "problem" is, I'm calculating this hash once for each endpoint, which doesn't make sense. If there was a hook in the transformer class, I'd be able to use that instead, and do this only once. So that would be another benefit.

pfefferle:
@janboddez what do you think about https://github.com/Automattic/wordpress-activitypub/pull/645/commits/94b2c6f9f405acabe68272750a67f9c7ea93cc9e ???

janboddez:
That looks like it could work! I'm not sure what would be more efficient, doing it there or in the "transformer" or elsewhere (like, into the "inbox loop") ...?

I think for _my purposes_, this would work. (There's still the question of conneg, of course, but like I said, that _could_ be dealt with elsewhere.)

janboddez:
Can we add an `activitypub_the_content` filter at the very end? :-)

PR #650, opened 2024-01-11
Update class-activitypub.php

janboddez:
Adds an `exit` after the redirect, as per https://developer.wordpress.org/reference/functions/wp_safe_redirect/.

> Note: [wp_safe_redirect()](https://developer.wordpress.org/reference/functions/wp_safe_redirect/) does not exit automatically, and should almost always be followed by a call to `exit;`:

## Proposed changes:

* See description. :-)

### Other information:

## Testing instructions:

N/A


pfefferle (approved):
LGTM

PR #661, opened 2024-01-15
A better default content handling based on the Object Type

pfefferle:
Mastodon improved it's default handling of non Note (or unknown) Activities, so we should prepare to remove the shortcodes. 

## Proposed changes:

* Better media handling based on the Object Type
* Set title and summary based on the Object Type

/cc @Menrath 

### Other information:

## Testing instructions:

* Go to '..'
*


pfefferle:
@janboddez I would love to get your feedback on this. I plan to remove the shortcodes in the near future and have a better default setup.

janboddez:
Ha, this is interesting, I'll try to have a look later.

---

Somewhat off-topic (because it relates to what _I_ am doing):

I personally still use post types. However, I might end up mapping these to formats so I can support both Note and Article. 🤔 Currently everything "is" a Note and I filter the content (title or no title) based off the post type.

Not sure if a filter to override/bypass the type would be handy? Guessing you can also end up with "too many" filters this way. And ... only devs use them, they're mostly useless to everyone else.

Maybe having (only) the entire object filterable would be a nice alternative? I currently use `activitypub_activity_object_array` for that (I sometimes override the `inReplyTo` or `to` and `cc` properties) but it is a bit messy. Then again, it's not really your or the plugin's problem.

---

Back on-topic:

I would love it if Notes (_if_ they correspond with status or aside post formats) did not have a title, but Articles did.

Not sure how this could work though, given that nearly everybody is probably converting Articles to Notes, _and_ that their template is set up for posts (i.e., Articles).

Eventually getting rid of the template altogether and just having a solid default based on post formats sounds really nice, though. Someone with PHP knowledge could always use the content filter to fine-tune more.

DeadSuperHero:
Following up from #678:

It might be interesting to offer a variety of different types, that are compatible with existing Fediverse platforms:
* `Note`: Mastodon
* `Article`: Plume
* `Video`: PeerTube
* `Audio`: Funkwhale

It also might be useful to offer the ability to customize the output for each type, similar to what we already have. For a Podcast post type, for example, maybe it makes the most sense to offer a `Note` activity with an excerpt, link, and media attachment for the audio.

pfefferle:
> It might be interesting to offer a variety of different types, that are compatible with existing Fediverse platforms

Does that mean, that you propose to send different versions of the same post depending on the receiver platform?

DeadSuperHero:
> Does that mean, that you propose to send different versions of the same post depending on the receiver platform?

That's an interesting idea, a concept I have thought about before, but no. Each WordPress post would federate out to all followers as the designated type specified in WordPress-ActivityPub settings. 

If the schema maps up to something a follower can read, it will be up to the follower's platform to attempt to render it. It's the type is unsupported and the platform doesn't fall back to simply rendering a note with a link, the platform instead will not render anything.

In other words, if I post an `Event` activity type to Mastodon, the default Mastodon behavior would display it as a `Note`. If I post the same `Event` activity and someone on Friendica sees it, they will be able to interact with it natively. If a PeerTube follower receives that same `Event`, the user will see nothing.

pfefferle:
Ok, that's what the post is about, but I am not sure about Audio/Video/Image... initially I thought about mapping these types to the corresponding ActivityPub type... But an Audio-Post is not really an Audio Activity, it is more a Note with an Audio Attachment...

DeadSuperHero:
That's okay, as long as we can choose a `Note` and specify a media attachment, I'm totally fine with that.

It was more of a suggestion that we should try to support more activity types than `Note` and `Article` in the future!

pfefferle:
@Menrath is also already working on Events btw.: https://github.com/Automattic/wordpress-activitypub/blob/master/includes/activity/extended-object/class-event.php

pfefferle:
Maybe we could map attachments to their corresponding Activity Type?!? https://github.com/Automattic/wordpress-activitypub/blob/master/includes/transformer/class-attachment.php

DeadSuperHero:
Yeah, that looks good to me!

PR #674, opened 2024-01-24
Enable Mastodon Apps: Implement the account endpoint for an external account

akirk:
This moves functionality from the Enable Mastodon Apps into ActivityPub. It allows retrieving the account of someone via ActivityPub with an API endpoint like `/api/v1/accounts/alex@kirk.at/`

The unit test currently needs network access, we should add the ability to mock the response.

### DIFF
--- a/integration/class-enable-mastodon-apps.php
+++ b/integration/class-enable-mastodon-apps.php
@@ -96,15 +96,15 @@ class Enable_Mastodon_Apps {
 		$account->display_name   = $data['name'];
 		$account->url            = $uri;
 		$account->note           = $data['summary'];
-		if ( isset( $data['icon']['type'] ) &&  isset( $data['icon']['url'] ) && 'Image' === $data['icon']['type']) {
-		$account->avatar         = $data['icon']['url'];
-		$account->avatar_static  = $data['icon']['url'];
+		if ( isset( $data['icon']['type'] ) && isset( $data['icon']['url'] ) && 'Image' === $data['icon']['type'] ) {
+			$account->avatar         = $data['icon']['url'];
+			$account->avatar_static  = $data['icon']['url'];
 		}
 		if ( isset( $data['image'] ) ) {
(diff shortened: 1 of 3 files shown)


integration/class-enable-mastodon-apps.php
```diff
@@ -71,4 +83,54 @@ function ( $item ) {
...
+		}
+
+		$acct = Webfinger_Util::uri_to_acct( $uri );
+		$data = get_remote_metadata_by_actor( $uri );
+
+		if ( ! $data ) {
+			return $user_data;
+		}
+
+		$account = new Account();
```
pfefferle, line 113: To not always replace the complete we should re-use `$user_date` if from type Account. What do you think?

```suggestion
        if ( $user_data instanceOf 'Account' ) {
         		$account = $user_data;
        } else {
        		$account = new Account();
        }
```

pfefferle (dismissed):
thanks @akirk

integration/class-enable-mastodon-apps.php
```diff
@@ -71,4 +83,58 @@ function ( $item ) {
...
+
+	/**
+	 * Resolve external accounts for Mastodon API
+	 *
+	 * @param Enable_Mastodon_Apps\Entity\Account $user_data The user data
+	 * @param string                              $user_id   The user id
+	 *
+	 * @return Enable_Mastodon_Apps\Entity\Account The filtered Account
+	 */
+	public static function api_account_external( $user_data, $user_id ) {
```
pfefferle, line 95: what about renaming it to `api_account` and to extend also the internal accounts with `acct` and follower counts? Or should we do that in a different hook?

pfefferle (approved):
👍

integration/class-enable-mastodon-apps.php
akirk: Added that second hook in https://github.com/Automattic/wordpress-activitypub/pull/676

PR #677, opened 2024-01-25
Fix #621

### LINKED ISSUE #621: Follow timestamp gets overwritten with last update timestamp
### Quick summary

As the title says, the timestamp for "followed" is the same as the timestamp for "last updated" (or wrongly displayed?).  See screenshot.  I (the entry at the bottom) followed right after installing about a week before, but now it states that I follow since yesterday only. This "reset" happened for all entries, so it's impossible now to check out when people started following.

### Steps to reproduce

- install plugin
- wait some days for some followers
- review followers tab

followed and last updated shouldn't be the same.

### What actually happened

followed and last updated are the same

### Impact

All

### Available workarounds?

### Logs or notes

Screenshot 2023-12-24 130503

Wordpress version: 6.4.2
Plugin version: 1.3.0

janboddez:
Fixes #621.

## Proposed changes:

* By adding a `post_date` argument to `wp_insert_post()`, for existing posts only (!), the `post_date` (shown in the Followers table as "Followed") and `post_date_gmt` columns no longer get set to the current datetime.

### Other information:

## Testing instructions:

* Apply the patch.
* (Optionally) run something like `UPDATE wp_posts SET post_date = '2024-01-01 00:00:00', post_date_gmt = '2023-12-31 23:00:00' WHERE post_type = 'ap_follower';`. (This'll obviously be wrong, but the current datetimes are too. Also, you may want to apply the right GMT offset for your site!)
* Somehow trigger the `activitypub_update_followers` cron a couple times (and, if possible, ensure the followers actually get updated by, e.g., temporarily having `Followers::get_outdated_followers()` return all of them).
* Go to `/wp-admin/users.php?page=activitypub-followers-list`.
* Notice the "followed" dates are no longer being overwritten.

If you don't want to "force" an update: it is possible to wait long enough, too.

Either way, it seems important to somehow pass the follower's existing `post_date` when updating them!

Notice that this `post_date` ≠ `$this->get_published()`. Because at least in the context of the `Follower` class, `get_published()` seems to return the date the follower account was created (on their remote instance)!

Did not update any tests: I'm assuming this isn't currently covered, as the bug would've probably been noticed, but also because I'm not too familiar them!

### DIFF
--- a/includes/model/class-follower.php
+++ b/includes/model/class-follower.php
@@ -162,19 +162,27 @@ class Follower extends Actor {
 			}
 		}
 
+		$post_id = $this->get__id();
+
 		$args = array(
-			'ID'            => $this->get__id(),
-			'guid'          => esc_url_raw( $this->get_id() ),
-			'post_title'    => wp_strip_all_tags( sanitize_text_field( $this->get_name() ) ),
-			'post_author'   => 0,
-			'post_type'     => Followers::POST_TYPE,
-			'post_name'     => esc_url_raw( $this->get_id() ),
-			'post_excerpt'  => sanitize_text_field( wp_kses( $this->get_summary(), 'user_description' ) ),
-			'post_status'   => 'publish',
-			'meta_input'    => $this->get_post_meta_input(),
-			'post_date'     => get_date_from_gmt( $this->get_published() ), // Or something.
-			'post_date_gmt' => $this->get_published(),
+			'ID'           => $post_id,
+			'guid'         => esc_url_raw( $this->get_id() ),
+			'post_title'   => wp_strip_all_tags( sanitize_text_field( $this->get_name() ) ),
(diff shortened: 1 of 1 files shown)


janboddez:
Note that I _again_ set `$post_id`, to ensure it is set to the latest and greatest value.

I wasn't too sure what `$this->get__id()` returns for new posts (could be `0`, could be `false` ... I couldn't really tell). so I decided to do what core does to decide whether we're dealing with and update or a mere insert, by checking for `empty( $post_id )`.

And then I get the "raw" `post_date`, which to me seemed safer than, you know, `get_the_date( $this->get__id() )`.

And so on. But feel free to make any edits; just wanted to "err on the side of caution."

pfefferle (approved):

pfefferle:
Thanks @janboddez

Issue #673, opened 2024-01-23
All images uploaded to blog are immediately published to Mastodon without being added to a post


suramyatomar:
### Quick summary

Whenever I upload an image to the Blog using the Upload files functionality in Wordpress Add Media page, the system automatically publishes it to Mastodon. It is happening to all images being added to the site even before they are inserted into a post.

Example: https://mastodon.social/@blog@www.suramya.com/111804854104552673

### Steps to reproduce

1. Start a new Blog post using the Classic Editor
2. Click on 'Add Media' button
3. Click on 'Upload Files' tab
4. Upload an Image

### What you expected to happen

Nothing should happen, because the image is not added to any post yet

### What actually happened

The image is immediately published on Mastodon: 
![Screenshot_20240123_161858](https://github.com/Automattic/wordpress-activitypub/assets/35488010/8b1d7ac9-8839-4c31-9ff4-1736251bdb89)

### Impact

All

### Available workarounds?

No but the platform is still usable

### Logs or notes

Self Hosted Wordpress version: 6.4.2
ActivityPub Version: 2.0.1

janboddez:
Do you perhaps have "media" enabled? Under wp-admin/options-general.php?page=activitypub&tab=settings

suramyatomar:
Yes, I did. Unchecking it seems to have resolved the issue. However, I have not changed the settings in a few months but didn't hit this issue till I upgraded to ver 2.x

pfefferle:
@suramyatomar yes, the earlier versions had a bug in the media federation and that was fixed in 2.0.0 :)

suramyatomar:
Ah ok.
I will recommend you add a small note to the settings page stating that enabling this value (media ) will post all images you upload to the site immediately to Mastodon.

pfefferle:
The thing is, that this list will be auto generated from the list of post types. I will see if there is a description I could use.

docpop:
I had the same issue as Suramya and was surprised to see my images being published to my Mastodon feed. I have turned off "media" and "pages" from the "Supported post types", but I wanted to leave a suggestion. When I enabled those options, a year ago, I thought they enabled media to show up when I publish a post. I didn't expect them to auto-publish anytime I uploaded an image or new page. So I suggest changing the text from 
**Enable ActivityPub support for the following post types:**
to
**Automatically publish to the selected options to Mastodon**
[...]

PR #681, opened 2024-01-30
improve the settings screen

pfefferle:
* change description
* add default description for Attachments

see #673

Screenshot 2024-01-30 at 12 54 40

What do you think @suramyatomar and @docpop ???


docpop:
This seems good to me. I like the changes.

pfefferle:
Thanks @docpop

PR #682, opened 2024-01-31
fix ActivityPub `@context`

pfefferle:
The `call_user_func` version seem to not work properly (returns `null` in this case https://github.com/Automattic/wordpress-activitypub/blob/master/includes/handler/class-follow.php#L98) in this case and this seems to break the follow flow.

/cc @Menrath

### DIFF
--- a/includes/activity/class-activity.php
+++ b/includes/activity/class-activity.php
@@ -173,10 +173,13 @@ class Activity extends Base_Object {
 	 */
 	public function get_json_ld_context() {
 		if ( $this->object instanceof Base_Object ) {
-			// Without php 5.6 support this could be just: 'return  $this->object::JSON_LD_CONTEXT;'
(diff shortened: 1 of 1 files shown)


includes/activity/class-activity.php
```diff
@@ -173,10 +173,13 @@ public function set_object( $object ) {
 	 */
 	public function get_json_ld_context() {
 		if ( $this->object instanceof Base_Object ) {
-			// Without php 5.6 support this could be just: 'return  $this->object::JSON_LD_CONTEXT;'
-			return call_user_func( array( get_class( $this->object ), 'CONTEXT' ) );
```
Menrath, line 177: Oh! Shouldn't it have been 'JSON_LD_CONTEXT' anyway? :/

pfefferle: yes, but this has not solved it.

PR #683, opened 2024-02-01
re-add vary header

### LINKED ISSUE #580: WordPress sometimes returns JSON to a Webbrowser instead of the HTML page.
### Quick summary

Sometimes and for some users WordPress returns a JSON object to a regular browser request interested of the normal website.

This (sometimes) persists on reload
signal-2023-11-27-21-53-44-776

### Steps to reproduce

No idea 🤷

The normal website.

### What actually happened

I see JSON in the browser

### Impact

Some (< 50%)

### Available workarounds?

### Logs or notes

_No response_

pfefferle:
helps with #580

@timnolte & @snarfed what would you prefer? opt-in or opt-out?

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ The plugin uses PHP Constants to enable, disable or change its default behaviour
 * `ACTIVITYPUB_DISABLE_INCOMING_INTERACTIONS` - Block incoming replies/comments/likes. Default: `false`.
 * `ACTIVITYPUB_DISABLE_OUTGOING_INTERACTIONS` - Disable outgoing replies/comments/likes. Default: `false`.
 * `ACTIVITYPUB_SHARED_INBOX_FEATURE` - Enable the shared inbox. Default: `false`.
+* `ACTIVITYPUB_SEND_VARY_HEADER` - Disable to not send the `Vary: Accept` header. Default: `true`.
 
 ### Where can you manage your followers? ###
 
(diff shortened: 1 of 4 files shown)


snarfed:
Woo, thank you @pfefferle!

Personally, I prefer defaulting to correctness over efficiency - "make it work, then make it right, then make it fast" - so I like this as is. You all know the WordPress ecosystem much better than me, so I defer to you!

akirk (approved):
I was "pro" this all along :) #305

timnolte:
Has this been tested with Nginx caching and an already configured `Vary` header?

My concern is this all of a sudden causing caching data storage or breaking caching altogether on a site, which can cost users money or downtime. I'd personally choose opt-in until we have enough testing to confirm that opt-out is OK.

pfefferle:
Thanks for sharing your experience @lencioni !

pfefferle:
@snarfed we aleady had a talk with cloudflare because of that issue... Sadly the `Vary` header is not the solution to all conneg problems...

timnolte:
@snarfed are we certain that Mastodon servers use the exact same endpoint for both JSON and other/HTML output? I'm inclined to think that it is actually the Mastodon client being used that is rendering the JSON response and it's not just presenting for example a raw HTML version of the content. I think this is where WordPress is different than a Mastodon server where there is a separate backend server and frontend Mastodon client. If the WordPress sites was, for example headless, where an HTML client was receiving the JSON always and rendering it as HTML for visitors then this might more closely resemble Mastodon.

snarfed:
Good question! Mastodon and at least some fediverse servers do use different canonical URLs for AP (JSON) objects vs user-facing HTML. However, they still also do full conneg on the HTML URLs, and redirect AP requests to the AP URLs, since copy/pasting a user-facing URL into your instance's search is a core UX workflow in the fediverse.

It's interesting to consider a fediverse webapp client using AP as its API to its own backend. AP S2S (server to server) isn't suited to that, C2S (client/server) definitely is, but barely any fediverse servers implement or use C2S. 🤷

PR #685, opened 2024-02-05
Add "Move Account" functionality

### LINKED ISSUE #654: "Move account" when deactivating per user accounts
### What

I deactivated the user specific ap-accounts and wen with one account for the whole blog only (it is a single user blog).
I would love to send the followers of tomate@jascha.wtf the message, that the account moved that the followers change automatically to the new account (like when you moving account in mastodon or other ap-platforms). 

### Why

When I started with the plugin there was no user for the whole blog and now there is.

### How

Like moving account at mastodon.

pfefferle:
See:

* https://docs.joinmastodon.org/user/moving/
* https://blog.joinmastodon.org/2019/06/how-to-migrate-from-one-server-to-another/

Fix #654

## Usecases

* Move from Mastodon to WordPress
* Move from WordPress to Mastodon
* Change Blog-Domain
* Rename Actor-URL or Change [Actor-ID (URI)](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-id)
* ...

/cc @jessLundie


includes/class-move.php
```diff
@@ -0,0 +1,98 @@
...
+
+	/**
+	 * Move an ActivityPub account from one location to another.
+	 *
+	 * @param string $from The current account URL.
+	 * @param string $to   The new account URL.
+	 *
+	 * @return int|bool|\WP_Error The ID of the outbox item or false or WP_Error on failure.
+	 */
+	public static function account( $from, $to ) {
```
pfefferle, line 27: verify that `$from` and `$to` are URIs.

includes/class-move.php
```diff
@@ -0,0 +1,98 @@
...
+	 *
+	 * @return array the extended actor profile
+	 */
+	public static function extend_actor_profiles( $actor, $id, $user ) {
+		// Check if the user is a valid user object.
+		if ( ! $user instanceof \Activitypub\Model\User ) {
+			return $actor;
+		}
+
+		$move_to = \get_user_option( 'activitypub_move_to', $user->get__id() );
```
pfefferle: Is it possible to use a `user_option` for ID `0`?

pfefferle: See: https://github.com/Automattic/wordpress-activitypub/pull/1389

obenland (approved):
Let's merge and iterate.

PR #688, opened 2024-02-08
Fix "Undefined variable $type in includes/class-scheduler.php on line 191"

janboddez:
Fixes an "undefined variable `$type`" warning.

## Proposed changes:

* Replace `if ( ! $type )` (which throws a warning since I think PHP 8) with `if ( empty( $type ) )`

### Other information:

## Testing instructions:

* I got this when deleting and restoring comments
* Maybe when the old status was `trash` and the new one `pending` or whatever it's called
* Either way, there are scenarios where `$type` does not get initialized, ever

### DIFF
--- a/includes/class-scheduler.php
+++ b/includes/class-scheduler.php
@@ -193,7 +193,7 @@ class Scheduler {
 			$type = 'Delete';
 		}
 
-		if ( ! $type ) {
+		if ( empty( $type ) ) {
 			return;
 		}
 


janboddez:
Could've also used `if ( ! isset( $type ) )`. But in _this specific case_ there's no difference, because `$type` is _either_ undefined _or_ a string.

Or you could set a default value (`$type = false;` or something) _before_ the if/else statements. Also okay.

pfefferle (approved):

PR #689, opened 2024-02-09
Add/remote reply

pfefferle:
To let users reply to federated comments, we have to direct them to their Mastodon (or any other fediverse) instance. Therefore we need a lightbox like we have for "following":

Screenshot 2024-02-09 at 16 39 31

That open if a user clicks the "reply-to" link:

Screenshot 2024-02-09 at 16 42 13

@mattwiebe I have already built the API endpoint and a dummy "Reply on the Fediverse" text (link is missing). Can you support with the lightbox logic?


mattwiebe:
Apologies for not getting to this sooner.

What do I need to do to get the comment to show the "Reply on the fediverse" link? This was a comment that I made on a local test site. I guess it needs to be a comment that came in from the fediverse?

Screenshot 2024-02-16 at 13 51 36

pfefferle:
> Apologies for not getting to this sooner.

No problem :)

> What do I need to do to get the comment to show the "Reply on the fediverse" link? This was a comment that I made on a local test site. I guess it needs to be a comment that came in from the fediverse?

Yes, it must be a comment from the fediverse and you have to be logged out.

To simulate that, you can simply add a comment-meta with the key: "protocol" and the value: "activitypub".

pfefferle:
@mattwiebe I gave it a try and implemented a working UI based on your follow me widget.

/cc and thanks for your help @jeherve and @akirk

includes/class-comment.php
```diff
@@ -335,4 +346,29 @@ public static function generate_id( $comment ) {
 			\trailingslashit( site_url() )
 		);
 	}
+
+	/**
+	 * Enqueue scripts for remote comments
+	 */
+	public static function enqueue_scripts() {
```
mattwiebe, line 353: Let's be more selective here: only print these scripts if we're on a single page with comments open.

includes/class-comment.php
```diff
@@ -335,4 +346,29 @@ public static function generate_id( $comment ) {
...
+			$assets = require_once $asset_file;
+
+			\wp_enqueue_script(
+				$handle,
+				\plugins_url( 'build/remote-reply/index.js', __DIR__ ),
+				$assets['dependencies'],
+				'1.0',
+				true
+			);
+			\wp_add_inline_script( $handle, $js, 'before' );
```
mattwiebe, line 371: Need to add
```php
wp_enqueue_style( 'wp-components' );
```

For the Dialog to appear properly.

src/remote-reply/remote-reply.js
```diff
@@ -0,0 +1,117 @@
...
+	);
+}
+
+export default function RemoteReply( { selectedComment, commentId } ) {
+	const [ isOpen, setIsOpen ] = useState( false );
+	const title = __( 'Remote Reply', 'activitypub' );
+
+	return(
+		<>
+			<a href="javascript:;" className="comment-reply-link activitypub-remote-reply__button" onClick={ () => setIsOpen( true ) } >
```
mattwiebe, line 103: My devtools were complaining about this URL. But you still want your mouse cursor to behave as expected for a link, despite it not taking you anywhere. A few thoughts:

1. Firstly, we could just use `selectedComment` as the href? Graceful degradation?
2. If you don't want a link at all, let's try with the `Button` component:

```js
 // You're already importing `Button` from `@wordpress/components` above so no need to add it
<>
			<Button isLink className="comment-reply-link activitypub-remote-reply__button" onClick={ () => setIsOpen( true ) } >
			{ is Open && ( // ...
```

includes/class-comment.php
```diff
@@ -335,4 +346,29 @@ public static function generate_id( $comment ) {
...
+		$asset_file = ACTIVITYPUB_PLUGIN_DIR . 'build/remote-reply/index.asset.php';
+
+		if ( \file_exists( $asset_file ) ) {
+			$assets = require_once $asset_file;
+
+			\wp_enqueue_script(
+				$handle,
+				\plugins_url( 'build/remote-reply/index.js', __DIR__ ),
+				$assets['dependencies'],
+				'1.0',
```
mattwiebe, line 368: Try using `$assets['version]` here as well

src/remote-reply/index.js
```diff
@@ -0,0 +1,16 @@
...
+let id = 1;
+function getUniqueId() {
+	return `activitypub-remote-reply-link-${ id++ }`;
+}
+
+domReady( () => {
+	// iterate over a nodelist
+	[].forEach.call( document.querySelectorAll( '.activitypub-remote-reply' ), ( element ) => {
+		const attrs = JSON.parse( element.dataset.attrs );
+		render( <RemoteReply { ...attrs } id={ getUniqueId() } useId={ true } />, element );
```
mattwiebe, line 14: React has recently deprecated the `React.render` method in favour of `createRoot`. It would now be something like 

```js
import { createRoot } from '@wordpress/element`;

// line instead of `render( <RemoteReply //etc`
createRoot( element).render( <RemoteReply { ...attrs } id={ getUniqueId() } useId={ true } )
```

(I need to go update other uses too)

mattwiebe (changes requested):
This is an awesome start and not far away at all!

Left a few comments that should be addressed but we should be able to ship this soon.

includes/class-comment.php
```diff
@@ -335,4 +346,33 @@ public static function generate_id( $comment ) {
...
+			$assets = require_once $asset_file;
+
+			\wp_enqueue_script(
+				$handle,
+				\plugins_url( 'build/remote-reply/index.js', __DIR__ ),
+				$assets['dependencies'],
+				$assets['version'],
+				true
+			);
+			\wp_enqueue_style( 'wp-components' );
```
mattwiebe: I missed that we had some custom styles for this, instead of just adding `wp-components`.

So something like

```php
\wp_enqueue_style( 
  'activitypub-remote-reply',
  \plugins_url( 'build/remote-reply/style-index.css', __DIR__ ),
  [ 'wp-components' ],
  $assets['version']
);

includes/class-comment.php
pfefferle: Thanks 🫣

mattwiebe (approved):
I left one change but I just went and added it myself. I do think that we could be DRY-er between the Follow vs Reply popups but I am a bigger fan of "it works" than "it's DRY."

We can fix any problems in follow-up PRs, well done on your first React component! 🌟

PR #698, opened 2024-02-16
Prevent multiple scheduled jobs for a post with locking

mattwiebe:
On dotcom we've seen one site with a very large `cron` option and this is one possible path to reducing that. This will use postmeta to lock a post, the lock being cleared inside the cron job itself.

We could run into an issue here due to passing the post object into the cron job's data rather than fetching it at runtime. As in, update 1 (`U1`) is performed, adding a cron job and postmeta lock. Update 2 (`U2`) is performed before the cron job fires, detects the postmeta lock, and does not add a redundant cron job. After this, the cron job fires and removes the postmeta lock, but it is relying on the post data from `U1`, not the most recent `U2`. 

(I did attempt passing a `post_id` rather than a full post before realizing that I'd need to refactor more extensively due to `WP_Comment`s also being passed to `activitypub_send_activity` and then I decided on this MVP instead.)

Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


pfefferle:
Closed in favour of #700

thanks a lot for the fast proposal @mattwiebe

PR #706, opened 2024-03-04
Enqueue reply assets only if there is anything to reply to on the page

### LINKED ISSUE #705: `activitypub-remote-reply` script enqueues 29 dependencies on all `is_singular()` requests
### Quick summary

Introduced in #689 the `activitypub-remote-reply` script handle now enqueues itself with 29 dependencies:

> wp-polyfill-inert, regenerator-runtime, wp-polyfill, wp-hooks, wp-i18n, wp-url, wp-api-fetch, react, react-dom, wp-dom-ready, wp-a11y, wp-deprecated, wp-dom, wp-escape-html, wp-element, wp-is-shallow-equal, wp-keycodes, wp-priority-queue, wp-compose, moment, wp-date, wp-html-entities, wp-primitives, wp-private-apis, wp-redux-routine, wp-data, wp-rich-text, wp-warning, wp-components

This happens on posts and pages that (1) don't support comments, (2) don't have comments enabled.

This also adds a combined total of almost 2MB of potentially unused JS to all of those requests. 

### Steps to reproduce

1. Enable the plugin.
2. Visit any single post or page and confirm that all of the mentioned dependencies are enqueued in the HTML source code.

1. The reply script is enqueued only on posts and pages that support comments and have comments enabled.
2. The script is lazy-loaded only when the user clicks on the call-to-action link/button.

### What actually happened

The mentioned script and style gets enqueued on all post and page views. The enqueued dependency bundle size is incredibly large.

### Impact

Most (> 50%)

### Available workarounds?

Yes, difficult to implement

### Logs or notes

_No response_

kasparsd:
Fixes #705 

## Proposed changes:

* Enqueue the reply assets only if the DOM element `.activitypub-remote-reply` was ever needed for the current request.

### Other information:

## Testing instructions:

* Create a new post with comments enabled.
* View the single post template and confirm that the `activitypub-remote-reply` script and styles are not enqueued in the HTML.
* Leave a comment.
* Confirm that now `activitypub-remote-reply` script and styles are enqueued because the `<div class="activitypub-remote-reply"></div>` is now rendered on the page.

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -14,6 +14,14 @@ use function Activitypub\is_user_disabled;
  */
 class Comment {
 	/**
+	 * Enqueue reply assets only if the reply feature is needed
+	 * for the current request.
+	 *
+	 * @var bool
+	 */
+	private static $remote_reply_assets_needed = false;
+
+	/**
 	 * Initialize the class, registering WordPress hooks
 	 */
 	public static function init() {
@@ -39,6 +47,8 @@ class Comment {
 		if ( self::are_comments_allowed( $comment ) ) {
 			return $link;
 		}
+
+		self::$remote_reply_assets_needed = true; // Request script and styles for remote reply feature.
 
 		$attrs = array(
 			'selectedComment' => self::generate_id( $comment ),
@@ -351,7 +361,11 @@ class Comment {
 	 * Enqueue scripts for remote comments
 	 */
 	public static function enqueue_scripts() {
-		if ( ! is_singular() ) {
+		$assets_needed = \apply_filters(
+			'activitypub_remote_reply_assets_needed',
+			self::$remote_reply_assets_needed
+		);
+		if ( ! $assets_needed ) {
 			return;
 		}
 		$handle     = 'activitypub-remote-reply';


includes/class-comment.php
```diff
@@ -40,6 +48,8 @@ public static function comment_reply_link( $link, $args, $comment ) {
 			return $link;
 		}
 
+		self::$remote_reply_assets_needed = true; // Request script and styles for remote reply feature.
```
kasparsd, line 51: Since `build/remote-reply/index.js` depends on `.activitypub-remote-reply` being present in the DOM, it is safe to assume that the script is only needed if this filter has been triggered.

mattwiebe: I was initially dubious that `comment_reply_link` would run earlier than `enqueue_scripts`, but it appears to be the case!

mattwiebe (approved):
Great contribution! Definitely don't want load all those assets if they're not needed. 

I'd really like to migrate these dialogs to something in the [new Interactivity API](https://make.wordpress.org/core/2024/03/04/interactivity-api-dev-note/) to hopefully be a bit more lightweight.

PR #709, opened 2024-03-12
Generate a WP_User for the Blog and Application User

pfefferle:
WIP!

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


Menrath:
I can imagine the improvements of this, as the current implementation took a lot of time to properly understand it. Thanks for working on this. I would ask for one little favour though: if you could find the time to communicate the motivation behind that PR (again in written words) that would be really awesome! :)

pfefferle:
@Menrath I had/have to rethink this idea, because it won't work on WordPress multi site, because users are global there.

PR #712, opened 2024-03-21
Comment reply federation: support `is_single_user` sites

### LINKED ISSUE #711: Federate Replies with User mode disabled
### Quick summary

The [`are_comments_allowed()` check](https://github.com/Automattic/wordpress-activitypub/blob/15ae5f68ed775ccd5475f53a1b5436a8889e39c0/includes/class-comment.php#L88) will prevent a logged-in user from replying to comments on a site with only the Blog User enabled. 

### Steps to reproduce

1. Add `define( 'ACTIVITYPUB_SINGLE_USER_MODE', true );` to your `wp-config.php`
2. Reply to a federated post from a Mastodon (or similar) service.
3. Only see the "Reply on the Fediverse" link for the comment, not allowing me to make a federated reply from WP

As a logged-in user with sufficient rights, I should be able to reply to the comment and have it federated as the Blog User.

### What actually happened

I only see the "Reply on the Fediverse" option.

### Impact

Some (< 50%)

### Available workarounds?

### Logs or notes

_No response_

mattwiebe:
Fixes #711 

## Proposed changes:
Federate as the Blog User on `is_single_user` sites.

### Other information:

## Testing instructions:
1. Add `define( 'ACTIVITYPUB_SINGLE_USER_MODE', true );` to your `wp-config.php`
2. Reply to a federated post from a Mastodon (or similar) service.
3. When logged in with a user with the `activitypub` capability, you should be able to reply to the post and have it federate - you should receive a notification on the service you used to make the comment in step 2.

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -2,9 +2,11 @@
 
 namespace Activitypub;
 
+use Activitypub\Collection\Users;
 use WP_Comment_Query;
 
 use function Activitypub\is_user_disabled;
+use function Activitypub\is_single_user;
 
 /**
  * ActivityPub Comment Class
@@ -83,6 +85,11 @@ class Comment {
 
 		if ( ! $current_user ) {
 			return false;
+		}
+
+		if ( is_single_user() && \user_can( $current_user, 'publish_posts' ) ) {
+			// On a single user site, comments by users with `publish_posts` will be federated as the blog user
+			$current_user = Users::BLOG_USER_ID;
 		}
 
 		$is_user_disabled = is_user_disabled( $current_user );
@@ -188,6 +195,11 @@ class Comment {
 		// comments without user can't be federated
 		if ( ! $user_id ) {
 			return false;
+		}
+
+		if ( is_single_user() && \user_can( $user_id, 'publish_posts' ) ) {
+			// On a single user site, comments by users with `publish_posts` will be federated as the blog user
+			$user_id = Users::BLOG_USER_ID;
 		}
 
 		$is_user_disabled = is_user_disabled( $user_id );
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -114,7 +114,7 @@ class Comment {
 		}
 
 		if ( is_single_user() && \user_can( $current_user, 'publish_posts' ) ) {
-			// On a single user site, comments by users with the `activitypub` capability will be federated as the blog user
+			// On a single user site, comments by users with the `publish_posts` capability will be federated as the blog user
 			$current_user = Users::BLOG_USER_ID;
 		}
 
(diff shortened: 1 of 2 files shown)


pfefferle:
Is this something we should "fix" in the plugin or only for .com? The first version of comment federation had this behaviour you re-implemented, but some people that where irritated by that.

(I try to find some of the feedback and comments)

mattwiebe:
> The first version of comment federation had this behaviour you re-implemented, but some people that where irritated by that.

In this case it only federates that way when the blog user is the only thing active, seems like that would be suitable but I'd be curious to see what the negative feedback was.

mattwiebe:
@pfefferle can we get this merged? I don't think we need a special mode for dotcom, since this will simply apply for everyone who has only the Blog User enabled. Otherwise, users of that mode will never be able to federate comment replies.

includes/transformer/class-comment.php
```diff
@@ -30,6 +30,10 @@ class Comment extends Base {
 	 * @return int The User-ID of the WordPress Comment
 	 */
 	public function get_wp_user_id() {
+		if ( is_single_user() && \user_can( $this->wp_object->user_id, 'activitypub' ) ) {
+			// Single user sites are federated with the Blog User.
+			return Users::BLOG_USER_ID;
+		}
```
pfefferle, line 36: This should not be needed, because it is handled by the Dispatcher, the same way it is handled for posts: https://github.com/Automattic/wordpress-activitypub/blob/master/includes/class-activity-dispatcher.php#L46

pfefferle (changes requested):

includes/class-comment.php
```diff
@@ -111,6 +113,11 @@ public static function are_comments_allowed( $comment ) {
 			return false;
 		}
 
+		if ( is_single_user() && \user_can( $current_user, 'activitypub' ) ) {
```
pfefferle, line 116:
```suggestion
		if ( is_single_user() && \user_can( $current_user, 'publish_posts' ) ) {
```

Maybe we should keep checking `publish_posts`? If someone uses the blog-only mode, he might not care about setting up the caps properly?!?

includes/class-comment.php
```diff
@@ -216,6 +223,11 @@ public static function should_be_federated( $comment ) {
 			return false;
 		}
 
+		if ( is_single_user() && \user_can( $user_id, 'activitypub' ) ) {
```
pfefferle, line 226:
```suggestion
		if ( is_single_user() && \user_can( $user_id, 'publish_posts' ) ) {
```

Maybe we should keep checking `publish_posts`? If someone uses the blog-only mode, he might not care about setting up the caps properly?!?

includes/transformer/class-comment.php
mattwiebe: Oh, love to remove code! 👍🏻

mattwiebe:
These latest changes confirmed working in prerelease on dotcom.

pfefferle (approved):

PR #715, opened 2024-03-25
Federated replies: improved UX for "your reply will federate"

mattwiebe:
## Proposed changes:

* Change comment reply link from "Reply" to "Reply with federation"
* Also adds a `title` attribute that currently says "is comment was received from the fediverse and your reply will be sent to the original author"

### Other information:
Screenshot 2024-03-25 at 16 26 54

## Testing instructions:
Try this on a post that has received a reply from the fediverse. The reply link, for a logged-in user who will be able to federate, will be changed like the screenshot above

* Go to '..'
*

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -75,7 +75,7 @@ class Comment {
 	 * to the original author.
 	 *
 	 * @param string $link The HTML markup for the comment reply link.
-	 * @param array  $args An array of arguments overriding the defaults.
+	 * @param array  $args The args provided by the `comment_reply_link` filter.
 	 *
 	 * @return string The modified HTML markup for the comment reply link.
 	 */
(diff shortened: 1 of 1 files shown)


includes/class-comment.php
```diff
@@ -45,6 +45,11 @@ public static function init() {
 	 */
 	public static function comment_reply_link( $link, $args, $comment ) {
 		if ( self::are_comments_allowed( $comment ) ) {
+			$user_id = get_current_user_id();
+			if ( $user_id && self::was_received( $comment ) && \user_can( $user_id, 'publish_posts' ) ) {
```
pfefferle, line 49: use `activitypub` cap instead of `publish_posts`

```suggestion
			if ( $user_id && self::was_received( $comment ) && \user_can( $user_id, 'activitypub' ) ) {
```

pfefferle (changes requested):

pfefferle:
nice!

PR #717, opened 2024-04-04
Enable Mastodon Apps: Add remote outbox fetching

akirk:
When in the Enable Mastodon Apps the statuses for a remote account are requested, this patch adds the ability to view their posts by fetching them via ActivityPub.

There are two filters needed:
1. The EMA plugin asks the `mastodon_api_get_posts_query_args` filter if it knows an `author` that is a string. If it can be fetched via Webfinger, it rewrites it to an `activitypub` arg,
2. Which is then picked up by the `api_statuses_external` function which then looks for the outbox and queries its first page, converting the ActivityPub activity into a `Enable_Mastodon_Apps\Entity\Status` along the way.

Right now this has support for attachments, we could also add support for mentions.


integration/class-enable-mastodon-apps.php
```diff
@@ -233,4 +248,96 @@ public static function api_search( $search_data, $request ) {
...
+			return $statuses;
+		}
+
+		$data = get_remote_metadata_by_actor( $args['activitypub'] );
+
+		if ( ! $data || is_wp_error( $data ) || ! isset( $data['outbox'] ) ) {
+			return $statuses;
+		}
+
+		$outbox = get_remote_metadata_by_actor( $data['outbox'] );
```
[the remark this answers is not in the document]
akirk: A cached HTTP request to the URL using `Accept: application/activity+json`. Works but I agree that it's misusing this.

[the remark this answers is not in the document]
pfefferle: you could use https://github.com/Automattic/wordpress-activitypub/blob/master/includes/class-http.php#L74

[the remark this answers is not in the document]
akirk: Ok, I introduced caching to `\Activitypub\HTTP::get()`

includes/class-http.php
```diff
@@ -108,6 +120,14 @@ public static function get( $url ) {
 
 		\do_action( 'activitypub_safe_remote_get_response', $response, $url );
 
+		if ( $cached ) {
+			$cache_duration = $cached;
+			if ( ! is_int( $cached ) ) {
```
pfefferle, line 125:
```suggestion
			if ( ! is_int( $cache_duration ) ) {
```

includes/class-http.php
```diff
@@ -108,6 +120,14 @@ public static function get( $url ) {
...
+				$cached = HOUR_IN_SECONDS;
```
pfefferle, line 126:
```suggestion
				$cache_duration = HOUR_IN_SECONDS;
```

integration/class-enable-mastodon-apps.php
```diff
@@ -233,4 +249,104 @@ public static function api_search( $search_data, $request ) {
...
+			return $statuses;
+		}
+
+		$data = get_remote_metadata_by_actor( $args['activitypub'] );
+
+		if ( ! $data || is_wp_error( $data ) || ! isset( $data['outbox'] ) ) {
+			return $statuses;
+		}
+
+		$response = \Activitypub\Http::get( $data['outbox'], true );
```
pfefferle, line 276: Can you add that as `use Activitypub\Http` to the head, to not have namespaces in the code?

pfefferle (commented):
Some small comments

integration/class-enable-mastodon-apps.php
```diff
@@ -233,4 +250,104 @@ public static function api_search( $search_data, $request ) {
...
+					$status->media_attachments = array_map(
+						function ( $attachment ) {
+							$media_attachment = new Media_Attachment();
+							$media_attachment->id = Mastodon_API::remap_url( $attachment['url'], $attachment );
+							$media_attachment->type = strtok( $attachment['mediaType'], '/' );
+
+							$media_attachment->url = $attachment['url'];
+							$media_attachment->preview_url = $attachment['url'];
+							$media_attachment->description = $attachment['name'];
+							$media_attachment->blurhash = $attachment['blurhash'];
```
pfefferle, line 331: should we check if the keys exist? WordPress for example does not provide blurhashes yet?!?

akirk: We could also use a default blurhash `LaS5;bKg%z,IofafjtkC.lw1RjNt` (see https://blurha.sh/):
Screenshot 2024-04-12 at 12 11 01

integration/class-enable-mastodon-apps.php
```diff
@@ -127,8 +127,9 @@ public static function api_account_add_followers( $account, $user_id ) {
 			}
 		}
 
-		$account->acct = $user->get_webfinger();
+		$account->acct = $user->get_preferred_username();
```
pfefferle, line 130: shouldn't this be a WebFinger acct?

akirk: No, because it's a local account, it should not have a server name, otherwise your identity will show as `@user@server@server`.

pfefferle: ah, I see!

pfefferle (approved):
LGTM!

pfefferle (approved):

PR #719, opened 2024-04-05
Mask version number

pfefferle:
Show only `<major>.<minor>` and hide `<patch>` and beta infos


mattwiebe (approved):
Works nicely in my testing, this is a good improvement

PR #724, opened 2024-04-09
fix outbox collection for blog and application user

Menrath:
## Little fix:
The outbox for the blog and the application user is currently empty, even if there are posts of a post type that have ActivityPub support enabled and the blog user is activated.


pfefferle (approved):
Thanks!

PR #727, opened 2024-04-11
Add possibility to disable scheduling for certain posts/comments

pfefferle:
/cc @janboddez @jordesign

### DIFF
--- a/includes/class-scheduler.php
+++ b/includes/class-scheduler.php
@@ -142,6 +142,11 @@ class Scheduler {
 			return;
 		}
 
+		// check if the Post should be scheduled for sending
+		if ( ! apply_filters( 'activitypub_schedule_send_post', true, $post, $type ) ) {
+			return;
+		}
+
 		$hook = 'activitypub_send_post';
 		$args = array( $post->ID, $type );
 
@@ -191,6 +196,11 @@ class Scheduler {
 
 		// check if comment should be federated or not
 		if ( ! should_comment_be_federated( $comment ) ) {
+			return;
+		}
+
+		// check if the Post should be scheduled for sending
+		if ( ! apply_filters( 'activitypub_schedule_send_comment', true, $comment, $type ) ) {
 			return;
 		}
 


pfefferle:
Oh, that's a valid point!

pfefferle:
Waiting for your PR @janboddez ;)

janboddez:
> Waiting for your PR @janboddez ;)

Haha! I was _so_ close but then I couldn't get the latest `master` running (admittedly, I didn't try very hard either) ...

Maybe this upcoming week, who knows!

pfefferle:
I changed the user handling! Maybe the migration does not work properly?!? It now adds a capability "activitypub", this way you can better decide which user you want to allow and which not!

janboddez:
Very first try: https://github.com/Automattic/wordpress-activitypub/compare/master...janboddez:wordpress-activitypub:improve-transformer

It does not yet work, or rather, it returns an empty screen for posts that don't "support" ActivityPub. Like so: https://jan.boddez.net/notes/f8dab23f83?activitypub

I prevent posts from being scheduled like so:
```
<?php

add_filter( 'activitypub_transformer', function ( $transformer, $obj ) {
  if ( $obj instanceof \WP_Post && in_category( 'disable-activitypub', $obj ) ) {
    return new \WP_Error( 'unsupported_object', __( 'This object does not support ActivityPub.', 'activitypub' ) );
  }

  return $transformer;
}, 10, 2 );
```

And, as mentioned, rather than _skip_ conneg, it just _breaks_ "ActivityPub requests" for these posts. Right now at least. But it's something. And they're very minor changes. In fact, I didn't add anything but "better" (well, _different_) error handling.

I'm also worried that a post might get scheduled before Gutenberg/the REST API has saved the, in this case, category. But that's another issue. Haven't really tested this, yet (I made sure to add the category, and save the post as draft, _before_ publishing.)

janboddez:
I get that this isn't as user-friendly as a per-post setting, but it's something. Site owners could use this to prevent certain categories, post formats, etc. from being federated.

Also noticed the `activitypub` custom field got set to, I think, `federated`. I may have to address that, too.

janboddez:
> Also noticed the `activitypub` custom field got set to, I think, `federate`. I may have to address that, too.

Ha, it does that because it is updated before scheduling. I'd have to look up what it's used for exactly, but may want to set it only if a post _really did federate_.

janboddez:
OK, so conneg can be fixed by moving the `is_wp_error( $transformer )` check to the `redirect_template` callback.

But I found another challenge: the post still shows in outboxes.

janboddez:
@pfefferle See https://github.com/Automattic/wordpress-activitypub/pull/731, where I merely add a couple error checks, which, as a side effect, allow excluding posts from being federated (both outgoing requests and attempts to fetch them from the Fediverse).

janboddez:
Just wanted to add that I now do use `activitypub_send_activity_to_followers` instead, which sort of does the same thing---right?

Together with an early `template_include` (to remove the conneg filter for certain posts) and custom `pre_get_posts` (for the outboxes and stuff) filter, it lets me disable federation all the way (for certain categories, at least).

Issue #726, opened 2024-04-10
Enable Authorized Fetch by default


moyitpro:
### What

Authorized fetch should be enabled by default as more servers enable it to make blocks more effective against Meta Threads. On servers that have Authorized fetch enabled, if a WordPress blog doesn't enable it, that server won't load the post nor the user can follow it.

### Why

By enabling authorized fetch enabled, servers that have it enabled can view and interact with the post. It may have performance implications as every request needs to be signed, but probably not.

### How

Set the following:
\defined( 'ACTIVITYPUB_AUTHORIZED_FETCH' ) || \define( 'ACTIVITYPUB_AUTHORIZED_FETCH', false );
to true.

Post shown on an instance that has Authorized Fetch enabled:
<img width="622" alt="2024-04-10_19-07-30" src="https://github.com/Automattic/wordpress-activitypub/assets/161183/7ebff6f0-6500-4c97-ac58-8b27d7e43def">

There should be a configurable option from the ActivityPub settings to enable/disable it, and not PHP constants, which I find doesn't work to enable it.

pfefferle:
I think we have to improve the authorized fetch feature first. The current implementation only limits the access to the API endpoints, but it still shows all profile informations and posts.

mediaformat:
@pfefferle what would you think about enabling Authorized fetch via a UI setting, with appropriate warning message and affordance? 

I could submit a PR for this.

PR #728, opened 2024-04-16
Improve AUTHORIZED_FETCH

pfefferle:
Improves Authorized-Fetch #726 

And refactors Application handling.

/cc @mediaformat @Menrath


pfefferle:
Should we add a setting to enable/disable `AUTHORIZED_FETCH`?

mattwiebe:
> Should we add a setting to enable/disable `AUTHORIZED_FETCH`?

Seems like we should, it's already an option for Mastodon admins IIRC? Since there may be some performance and compatibility tradeoffs, we should spell those out

activitypub.php
```diff
@@ -50,7 +50,7 @@
  * Initialize REST routes.
  */
 function rest_init() {
-	Rest\Users::init();
+	Rest\Actors::init();
```
Menrath, line 53: So it did happen after all! :) I personally think this is a really welcome change.

pfefferle: this change is dedicated to you! :)

mediaformat:
I agree with @mattwiebe we should clearly articulate the performance trade offs. 

Perhaps the option shouldn't just be a plain checkbox on the settings page, but inside a modal (confirmation) or disclosure element to make it explicit.

mediaformat:
@pfefferle it looks fine, i can test if needed

includes/class-activitypub.php
```diff
@@ -107,6 +107,7 @@ public static function render_json_template( $template ) {
 
 		if ( \is_author() ) {
 			$json_template = ACTIVITYPUB_PLUGIN_DIR . '/templates/author-json.php';
+			$template      = ACTIVITYPUB_PLUGIN_DIR . '/templates/author-json.php';
```
mediaformat, line 110: This ends up enabling unsigned get requests for actors

includes/class-activitypub.php
```diff
@@ -115,9 +116,37 @@ public static function render_json_template( $template ) {
...
+								)
+							)
+						);
+
+						return $activity;
+					}
+				);
+
 				// fallback as template_loader can't return http headers
 				return $template;
```
mediaformat, line 151: Only requests for the Application actor should be allowed to bypass signature verification (to prevent a deadlock loop).

The current code sends a 401 and the whole actor object!

mediaformat: Also, I'm not sure the application actor needs a non-REST profile!?

So the above mentioned signature-bypass should probably only apply in the REST Server class.

(Unless I am missing something :-)

mediaformat (changes requested):
Actor profiles are available with unsigned requests, only Application actor should be available unsigned.

includes/class-activitypub.php
pfefferle: I thought about having them available but returning only a limited set of data then? (See some lines below)

Wasn't sure about the current best practice.

includes/class-activitypub.php
pfefferle: The whole object? It should only send a limited subset with the signature!?! Can you verify this? But I can remove it completely if the Application is enough.

madcitymultimedia (approved):

includes/class-activitypub.php
mediaformat: I believe the current best practice is outlined [here](https://swicg.github.io/activitypub-http-signature/#instance-actor):
> [...] servers with authorized fetch enabled often use an [instance actor](https://seb.jambor.dev/posts/understanding-activitypub-part-4-threads/#the-instance-actor) to sign object fetches. **This is generally a "server-level" actor, separate from any normal user, that doesn't require an HTTP Signature to be fetched**. This breaks the loop of fetching each actor back and forth to validate their signatures.

includes/class-activitypub.php
mediaformat: Right, not the whole actor, but this would prevent enforcing a server level block.

pfefferle:
@mediaformat resolved!

Issue #631, opened 2024-01-04
Feature Request (and bug?): validate posts (and comments?) before they are sent


janboddez:
### What

As [mentioned elsewhere](https://github.com/Automattic/wordpress-activitypub/issues/628#issuecomment-1875066820), I'm not sure posts are sufficiently validated before being federated. ~~(I may be wrong, though!)~~

In another plugin I created, I also allow scheduling of HTTP POST requests in a very similar manner, and I basically check (once more!) that a post is still published, isn't password protected, and so on, right before it is being sent out.

~~Now, I may be overthinking this, or it may already be in one of the `Transformer` classes that I've missed. In that case, feel free to close this issue.~~

### Why

If users use _actual cron_ and have WP-Cron disabled as (used to be?) suggested by the plugin, HTTP POST requests (to followers' outboxes) could be delayed by as much as, e.g., 5 or even 10 minutes. (And even if the cron job runs every minute or if they do rely on WP-Cron, there might similarly be race conditions or something.)

A post may have changed sufficiently between them clicking "Publish" for the very first time and it being sent out. Seems to me that all these checks (plus, possibly, checks for mentions/tags/etc.) have to be re-run (or even delayed to the point) when the post is _actually_ going to get sent.

(Also, if you ever wanted to offer a checkbox or so to opt-out and it was going to be a _classic_ meta box [and not Gutenberg-powered], there's this "weird" behavior where Gutenberg will _first_ publish the post [triggering the cron event to be scheduled] but only _then_ allow processing of the meta box's values.)

TL;DR: "race" conditions.

### How

Same way you do, e.g.: https://github.com/Automattic/wordpress-activitypub/blob/4c297acfd454a13260900ff0f97a9e3f425e960c/includes/class-activity-dispatcher.php#L74

Except run additional checks.

Issue #668, opened 2024-01-18
Feature Request: Use more appropriate hook for scheduling activities


janboddez:
### What

Related: the second half of [this comment here](https://github.com/Automattic/wordpress-activitypub/issues/667#issuecomment-1899314788).

Long story short: I noticed quite some inconsistent "tag" behavior, probably related to my (experimental!) "limiting updates."

Reminds me a whole lot of what I noticed developing a plugin that similarly relied on `transition_post_status`. The "wrong" combination of meta boxes and the block editor could really throw things off.

Therefore, I was wondering if it wouldn't be safer to use a later hook, like, [`rest_after_insert_{$post->post_type}`](https://github.com/WordPress/wordpress-develop/blob/0c7b78e99da9094a72a050c32657d084526b37dd/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L772), _for Gutenberg posts_. (Which is like opening a whole other can of worms, but ...)

### Why

Depending on whether meta boxes are visible in the block editor, and on whether you use `DISABLE_WP_CRON`, and on whether you use the block editor at all, WordPress behaves somewhat differently, and I think this affects what types of activities are sent when.

Now I don't expect this to get "solved" anytime soon as most people won't notice or suffer from this. Because _most of the time_ a scheduled event won't run until _somewhat_ later in time.

Thing is, there's a _super slight chance_ (?) that not all of a post's metadata is processed when an activity is sent out. A kind of race condition. (Yet, even if that happens, a subsequent "update" will _almost_ always fix it.)

### How

_If you were to ever_ "limit" updates to prevent "waste," it may become needed to either delay federation more or use a "better" hook for certain scenarios. One could probably use `REST_REQUEST` to set up the "correct" hook (e.g., either `rest_after_insert*` or `transition_post_status` or similar.

Either way, thought it would be "good" to have this written down somewhere.

janboddez:
Another thing I forgot to mention is that as soon as meta boxes are being used in combination with the Gutenberg editor, that `transition_post_status` is run twice.

That was [probably the reason for my tags not being sent](https://github.com/Automattic/wordpress-activitypub/issues/667#issuecomment-1899314788), in combination with my "personal solution" (an "add-on plugin") to #628.

First time, `post_status` moves to `publish`, but tags aren't saved yet. Create event is scheduled.

[...]

PR #731, opened 2024-04-20
Improve transformer

janboddez:
Adds "improved" error handling to code that uses the `Transformer` class(es). (Really, all it does is add a bunch of `is_wp_error( $transformer )` checks!)

(The current version of the plugin has some checks in place, but not enough to prevent fatal errors down the line, _should one ever attempt to pass an invalid object or transformer_ [to the rest of the code].)

As a bonus, it allows _somewhat experienced PHP developers_ to _exclude_ certain posts (and, potentially, comments) from federation. (Which is an oft-requested feature!)

## Proposed changes:
* Add better "transformer error" handling to JSON templates and ActivityPub dispatcher
* Add better "transformer error" handling to outbox and collection endpoint
    * As a result, "method chaining" is no longer possible (I think); I'd suggest using the "Factory" everywhere, make things a bit simpler
        * If this is somehow not an option, then `Activitypub\Transformer\Base::transform()` has to become filterable, too (to prevent "unwanted" posts from showing in outboxes)
            * This PR currently includes these changes, but they could be scrapped if this function is no longer used
        * If it is, maybe deprecate this function in favor of `Factory::get_transformer()`? 😬

### Other information:
* As mentioned above, this _could_, potentially, be used to _exclude posts from federation_.
* While I see some challenges (related to [#668](https://github.com/Automattic/wordpress-activitypub/issues/668)) using the filter _for that purpose_, that's ... a different issue, really. (Probably should not make things this confusing ... but I wanted to mention it regardless.)
* The only "real" downside I see is that by _allowing outboxes and collections to be filterable_, that the per-page counts will no longer "be correct" (But only _if such a filter is actually in use_. Also, I'm not sure if any software currently relies on outboxes, or if they expect a page to hold a certain number of items _always_.)
    * Reason for this is filtering happens _after_ fetching the items from the database ... (There's currently no way to filter "outbox _queries_." Unless maybe we use the `federated` custom field for that.)
    * So that _could_ be a reason to leave (only) the outbox/collection endpoints as they currently are.
 
- [ ] I did not run or add automated tests (no Docker on this machine, although I don't expect problems, the code does not change the plugin's default behavior) 😬
- [x] I absolutely tested in production

## Testing instructions:
I'm going to just outline what I did (and do):
- Add an mu-plugin like this one:
```
<?php

add_filter( 'activitypub_transformer', function ( $transformer, $obj ) {

	if ( $obj instanceof \WP_Post && in_category( 'disable-activitypub', $obj ) ) {
		return new \WP_Error( 'unsupported_object', __( 'This object does not support ActivityPub.', 'activitypub' ) );
	}

	return $transformer;

}, 10, 2 );
```
- Create a new post, and make sure it has the `disable-activitypub` category
- Save it as draft, to _ensure Gutenberg has applied this category_ before "federation" is scheduled
- Publish the post
- Watch _no_ outgoing POST requests happen
- Finally, try searching for the post's URL in, e.g., Mastodon. Nothing should show up.
- Try visiting the `?activitypub` version of the posts, in your own browser. You should "simply" see the post, _not_ JSON.
- If you used such a mu-plugin (i.e., an `activitypub_transformer` filter that does _not_ return an actual transformer) with the currently released version of the plugin, you'd instead get a fatal PHP error.

The mu-plugin can "easily" be adapted to "exclude" post formats, posts with certain custom fields, and so on. (The only requirement is that these are stored before scheduling runs, which _could be tricky when you use the block editor_. As per the linked issue above.)

### DIFF
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -110,6 +110,14 @@ class Activitypub {
 		} elseif ( is_comment() ) {
 			$json_template = ACTIVITYPUB_PLUGIN_DIR . '/templates/comment-json.php';
 		} elseif ( \is_singular() ) {
+			global $post;
+
+			// So... this gets done in the template, too, but then it's too late.
+			$transformer = \Activitypub\Transformer\Factory::get_transformer( $post );
+			if ( \is_wp_error( $transformer ) ) {
+				return $template;
+			}
+
 			$json_template = ACTIVITYPUB_PLUGIN_DIR . '/templates/post-json.php';
 		} elseif ( \is_home() ) {
 			$json_template = ACTIVITYPUB_PLUGIN_DIR . '/templates/blog-json.php';
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -5,6 +5,7 @@ use Exception;
 use Activitypub\Signature;
 use Activitypub\Collection\Users;
 use Activitypub\Collection\Followers;
+use Activitypub\Transformer\Factory;
 
 use function Activitypub\is_comment;
 use function Activitypub\sanitize_url;
@@ -108,12 +109,18 @@ class Activitypub {
 		if ( \is_author() ) {
 			$json_template = ACTIVITYPUB_PLUGIN_DIR . '/templates/author-json.php';
 		} elseif ( is_comment() ) {
+			$comment     = \get_comment( \get_query_var( 'c', null ) );
+			$transformer = Factory::get_transformer( $comment );
+
+			if ( \is_wp_error( $transformer ) ) {
+				return;
+			}
+
 			$json_template = ACTIVITYPUB_PLUGIN_DIR . '/templates/comment-json.php';
 		} elseif ( \is_singular() ) {
 			global $post;
 
-			// So... this gets done in the template, too, but then it's too late.
-			$transformer = \Activitypub\Transformer\Factory::get_transformer( $post );
+			$transformer = Factory::get_transformer( $post );
 			if ( \is_wp_error( $transformer ) ) {
 				return $template;
 			}
--- a/includes/rest/class-collection.php
+++ b/includes/rest/class-collection.php
@@ -169,7 +169,13 @@ class Collection {
 		);
 
 		foreach ( $posts as $post ) {
-			$response['orderedItems'][] = Post::transform( $post )->to_object()->to_array( false );
+			$post = Post::transform( $post );
+
+			if ( \is_wp_error( $post ) ) {
+				continue;
+			}
+
+			$response['orderedItems'][] = $post->to_object()->to_array( false );
 		}
 
 		$rest_response = new WP_REST_Response( $response, 200 );
--- a/includes/rest/class-collection.php
+++ b/includes/rest/class-collection.php
@@ -1,13 +1,12 @@
 <?php
 namespace Activitypub\Rest;
 
-use WP_Error;
 use WP_REST_Server;
 use WP_REST_Response;
-use Activitypub\Transformer\Post;
 use Activitypub\Activity\Actor;
 use Activitypub\Activity\Base_Object;
 use Activitypub\Collection\Users as User_Collection;
+use Activitypub\Transformer\Factory;
 
 use function Activitypub\esc_hashtag;
 use function Activitypub\is_single_user;
@@ -169,13 +168,13 @@ class Collection {
 		);
 
 		foreach ( $posts as $post ) {
-			$post = Post::transform( $post );
+			$transformer = Factory::get_transformer( $post );
 
-			if ( \is_wp_error( $post ) ) {
+			if ( \is_wp_error( $transformer ) ) {
 				continue;
 			}
 
-			$response['orderedItems'][] = $post->to_object()->to_array( false );
+			$response['orderedItems'][] = $transformer->to_object()->to_array( false );
 		}
 
 		$rest_response = new WP_REST_Response( $response, 200 );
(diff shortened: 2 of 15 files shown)


janboddez:
In the future, one could envision a per-post checkbox to disable federation, and store its value in a custom field. You'd then "simply" have to look for that in an `activitypub_transformer` callback.

(The issue there would be to get Gutenberg to save the checkbox value _before_ ActivityPub requests are scheduled, which what #668, even if it currently isn't a problem, is all about.)

It could even be used to fix #631. (In that issue, I describe being able to federate a _private post_, simply because, well, it was made private after it got scheduled but before federation happened.) "Simply" use a `activitypub_transformer` callback to ensure a post isn't private.

includes/rest/class-collection.php
```diff
@@ -169,7 +168,13 @@ public static function featured_get( $request ) {
 		);
 
 		foreach ( $posts as $post ) {
-			$response['orderedItems'][] = Post::transform( $post )->to_object()->to_array( false );
+			$transformer = Factory::get_transformer( $post );
```
pfefferle, line 171: If we check it here, do we have to check it again in the template?

includes/transformer/class-base.php
```diff
@@ -30,9 +31,37 @@ abstract class Base {
 	 *
 	 * @param WP_Post|WP_Comment $wp_object The WordPress object
 	 *
-	 * @return Base_Object
+	 * @return Base
 	 */
-	public static function transform( $object ) {
+	public static function transform( $object ) { // phpcs:ignore Universal.NamingConventions.NoReservedKeywordParameterNames.objectFound
```
pfefferle, line 36: Do we need this filterable too? You can filter it already on the Factory level!

https://github.com/Automattic/wordpress-activitypub/blob/d94d064acbc42d371065f3d3839cd949eee0b7c7/includes/transformer/class-factory.php#L50

janboddez: The collection and outbox endpoints don't use the Factory, as far as I can tell. That's why I added it in the `Base` class as well.

But: I'd actually suggest to always use the Factory (and not `Post::transform()`, which also returns a transformer).

Before, you'd do `Post::transform( $post )->to_object()`, but if you're going to check that `Post::transform()` really _does_ return a transformer, then you can't "chain" anyways. (And even then, you could've also done `Factory::get_transformer( $post )->to_object()`. But I'm guessing this code "grew organically" and that's how it ended up.)

So maybe I should just replace `$transformer = Post_Transformer::transform( $post );` with `$transformer = Factory::get_transformer( $post );`. Then the code would be the same always and we can drop the filter from the `Base` class (and always use the Factory instead).

pfefferle: I see, I agree with you on always using the transformer! Maybe force it, by protecting the `__construct`.

pfefferle: Maybe we can keep the chain and have the sanity check in the `to_*` methods?

janboddez:
> This has one "side effect," potentially. If a post's transformer is somehow invalid (e.g., because it was filtered and a site author wanted to exclude certain posts from "federation"), then it'll be skipped, while it was returned by the query. So your "page count" wouldn't match the actual number of items on that "page." I don't think this is a huge issue, though. (Still, if we could somehow exclude items during the "query," rather than skip over them while looping, that'd be even better.)

This could be fixed by allowing the `get_post()` arguments of the outbox and collection queries to be filterable.

If someone wanted to disable a certain post format, they need to disable active federation but also conneg for the posts in question (this could be done, at least for now, by filtering the transformer as outlined above).

They would also have add a `tax_query` to the `get_posts()` used in the collection and outbox queries.

Anyway, this PR isn't so much about disabling federation; it just catches possible errors. :-D

includes/class-activitypub.php
```diff
@@ -108,8 +109,22 @@ public static function render_json_template( $template ) {
 		if ( \is_author() ) {
 			$json_template = ACTIVITYPUB_PLUGIN_DIR . '/templates/author-json.php';
 		} elseif ( is_comment() ) {
+			$comment     = \get_comment( \get_query_var( 'c', null ) );
```
pfefferle, line 112: @janboddez we now check that twice... here and in the template...

pfefferle: @janboddez we also check this in the template, so we do the transformation twice! Maybe we can remove the check here or at least simplify it?!?

pfefferle (changes requested):

janboddez:
OK, I'm not too familiar with GitHub reviews and (un)resolving conversations, but:

- I removed the check from the `Base::transform()` function. This also means _that it should probably never be used anymore_. (I think it's safe to remove, in fact, as I replaced all calls with `Factory::get_transformer()` instead, which should be 100% compatible.)
- I removed the checks from the templates. That means the transformer still gets called twice, but it's (for now) the only way I see. (The "trick" I used to test still works, ;-) https://jan.boddez.net/notes/f8dab23f83?activitypub)

pfefferle:
It thought more about removing it from the "controller" and simply load the template and throw an error there if there is one? this way we have to only call it once! To hacky?

janboddez:
> I also thought about using the cache or even a transient, but I think that will be way worse in terms of performance.

Oh, maybe we can use that trick where you call a helper function (`\Activitypub\current_transformer()` or something, just thinking out loud) to set a `static` variable, and then call it without an argument later on to retrieve the value.

Since you know the transformer is indeed valid and the template is about to be loaded, it should be absolutely safe to store it for the rest of the request. That way you'd instantiate only one transformer, and you could do it before the template actually loads.

pfefferle:
Like the singleton approach?!?

janboddez:
> Like the singleton approach?!?

That would cause collections to fail, I think. More like this: https://github.com/Automattic/wordpress-activitypub/blob/f402e842ff20cd06a1008a20a70c6796ddb3dc56/includes/functions.php#L859-L873

pfefferle (approved):

PR #740, opened 2024-04-25
Only send Mentions to comments in the direct hierarchy

mattwiebe:
We will currently send an @-mention to every commenter on a given post, regardless of whether or not they were in the direct hierarchy of the comment being left. 

This patch will instead traverse up the comment tree in question and only send an `@-mention` to commenters in the direct tree we are on.

### Other information:

## Testing instructions:
 
1. Have a publicly accessible post that you comment against using a 3rd party Actor. 
2. Leave a top-level comment on that post as your own user. Without this patch, it will `@-mention` the user who you are not directly replying to, with this, it will not.
3. Make a direct reply to the received comment in 1). With or without this patch, that user should receive an `@-mention`.


pfefferle:
Too noisy??

pfefferle (dismissed):

mattwiebe:
> Too noisy??

I think so! Let's imagine this comment hierarchy, where `B` and `C` are both comments received from fediverse Actors on post `A`:

```
    A
   / \
  B   C
 /
D
```

If I as the blog author leave comment `D`, it will `@-ping` the Actor who commented in `C`, even though that author had nothing to do with that comment thread. Only `B` should be pinged. This will get much noisier in a much busier comment section.

PR #742, opened 2024-04-29
Add generic get_comment_ancestors function

pfefferle:
based in the `get_post_ancestors` method!

I have to test that a bit more, but I tried to adapt the `get_post_ancestors` to not have to use a limit!

What do you think @mattwiebe

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -854,3 +854,42 @@ function get_masked_wp_version() {
 
 	return implode( '.', $version );
 }
+
+/**
+ * Retrieves the IDs of the ancestors of a comment.
+ *
+ * Adaption of `get_post_ancestors` from WordPress core.
+ *
+ * @see https://developer.wordpress.org/reference/functions/get_post_ancestors/
+ *
+ * @param int|WP_Comment $comment Comment ID or comment object.
+ *
+ * @return WP_Comment[] Array of ancestor Comments or empty array if there are none.
+ */
+function get_comment_ancestors( $comment ) {
+	$comment = get_comment( $comment );
+
+	// phpcs:ignore Universal.Operators.StrictComparisons.LooseEqual
+	if ( ! $comment || empty( $comment->comment_parent ) || $comment->comment_parent == $comment->comment_ID ) {
+		return array();
+	}
+
+	$ancestors = array();
+
+	$id          = $comment->comment_parent;
+	$ancestors[] = $comment;
+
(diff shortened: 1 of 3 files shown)


dshanske:
Why not submit the function or idea to Core as well? Might be worth inclusion.

includes/functions.php
```diff
@@ -854,3 +854,42 @@ function get_masked_wp_version() {
...
+	// phpcs:ignore Generic.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
+	while ( $ancestor ) {
+		// Loop detection: If the ancestor has been seen before, break.
+		// phpcs:ignore Universal.Operators.StrictComparisons.LooseEqual
+		if ( empty( $ancestor->comment_parent ) || ( $ancestor->comment_parent == $comment->comment_ID ) || in_array( $ancestor->comment_parent, $ancestors, true ) ) {
+			break;
+		}
+
+		$ancestor    = \get_comment( $comment->comment_parent );
+		$ancestors[] = $ancestor;
```
mattwiebe, line 891: We don't know yet here if `$ancestor` returned a comment. If not, we should break and not append to the array.

includes/functions.php
```diff
@@ -854,3 +854,42 @@ function get_masked_wp_version() {
...
+ *
+ * @param int|WP_Comment $comment Comment ID or comment object.
+ *
+ * @return WP_Comment[] Array of ancestor comments or empty array if there are none.
+ */
+function get_comment_ancestors( $comment ) {
+	$comment = get_comment( $comment );
+
+	// phpcs:ignore Universal.Operators.StrictComparisons.LooseEqual
+	if ( ! $comment || empty( $comment->comment_parent ) || $comment->comment_parent == $comment->comment_ID ) {
```
mattwiebe, line 873: The `in_array` check here and in the loop won't work because we're appending the full comment to the `$ancestors` array, while `get_post_ancestors()` only uses the post IDs.

mattwiebe (changes requested):
Love the idea of taking inspiration from `get_post_ancestors()`! 

I noticed a couple of potential problem points in the code, see detailed inline comments.

includes/functions.php
pfefferle: should we also only use IDs to be consistent?

mattwiebe: Probably for the best, we can grab the full comments when needed.

pfefferle: done

mattwiebe (approved):
:shipit:

PR #749, opened 2024-05-01
Enable Mastodon Apps Integration: Add remote context

akirk:
When clicking on a remote status, this will fetch the replies to this status via ActivityPub and return them in the context API call.

### DIFF
--- a/integration/class-enable-mastodon-apps.php
+++ b/integration/class-enable-mastodon-apps.php
@@ -343,13 +343,8 @@ class Enable_Mastodon_Apps {
 			return $statuses;
 		}
 
-		$response = Http::get( $data['outbox'], true );
-		if ( is_wp_error( $response ) || wp_remote_retrieve_response_code( $response ) !== 200 ) {
-			return $statuses;
-		}
-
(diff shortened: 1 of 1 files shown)


integration/class-enable-mastodon-apps.php
```diff
@@ -300,67 +366,53 @@ public static function api_statuses_external( $statuses, $args ) {
...
-				$status->content    = $object['content'];
-				$status->account    = $account;
+		return $activitypub_statuses;
+	}
 
-				if ( ! empty( $object['inReplyTo'] ) ) {
-					$status->in_reply_to_id = $object['inReplyTo'];
-				}
+	public static function api_get_replies( $context, $post_id, $url ) {
+		$response = Http::get( $url, true );
```
pfefferle, line 378: should we us https://github.com/Automattic/wordpress-activitypub/blob/74403ff37f30f1cdd28a48d7b7993900bf370718/includes/class-http.php#L166 instead?

akirk: Nice, ok!

pfefferle (approved):

PR #751, opened 2024-05-02
Improve attachment handling

pfefferle:
This should reduce the load of the fediverse crawlers after federating a post.

Thanks @tellyworth for the idea!


mattwiebe (approved):
LGTM!

PR #759, opened 2024-05-12
Enable Mastodon Apps: Add support for fetching external statuses without replies

akirk:
Before, we'd just fetch any statuses. Now this supports paging and distinguishes between replies and non-replies.


pfefferle (approved):
LGTM

PR #762, opened 2024-05-17
Add extra fields

### LINKED ISSUE #810: Actor `attachment` objects are currently invalid per the spec
### Quick summary

Hi there, the code [1] that produces the `attachment` array for actors seems to be invalid according to the [spec for `attachment`](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-attachment)

Instead of producing `PropertyValue` items (which are neither valid `Object` or `Link` items), it should instead probably produce `Link` items (here is the [spec for `Link`](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-link))

[1] https://github.com/Automattic/wordpress-activitypub/blob/4397ae95c8b7e6f9a44283a84ec99be91629257a/includes/model/class-user.php#L235

### Steps to reproduce

Send an AP request here: https://fediversereport.com/author/laurenshof/

Valid `Object` or `Link` in the `attachment` values

### What actually happened

Invalid `Object` or `Link` in the `attachment` values

### Impact

All

### Available workarounds?

### Logs or notes

_No response_

pfefferle:
Make extra fields editable

Fix #810 

Screenshot 2024-07-15 at 11 37 18
Screenshot 2024-07-15 at 11 37 33


includes/class-activitypub.php
```diff
@@ -463,13 +465,43 @@ private static function register_post_types() {
 			)
 		);
 
+		\register_post_type(
+			'ap_extrafield',
+			array(
+				'labels'           => array(
+					'name'          => _x( 'Extra fields', 'post_type plural name', 'activitypub' ),
```
mattwiebe, line 472: I think that "Extra fields" is too generic? Should we qualify it, as in "ActivityPub extra Field" instead?

includes/class-activitypub.php
```diff
@@ -478,4 +510,17 @@ public static function user_register( $user_id ) {
...
+	/**
+	 * Filter the query to only show the current author's extra fields
+	 *
+	 * @param WP_Query $query The WP_Query instance (passed by reference).
+	 *
+	 * @return void
+	 */
+	public static function filter_get_posts_query( $query ) {
+		if ( $query->get( 'post_type' ) === 'ap_extrafield' ) {
+			$query->set( 'author', get_current_user_id() );
```
mattwiebe, line 523: I'm trying to understand why we need this? The `author` field is already specified in the only two `WP_Query` instances trying to get this data.

includes/model/class-user.php
```diff
@@ -233,9 +233,44 @@ public function get_endpoints() {
 	 * @return array The extended User-Output.
 	 */
 	public function get_attachment() {
-		$array = array();
+		$extra_fields = new WP_Query(
+			array(
+				'post_type'      => 'ap_extrafield',
+				'posts_per_page' => -1,
```
mattwiebe, line 239: You can use `'nopaging' => true` in place of -1 on `posts_per_page` for a bit more clarity.

mattwiebe (commented):
Having these available will be nice! Left a few comments to clarify. 

And it would be good to have these available for the Blog <strike>User</strike> as well!

includes/class-activitypub.php
pfefferle: I thought this might be enough, because it is hidden in the menu and the only entry point is the ActivityPub section in the user settings, so it should be clear that this is ActivityPub related. But I have no strong opinion on that and it is fine to change it.

includes/class-activitypub.php
pfefferle: This shows only the extra field of a user and hides the extra fields from the others in the wp-admin pages.

includes/class-activitypub.php
mattwiebe: Good points! Let's leave it for now, we can always revisit later

includes/class-activitypub.php
mattwiebe: Ok! Then it would probably be good to either 1) check that `author` isn't already set before setting it, or 2) only add this filter on that page? Again, not that important, we can revisit if it's a problem.

mattwiebe (approved):
🚢

includes/class-activitypub.php
pfefferle: Makes sense! I moved the filters around a bit.

akirk (commented):
Screenshots in the PR description would be nice!

As we discussed, it'd be nice to have some form of migration and do display the defaults here:
Screenshot 2024-06-04 at 10 20 39

To me it was surprising to arrive in the Gutenberg editor when I clicked "Add new", it felt like a bug. What about just adding another row for quickly adding a new entry and the Edit link gives you full-text editing?

Mockup:
Screenshot 2024-06-04 at 10 23 40
```html
<p>
<input class="regular-text" type="text" size="5" value="" placeholder="Field">
<input class="regular-text" type="text" value="" placeholder="Value">
</p>
```

pfefferle:
Should now also support [FEP-fb2a](https://codeberg.org/fediverse/fep/src/branch/main/fep/fb2a/fep-fb2a.md):

```JSON
"attachment": [
  {
    "type": "PropertyValue",
    "name": "Pronouns",
    "value": "<p>he/him/his</p>"
  },
  {
    "type": "Note",
    "name": "Pronouns",
    "content": "<p>he/him/his</p>"
  },
  {
    "type": "PropertyValue",
    "name": "Blog",
    "value": "<p><a rel=\"me noopener\" title=\"http://localhost:8076/\" target=\"_blank\" href=\"http://localhost:8076/\">localhost</a></p>"
  },
  {
    "type": "Link",
    "name": "Blog",
    "href": "http://localhost:8076/",
    "rel": [
      "me",
      "noopener"
    ]
  }
]
```

/cc @johnspurlock

jeherve:
> To me it was surprising to arrive in the Gutenberg editor when I clicked "Add new", it felt like a bug.

Coming in late to this discussion, to say that I experienced the same thing. To be honest, I'm still not sure I added the new field correctly. I created a new entry, with a title and just the link in the post content, but that didn't seem to work. I had to go back and add text with a link on it for it to appear in the profile page.

I like the idea of making this editable directly through the Profile page.

pfefferle:
> Coming in late to this discussion, to say that I experienced the same thing. To be honest, I'm still not sure I added the new field correctly. I created a new entry, with a title and just the link in the post content, but that didn't seem to work. I had to go back and add text with a link on it for it to appear in the profile page.

This might be a bug, I will check this.

> I like the idea of making this editable directly through the Profile page.

We can add that feature, but I still like having the Block editor to help with text and HTML. The extra fields are not limited to URLs, so why not profit from the build in features of WordPress?

jeherve:
> The extra fields are not limited to URLs, so why not profit from the build in features of WordPress?

This is true. Maybe we could display a small rich text editable input field on the profile page?

pfefferle:
@jeherve Auto-Link comes with #849 btw.

jeherve:
Thank you!

PR #764, opened 2024-05-21
Always fall back to `get_by_username`

pfefferle:
Always fall back to `get_by_username`, even if `get_by_id` and `get_by_resource` fail.

This would allow us to also handle WebFinger resources without `acct:` scheme and still fall back to emails as user-names.


akirk (approved):
Nice!

PR #775, opened 2024-06-03
better conneg handling

### LINKED ISSUE #743: Comment JSON never loaded?
### Quick summary

An `application/activity+json` request for a comment seems to return the homepage's HTML instead.

### Steps to reproduce

1. Use cURL to request a comment's Activity Streams representation. E.g., `curl -H "Accept: application/activity+json" https://jan.boddez.net/?c=2133`.
2. See the homepage HTML being returned.

That the comment's JSON is returned.

### What actually happened

Got HTML for the homepage instead.

I quickly gathered some debug logging think the problem may be here: https://github.com/Automattic/wordpress-activitypub/blob/119bab3cb40f65095d1c8b59f40157aeb9844768/includes/class-activitypub.php#L104

Because a comment ID looks like the home URL but I've got the "blog user" disabled, it stops here (for the example request above).

I think the "comment check" should precede this condition.

### Impact

Most (> 50%)

### Available workarounds?

### Logs or notes

_No response_

pfefferle:
Should fix #743

/cc @janboddez

### DIFF
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -9,6 +9,7 @@ use Activitypub\Collection\Followers;
 use function Activitypub\is_comment;
 use function Activitypub\sanitize_url;
 use function Activitypub\is_local_comment;
+use function Activitypub\is_user_type_disabled;
 use function Activitypub\is_activitypub_request;
 use function Activitypub\should_comment_be_federated;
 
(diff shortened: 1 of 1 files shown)


akirk (approved):
Now my comment loads as ActivityPub even when I have the blog user disabled.

PR #752, opened 2024-05-02
Improve lemmy compatibility

pfefferle:
Improve the compatibility with lemmy! Thanks @Nutomic for the feedback https://socialhub.activitypub.rocks/t/discourse-wordpress-integration/4162/13?u=pfefferle

PR #778, opened 2024-06-07
Lemmy seems to still check the charset case sensitive

pfefferle:
Lemmy seems to still check for a lowercase charset.

@Nutomic what about your PR for check case insensitive? https://github.com/Automattic/wordpress-activitypub/pull/752

I tested on https://voyager.lemmy.ml/ and on https://lemmy.ml/

From the spec:

> The charset attribute specifies the [character encoding](https://encoding.spec.whatwg.org/#encoding) used by the document. This is a [character encoding declaration](https://html.spec.whatwg.org/multipage/semantics.html#character-encoding-declaration). If the attribute is present, its value must be an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "utf-8".

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/rest/class-actors.php
+++ b/includes/rest/class-actors.php
@@ -91,7 +91,7 @@ class Actors {
 		$json = $user->to_array();
 
 		$rest_response = new WP_REST_Response( $json, 200 );
-		$rest_response->header( 'Content-Type', 'application/activity+json; charset=' . get_option( 'blog_charset' ) );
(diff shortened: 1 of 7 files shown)


Nutomic:
Oh it looks like I messed up the PR and didnt actually make the check case insensitive.

https://github.com/LemmyNet/activitypub-federation-rust/pull/114

pfefferle:
Fixed in lemmy! thanks @Nutomic

PR #781, opened 2024-06-10
More efficient `WP_User_Query` with `count_total=false`

mattwiebe:
We have some extra protections on dotcom for counting users in User queries, since we have a lot of users. Since we don't the total count anyway, we can make our query more efficient for everyone by skipping it :)

## Proposed changes:

* Add `count_total => false` to our `WP_User_Query` instances


pfefferle (approved):
good point! thanks!

PR #788, opened 2024-06-23
Enable Mastodon Apps: support profile editing, blog user

mattwiebe:
Below is a screen recording of me using this PR to drive a Mastodon app, providing full profile editing support through 3rd party apps. (still requires https://github.com/akirk/enable-mastodon-apps/pull/167 to work for Extra Fields)

https://github.com/user-attachments/assets/bd6aa0b9-7fb7-4771-b627-2c8edbb92d99

## Proposed changes:
* Fully support the profile editing route: name, bio, avatar, header, extra fields
* Use the Blog User account in the app when only the blog user is enabled

### Other information:

## Testing instructions:

* Go to '..'
*


pfefferle:
I think it is not a good practice to directly save informations that are added by a `setter`?!?

mattwiebe:
> I'm not sure if it's a good idea to store the information directly in the `setter`!? I think that's not what a model/bean is meant to be!?

Can you explain more? I don't follow. I've used Models to coordinate CRUD layers before, but always open to learning a better way of doing things!

pfefferle:
I am sorry, my brain is still a but fuzzy!

It is not about the CRUD idea, it is about directly saving the data to the DB. Normally `getters` and `setters` are to get and set object attributes. I know that we are not consistent yet, because the `getters` also directly load stuff from the DB (maybe we have to refactor this too), but the `from_array` or `from_json` methods do use the [`set_` methods](https://github.com/Automattic/wordpress-activitypub/blob/5e765e0b6254a76f76337a1a7e478461c8ffbd1d/includes/activity/class-base-object.php#L618) to prefill the objects and this might cause strange side effects if we write directly to the db.

Maybe we should use this feature to refactor the code a bit and to consistently use attributes and maybe have a [`save` function](https://github.com/Automattic/wordpress-activitypub/blob/5e765e0b6254a76f76337a1a7e478461c8ffbd1d/includes/model/class-follower.php#L143), that finally persists things.

integration/class-enable-mastodon-apps.php
```diff
@@ -34,6 +37,72 @@ public static function init() {
...
+
+		if ( ! $user || is_wp_error( $user ) ) {
+			return;
+		}
+
+		if ( isset( $data['avatar'] ) ) {
+			$icon_id = (int) $data['avatar'];
+			$attachment = \get_post( $icon_id );
+			if ( $attachment && 'attachment' === $attachment->post_type ) {
+				l( 'set_icon' );
```
pfefferle, line 75: this should cause errors outside of the wp.com context.

integration/class-enable-mastodon-apps.php
```diff
@@ -26,14 +29,84 @@ class Enable_Mastodon_Apps {
...
+	}
+
+	/**
+	 * Map user to blog if user is disabled
+	 *
+	 * @param int $user_id The user id
+	 *
+	 * @return int The user id
+	 */
+	public static function maybe_map_user_to_blog( $user_id ) {
```
pfefferle, line 52: is this working?

includes/model/class-user.php
```diff
@@ -97,6 +97,17 @@ public function get_name() {
 		return \esc_attr( \get_the_author_meta( 'display_name', $this->_id ) );
 	}
 
+	/**
+	 * Set the User-Name.
+	 *
+	 * @param string $name The User-Name.
+	 * @return bool       True if the User-Name was updated, false otherwise.
+	 */
+	public function save_name( $name ) {
```
pfefferle, line 106: I would prefer to move the save functions to the EMA-API class instead then, because they are very specific and I do not know if we can re-use them in other parts of the code.

pfefferle:
@mattwiebe There is no way you can login as the blog author (yet), or do you have found a way?

The other thing: Some of the informations seem to be generic and not related to ActivityPub, so why not move them to Enable-Mastodon-Apps instead? Like `save_name` and `save_description` for example.

includes/model/class-user.php
```diff
@@ -146,7 +176,30 @@ public function get_icon() {
...
+		$maybe_id = (int) $icon;
+		if ( $maybe_id ) {
+			$image = wp_get_attachment_image( $maybe_id, 'full' );
+			if ( ! $image ) {
+				return false;
+			}
+			$icon = wp_get_attachment_url( $maybe_id );
+		}
+
+		$option_name = 'activitypub_user_icon_' . $this->_id;
```
pfefferle, line 189: For signatures we use `$user->user_login` instead, because the ID can change on migrations!?!

integration/class-enable-mastodon-apps.php
mattwiebe: It is! I use it inside a couple of our callbacks to change somebody's WP User ID to the Blog User ID only when the user type is disabled and the blog is enabled. In this way a user who is logged in is doing what the plugin supports: only the Blog User. It does not allow us to log in as the Blog User when both Blog and User types are enabled, but that seems to be a much smaller problem overall.

Screenshot 2024-06-26 at 12 48 08

includes/model/class-user.php
mattwiebe: For the User type I completely agree, as long as we and EMA are both using the same attributes

includes/model/class-user.php
mattwiebe: That seems like a pretty good idea. For dotcom Simple -> Atomic migrations we do maintain the User ID although that may not always be the case when we do something like a WXR Import

mattwiebe:
> @mattwiebe There is no way you can login as the blog author (yet), or do you have found a way?

Yup there is! See above, only when only the Blog User is active.

> The other thing: Some of the informations seem to be generic and not related to ActivityPub, so why not move them to Enable-Mastodon-Apps instead? Like `save_name` and `save_description` for example.

Those easily could, and should be supported in EMA itself, but it doesn't yet have a concept of header or avatar. The latter is still managed in Core through Gravatar, and there is only the (deprecated in Block Themes) header_image for the whole site. So perhaps EMA supports name and description, but we do avatar, header, and soon, extra fields

includes/model/class-user.php
pfefferle: But we should be consistent! I am already having a hard time with this login key on the deletion feature. If we decide to use the id instead then I will migrate that stuff!

includes/model/class-blog.php
```diff
@@ -368,4 +377,36 @@ public function get_indexable() {
...
+	 *                     - name: The User-Name.
+	 *                     - summary: The User-Description.
+	 *                     - icon: The User-Icon.
+	 *                     - header: The User-Header-Image.
+	 * @return bool True if the attribute was updated, false otherwise.
+	 */
+	public function save( $key, $value ) {
+		switch ( $key ) {
+			case 'name':
+				return \update_option( 'blogname', $value );
```
akirk, line 395: That seems a bit dangerous, you might not realize that you're changing the whole blog name. Should this just be an override?

includes/model/class-blog.php
```diff
@@ -368,4 +377,36 @@ public function get_indexable() {
...
+			case 'summary':
+				return \update_option( 'blogdescription', $value );
```
akirk, line 397: Same as above, maybe rather with an override?

includes/model/class-blog.php
mattwiebe: It's tricky. Once we introduce an override, now you have to go into wp-admin to also change the blogname, if that's what you want to do, and/or being able to delete the override, falling back to the blogname, which you can now only edit in wp-admin. If you want a consistent name, you have to go to wp-admin. I'm thinking more of somebody who never wants to go there at all.

I'm erring on the side of consistency, not on the side of safety. And if somebody doesn't like the change they made, they can go back and change it again, without having to go to wp-admin.

At least that's the kind of thinking I was doing when I made it behave this way

mattwiebe:
Now that https://github.com/akirk/enable-mastodon-apps/pull/157 is merged, my plan is to first get #838 merged, and then rework this PR to fully take advantage of 1) Header Images and 2) Extra Fields. So close!

includes/model/class-user.php
```diff
@@ -150,20 +158,15 @@ public function get_icon() {
 
 	public function get_image() {
 		$header_image = get_user_option( 'activitypub_header_image', $this->_id );
-		$image_url    = null;
-
-		if ( $header_image ) {
```
pfefferle, line 155: I think we can't remove the `\wp_get_attachment_url( $header_image );` because the wp-admin settings only saves the attachment-id. If the EMA plugin is doing it differently, we have to unify code.

mattwiebe: I think I found that we were saving just the URL, but I'll circle back and check

includes/model/class-user.php
```diff
@@ -277,4 +280,50 @@ public function get_indexable() {
...
+	 * @param string $key The attribute to update.
+	 * @param mixed $value The new value.
+	 *                     Possible values:
+	 *                   - name: The User-Name.
+	 *                   - summary: The User-Description.
+	 *                   - icon: The User-Icon.
+	 *                   - header: The User-Header-Image.
+	 * @return bool True if the attribute was updated, false otherwise.
+	 */
+	public function save( $key, $value ) {
```
pfefferle, line 296: I would rename that, so that we will be able to have a generic `save` in the future, that saves all attributes to the DB.

mattwiebe: Happy to have dedicated functions. I went in an `update_key` direction with it.

includes/model/class-user.php
mattwiebe: You were right!

includes/model/class-user.php
pfefferle: Sorry for maybe not being precise enough: My proposal was to simply not use the generic `save` function, but something different instead. I am fine with having one function with `key` => `value` params, but I am also fine with dedicated functions :)

mattwiebe:
> Now that [akirk/enable-mastodon-apps#157](https://github.com/akirk/enable-mastodon-apps/pull/157) is merged, my plan is to first get #838 merged, and then rework this PR to fully take advantage of 1) Header Images and 2) Extra Fields. So close!

Only Extra Fields remains of that. We could probably just ship Extra Fields in a subsequent PR, but otherwise I'll wrap this up early next week.

mattwiebe:
Extra Fields support went quicker than expected! 

We will require https://github.com/akirk/enable-mastodon-apps/pull/167 to test. I will come back and try to form some testing instructions later.

includes/collection/class-extra-fields.php
```diff
@@ -231,6 +238,64 @@ function( $rel ) {
...
+	}
+
+	private static function make_paragraph_block( $content ) {
+		if ( ! site_supports_blocks() ) {
+			return $content;
+		}
+		return '<!-- wp:paragraph --><p>' . $content . '</p><!-- /wp:paragraph -->';
+	}
+
+	public static function set_extra_fields_from_mastodon_api( $user_id, $fields ) {
```
pfefferle, line 262: I think we should have all "enable mastodon apps" functions in the integration file and not mix them up with the "core" files!?!

mattwiebe: Yup sounds good to me

pfefferle (changes requested):

includes/class-admin.php
```diff
@@ -113,7 +113,11 @@ public static function admin_notices() {
 		if ( 'edit' === $current_screen->base && Extra_Fields::is_extra_fields_post_type( $current_screen->post_type ) ) {
 			?>
 			<div class="notice" style="margin: 0; background: none; border: none; box-shadow: none; padding: 15px 0 0 0; font-size: 14px;">
-				<?php esc_html_e( 'These are extra fields that are used for your ActivityPub profile. You can use your homepage, social profiles, pronouns, age, anything you want.', 'activitypub' ); ?>
+				<?php
+					esc_html_e( 'These are extra fields that are used for your ActivityPub profile. You can use your homepage, social profiles, pronouns, age, anything you want.', 'activitypub' );
+					echo '<br />';
+					esc_html_e( 'Note that Mastodon clients will only show four fields.', 'activitypub' );
```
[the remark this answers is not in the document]
mattwiebe: Nice find, I foolishly believed something I read in an API doc rather than testing things :D

PR #797, opened 2024-07-05
Update dependencies

pfefferle:
* Drop PHP 5.6 support

See compatibility matrix: https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/


pfefferle:
We could also drop PHP 7.0!?

mattwiebe (approved):
:shipit:

PR #801, opened 2024-07-09
Update class-create.php

janboddez:
Think I came across this before ... Either way, in the original code, `$reaction` has literally just been set to `null`, so running `is_wp_error( $reaction )` makes little sense. `$state`, on the other hand, _can be_ an integer, or false, or a `WP_Error`.

## Proposed changes:

* Check `$state` (either false or the result of `wp_new_comment()` isn't an error.

## Testing instructions:

* Mostly ... not applicable?


pfefferle (approved):
thanks @janboddez

Issue #235, opened 2022-12-28
Show likes from fediverse


jaschaurbach:
I have the webmention plugin installed and it shows the Likes to toots I made from my day-to-day account where I have a link to a blogpost inside. 

Likes made to the post of the blog are shown nowhere (which is a pitty). It would be great if the plugin could store likes received on the post the same way as the webmentions plugin which would display them as well as the likes.

bekopharm:
Same for shares :+1:

lapineige:
I would say it's every more interesting to know how many times (and by who) a content was shared :)

bekopharm:
!keep

gaviscapes:
Is there any progress on this? Very interested in this feature.

pfefferle:
Keep

janboddez:
On my own site, I just added like and "repost" ("Announce") support by means a couple custom `activitypub_inbox_{$type}"` callbacks.

Also added an `undo` callback, so that undoing the like/repost removes the "comment."

Since it's all implemented as action hooks, it would be very easy to add this to the ActivityPub plugin itself.

I do set some dummy content (in case I ever disable my site's "facepile" solution) and a couple comment meta fields allowing likes and reposts to be picked up by said "facepile" block. Maybe there's a better way (like "comment types," the way the Webmention plugin does it).

I think that's the main "challenge": deciding on a meta field "structure." It would be easy enough to update my block to extend compatibility to whichever solution ends up being implemented, so I don't really have a strong preference.

If there's any interest, I wouldn't mind opening a PR.

Also considering a (dynamic) Gutenberg block that would simply display a like and repost (filterable, preferably, so it can be made compatible with, e.g., webmentions) counter.

pfefferle:
I would love to see a PR for that!!!

@janboddez btw. thanks for all your work/support so far!

pfefferle:
Keep

maltfield:
@janboddez have you also seen #929?

 * https://github.com/Automattic/wordpress-activitypub/issues/929

If you're already implementing a PR for this for "Like", would you also consider implementing "DisLike"?

pfefferle:
@maltfield this is "only" for posts for now, it is not possible to up/down vote comments yet.

I would love to have a more generic approach that it will be possible to implement new activities with simple hooks and filters.

PR #804, opened 2024-07-10
Add like handler

janboddez:
Adds basic support for incoming likes (see #235).

Probably needs to go through a couple more iterations before it could be merged! For instance, the ActivityPub plugin should maybe also get a `pre_get_comments` filter, similar to the Webmention plugin, so that these "likes" don't show up among "regular comments." Should be "rather easy" to implement.

But it's a "first draft."

I originally wanted to also add "repost" support but noticed an announce handler already exists. I'm going to have to check that out later. Maybe in a different PR.

Edit: This won't actually _display_, e.g., a facepile. Though it currently is compatible with IndieBlocks' Facepile block. I can imagine you'd want to eventually add a block of your own ...

## Proposed changes:
* Adds a incoming like handler, and extends the existing undo handler to handle undoing of previous likes.

Note: I kept some of the actions from other reaction handlers and also set some comment meta to be able to later on display the like in a "facepile" kind of way. I also set some "dummy" content; feel free to change it in order to have it behave the same as, e.g., the Webmention plugin.

### Other information:

## Testing instructions:

* Upload patched version
* Search for a local URL from, e.g., your Mastodon instance
* Like the corresponding post
* Watch the "like" appear as a new comment with comment type "like"
* Undo like
* Watch the "like" comment move to trash


janboddez:
> [T]he ActivityPub plugin should maybe also get a pre_get_comments filter, similar to the Webmention plugin, so that these "likes" don't show up among "regular comments."

On my personal site I don't need this, as I have my own such filter and block. But I thought I'd mention it.

pfefferle:
@janboddez thanks a lot!

Do you think we need a UI/setting for that? It might be confusing if "Like" comments will simply show up in the comment section after the next update!?!

janboddez:
Yes, I do think you will need a `pre_get_comments` filter to hide likes from the front end. And only show them through a shortcode or block (similar to IndieBlocks' Facepile block, or the Followers block), which could be added in a later version.

I'm not sure about showing/hiding them in WP-Admin. It could be confusing to see "comments" in WP-Admin but not see them show up anywhere on the site. Hmm, maybe we need a completely separate WP-Admin (sub)menu for "likes"?

pfefferle:
@janboddez I refactored the code a bit and also added Announce support!

What do you think?

I am not very happy with the `add_reaction` method in the `Interactions` class, because I think this could be simplified a lot, but it should work for now!

Maybe we also hide it behind a feature flag, to run some tests and think about a nicer UI?

/cc @mattwiebe

janboddez:
Cool!

Feature flag (like a PHP constant?) seems like a good idea! I can imagine most people wanting to have this disabled or at least not show the likes/reposts until there's something like a Facepile block (or some other way of displaying likes/reposts, like a counter of sorts). Not everyone will want to install IndieBlocks either, haha!

Are the `register_comment_type()` calls compatible with, e.g., the Webmention plugin? Won't it complain if the same type is registered "twice"? (Just asking, I honestly have no clue.)

pfefferle:
> Feature flag (like a PHP constant?) seems like a good idea! I can imagine most people wanting to have this disabled or at least not show the likes/reposts until there's something like a Facepile block (or some other way of displaying likes/reposts, like a counter of sorts). Not everyone will want to install IndieBlocks either, haha!

Yes!

> Are the register_comment_type() calls compatible with, e.g., the Webmention plugin? Won't it complain if the same type is registered "twice"? (Just asking, I honestly have no clue.)

Kind of both! The ActivityPub version is stripped down a lot, so they are not completely identical, but they use the same vocabulary, so that there is no difference between a Webmention `Like` and an ActivityPub `Like` in the end.

pfefferle (approved):

janboddez:
Silly question: Do we also need some custom CSS to make likes look "good" in WP's dashboard? Like in https://github.com/pfefferle/wordpress-webmention/issues/396.

Didn't yet have a chance to test, but basically the question is: How do [likes and reposts] look (in the dashboard) with the Webmention plugin turned off?

pfefferle:
The Webmention Plugin does not change the UI of the dashboard!

janboddez:
Oh, I thought we added a [bit of CSS to make like and reposts avatars behave](https://github.com/pfefferle/wordpress-webmention/blob/main/assets/css/admin.css).

pfefferle:
You are totally right! I will add that!

Issue #768, opened 2024-05-31
Feature Request: Be able to add a bio when profile type is set to ‘enable blog’.


ethicalrevolution:
### What

Brilliant plugin. 

In Settings -> Profiles -> Enable profiles by type …When ‘Enable blog’ is selected so that the blog becomes an ActivityPub profile, the bio shown to followers on Fediverse platforms such as Mastodon is the website description in Wordpress.

It would be great if there was a way I could change the bio without changing the Wordpress website description.

e.g. I want my website description to still read, “An awesome website about [whatever]” for viewers on the website, search engines etc. …but I want the ActivityPub profile to display something different such as “Please be aware that messages sent from this account are automatically posted from [website]. Replies will be received as comments on the website, but any mentions and DMs that are not replies to posts made by this account will not be received.”

### Why

The bio I want to display for people following on Mastodon is not the same as the info I want to show as the website description in Wordpress. 

I don’t want people to interact with the profile on ActivityPub and think they’re being ignored, which is what happens if someone tags the account in a message that is not a direct reply to a message that the profile has sent from Wordpress.

pfefferle:
This is already possible in the latest version!

Issue #584, opened 2023-11-28
Profile Editor


mattwiebe:
The first thing most people want to do when setting up a social media profile is to edit/customize that profile. Currently the ActivityPub plugin draws its profile settings from various places in WP, and it neither helps make those connections, nor does it collect them into a single convenient place.

This project will create a consolidated UI for editing ActivityPub profiles in the plugin. It will probably also be a deluxe drop-in replacement for the Follow Me block, although that will need to be tested in practice vs creating a new Profile block.

To illustrate, here is a Mastodon profile view of my mattwiebe.blog profile

![image](https://github.com/Automattic/wordpress-activitypub/assets/195089/2ceaa9b4-754e-466e-b180-f37cbf3778f2)

It's not clear to a user where any of those settings are being drawn from. This makes sense from a WP integration perspective, but not from a user's perspective.

A few notes:

1. This can live in both Settings -> ActivityPub and also become a kind of supercharged version of the Follow Me block. The latter is currently read-only, but would become read/write and allow support for a Header image (this would effectively allow using the Follow Me block for your Site's header).
2. Header images have various problems:
    * Core header images are unsupported in block themes
    * users only use the whole-blog header image, no per-user settings
3. We should allow overriding the default `blogname` and `blogdescription` title/tagline stuff with specific ActivityPub Profile content.
4. Allow adding extra profile links

cogdog:
This would be extremely helpful! 

Not likely related, but the odd New York, New York username style of a WordPress.com Activity pub enabled site seems clunky, speaking for my test instance @cogdoglab.wordpress.com@cogdoglab.wordpress.com  why the dupe naming?

pfefferle:
this is because most of the platforms only show the informations before the middle "@" `@cogdoglab.wordpress.com` and we wanted to represent the blog, so the domain made a lot of sense for us.

pfefferle:
the plugin now supports image and description editing and because inline editing is no longer a priority, I will close this for now.

@mattwiebe please re-open if you think this is still relevant!

PR #816, opened 2024-07-18
Add "Header Image" support

pfefferle:
Add "Header Image" support based on WordPress Site-Icon JS

Implements: #505 #768

Related: #584

See Blog-Settings:

Screenshot 2024-07-18 at 11 25 01

See Author-Settings:

Screenshot 2024-07-19 at 15 21 24

### DIFF
--- a/assets/css/activitypub-admin.css
+++ b/assets/css/activitypub-admin.css
@@ -34,10 +34,10 @@
 
 .activitypub-settings-tabs-wrapper {
 	display: -ms-inline-grid;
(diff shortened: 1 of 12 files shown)


pfefferle:
@mattwiebe This would be a very easy version of a profile editor with all missing item. We can also re-use the fields for the `Enable Mastodon Apps` Profile Editing.

mattwiebe (approved):
This is a terrific addition! I tested as a user and it works well. Haven't reviewed the code much but nothing concerning when giving it a quick browse.

PR #817, opened 2024-07-19
Add link header

### LINKED ISSUE #626: Feature Request: `<link rel="alternate" type="application/activity+json">`
### What

Consider adding the link rel=alternate element as you can see in Mastodon/Misskey, e.g. in https://fosstodon.org/@krosylight:

```html
<link href='https://fosstodon.org/@krosylight.rss' rel='alternate' type='application/rss+xml'>
<link href='https://fosstodon.org/users/krosylight' rel='alternate' type='application/activity+json'>
```

### Why

This element advertises that the page can be read in ActivityPub form too. Browser extensions can detect it and reopen it on Mastodon for example.

### How

_No response_

pfefferle:
Fixes #626

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


includes/class-activitypub.php
```diff
@@ -131,12 +131,66 @@ public static function render_json_template( $template ) {
...
+			return;
+		}
+
+		// only add self link to author pages...
+		if ( is_author() ) {
+			if ( is_user_disabled( get_queried_object_id() ) ) {
+				return;
+			}
+		} elseif ( is_singular() ) { // or posts/pages/custom-post-types...
+			if ( ! \post_type_supports( \get_post_type(), 'activitypub' ) ) {
```
Menrath, line 155: Maybe this should be updated to also include whether a specific post is disabled for ActivityPub.

includes/class-activitypub.php
```diff
@@ -131,12 +131,66 @@ public static function render_json_template( $template ) {
...
+		);
+	}
+
 	/**
 	 * Custom redirects for ActivityPub requests.
 	 *
 	 * @return void
 	 */
 	public static function template_redirect() {
+		self::add_headers();
```
Menrath, line 192: I can not understand why these function is called within the 'template_redirect' hook/function.

Menrath (commented):
This came up while reviewing #1148

includes/class-activitypub.php
pfefferle: this will already be checked in the `post-json.php`... no need to do it twice.

includes/class-activitypub.php
pfefferle: It was a good combination between the wp_query is already processed an the headers are not yet sent.

do you see a better hook?

PR #819, opened 2024-07-23
Add "reply context" support

pfefferle:
This adds basic support for "reply to a fediverse post" using a block.

Thanks for pairing with me on that @mattwiebe !!

## Editor

Screenshot 2024-07-23 at 19 35 54

## Frontend

Screenshot 2024-07-23 at 19 38 40

## Fediverse

```JSON
{
    "@context": [
        "https:\/\/www.w3.org\/ns\/activitystreams",
        {
            "Hashtag": "as:Hashtag"
        }
    ],
    "id": "http:\/\/localhost:8076\/2024\/07\/hello\/",
    "type": "Article",
    "attachment": [],
    "attributedTo": "http:\/\/localhost:8076\/author\/test\/",
    "audience": "http:\/\/localhost:8076\/@blog",
    "content": "\u003Cp\u003ELorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.\u003C\/p\u003E",
    "name": "Hello",
    "inReplyTo": "https:\/\/mastodon.social\/@pfefferle\/112834543027990299",
    "published": "2024-07-23T16:12:22Z",
    "tag": [],
    "updated": "2024-07-23T17:35:44Z",
    "url": "http:\/\/localhost:8076\/2024\/07\/hello\/",
    "to": [
        "https:\/\/www.w3.org\/ns\/activitystreams#Public",
        "http:\/\/localhost:8076\/wp-json\/activitypub\/1.0\/actors\/1\/followers"
    ],
    "cc": []
}
```


mattwiebe:
I added some UX niceness:

1. If enter is pressed, a new empty block is inserted and focused
2. If the text input is empty, and delete or backspace are pressed, the block is deleted
3. When the `isUrl` check runs false, a more specific error is shown

and then I hadn't sent this comment, only to add:

A BOOKMARKLET ;)

pfefferle:
Just tested it and it works like a charm! This is awesome!

(I have some small naming feedbacks inline, but I am fine merging it as is)

build/reply-handle-new/block.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "reply-handler",
```
pfefferle, line 2: should we use `reply-intent` as (folder) name because this seems to be the commonly used name?

* https://developer.x.com/en/docs/twitter-for-websites/tweet-button/guides/web-intent
* https://developer.android.com/training/sharing/send

includes/class-blocks.php
```diff
@@ -13,6 +13,20 @@ public static function init() {
...
 		\add_action( 'wp_enqueue_scripts', array( self::class, 'add_data' ) );
 		\add_action( 'enqueue_block_editor_assets', array( self::class, 'add_data' ) );
+		\add_action( 'load-post-new.php', array( self::class, 'handle_in_reply_to_get_param' ) );
+	}
+
+	/**
+	 * Enqueue the reply handle script if the inReplyTo GET param is set.
+	 */
+	public static function handle_in_reply_to_get_param() {
+		if ( ! isset( $_GET['inReplyTo'] ) ) {
```
pfefferle, line 23: is it a good idea to have that in camel case?

mattwiebe: Happy to switch over, snake_case is more idiomatic in WP anyway :)

pfefferle:
I added a function to:

* get the actor of the `in-reply-to` post
* get the inbox of the `Actor` of that post
* add the inbox to the other inboxes before sending it to the audience

pfefferle:
@mattwiebe what do you think about `/wp-admin/tools.php`??

mattwiebe:
> @mattwiebe what do you think about `/wp-admin/tools.php`??

Looks nice! Maybe we can link over to it from the general setting page, I don't think people would necessarily go look there, but I like how it provides options and education.

mattwiebe (approved):
🚢 !

Issue #828, opened 2024-07-30
Feature Request: Allow line brackes for user biography and add links to hashtags, user accounts and webseites


danielhuesken:
### What

I would like to have support for line brakes in user biography. Also it will be nice if html for links to urls, hastags and users will automaticly added.

### Why

In Mastodon it is supported and helps me to have well displayed biography.

PR #831, opened 2024-07-30
Replace hashtags, urls and user for descriptions

danielhuesken:
See #828 URLs, Users and Hashtags are replaced with links for the descriptions.

Added a fallback when no tag is found to use a category

Feel free to change and add to a own function.


pfefferle:
There seem to be a glitch if you have a line break after an URL.

Example:

```
lore ipsum! #wordpress #activitypub

https://notiz.blog

@pfefferle@mastodon.social
```

HTML Output:

```HTML
<p>lore ipsum! <a rel="tag" class="hashtag u-tag u-category" href="http://localhost:8076/tag/wordpress/">#wordpress</a> #activitypub</p>\n<p><a href="https://notiz.blog/</p" target="_blank" rel="nofollow noopener noreferrer" translate="no"><span class="invisible">https://</span>notiz.blog</p<span class="invisible"></span></a>>\n<p><a rel="mention" class="u-url mention" href="https://mastodon.social/@pfefferle">@<span>pfefferle</span></a></p>\n
```

pfefferle:
@danielhuesken there seem to be still a problem. If you add hashtags or @-replies, the code tries to make these already links once again.

activitypub.php
```diff
@@ -32,6 +32,7 @@
 \defined( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS' ) || \define( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS', 3 );
 \defined( 'ACTIVITYPUB_HASHTAGS_REGEXP' ) || \define( 'ACTIVITYPUB_HASHTAGS_REGEXP', '(?:(?<=\s)|(?<=<p>)|(?<=<br>)|^)#([A-Za-z0-9_]+)(?:(?=\s|[[:punct:]]|$))' );
 \defined( 'ACTIVITYPUB_USERNAME_REGEXP' ) || \define( 'ACTIVITYPUB_USERNAME_REGEXP', '(?:([A-Za-z0-9\._-]+)@((?:[A-Za-z0-9_-]+\.)+[A-Za-z]+))' );
+\defined( 'ACTIVITYPUB_URLS_REGEXP' ) || \define( 'ACTIVITYPUB_URLS_REGEXP', '(http|https):\/\/([-a-zA-Z0-9@:;\/%_;\+.~#?&=]*)' );
```
pfefferle, line 35:
```suggestion
\defined( 'ACTIVITYPUB_URLS_REGEXP' ) || \define( 'ACTIVITYPUB_URLS_REGEXP', '(www|http:|https:)+[^\s]+[\w]' );
```

@danielhuesken what do you think of this REGEX instead? This would allow umlauts in GET params and the domain!?!

pfefferle:
@danielhuesken on the post, we add the Hashtags as Tags in the JSON: https://github.com/Automattic/wordpress-activitypub/blob/80ec65ad0a4d5228a1d3040ddc5b13443a8f4486/includes/transformer/class-post.php#L683

Maybe we can also do that for the users? This way, the user will be indexed for these tags on Mastodon et al.

activitypub.php
danielhuesken: @pfefferle looks good for me because of not change urls in html tags it works now.

includes/class-activitypub.php
```diff
@@ -585,10 +587,8 @@ public static function default_actor_extra_fields( $extra_fields, $user_id ) {
 				'post_status'    => 'publish',
 				'post_author'    => $user_id,
 				'post_content'   => sprintf(
-					'<!-- wp:paragraph --><p><a rel="me" title="%s" target="_blank" href="%s">%s</a></p><!-- /wp:paragraph -->',
-					\esc_attr( $url ),
-					$url,
-					\wp_parse_url( $url, \PHP_URL_HOST )
+					'<!-- wp:paragraph --><p>%s</p><!-- /wp:paragraph -->',
+					Urls::the_content( $url )
```
pfefferle, line 591: Is it even possible to post a text-url that will not be transformed to a link from the block editor?

includes/model/class-user.php
```diff
@@ -281,9 +281,9 @@ public function get_attachment() {
 					$tags->next_tag();
 				}
 
-				if ( 'A' === $tags->get_tag() ) {
+				if ( 'A' === $tags->get_tag() && $tags->get_attribute( 'href' ) ) {
 					$tags->set_bookmark( 'link' );
-					if ( ! $tags->next_tag() ) {
+					if ( ! $tags->next_tag() || 'SPAN' === $tags->get_tag() ) {
```
pfefferle, line 286: why the `SPAN` check here?

danielhuesken: When youn not check for `SPAN` it will ignore tags like `<a href="test.de><span>Tiltle</span></a>"`. So it will also work when you generate links like mastodon.

includes/class-activitypub.php
danielhuesken: In that case the URLS comes from the author or blog. They should be convertred without problems.

includes/model/class-user.php
pfefferle: Is `$tags->next_tag()` the next tag on the same level or one level down the DOM?

I thought it is on the same level as the Link!

danielhuesken: On my tests `$tags->next_tag()` is simply the next tag after the opening 'a' tag. 

When you use  `!$tags->next_tag('A')` you can be shure that there are not more `a` tags in the DOM. 
Depends on what the idea was behind :
Checking ther is only one `a` tag and getting the link from it, or also checking the the Tag has only text.

includes/class-activitypub.php
danielhuesken: Sidenot  the `<p>` tags in Mastodon are inside of a `<span>`.

includes/model/class-user.php
pfefferle: OK, then my initial code is wrong! My idea was to check if there is only one link in the text!

danielhuesken: OK, i can fix that in the PR. So that it is like that.

includes/class-activitypub.php
pfefferle: and `<p>` inside of `<span>` is not valid in terms of HTML5

includes/model/class-user.php
pfefferle: thanks @danielhuesken !

includes/class-activitypub.php
danielhuesken: Exactly

danielhuesken:
@pfefferle Looks like I did somthing worng on merging the last changes. I can redo my changes on a new branch/PR when it is helpfull.

danielhuesken:
frech PR is #849

PR #832, opened 2024-07-31
Fix undefined `get_sample_permalink`

mattwiebe:
Even after https://github.com/Automattic/wordpress-activitypub/commit/12cb73bcd678f3d2bde3c92f46a7dcb500e6e4f7 we are seeing fatal errors when saving drafts. 

This is also exposing an unintended problem with sending `Update` Activities for new drafts introduced in #780. I think we should only be sending those for previously-published posts?

## Proposed changes:
* Include `wp-admin/includes/post.php` when needed

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -99,13 +99,21 @@ class Post extends Base {
 	public function get_url() {
 		$post = $this->wp_object;
 
-		if ( 'trash' === \get_post_status( $post ) ) {
-			$permalink = \get_post_meta( $post->ID, 'activitypub_canonical_url', true );
-		} elseif ( 'draft' === \get_post_status( $post ) && \get_sample_permalink( $post->ID ) ) {
-			$sample    = \get_sample_permalink( $post->ID );
-			$permalink = \str_replace( array( '%pagename%', '%postname%' ), $sample[1], $sample[0] );
-		} else {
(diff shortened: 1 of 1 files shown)


mattwiebe:
> This is also exposing an unintended problem with sending `Update` Activities for new drafts introduced in #780. I think we should only be sending those for previously-published posts?

Addressed in #833

includes/transformer/class-post.php
```diff
@@ -99,13 +99,21 @@ public function get_id() {
...
 		$post = $this->wp_object;
 
-		if ( 'trash' === \get_post_status( $post ) ) {
-			$permalink = \get_post_meta( $post->ID, 'activitypub_canonical_url', true );
-		} elseif ( 'draft' === \get_post_status( $post ) && \get_sample_permalink( $post->ID ) ) {
-			$sample    = \get_sample_permalink( $post->ID );
-			$permalink = \str_replace( array( '%pagename%', '%postname%' ), $sample[1], $sample[0] );
-		} else {
-			$permalink = \get_permalink( $post );
+		switch ( \get_post_status( $post ) ) {
```
mattwiebe, line 102: Any time I see an `elseif` I immediately want to change it to a `switch` for clarity

pfefferle (approved):

PR #833, opened 2024-07-31
Only send `Update` for previously-published posts

mattwiebe:
In #780 we started sending `Update` Activities for drafts. But we should only be sending those if the post was previously published. The previous logic would always send for all draft status updates, which is probably spamming a bunch of instances with an update for a post that never existed.

## Proposed changes:
* Fix `Update` logic to only send for posts transitioning from `publish` to `draft`

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-scheduler.php
+++ b/includes/class-scheduler.php
@@ -142,8 +142,9 @@ class Scheduler {
 			$type = 'Create';
 		} elseif (
 			'publish' === $new_status ||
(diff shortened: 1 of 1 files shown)


pfefferle (approved):

pfefferle:
I am not sure why I haven't restricted it to "published" 🤔

PR #838, opened 2024-08-01
Add Blog extra fields

mattwiebe:
Following the pattern of User extra fields, we bring you: Blog extra fields!

Since we can't assign `post_author=0` directly in the post editor, we instead register a separate `ap_extrafield_blog` `post_type` for the Blog User.

I moved a bunch of stuff into an `Extra_Fields` class. This will come in handy for #788 😁 

Screenshot 2024-08-01 at 17 44 01

### Other information:

## Testing instructions:
1. Go to Settings -> ActivityPub -> Blog-Profile. 
2. The Extra Fields functionality should be there as shown above.
3. `curl -H 'Accept: application/activity+json' http://your.blog/@your.blog | jq .attachment` should reflect the state of the Blog Extra Fields. In the case of the screenshot above:

```json
[
  {
    "type": "PropertyValue",
    "name": "Blog",
    "value": "<p><a rel=\"me noopener\" title=\"http://wp.test/\" target=\"_blank\" href=\"http://wp.test/\">wp.test</a></p>"
  },
  {
    "type": "Link",
    "name": "Blog",
    "href": "http://wp.test/",
    "rel": [
      "me",
      "noopener"
    ]
  }
]
```

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -81,7 +81,6 @@ function plugin_init() {
 	\add_action( 'init', array( __NAMESPACE__ . '\Health_Check', 'init' ) );
 	\add_action( 'init', array( __NAMESPACE__ . '\Scheduler', 'init' ) );
 	\add_action( 'init', array( __NAMESPACE__ . '\Comment', 'init' ) );
-	\add_action( 'init', array( __NAMESPACE__ . '\Collection\Extra_Fields', 'init' ) );
 
 	if ( site_supports_blocks() ) {
 		\add_action( 'init', array( __NAMESPACE__ . '\Blocks', 'init' ) );
--- a/activitypub.php
+++ b/activitypub.php
@@ -81,6 +81,7 @@ function plugin_init() {
 	\add_action( 'init', array( __NAMESPACE__ . '\Health_Check', 'init' ) );
 	\add_action( 'init', array( __NAMESPACE__ . '\Scheduler', 'init' ) );
 	\add_action( 'init', array( __NAMESPACE__ . '\Comment', 'init' ) );
+	\add_action( 'init', array( __NAMESPACE__ . '\Collection\Extra_Fields', 'init' ) );
 
 	if ( site_supports_blocks() ) {
 		\add_action( 'init', array( __NAMESPACE__ . '\Blocks', 'init' ) );
--- a/includes/collection/class-extra-fields.php
+++ b/includes/collection/class-extra-fields.php
@@ -44,7 +44,7 @@ class Extra_Fields {
 			'show_in_rest'        => true,
 			'map_meta_cap'        => true,
 			'show_ui'             => true,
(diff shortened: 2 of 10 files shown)


mattwiebe:
Hmm must have borked something with those test failures but I'm EOD, will dig in tomorrow

pfefferle:
There seem to be something broken with the Unit-Tests

includes/model/class-user.php
```diff
@@ -246,73 +247,7 @@ public function get_endpoints() {
 	 */
 	public function get_attachment() {
 		$extra_fields = get_actor_extra_fields( $this->_id );
```
pfefferle, line 249: should we move the `get_actor_extra_fields()` global function to the collection class, to simplify things? I think after your re-write the global function makes no more sense?!?

includes/collection/class-extra-fields.php
```diff
@@ -0,0 +1,233 @@
...
+				);
+			}
+
+			$attachments[] = $attachment;
+		}
+
+		return $attachments;
+	}
+
+	public static function is_post_type( $post_type ) {
```
pfefferle, line 149:
```suggestion
	public static function is_valid_post_type( $post_type ) {
```

Should we make it more explicit?

pfefferle:
I like your code improvements a lot!

includes/collection/class-extra-fields.php
mattwiebe: `valid` doesn't quite sit right for me. Even though it's verbose, `is_extra_fields_post_type` may be the correct answer?

includes/model/class-user.php
mattwiebe: Yeah that's probably a good idea. Wanted to keep the helper function in place initially but it's not doing much for us any more.

activitypub.php
```diff
@@ -81,6 +81,7 @@ function plugin_init() {
 	\add_action( 'init', array( __NAMESPACE__ . '\Health_Check', 'init' ) );
 	\add_action( 'init', array( __NAMESPACE__ . '\Scheduler', 'init' ) );
 	\add_action( 'init', array( __NAMESPACE__ . '\Comment', 'init' ) );
+	\add_action( 'init', array( __NAMESPACE__ . '\Collection\Extra_Fields', 'init' ) );
```
pfefferle, line 84: I would love to avoid hooks in the collections. It's true that there is no real consistent way how classes with hooks are used, maybe we should discuss that on monday.

mattwiebe: Yeah great topic, let's dig into it.

mattwiebe:
Happy to address refactoring some of this in a followup PR, but it's good to go now I think.

pfefferle (approved):
I fixed an issue with a missing class and moved stuff around a bit, to not have an init function in the collection any more (I hope this is fine for you).

Looks like we can merge that now!

PR #845, opened 2024-08-09
Add DDEV developmet env

danielhuesken:
Devlopers can use DDEV for developmet


pfefferle:
🤔 I have no idea what to do with that.

How can I verify that this works?

danielhuesken:
You should install https://ddev.com/.
when you have it installed you start the env with: `ddev start`
First inssalation WordPress and plugins with: `ddev orchestrate` reinstall with `ddev orchestrate -f`
Run tests with `ddev runtests`
Insttalation is acssesbile under https://activitypub.ddev.site (only local)

There is also a plugin for PHPStorm for DDEV.

pfefferle:
What do you think @Automattic/fediverse ???

obenland:
@danielhuesken Is this something that can live locally?

danielhuesken:
@obenland for me it is fine to have it only locally.

PR #847, opened 2024-08-09
Improve Validation

pfefferle:
This PR is about a leaner validation method so that not every handler class has to be validated again and again.

## Proposed changes:

* Disable `ACTIVITYPUB_DISABLE_INCOMING_INTERACTIONS` on Inbox level
* Improve `args` validation, to validate the Activity-Object as early as possible
* ...

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/handler/class-create.php
+++ b/includes/handler/class-create.php
@@ -21,6 +21,13 @@ class Create {
 			10,
 			3
 		);
+
+		\add_filter(
+			'activitypub_validate_object',
+			array( self::class, 'validate_object' ),
+			10,
+			3
+		);
 	}
 
 	/**
@@ -33,10 +40,6 @@ class Create {
 	 * @return void
 	 */
 	public static function handle_create( $array, $user_id, $object = null ) {
-		if ( ! isset( $array['object']['id'] ) ) {
-			return;
-		}
-
 		// check if Activity is public or not
 		if ( ! is_activity_public( $array ) ) {
 			// @todo maybe send email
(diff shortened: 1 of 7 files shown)


mattwiebe (approved):
I don't understand enough about validation challenges, but this approach and your reasoning seems sound.

includes/handler/class-create.php
```diff
@@ -21,6 +21,13 @@ public static function init() {
 			10,
 			3
 		);
+
+		\add_filter(
```
Menrath, line 25: @pfefferle Am I getting something wrong or is this filter added every time? That would also mean, that if one adds a custom handler for a `Create` activity one must also add a custom validation function that completely overrides (is applied later, higher priority) the one in this class (if for example one does not need `inReplyTo` set).

pfefferle: good point!!!

Menrath: @akirk Does this not prevent your friends plugin from receiving any creates to your create handler?

akirk: Maybe I am reading this wrong but I am listening on `activitypub_inbox_create` which should still be called normally, no?

https://github.com/akirk/friends/blob/199af50ae54ee82e8cc77b2311a7149fb1eea8b2/feed-parsers/class-feed-parser-activitypub.php#L45

Menrath: @akirk This is what my original question was about. I am not totally sure. I have been studying these lines of your code recently, before starting implementing my event-import feature where [I also added an action for `activitypub_inbox_create`](https://code.event-federation.eu/Event-Federation/wordpress-event-bridge-for-activitypub/src/commit/16762b2b3140759e963faab10db36e47b8023f7c/includes/activitypub/handler/class-create.php#L22). But within my setup the "aggressive" filter here makes incoming `Create` activities not even reach the handler in `activitypub_inbox_create` when for example the `object` within the `Create` does not have `inReplyTo` set. So I was wondering, if I am missing something else, or something is special in my setup etc.

PR #849, opened 2024-08-10
Replace hashtags, urls and mentions in summary with links

danielhuesken:
See #828 URLs, Ment ions and Hashtags are replaced with links for the summarys.

### Other information:

Replaces the messed up PR #831

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # ActivityPub #
-**Contributors:** [automattic](https://profiles.wordpress.org/automattic/), [pfefferle](https://profiles.wordpress.org/pfefferle/), [mediaformat](https://profiles.wordpress.org/mediaformat/), [mattwiebe](https://profiles.wordpress.org/mattwiebe/), [akirk](https://profiles.wordpress.org/akirk/), [jeherve](https://profiles.wordpress.org/jeherve/), [nuriapena](https://profiles.wordpress.org/nuriapena/), [cavalierlife](https://profiles.wordpress.org/cavalierlife/)  
-**Tags:** OStatus, fediverse, activitypub, activitystream  
-**Requires at least:** 5.5  
-**Tested up to:** 6.6  
-**Stable tag:** 3.1.0  
-**Requires PHP:** 7.0  
-**License:** MIT  
-**License URI:** http://opensource.org/licenses/MIT  
+**Contributors:** [automattic](https://profiles.wordpress.org/automattic/), [pfefferle](https://profiles.wordpress.org/pfefferle/), [mediaformat](https://profiles.wordpress.org/mediaformat/), [mattwiebe](https://profiles.wordpress.org/mattwiebe/), [akirk](https://profiles.wordpress.org/akirk/), [jeherve](https://profiles.wordpress.org/jeherve/), [nuriapena](https://profiles.wordpress.org/nuriapena/), [cavalierlife](https://profiles.wordpress.org/cavalierlife/)
+**Tags:** OStatus, fediverse, activitypub, activitystream
+**Requires at least:** 5.5
+**Tested up to:** 6.6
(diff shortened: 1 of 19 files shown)


pfefferle:
@danielhuesken Thanks a lot, this change is awesome!

There is one last thing about the Extra-Fields generator. I think this is nothing that came with your change but maybe we could/should fix it with this PR.

If you have a text like:

```
hallo https://example.com hallo
```

it translates to:

```
{
  "type": "PropertyValue",
  "name": "Test",
  "value": "\u003Cp\u003Ehallo \u003Ca href=\u0022https:\/\/pfefferle.org\u0022 rel=\u0022nofollow\u0022\u003Ehttps:\/\/example.com\u003C\/a\u003E hallo\u003C\/p\u003E"
},
{
  "type": "Link",
  "name": "Test",
  "href": "https:\/\/example.com",
  "rel": [
    "nofollow"
  ]
}
```

The property value is fine, but shouldn't the second one be a `Note` instead of a `Link`, because it is not only a link but also some text?

includes/collection/class-extra-fields.php
```diff
@@ -38,10 +39,11 @@ public static function get_actor_fields( $user_id ) {
 
 	public static function fields_to_attachments( $fields ) {
 		$attachments = array();
+		\add_filter( 'activitypub_urls_rel_me', '__return_true' );
```
pfefferle, line 42: can we assume that a link in the extra fields is always a `me` link?

pfefferle: Mastodon is doing it like that, so maybe it is the way to go! never mind, forget what I said :)

danielhuesken:
@pfefferle I currently thinking about striping tags form the extra fields HTML. See https://docs.joinmastodon.org/spec/activitypub/#sanitization

Maybe we can also strip p tags. And maybe it is easier to detect that only one link ist there. 
What did you think?

pfefferle:
I am not sure about stripping HTML... I tried to not limit the plugin to only match the Mastodon spec, that's why we also removed stripping/sanitization from the post-content. Friendi.ca and others are very good in supporting HTML!

And do you have an idea how to check if there is only one Link in the content?

danielhuesken:
@pfefferle I implemented a check. Maybe will not help in all cases but will improve it for your case.

pfefferle:
@danielhuesken this is awesome! thanks so much!

pfefferle (approved):

pfefferle:
@danielhuesken do you think it is ready to merge?

danielhuesken:
Yes, I have nothing to add.

README.md
```diff
@@ -115,6 +115,7 @@ The plugin uses PHP Constants to enable, disable or change its default behaviour
 * `ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS` - Change the number of attachments, that should be federated. Default: `3`.
 * `ACTIVITYPUB_HASHTAGS_REGEXP` - Change the default regex to detect hashtext in a text. Default: `(?:(?<=\s)|(?<=<p>)|(?<=<br>)|^)#([A-Za-z0-9_]+)(?:(?=\s|[[:punct:]]|$))`.
 * `ACTIVITYPUB_USERNAME_REGEXP` - Change the default regex to detect @-replies in a text. Default: `(?:([A-Za-z0-9\._-]+)@((?:[A-Za-z0-9_-]+\.)+[A-Za-z]+))`.
+* `ACTIVITYPUB_URL_REGEXP` - Change the default regex to detect urls in a text. Default: `(http|https):\/\/([-a-zA-Z0-9@:;\/%_;\+.~#?&=]*)`.
```
danielhuesken, line 118: @pfefferle can you change that. The regex in the readme.md is not the same as defined.

pfefferle: good point! will do!

PR #851, opened 2024-08-12
Hide Bookmarklet if site does not support Blocks

pfefferle:
Maybe fixes https://wordpress.org/support/topic/bookmarklet-seems-to-do-nothing/

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -10,6 +10,7 @@ use Activitypub\Collection\Extra_Fields;
 use function Activitypub\is_comment;
 use function Activitypub\sanitize_url;
 use function Activitypub\is_local_comment;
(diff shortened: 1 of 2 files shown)


mattwiebe (changes requested):
Let's hide where it appears in tools.php as well.

Should we leave a message stating what they're missing out on by not having the block editor! 😇 😈

pfefferle:
The PR also hides it on the tools page already!

Instead of having a message only for the bookmarklet, maybe we have a general feature, that the functionality is only limited without the Block-Editor?

mattwiebe (approved):
Ohhhh I was testing this completely wrongly before. Good change!

PR #854, opened 2024-08-13
Header Images: fix cropping for users without the 'customize' capability

### LINKED ISSUE #846: Can't add Header Image to authors - cropping fails
### Quick summary

When trying to add an image as "Header Image" for an AP author, cropping fails "_There has been an error cropping your image._" and no image is added, even if the image already has the proper ratio.

### Steps to reproduce

Logged to WordPress as the Author/Editor, declared as the AP author for my blog (**not the blog-profile then**).

From the User Profile, I can choose an existing image from Media or upload a new one.
After selecting the desired image, it forces to crop it, probably to respect Mastodon's ratio (1500*500px), even if image already has the proper ratio (appears fully selected, but have to validate the selection).

When validating the cropping selection, it fails with an error message "_There has been an error cropping your image._".
Operation has to be aborted, no header image added.

PS: new uploads appear in Media.

Behavior reproduced on 2 different blogs.

1 - cropping works if needed
2 - is it possible to validate images that already have the proper ratio, without the need to recreate a cropped image?
3 - add information around the "Header Image" area in profile, to specify the max image size.

### What actually happened

No Header Image, process interrupted.

PS : I suppose (hope) roles such as Author or Editor have the WordPress right to crop image...

### Impact

Some (< 50%)

### Available workarounds?

### Logs or notes

WordPress and plugins up-to-date to latest versions.

mattwiebe:
I think I found a Core bug here. There's absolutely no reason for the ajax request to include `wp_customize=on`, the logic is entirely handled in the `wp_ajax_crop_image()` handler.

[This commit](https://github.com/WordPress/wordpress-develop/commit/9fe59ced578169572f00d14a065aed22da165cab) appears to have moved the logic from the Customizer into a more general purpose context while [leaving the constraining `wp_customize=on` parameter](https://github.com/WordPress/wordpress-develop/commit/9fe59ced578169572f00d14a065aed22da165cab#diff-ce553f4ec6128c3384da17e519b3b0390203c996681ef1413ecf4b600fac5673R25).

I guess this could be a core patch

Fixes #846

## Proposed changes:

Make a custom controller but without `wp_customize=on`

### Other information:

## Testing instructions:

1. Be logged in as an author-level user (or any other user who can create posts but not edit others' posts)
2. Ensure that this user is ActivityPub-enabled
3. Go to profile and attempt to upload and crop an image. Any image uploaded or chosen from the library should work (the library will now only show the author user's own posts)

### DIFF
--- a/assets/js/activitypub-header-image.js
+++ b/assets/js/activitypub-header-image.js
@@ -39,7 +39,6 @@
 				cropDetails.dst_height = control.params.flex_height ? control.params.width  / ratio : control.params.height;
 			}
 
-
 			return wp.ajax.post( 'crop-image', {
 				// where wp_customize: 'on' would be in Core, for no good reason I understand.
 				nonce: attachment.get( 'nonces' ).edit,
@@ -109,6 +108,11 @@
 	 */
 	$chooseButton.on( 'click', function () {
 		var $el = $( this );
+		var userId = $el.data( 'userId' );
+		var mediaQuery = { type: 'image' };
+		if ( userId ) {
+			mediaQuery.author = userId;
+		}
 
 		// Create the media frame.
 		frame = wp.media( {
@@ -122,7 +126,7 @@
 			states: [
 				new wp.media.controller.Library( {
 					title: $el.data( 'choose-text' ),
-					library: wp.media.query( { type: 'image' } ),
+					library: wp.media.query( mediaQuery ),
 					date: false,
 					suggestedWidth: $el.data( 'size' ),
 					suggestedHeight: $el.data( 'size' ),
--- a/assets/js/activitypub-header-image.js
+++ b/assets/js/activitypub-header-image.js
@@ -14,7 +14,43 @@
 		$headerImagePreview = $( '#activitypub-header-image-preview' ),
 		$hiddenDataField = $( '#activitypub_header_image' ),
 		$removeButton = $( '#activitypub-remove-header-image' ),
-		frame;
+		frame,
+		ImageCropperNoCustomizer;
+
+	/**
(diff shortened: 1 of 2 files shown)


assets/js/activitypub-header-image.js
```diff
@@ -14,7 +14,43 @@
...
+			// Constrain flexible side based on image ratio and size of the fixed side.
+			} else {
+				cropDetails.dst_width  = control.params.flex_width  ? control.params.height * ratio : control.params.width;
+				cropDetails.dst_height = control.params.flex_height ? control.params.width  / ratio : control.params.height;
+			}
+
+
+			return wp.ajax.post( 'crop-image', {
+				// where wp_customize: 'on' would be in Core, for no good reason I understand.
+				nonce: attachment.get( 'nonces' ).edit,
```
pfefferle, line 45: this seem to still return `false` as nonce in my cases.

mattwiebe: Dang, this should be fun to debug, since it now works perfectly for me with an `author`-level user. Hmmmmmm…

mattwiebe:
I fixed the remaining problem in https://github.com/Automattic/wordpress-activitypub/pull/854/commits/3000bf52cb0e17b74edcf12310f51aec2bc89f08 - you can't crop an image that you didn't upload when you can't edit others' posts, [failing this check that supplies the nonces](https://github.com/WordPress/wordpress-develop/blob/ce571018a9d72fb513b803599cba21037fec68ed/src/wp-includes/media.php#L4479-L4483).

This is definitely ready for testing again @pfefferle

pfefferle (approved):
works like a charm!

Issue #848, opened 2024-08-10
Feature Request: single-post based template metabox


talksina:
### What

Give the possibility to change Fediverse output for every single post, and enrich hashtag/taxonomy choice.

### Current behaviour
Activitypub -> settings -> template. You add ap_title ap_excerpt ap_content ap_hashtags ... parameters and this model works for _every_ post you publish. But what if you want to customize single posts with custom content?
Hashtags: the ap_hashtags parameter picks up the "post_tag" taxonomy and there is no way to choose another one if available. 
### Requested behaviour:
Settings -> activitypub -> template has not to change. That is the standard model.
But, to the post editing window, add one "ActivityPub" dialog box (meta box?) where you can customize the parameters: 
- hashtags: choose preferred terms taxonomy. if current post does not have tags but "topics", let me choose the "topics" taxonomy to map as hashtags so the ap_hashtags code picks "topics" up.
- message: it is pre-compiled with the settings given in the main activitypub settings page, but you could even add a custom text with custom hashtags if you want to, if you want to give a more engaging text in the activitypub message rather than it is in the excerpt. 
- For example JetPack has it, its metabox gives you a message field where you type what social networks will then see. Including MENTIONS. 

### Why

Putting hashtags and @-mentions in the excerpt (WordPress default excerpt) is not friendly for search engines and even for screen reader usage, especially if you are reading from a third-party app or from the home page. And as I tested, it does not work, using ap_excerpt activitypub ignores the tags and mentions I place there.

It is not necessary to tag everything for ActivityPub's hashtag reasons. 
For example, in a post regarding HIV-related topics I could place the hashtag "#WorldAidsDay" or "#ZeroHivStigmaDay" related to a specific yearly event. It makes no sense to have a WordPress tag, or whatever taxonomy element, for something that goes into a WordPress post published once a year.
So, what I'd like to be able to do is decide for each post to customize what will be posted into federated platforms.
Hope it is clear, otherwise @pfefferle knows how and where to contact me.

pfefferle:
Thanks for the request @talksina 👋

This feels very much like we did it over the last years with cross-posting. We had to do it, because it was not possible to add real tags/images/@-replies/..., so that we had to cheat a bit with templates.

But with ActivityPub it is possible to re-use everything that WordPress-Core supports, so why should we reply on the templates instead of simply add tags (like they are meant to be), @-replies and images and the plugin tries to map it best possible to ActivityPub?!?

[...]

Issue #556, opened 2023-11-09
Feature Request: Support for "more" delimiter


MatzeKitt:
### What

I just came across a thing I don’t like for publishing the content of my posts. Would it be possible to add support for the “more” delimiter/block, that means that just the content above this delimiter is being posted in the Fediverse?

I tested `[ap_content]` as well as `[ap_excerpt length="400"]` and `[ap_excerpt]`.

### Why

I would find it more natural if it has the same functionality as WordPress itself so that you post only the content until the “more” delimiter.

And it’s a default WordPress functionality.

mdingemanse:
What may be helpful to know (which was not clear to me from the documentation) is that ActivityPub's `[ap_excerpt]` does support `the_excerpt()` if you supply it, i.e. the separate excerpt you can provide for each post in WordPress.

tomfinitely:
These seems like reasonable requests from a functionality standpoint, @MatzeKitt. Weighing in because I don't want to see this get closed.

MatzeKitt:
I still have it on my list to create a pull request for it, but hadn’t yet the time to do so. 🥲

pfefferle:
Is implemented now!

PR #856, opened 2024-08-14
Support "read more" for Activity-Summary

pfefferle:
See #848 also #556

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1149,7 +1149,7 @@ function generate_post_summary( $post, $length = 500 ) {
 	$content       = \sanitize_post_field( 'post_content', $post->post_content, $post->ID );
 	$content_parts = \get_extended( $content );
 
-	$excerpt_more = \apply_filters( 'activitypub_excerpt_more', '[&hellip;]' );
+	$excerpt_more = \apply_filters( 'activitypub_excerpt_more', '[…]' );
 	$length       = $length - strlen( $excerpt_more );
 
 	// Check for the <!--more--> tag.
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1116,3 +1116,46 @@ function enrich_content_data( $content, $regex, $regex_callback ) {
 
 	return $content_with_links;
 }
+
+function generate_post_summary( $post, $length = 500 ) {
+	$post = get_post( $post );
+
+	if ( ! $post ) {
+		return '';
+	}
+
+	$content = \get_post_field( 'post_excerpt', $post->ID );
+
+	if ( $content ) {
+		return $content;
+	}
+
+	$content       = \get_post_field( 'post_content', $post->ID );
+	$content_parts = \get_extended( $content );
+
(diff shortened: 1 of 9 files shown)


includes/functions.php
```diff
@@ -1116,3 +1116,47 @@ function enrich_content_data( $content, $regex, $regex_callback ) {
 
 	return $content_with_links;
 }
+
+function generate_post_summary( $post, $length = 500 ) {
```
MatzeKitt, line 1120: Missing method description

includes/functions.php
```diff
@@ -1116,3 +1116,47 @@ function enrich_content_data( $content, $regex, $regex_callback ) {
...
+
+	if ( $content ) {
+		return $content;
+	}
+
+	$content       = \sanitize_post_field( 'post_content', $post->post_content, $post->ID );
+	$content_parts = \get_extended( $content );
+
+	$excerpt_more = \apply_filters( 'activitypub_excerpt_more', '[&hellip;]' );
+	$length       = $length - strlen( $excerpt_more );
```
MatzeKitt, line 1137: Wouldn’t this be 500 - 10, since it counts characters, whereas `[&hellip;]` in the end only represents 3 characters ([…])?

includes/functions.php
```diff
@@ -1116,3 +1116,47 @@ function enrich_content_data( $content, $regex, $regex_callback ) {
...
+
+	// Check for the <!--more--> tag.
+	if (
+		$content_parts &&
```
MatzeKitt, line 1141: This is redundant as `get_extended()` always returns an array.

includes/transformer/class-post.php
```diff
@@ -850,6 +834,9 @@ protected function get_post_content_template() {
 				break;
 			default:
 				$template = \get_option( 'activitypub_custom_post_content', ACTIVITYPUB_CUSTOM_POST_CONTENT );
+				if ( ! $template ) {
+					$template = ACTIVITYPUB_CUSTOM_POST_CONTENT;
+				}
```
MatzeKitt, line 839: Alternative oneliner:
```php
$template = \get_option( 'activitypub_custom_post_content', ACTIVITYPUB_CUSTOM_POST_CONTENT ) ?: ACTIVITYPUB_CUSTOM_POST_CONTENT;
```

includes/functions.php
pfefferle: 🤔

so should we use `[...]` instead?

MatzeKitt: Should be working fine. Entities should only be used if the correct sign cannot. So, as long as you use the correct hellip sign (it’s a single one: `…`), this should be fine and the correct count for `strlen`

MatzeKitt (commented):
Looks good to me. 🙂

MatzeKitt (approved):

PR #857, opened 2024-08-14
fix link detection for extra fields when spaces after the link and fi…

danielhuesken:
I found 2 little problems after the code changes in PR #849 
When after an a tags some spaces they will not be detected for Link
When 2 a tags in the content it will get the first as Link and not use the note.

This PR fixes it.


pfefferle (approved):

mattwiebe:
Thanks @danielhuesken !

PR #860, opened 2024-08-15
Add support for The Events Calendar

danielhuesken:
This PR adds a transformer for [The Events Calendar](https://wordpress.org/plugins/the-events-calendar/) Plugin.
So event activity will be implemented for that post type.

### DIFF
--- a/includes/activity/extended-object/class-event.php
+++ b/includes/activity/extended-object/class-event.php
@@ -24,6 +24,7 @@ class Event extends Base_Object {
 		array(                                   // The keys here override/extend the context even more.
 			'pt'                            => 'https://joinpeertube.org/ns#',
 			'mz'                            => 'https://joinmobilizon.org/ns#',
+			'sc'                            => 'http://schema.org#',
 			'status'                        => 'http://www.w3.org/2002/12/cal/ical#status',
 			'commentsEnabled'               => 'pt:commentsEnabled',
 			'isOnline'                      => 'mz:isOnline',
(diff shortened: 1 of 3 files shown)


pfefferle:
Oh, this is awesome!

What about enabling Events by default?

```
\add_post_type_support( 'tribe_events, 'activitypub' );
```

pfefferle:
Is this already ready to merge?

Can you give some advise how to test it properly?

danielhuesken:
I think it is not ready for merge. I had not time for test it deeply.
For Testing you need:
- https://wordpress.org/plugins/the-events-calendar/
- https://wordpress.org/plugins/event-tickets/
- https://theeventscalendar.com/products/wordpress-events-calendar/ (Pro for virtuell events (Payed))
With that create an event with and without tickets.

pfefferle:
I will convert it to a draft, if you think the PR is ready, please click on "Ready for review" :)

Maybe @Menrath can help with Events!?!

Menrath:
@danielhuesken We are also working on supporting various Event plugins.

https://code.event-federation.eu/Event-Federation/wordpress-activitypub-event-extensions/src/branch/main/includes/activitypub/transformer/class-tribe.php

We are also automatically enabling the support for the events post type when our plugin is activated. That seems reasonable, when somebody installs our plugin (though it took us quite some time to do so). We also added admin notifications on several pages if it's not the case.

We are currently in contact with more than 10 event organizers we are evaluating our work with, and from what information I'd already gathered, automatically enabling certain post types for ActivityPub should not be done, without proper admin notification, I will write a blog post about this soon.

You transformer class looks good to me, actually more advanced that ours, but ours also a WIP.

We are currently working on some common settings for event post types transformation: e.g., a custom HTML summary with is composed by the events meta-data and a WordPress category to Event-Category mapper. Within events its good if they can be grouped when listed somewhere, some default groups according to some standardization attempts exist.

I would be glad to also merge some of your improvents in our project. We plan to release our plugin in November.

@pfefferle In the end it's your decision whether to merge this into this plugin. My opinion is that this part of the code would be best at home directly at the Events Calendar Codebase. The ActivityPub Plugin providing a FrameWork for applying this seems enough for me, or those integrations should be enabled only by an extra setting.

danielhuesken:
@Menrath Hello, thanks for you feedback. I only wirte the code to see how it works and I have a small project where The Events Calendar plugin is used, but currently not planned to use Activitypub.

So from my side takeover from the code what you would like and maybe in future i will use your plugin ;)
Is that also fine for you @pfefferle ?

pfefferle:
I am totally fine to have that in the Events Bridge instead of Core! I will close this and feel free to take over the code or re-open if you changed your mind!

Issue #864, opened 2024-08-19
Feature Request: Remote interaction endpoint


mediaformat:
### What

I would like to be able to interact with remote posts using existing federation affordances.

https://github.com/user-attachments/assets/c12f7e4c-3cea-46a1-b66d-dff329c9df76

### Why

Better interop across federated platforms

### How

Adding an `authorize_interaction` endpoint which could populate existing the reply intent 

`https://exemple.com/authorize_interaction?uri=URI_DISTANT_TOOT` -> `https://mediaformat.org/wp-admin/post-new.php?in_reply_to=URI_DISTANT_TOOT`

the ostatus template should also be available via the user webfinger:
```
...
'rel' => 'http://ostatus.org/schema/1.0/subscribe',
'template' => 'https://'.$domain.'/authorize_interaction?uri={uri}',
```

PR #866, opened 2024-08-19
Simple Remote-Reply endpoint.

pfefferle:
implements #864

Adds the remote-reply feature to WebFinger using the OStatus reply feature `http://ostatus.org/schema/1.0/subscribe`.

Also adds a hook, so that maybe the Friends plugin can hook into the Remote-Follow feature:

`\do_action( 'activitypub_interactions_follow', $uri, $object );`

/cc @akirk

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -62,6 +62,7 @@ function rest_init() {
 	Rest\Comment::init();
 	Rest\Server::init();
 	Rest\Collection::init();
+	Rest\Interaction::init();
 
 	// load NodeInfo endpoints only if blog is public
 	if ( is_blog_public() ) {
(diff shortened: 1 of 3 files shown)


mediaformat (commented):
1st try I was sent to the login page, however after that I was returned to `/wp-admin`
2nd try (already logged-in) it worked like a charm!

pfefferle:
Thanks for testing it @mediaformat !

I also checked the endpoint without a user session and the redirect after the login worked nicely for me!?!

Do you have a second factor or something that might interfere with the redirect?

mediaformat:
No 2nd factors, but I can try it again!

pfefferle:
any update on your tests @mediaformat ?

PR #867, opened 2024-08-19
A CLI extention to help with manual updates, deletes and a self-destruct feature

pfefferle:
TBD.


mattwiebe (approved):
Looks like a great basis for getting up to speed with `tootctl`!

We can keep iterating.

PR #874, opened 2024-08-28
Fix: wrong followers URL in "to" attribute of posts

Menrath:
This is more like an issue bug report with my status on what should happen.

The to attribute of a object like a Note/Post always contains the followers URL of the post author, even if they are not active.

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -4,18 +4,16 @@ namespace Activitypub\Transformer;
 use WP_Post;
 use Activitypub\Shortcodes;
 use Activitypub\Model\Blog;
-use Activitypub\Transformer\Base;
 use Activitypub\Collection\Users;
+use Activitypub\Transformer\Base;
 
 use function Activitypub\esc_hashtag;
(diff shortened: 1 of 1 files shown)


pfefferle:
@Menrath thanks a lot! I added some changes, to remove complexity. What do you think?

pfefferle (approved):

Menrath:
The changes look good to me! And actually I remembered some cases, where I thought something like the `activitypub_actor` property would be useful!

One last thought I had, whether having the `#Public` set by default could be improved. Sure a transformer currently should not get called, when the post is not public. But we can keep that in might or leave a comment.

Oh like I miss the WordPress coding style (greetings from a Moodle conference!).

Oh one last thought: I stumbled upon some cases where I needed to call the same getter function multiple times. So the idea was of adding checks if `$this->activitypub_object->property` is set. Or add this check in general in the setter, but in the end my PHP knowledge is still too non-substantial to have an idea what the real performance impact might be, and in the end I thought lets keep it simple.

Issue #138, opened 2022-02-23
Does this plugin support the "replies" collection?


johnspurlock:
Sending an ActivityPub request to
 https://www.draky.net/wordpress/2019/11/15/bonjour-tout-le-monde/

which has this plugin installed, returns good looking AP, but without the "replies" collection 
  https://www.w3.org/TR/activitystreams-vocabulary/#dfn-replies

This makes it impossible to enumerate the comment thread.

I'm wondering if it's implemented?

Thanks
\- John

pfefferle:
It is not **yet** implemented.

Do you know a platform that is already consuming the replies-collection?

johnspurlock:
Ah, ok - thanks for getting back.

There is a new open standard for podcast episode comments that relies on the existing basic Fediverse protocol (ActivityPub, Webfinger, http signatures for auth) for sending and displaying replies to a root comment post associated with that episode.

Info site here: https://podcastsocial.org/
Underlying [rss feed tag spec](https://github.com/Podcastindex-org/podcast-namespace/blob/main/proposal-docs/social/social.md#socialinteract-element), and open source components to make it easier for existing podcast apps to [send](https://minipub.dev/) and [display](https://www.npmjs.com/package/threadcap) them.

For displaying comments, it relies on being able to enumerate comments by parsing the standard `replies` collections 
 recursively.  Mastodon supports paged collections in its `replies` implementation, but replies can also be modeled inline if that's easier.

For sending comments, it relies on standard s2s federation `inReplyTo` the target object.  Is this supported?  I'm getting a 406 error (see below for the gory details) when POSTing to the inbox url using my custom software.  Replying from a Mastodon instance does not seem to work either.

I was really excited when I saw your plugin, the idea of a podcaster being able to use a WordPress site as the venue for their comments instead of, say, Mastodon, is that WP is already very popular in the podcasting community, and it would presumably provide additional moderation options for these incoming comments.

```
POST https://example.com/wp-json/activitypub/1.0/users/1/inbox
date: Thu, 24 Feb 2022 14:36:26 GMT
signature: keyId="https://sender.com/path/to#main-key",algorithm="rsa-sha256",headers="(request-target) host date digest",signature="<base64>"
digest: SHA-256=<base64>
content-type: application/activity+json

{
  "@context": "https://www.w3.org/ns/activitystreams",
  "id": "https://sender.com/path/to/activity",
  "type": "Create",
  "actor": "https://sender.com/path/to/actor",
[...]

PR #876, opened 2024-09-01
Add replies collection

Menrath:
Add [replies collection](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-replies) to posts and federated comments.

**Solves Issue** https://github.com/Automattic/wordpress-activitypub/issues/138

### Description
The replies collection offer a neat way to sync replies over multiple instances. Mostly on demand and sometimes also client side on demand. They are not yet frequently used because of the moderation issues that arise with them. But since WordPress already got a good approve-system for comments this should not be too problematic here.

### Rest Endpoint
To make sure the ID's of the collection and the collections page are resolveable.
It also adds a new rest endpoint `/activitypub/1.0/(posts|comments)/<id>/replies` which returns a `Collection`. Adding the parameter `?page=0` returns the CollectionPage directly.
I decided not to use something like `<the_permalink>/replies` because it is only needed for `activity+json` and has no frontend view. Seemed consistent to me.

### Unit tests:
I started writing new unit tests, but was not sure how deep to test this. I'm happy for some discussions.

### Manual testing:
Write federated comments for a post, or receive comments and make sure those comments are approved.
When querying the ActivbityPub's JSON representation of the post the replies key exists, e.g.:

```json
...
 "replies": {
    "id": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies",
    "type": "Collection",
    "first": {
      "id": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies?page=0",
      "type": "CollectionPage",
      "partOf": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies",
      "items": [
        "https://mz.lan/comments/8be36040-5a02-4335-917f-4c5dcc988e1f",
        "https://mz.lan/comments/de26c344-8d7d-400f-a8f5-5b89307bb72a",
        "https://mz.lan/comments/abce32af-a59e-41c6-8b47-a8925980f1ef",
        "https://wp.lan/?c=13",
        "https://mz.lan/comments/fa3d2ee2-147b-4eee-b682-98b0b7115603"
      ]
    }
  },
...
```

Changing the global option `comments_per_page` to for example "3" would change the response to this:

```json
...
"replies": {
    "id": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies",
    "type": "Collection",
    "first": {
      "id": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies?page=0",
      "type": "CollectionPage",
      "partOf": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies",
      "items": [
        "https://mz.lan/comments/8be36040-5a02-4335-917f-4c5dcc988e1f",
        "https://mz.lan/comments/de26c344-8d7d-400f-a8f5-5b89307bb72a"
      ],
      "next": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies?page=1"
    }
  },
...
```

Note that the items only contains 2 IDs. This is because the code currently uses core functions to query the paginated comments and later filters out the local-only ones. Because we have more pages also the key `next` is present.

It (`https://wp.lan/wp-json/activitypub/1.0/posts/49/replies?page=1`) resolves to:

```json
**https://wp.lan/wp-json/activitypub/1.0/posts/49/replies?page=1**
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    {
      "Hashtag": "as:Hashtag"
    }
  ],
  "id": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies?page=1",
  "type": "CollectionPage",
  "partOf": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies",
  "items": [
    "https://mz.lan/comments/abce32af-a59e-41c6-8b47-a8925980f1ef",
    "https://wp.lan/?c=13"
  ],
  "next": "https://wp.lan/wp-json/activitypub/1.0/posts/49/replies?page=2"
}
```

**Unapproved or local only comments should never be listed.**

### Discussion points
- This draft has always functions handle both replies for comments and posts rather than splitting up the code in the beginning. I just chose this construct to get to a working prototype and I already fear that there might be good arguments to do it differently.
- The code which gets the federated comment IDs: https://github.com/Menrath/wordpress-activitypub/blob/replies_collection/includes/collection/class-replies.php#L108 These lines are duplicated or have very similar occurences, more than once in the code.
- Maybe use a direct meta-query to get all approved-federated comments in a paginated way.
- I am happy to hear any ideas and thoughts.
- Mastodon for example lists all self-replies on the first page of the collection, and external replies later, this would also be possible, to ensure threads are displayed. Threads are possible now also in WordPress, but maybe not that important.
- Mastodon includes local replies fully, not only the ID. Within WordPress this in theory could get difficult because of the JSON-LD context (different post types with different ActivityPub object types at worst).


includes/collection/class-replies.php
```diff
@@ -0,0 +1,175 @@
...
+use WP_Comment;
+use WP_Error;
+
+use function Activitypub\is_local_comment;
+use function Activitypub\get_rest_url_by_path;
+
+/**
+ * Class containing code for getting replies Collections and CollectionPages of posts and comments.
+ */
+class Replies {
```
pfefferle, line 14: I think we should merge that class with the interactions collection. I have no emotions about which one to use, so we can wait until this is merged and migrate the interactions class afterwards. Just wanted to mention it.

I think maybe `Replies` is really the better choice!

pfefferle:
Awesome 😍

pfefferle:
@Menrath Is this ready to merge at least for a first attempt?

Menrath:
I would love to see some more unit tests, but besides that I feel from my point of view I have nothing to add atm :)

includes/collection/class-replies.php
Menrath: I prefer smaller files and think having this under collection/ is a good place.

includes/rest/class-collection.php
```diff
@@ -225,4 +293,25 @@ public static function request_parameters() {
...
+
+	/**
+	 * The supported parameters
+	 *
+	 * @return array list of parameters
+	 */
+	public static function request_parameters_for_replies() {
+		$params = array();
+
+		$params['type'] = array(
```
pfefferle, line 305: You can use an `enum` for that, so that you do not have to check it in the action.

See: https://github.com/Automattic/wordpress-activitypub/blob/8bff92d26020c575782f50740dece4f6c6ed7247/includes/rest/class-followers.php#L134

includes/rest/class-collection.php
```diff
@@ -225,4 +293,25 @@ public static function request_parameters() {
...
+			'required' => true,
+			'type' => 'string',
```
pfefferle, line 307:
```suggestion
			'type' => 'string',
			'enum' => array( 'post', 'comment' ), 
```

includes/collection/class-replies.php
pfefferle: I agree that this should be a collection and I will think about the "small" files :)

includes/collection/class-replies.php
```diff
@@ -0,0 +1,175 @@
...
+
+	/**
+	 * Get the replies collection.
+	 *
+	 * @param WP_Post|WP_Comment $wp_object
+	 * @param int $page
+	 *
+	 * @return array An associative array containing the replies collection without JSON-LD context.
+	 */
+	public static function get_collection( $wp_object ) {
```
pfefferle, line 81: couldn't this be included in `get_collection_page` with `$page = 0`???

pfefferle: @Menrath friendly ping ☺️

Menrath (commented):
Seems all good to me.

includes/collection/class-replies.php
```diff
@@ -0,0 +1,175 @@
...
+	 */
+	private static function get_activitypub_comment_ids( $comments ) {
+		$comment_ids = array();
+		// Only add external comments from the fediverse.
+		// Maybe use the Comment class more and the function is_local_comment etc.
+		foreach ( $comments as $comment ) {
+			if ( is_local_comment( $comment ) ) {
+				continue;
+			}
+			$comment_meta = \get_comment_meta( $comment->comment_ID );
```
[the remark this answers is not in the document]
Menrath: @pfefferle Should someone have a look at this before this PR gets merged? I could find time in the next days.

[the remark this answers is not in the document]
pfefferle: but be aware that there are two different usecases:

1. We want to have the ID with URL as fallback.
2. We want to have the URL with ID as fallback.

pfefferle (approved):

includes/collection/class-replies.php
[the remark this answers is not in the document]
Menrath: https://github.com/Automattic/wordpress-activitypub/pull/876/commits/769e434f02b7acdbe641ca60c5488b5b7c90e64c added a static functions to the Comment util class, that tries to get the link via the comments meta.  At some points I thought better error-catching could be possible, but I think this is beyond the scope of this PR. We could also just merge the current state, and revert the commit and add another PR for refactoring the code for comment related stuff.

pfefferle (approved):

pfefferle:
thanks a lot @Menrath !

Menrath:
@pfefferle You're welcome! :) I also really like the last changes, splitting the functionality into the `get_source_url` and `get_source_id`. It's just easier to read.

PR #885, opened 2024-09-09
Improve Accessibility for Reply-Context

### LINKED ISSUE #875: Screen readers badly render links in "federated reply" block
### Quick summary

@pfefferle when using the "reply to" bookmarklet, at beginning of a WordPress post, an unpleasant output for screen readers appears - for example "↬mastodon.uno/@uaar/113039993829553640"
It is the direct link at the post/conversation we are replying to and it's correct, but the "a href" HTML tag has no link text! So, for fediverse newbies it's not easy to figure it is a post's reply because the screen reader reads the alpha-numeric link and a user would never click something "anonymous"; it seems like a malicious code.

### Steps to reproduce

1. open any Mastodon/Fediverse post;
2. use the "reply from [blog]" bookmarklet so that it creates a new post;
3. the issue can be verified also in previewing the post after having saved it into draft.

The "federated reply" block should look, in front-end, like this:
"this post is in reply to: [mentioned user] conversation: view original thread" - and text string in the block should be editable from within the block, so that a user can introduce it as they want (I mean the link text and "this post is in reply to"...).

### What actually happened

Now just an emoji and the post's alphanumeric link/code appear. Alone, with no explanation on what it is.

### Impact

All

### Available workarounds?

### Logs or notes

ActivityPub version 3.0.1, WordPress 6.6.1, Gutenberg 19.1. This issue has no workarounds and prevents me from replying to conversations. - Not even verified if clicking the link opens a new window. If it doesn't and takes me out of the blog, it's even more serious.

pfefferle:
Fixes #875

### DIFF
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -163,7 +163,7 @@ class Blocks {
 		return apply_filters(
 			'activitypub_reply_block',
 			sprintf(
-				'<p><a title="%2$s" href="%1$s" class="u-in-reply-to" target="_blank">%3$s</a></p>',
+				'<p><a title="%2$s" aria-label="%2$s" href="%1$s" class="u-in-reply-to" target="_blank">%3$s</a></p>',
 				esc_url( $attrs['url'] ),
 				esc_attr__( 'This post is a response to the referenced content.', 'activitypub' ),
 				// translators: %s is the URL of the post being replied to.


pfefferle:
@talksina does this solve your problem?

talksina:
I hoped so, but it seems not.

pfefferle:
@talksina are you referring to the change or the latest update?

talksina:
I tried latest ActivityPub update downloaded from WordPress repository with traditional updates.
and weird code still appears. I don't know if installing your pull requests or stuff posted above, no idea on how to use it
be patient with me - I am quite new on github about testing code and so on

PR #888, opened 2024-09-09
Added documentation for setup behind reverse proxy with Apache

jjschwarz:
Fixes #

## Proposed changes:

* Added information about use behind an Apache reverse proxy discovered in case 879


pfefferle (approved):

pfefferle:
Thanks a lot @jjschwarz

PR #892, opened 2024-09-12
Follow Me/Followers: add `inherit` mode for dynamic templating

mattwiebe:
We received this piece of feedback from our friends at Woo:

> A site owner should be able to configure a Widget so that the Widget tries to extract the correct user from the page it is on. This way it could be used in Post-Footers or Author-Page-Headers

This accomplishes that goal.

## Proposed changes:
* Allow setting an `inherit` mode called **Dynamic User** on the Follow Me and Followers blocks
* It shows the associated user's ActivityPub profile or followers on 1) user pages and 2) individual posts (the author's profile/followers are shown)

### Other information:

It looks like this:

Screenshot 2024-09-12 at 15 13 09

I haven't thoroughly tested it yet, particularly with users that exist and have posts but are disabled.


includes/class-blocks.php
```diff
@@ -67,12 +67,50 @@ public static function register_blocks() {
...
+
+		// If we're on a user archive page, use that user's ID.
+		if ( is_a( $queried_object, 'WP_User' ) ) {
+			return $queried_object->ID;
+		}
+
+		// For a single post, use the post author's ID.
+		if ( is_singular() && is_a( $queried_object, 'WP_Post' ) ) {
+			return $queried_object->post_author;
+		}
```
akirk, line 110: This doesn't work in a query loop since the queried object will be the outer post. In my testing it was enough to just use `get_the_author()` which will have the right author inside the query loop and for the outer post.

```suggestion
		// For a single post, use the post author's ID.
		if ( is_a( $queried_object, 'WP_Post' ) ) {
			return get_the_author_meta( 'ID' );
		}
```

src/follow-me/edit.js
```diff
@@ -37,7 +39,18 @@ export default function Edit( { attributes, setAttributes } ) {
...
+			{ isInheritMode ? (
+				<Card>
+					<CardBody>
+						{ createInterpolateElement(
+							__( 'This <strong>Follow Me</strong> block will adapt to the page it is on, showing the user profile if on a user archive, or the post author on a single post. It will be <strong>empty</strong> on non-author pages.', 'activitypub' ),
+							{ strong: <strong /> }
+						) }
+					</CardBody>
+				</Card>
+			) : (
```
akirk, line 51: Could this show the actual block instead?

mattwiebe: There's nothing to show at this point because no user is selected. 

I thought people would be more likely to be confused by having one thing show in the Editor and something else on the frontend. I was also thinking of adding a fallback user selector and have that show, but again, the potential for confusion…

includes/class-blocks.php
mattwiebe: Great find! Hadn't thought about loop stuff at all yet

src/follow-me/edit.js
akirk: I am not sure I follow. I fully agree, it should show the same thing in the editor as on the frontend. But I believe this can be achieved for cases when we talk about posts. The query loop block shows different posts live inside the editor, for example. The follow block could show for the right author for each post.

akirk: Do you think it'd be reasonable to provide a preview when in a query loop? Good enough for a follow up, too, just wondering.

mattwiebe: Yeah I think making it show up in the loop would go a long way. I have no idea how to do that, should be interesting to learn. :)

pfefferle (dismissed):
works like a charm :)

mattwiebe:
https://github.com/Automattic/wordpress-activitypub/pull/892/commits/4ade1f5863bea440798dbdde1d0a15305f8809cd tweaked the behaviour for disabled users in the Followers block, or for any other case where a user_id produces a `WP_Error`. The Follow Me block was already bailing on errors, so now we're consistent there too.

pfefferle (approved):

pfefferle (approved):

PR #893, opened 2024-09-14
It seems that Mastodon has issues with HTML in summaries

### LINKED ISSUE #889: HTML code in excerpts seeing content from Mastodon
### Quick summary

When publishing a content on WordPress blog, ActivityPub returns HTML code instead of text, in Mastodon networks.

### Steps to reproduce

1. Try to go to: - mastodon . social / @network @plusbrothers . net
-I put spaces in order to avoid undesired effects here in Github
2. you should see last posts
3. in last posts, the excerpt (or first sentence before the "read more") is enclosed between < p > tags instead of seeing pure text. And even the "& numbers;" code instead of apostrophes.

only text should appear.

### What actually happened

pure html code is displayed

### Impact

All

### Available workarounds?

### Logs or notes

platform: mastodon. Plugin version: activitypub 3.2.2.

### LINKED ISSUE #28455: The summary for converted ActivityPub objects must be treated as html according to the ActivityStreams specification
### Steps to reproduce the problem

1. Retrieve a non-note object like an `Event` from with a HTML summary.
2. The resulting post will not be rendered/converted as HTML but as plain text.

### Expected behaviour

The summary should be treated as HTML by default. 

### Actual behaviour

The summary is treated as plain text. 

The following Screenshot shows what a event received from Gancio then looks like.
Screenshot_from_2023-12-20_01-20-36

### Detailed description

The feature introduced in https://github.com/mastodon/mastodon/pull/9823 is basically a great one, without Mastodon implementing custom renderings for ActivityPub objects like `Event` it has basic support for them. The actual convertion likely takes place here: https://github.com/mastodon/mastodon/blob/2463b53363a621c1e18223bda5c47a663707a22c/app/lib/activitypub/activity/create.rb#L371-L373.

The [ActivityStreams vocabulary specification](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-summary) states that 

> **Summary**: A natural language summarization of the object encoded as HTML. 

This will cause interoperability issues with future Gancio, Mobilizon and WordPress ActivityPub releases.

Altough the ActivityPub spec does not mention it, one might also think of accepting custom `MediaType` for the summary, to handle both plain/text and HTML summaries.

See https://framagit.org/les/gancio/-/issues/321#note_2038846

### Mastodon instance

All

### Mastodon version

Since https://github.com/mastodon/mastodon/pull/9823

### Technical details

_No response_

pfefferle:
Fixes: https://github.com/mastodon/mastodon/issues/28455

Until this is merged: https://github.com/mastodon/mastodon/pull/28629

Internal Fixes: #889

### DIFF
--- a/includes/class-hashtag.php
+++ b/includes/class-hashtag.php
@@ -28,9 +28,11 @@ class Hashtag {
 	 * @return array the activity object array
 	 */
 	public static function filter_activity_object( $object_array ) {
+		/* Removed until this is merged: https://github.com/mastodon/mastodon/pull/28629
 		if ( ! empty( $object_array['summary'] ) ) {
 			$object_array['summary'] = self::the_content( $object_array['summary'] );
 		}
+		*/
 
 		if ( ! empty( $object_array['content'] ) ) {
 			$object_array['content'] = self::the_content( $object_array['content'] );
--- a/includes/class-link.php
+++ b/includes/class-link.php
@@ -24,9 +24,11 @@ class Link {
 	 * @return array the activity object array
 	 */
 	public static function filter_activity_object( $object_array ) {
+		/* Removed until this is merged: https://github.com/mastodon/mastodon/pull/28629
 		if ( ! empty( $object_array['summary'] ) ) {
 			$object_array['summary'] = self::the_content( $object_array['summary'] );
 		}
+		*/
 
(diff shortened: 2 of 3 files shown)


obenland:
@pfefferle Is this ready to be reverted?

PR #894, opened 2024-09-14
Prepare `?p=<id>` and `?author=<id>` for ActivityPub

pfefferle:
This way we could replace the permalink with the ID of a post or an author, to work agains the Username and Permalink issue.

Maybe fixes the caching issues in WordPress!?!


pfefferle:
@mattwiebe @akirk @jeherve @mediaformat @Menrath Is that the way to go? What do you think about using the `example.org/?p=<ID>` instead of `example.org/permalink/` as ID?

Menrath:
I think that would be a good idea, if I remember rightly, we discussed this very thing on a call about a year ago!  :) 

Can you link to any discussion about the mentioned "to work agains the Username and Permalink issue". I am not completely sure, if I have the right issue in mind.

A suggestion:  what about replacing `author` with `activitypub_actor`, along with a clear comment on how the relationship currently works. This would give us more flexibility for certain features in the future.

So do you think `get_id` return the proposed structure and `get_url` should still return the permalink? Or should we use that structure for both?

pfefferle:
> Can you link to any discussion about the mentioned "to work agains the Username and Permalink issue". I am not completely sure, if I have the right issue in mind.

I am not sure if there is a discussion exactly about that, but in the end it is the same discussion that speaks against the permalink. People want to change names and if the pretty URL is also the ID, we always need to send a `Move` activity if a user want to change usernames.

> A suggestion: what about replacing author with activitypub_actor, along with a clear comment on how the relationship currently works. This would give us more flexibility for certain features in the future.

The nice thing about using `?author=` is, that all core WordPress functions (`is_author` for example) still work and we do not have to change too much code.

> So do you think get_id return the proposed structure and get_url should still return the permalink? Or should we use that structure for both?

`get_id` should return `example.org/?p=1` and `get_url` should return the pretty permaling `example.org/pretty/permalink`.

jeherve:
It makes sense to me at first, since it follows the existing "ugly permalinks" defaults from Core itself. I'm not familiar with the issues it solves though, so I'll let others weigh in on that :)

pfefferle:
@akirk the problem with the REST endpoint is, that the ID has to return the ActivityPub JSON when requested with the correct Accept header!

akirk:
Makes sense not to use it then! I just wanted to add it as an option so that we know it was considered.

mediaformat:
This is very welcome! 

Using `?p=<id>` for ActivityPub `id` makes the system more robust, allowing site admins to change permalink structute and site editors/authors to change post slugs without breaking existing federated interactions.

mattwiebe:
The only downside of this that I can see is that you will no longer to be able to paste the pretty permalink URL into eg a Mastodon search and get the result.

But it's overall more robust and does have the possibility of bypassing some caching layers, have there been any tests along those lines? Are we looking to address caching plugins, or Cloudflare, or?

pfefferle:
@mattwiebe 

> The only downside of this that I can see is that you will no longer to be able to paste the pretty permalink URL into eg a Mastodon search and get the result.

This will still be possible! I do not want to disable the Content-Negotiation, so people will still be able to search for `permalinks` and `?p=` links!

pfefferle:
> But it's overall more robust and does have the possibility of bypassing some caching layers, have there been any tests along those lines? Are we looking to address caching plugins, or Cloudflare, or?

Exactly! If a caching system would not be able to support Content-Negotiation, the user can simply disable the cache for `?p=*`! Then the user well indeed no longer be able to search for the URL in Mastodon, but it would at least work!

mattwiebe:
This is definitely sounding like a good way to go. What if any are the downsides for existing sites having all of their IDs updated in this way?

pfefferle:
@mattwiebe I thought about storing a date, when the plugin updates to this version, to only change the ID for posts that are newer than that date. does that make sense?

pfefferle:
Interesting point: https://mastodon.social/@by_caballero/113152023836926387

includes/class-migration.php
```diff
@@ -337,4 +341,31 @@ private static function update_options_key( $old, $new ) { // phpcs:ignore
...
+				'order'       => 'DESC',
+				'post_status' => 'publish',
+			)
+		);
+
+		if ( $latest_post ) {
+			$latest_post_id = $latest_post[0]->ID;
+		}
+
+		update_option( 'activitypub_last_legacy_post', $latest_post_id );
```
Menrath, line 369: Suggestion: maybe choose a very clear option name like: `activitypub_last_post_with_pretty_permalink_as_id`

mattwiebe: I thought that might be too long (49 chars) so I double-checked: `option_name` is `varchar(191)` and we're just fine :)

includes/class-migration.php
```diff
@@ -136,6 +136,10 @@ public static function maybe_migrate() {
 			self::migrate_from_2_6_0();
 		}
 
+		if ( version_compare( $version_from_db, '4.0.0', '<' ) ) {
+			self::migrate_to_4_0_0();
```
mattwiebe, line 140: Will this logic run for a new install? Does it need to be added to `add_default_settings()`?

pfefferle: Good point. I checked the code and in the case of a new install, it simply falls back to the default which is `0` and shows all IDs in the new format.

mediaformat:
Just a heads up that I came across a site that is linking to `?author=<id>` already, but being blocked by a plugin https://robertkingett.com/?author=1

pfefferle:
I can't believe that "hiding the username" is still one of the biggest security measures!!!

pfefferle:
I am a big friend of re-using as much as possible from what's already provided by WordPress, mostly to not produce dead links when deactivating the plugin!

pfefferle:
@mediaformat I checked the site and there is nearly everything blocked! The shields thingy also blocks the rest api and other endpoints, so the plugin wouldn't work anyways!

Menrath:
FYI: Has been referenced by us externally: https://code.event-federation.eu/Event-Federation/wordpress-activitypub/issues/5

A proposal that I would be interested to hear your opinion would be, whether marking certain methods for instance in the Post transformer `get_id`, `get_published` as final.

pfefferle:
@Menrath so that it can't be overwritten by childs?

Menrath:
> @Menrath so that it can't be overwritten by childs?

I really didn't want to be nitpicky. If should have felt weird seeing us discussing things "behind your back", I'll apologize!  The only reason certain issues over there exist because we are absolutely not sure whether it is worth bothering your time, we just used them as a public notepad of what we were thinking about. Anyway, my idea was to make it as easy as possible for potential new contributors and to also to make the code a resource that might be useful for learning things about ActivityPub. In the end when taking edge cases into account, we may not have thought of, this could also be counter productive. To put it in a nutshell: @pfefferle You are doing a great job in governing this project! I wish you a nice day :)

pfefferle:
No need to apologize! I am happy about all discussions and proposals that happen around the plugin, regardless where they happen :)

I have not seen the link to your repo, that's why I asked!

I think it is a valid request! I like the idea of preventing implementers for breaking things.

I agree with:

* id
* published
* updated

I am not sure about:

* url

includes/transformer/class-post.php
```diff
@@ -131,7 +131,19 @@ protected function get_actor_object() {
...
+		$last_legacy_id = (int) \get_option( 'activitypub_last_post_with_permalink_as_id', 0 );
+		$post_id        = $this->wp_object->ID;
+		$post_status    = \get_post_meta( $post_id, 'activitypub_status', true );
+
+		if (
+			'federated' !== $post_status &&
+			$post_id > $last_legacy_id
+		) {
+			// generate URI based on comment ID
+			return \add_query_arg( 'p', $post_id, \trailingslashit( \home_url() ) );
```
Menrath, line 130: Don't know if it may be of importance: Some popular plugins like "The Events Calendar" use a `guid` that differs from that scheme, for instance `https://wp.lan/?post_type=tribe_events&#038;p=8" for the post with ID=8.

Any ideas, why using the GUID could be better or worse than composing an URI based on the post ID? ...uuups... that phpcomment is wrong, its mentioning the comment ID.

pfefferle: @Menrath Good point.

I checked my local test env and my live system and the GUID is sadly not really used consistently!

* I see the permalink and the `?p=` version for Posts (no idea why WordPress is mixing it up)
* For Attachments it is the path to the Attachment file in the file system
* For pages it is the same as for Posts, sometimes it is the permalink, some times it is `?p=`

And what's even worse: Because the GUID includes the full domain, I found some old pages that still uses my old domain and `http` instead of `https`!

But even if you take aside the permalink and domain issue, because it might be a historical issue, the Attachments would break everything.

Another point is, that the GUID can be set by implementers (as we do with Followers).

pfefferle: @Menrath so I would stick with the current composing!?!

Menrath: I agree!  Thanks for pointing out the arguments here - I think we discussed some of them privately some time ago, but I couldn't remember almost any of the conclusions - anyway, great that the knowledge is now being shared and documented.

Menrath (approved):
Looks good to me. I can/will do real-world tests of this changes in the next days.

pfefferle:
thanks @Menrath

pfefferle:
@Menrath I tested profiles and posts with misskey, mastodon, nodebb and friendi.ca and everything worked fine so far!

No issues with query params so far!

pfefferle:
@mattwiebe @akirk @jeherve @mediaformat @Menrath I would love to get this merged! Does anyone have any concerns or has anyone had positive/negative effects when testing?

akirk:
I didn't test this but I am unsure how this should work.

For posts, it makes sense: As I understand the code you want to maintain the old full URL until the last post when migrating and from then on the URL with the post id.

But for authors, you're changing the ActivityPub id and maintaining it for old followers. So doest this mean that for Mastodon instances they will appear as two separate authors?

akirk:
Oh, I realise because of [aliases in Webfinger this should "just work."](https://github.com/Automattic/wordpress-activitypub/pull/894/files#diff-cd3a14c352f6adc591b05ac8f5a54e17eb2ba5e1482534963855992da3ced05cR81)

pfefferle:
@akirk 

> But for authors, you're changing the ActivityPub id and maintaining it for old followers. So doest this mean that for Mastodon instances they will appear as two separate authors?

I check if the user has followers, if so I do not change anything, if not it uses the "new" id: https://github.com/Automattic/wordpress-activitypub/pull/894/files#diff-e2183e3284e3a7161c943a36d71aa3b1becf63ffe7ed41afa0b16d4c16944f2eR368 and https://github.com/Automattic/wordpress-activitypub/pull/894/files#diff-1d75cd772f6ce6eec7428f320fbc02e1829fbf3e96194d74c451beb246fc50e7R116

mattwiebe (approved):
I think it's ready! One way to find out. :shipit:

Issue #858, opened 2024-08-14
Example handle


nickbohle:
https://github.com/Automattic/wordpress-activitypub/blob/dd8f363c354ca1a25b8c9a8d95d49f0981abd88a/src/shared/dialog.js#L83

I 've just added the Follow block to my home page and in my opinion, the example handle in the description is confusing. Example is: yourusername@example.com It misses the first AT, for example. Why not @YourUsername@server.com or @YourUsername@instance.com

pfefferle:
It does work with and without a leading `@`. The specification (WebFinger) is without the `@`, this is only a Mastodon thing.

But I agree, it might be consistent to add it. Thanks for the report.

pfefferle:
done

PR #898, opened 2024-09-23
fix missing `@` in Follow-Me block

### LINKED ISSUE #897: Missing leading @ symbol in username on the follow-me block
### Quick summary

In the edit view of the follow-me block the @ symbol is missing from the view, making the username/account look like an email address.
image

### Steps to reproduce

1. Add the follow-me block
2. Select a specific author
3. Compare the editor view with the front-end view of the block

I would expect full usernames in the block preview; e.g. @colin@cdevroe.com 

### What actually happened

Instead I see colin@cdevroe.com (which looks like an email address, because it is my email address!)

### Impact

All

### Available workarounds?

There is no user impact

### Logs or notes

I'm submitting a simple patch that _may_ address it. But I couldn't decide exactly where the best place to add the @ symbol in. So, I figured someone else should address it.

pfefferle:
fix #897 and #858


cdevroe:
👏

mattwiebe (approved):
🚢

PR #900, opened 2024-09-24
Content warnings! 

### LINKED ISSUE #757: Feature Request: Content Warnings
### What

I would like the option to add content warnings to posts when they are published via activitypub.

### Why

Many fediverse instances require NSFW content to be content-warned. In some circles it's also a social norm to tag certain topics (such as "mh-", "American Politics", "transphobia", or others).

I would appreciate if this plugin provided users (like me) the option to both better respect instance rules and also be good citizens of the fediverse.

### How

Some possible implementations:
- Via tags - "Posts tagged with tags X, Y or Z get content warnings", using the tag name as the CW text
- Via categories - "Posts in category X, Y or Z get a content warning", using the category name as the CW text
- A new field on posts?

I'm unsure what would be the best way to handle it, but I'd appreciate the option

mattwiebe:
We now support content warnings! They are per-post.

fix #757 #176 

## Proposed changes:

Adds a "Fediverse" settings section to the post editor that allows entering a content warning.

Screenshot 2024-09-24 at 16 12 06

The warning is used to set `sensitive: true` and populates the `summary` attribute.

### Other information:

## Testing instructions:

On a post, set the value as above. I entered a value of `warning!!!` and then by appending `?activitypub` to the post, I got this JSON:

```json
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    {
      "Hashtag": "as:Hashtag"
    }
  ],
  "id": "http://wp.test/2024/09/make-a-poast/",
  "type": "Note",
  "attachment": [],
  "attributedTo": "http://wp.test/author/mattwiebe/",
  "audience": "http://wp.test/@wp.test",
  "content": "<p>Poasty poast</p><p><a href=\"http://wp.test/2024/09/make-a-poast/\" class=\"status-link unhandled-link\">http://wp.test/2024/09/make-a-poast/</a></p>",
  "contentMap": {
    "en": "<p>Poasty poast</p><p><a href=\"http://wp.test/2024/09/make-a-poast/\" class=\"status-link unhandled-link\">http://wp.test/2024/09/make-a-poast/</a></p>"
  },
  "published": "2024-09-16T19:04:37Z",
  "summary": "warning!!!",
  "tag": [],
  "updated": "2024-09-24T14:05:10Z",
  "url": "http://wp.test/2024/09/make-a-poast/",
  "to": [
    "https://www.w3.org/ns/activitystreams#Public",
    "http://wp.test/wp-json/activitypub/1.0/actors/1/followers"
  ],
  "cc": [],
  "sensitive": true
}
```


includes/functions.php
```diff
@@ -1177,6 +1177,11 @@ function generate_post_summary( $post, $length = 500 ) {
 		return '';
 	}
 
+	$content_warning = get_content_warning( $post->ID );
```
pfefferle, line 1180: if you set it in the transformer, you do not also have to add it here.

I quite like the idea to set it in the transformer only, this we can still generate the summary!

includes/functions.php
```diff
@@ -1215,3 +1220,24 @@ function generate_post_summary( $post, $length = 500 ) {
...
+	if ( ! $post ) {
+		return false;
+	}
+
+	$warning = get_post_meta( $post->ID, 'activitypub_content_warning', true );
+	if ( empty( $warning ) ) {
+		return false;
+	}
+
+	return $warning;
```
pfefferle, line 1242: do we have to escape it in some way?

pfefferle:
Awesome!

includes/functions.php
pfefferle: If we would keep it in there, it would show the "content warning" also in the content field if someone would use the shortcode `[ap_excerpt]` for the content field.

includes/functions.php
```diff
@@ -1177,6 +1177,11 @@ function generate_post_summary( $post, $length = 500 ) {
 		return '';
 	}
 
+	$content_warning = get_content_warning( $post->ID );
+	if ( $content_warning ) {
+		return $content_warning;
+	}
+
```
pfefferle, line 1184:
```suggestion
```

pfefferle (approved):

mattwiebe:
@pfefferle thanks for taking this to the finish line!

PR #901, opened 2024-09-25
Anonymous function lint

mattwiebe:
This is what wpcom demands, apparently :)

### DIFF
--- a/includes/collection/class-extra-fields.php
+++ b/includes/collection/class-extra-fields.php
@@ -67,7 +67,7 @@ class Extra_Fields {
 		$attachments = array();
 		\add_filter(
 			'activitypub_link_rel',
-			function( $rel ) {
+			function ( $rel ) {
 				$rel .= ' me';
 
 				return $rel;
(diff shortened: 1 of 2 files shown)


pfefferle (approved):

obenland:
[It's part of Core's PHP Coding Standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#closures-anonymous-functions).

How would you feel about requiring them for this plugin?

pfefferle:
@obenland I thought we do that, I have no idea why this was not checked: https://github.com/Automattic/wordpress-activitypub/blob/master/phpcs.xml

obenland:
So many exceptions! 😱

I'm not sure, it's been a while since I tried setting that up. Maybe all warnings need to elevated to errors for them to be blocked from going in? https://github.com/Automattic/wpcomsh/blob/trunk/phpcs.xml.dist#L35-L36

pfefferle:
Yes, @mattwiebe worked quite some time on it, so that it is compatible with WP.com

mattwiebe:
Those exceptions are mostly mimicking stuff I found on the wpcom side :)

PR #903, opened 2024-09-26
Keep priority of Icons (only change fallbacks)

pfefferle:
and only replace the icon, because we can not assume that a logo is always square!?!

Fixes https://mastodon.social/@vsg_DE/113203784133921789

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/model/class-blog.php
+++ b/includes/model/class-blog.php
@@ -209,10 +209,6 @@ class Blog extends Actor {
 		// try site_logo, falling back to site_icon, first
 		$icon_id = get_option( 'site_icon' );
 
-		if ( ! $icon_id ) {
-			$icon_id = get_option( 'site_logo' );
-		}
-
 		// try custom logo second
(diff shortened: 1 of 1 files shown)


includes/model/class-blog.php
```diff
@@ -207,7 +207,11 @@ public function get_preferred_username() {
 	 */
 	public function get_icon() {
 		// try site_logo, falling back to site_icon, first
-		$icon_id = get_option( 'site_logo', get_option( 'site_icon' ) );
+		$icon_id = get_option( 'site_icon' );
+
+		if ( ! $icon_id ) {
+			$icon_id = get_option( 'site_logo' );
```
mattwiebe: We should probably just get rid of `site_logo` altogether?

This is effectively identical to the way it was the fallback in the previous iteration.

mattwiebe (approved):
:shipit:

PR #905, opened 2024-09-28
Suggestion: Adopt WordPress PHP Coding Standards

obenland:
I'm not sure if switching to Core WordPress coding standards is something y'all want to do. If you do, this could be a good start. The exceptions could be removed and fixed one-by-one in subsequent PRs.

If not, feel free to just close.

## Proposed changes:
* Updates phpcs config to use `WordPress` coding standards.
* Automated fixes for for files where possible.

PR #909, opened 2024-09-28
PHPCS: Enable NoReservedKeywordParameterNames sniff

obenland:
## Proposed changes:

* Replaced parameter names of reserved keywords with names that are not reserved.
* Removed sniff exception.

See #905.

### DIFF
--- a/includes/activity/class-activity.php
+++ b/includes/activity/class-activity.php
@@ -145,49 +145,49 @@ class Activity extends Base_Object {
 	 *
 	 * @see https://www.w3.org/TR/activitypub/#object-without-create
 	 *
-	 * @param array|string|Base_Object|Link|null $activity_object
+	 * @param array|string|Base_Object|Link|null $data Activity object.
 	 *
 	 * @return void
 	 */
-	public function set_object( $activity_object ) {
+	public function set_object( $data ) {
 		// convert array to object
-		if ( is_array( $activity_object ) ) {
-			$activity_object = self::init_from_array( $activity_object );
+		if ( is_array( $data ) ) {
+			$data = self::init_from_array( $data );
 		}
 
 		// set object
-		$this->set( 'object', $activity_object );
+		$this->set( 'object', $data );
 
-		if ( ! is_object( $activity_object ) ) {
+		if ( ! is_object( $data ) ) {
 			return;
 		}
 
 		foreach ( array( 'to', 'bto', 'cc', 'bcc', 'audience' ) as $i ) {
-			$this->set( $i, $activity_object->get( $i ) );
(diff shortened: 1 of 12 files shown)


includes/activity/class-activity.php
```diff
@@ -143,49 +143,49 @@ class Activity extends Base_Object {
 	 *
 	 * @see https://www.w3.org/TR/activitypub/#object-without-create
 	 *
-	 * @param string|Base_Objectr|Link|null $object
+	 * @param array|string|Base_Objectr|Link|null $activity_object
```
pfefferle:
```suggestion
	 * @param array|string|Base_Object|Link|null $activity_object
```

pfefferle: because it can be an array, string, ... maybe `$data` would be much better?!

includes/handler/class-follow.php
```diff
@@ -72,23 +72,23 @@ public static function handle_follow( $activity ) {
...
-	 * @param Activitypub\Model\Follower $follower The Follower object
+	 * @param string                     $actor           The Actor URL
+	 * @param array                      $activity_object The Activity object
+	 * @param int                        $user_id         The ID of the WordPress User
+	 * @param Activitypub\Model\Follower $follower        The Follower object
 	 *
 	 * @return void
 	 */
-	public static function send_follow_response( $actor, $object, $user_id, $follower ) {
+	public static function send_follow_response( $actor, $activity_object, $user_id, $follower ) {
```
pfefferle, line 82: because this is only an array, I would call it `$activity` to be consistent with the other code you changed above?!

integration/class-webfinger.php
```diff
@@ -23,49 +23,48 @@ public static function init() {
 	/**
 	 * Add WebFinger discovery links
 	 *
-	 * @param array   $array    the jrd array
-	 * @param string  $resource the WebFinger resource
-	 * @param WP_User $user     the WordPress user
+	 * @param array   $jrd    the jrd array
+	 * @param string  $entity the WebFinger resource
```
pfefferle:
```suggestion
	 * @param string  $uri the WebFinger resource
```

in the user class you named it `$uri` so I would use that here too, because of consistency.

includes/handler/class-follow.php
obenland: I would have liked to do that but there's a `$activity` variable set in line 108. Any preference?

pfefferle: good point! then let's keep this one as is for now

pfefferle (dismissed):

pfefferle (approved):

PR #911, opened 2024-09-29
PHPCS: Enable UnusedFunctionParameter sniff

obenland:
## Proposed changes:

* Removes unused function parameters and their docs.
* Adds a unit test for `at_title` shortcode to make sure that still works.
* Removes sniff exception.

See #905.

### DIFF
--- a/includes/class-shortcodes.php
+++ b/includes/class-shortcodes.php
@@ -392,7 +392,6 @@ class Shortcodes {
 
 		$datetime   = \get_post_datetime( $item );
 		$dateformat = \get_option( 'date_format' );
-		$timeformat = \get_option( 'time_format' );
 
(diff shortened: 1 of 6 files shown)


pfefferle:
won't that result in at least PHP warnings, because the functions are always called with these params, if they are used or not?

obenland:
Yeah, maybe this needs a different approach

obenland:
@pfefferle Can you double-check? I think they should just be ignored. I've also not been able to generate a warning in my testing.

pfefferle:
OK, the shortcode API says:

> Three parameters are passed to the shortcode callback function. You can choose to use any number of them including none of them.

pfefferle (approved):

Issue #71, opened 2019-10-21
Probably Problems with Space in Username


waweic:
My Wordpress-Username has the following format: "My Username"
Thus, the generated Fediverse ID is "My Username@mydomain.tld", whereas the link to it is "https://mydomain.tld/author/my-username/".
My Problem is now that Mastodon won't follow this account and I suspect the space in the username to be the problem. Would it be possible to also user the "my-username"-part in the Fediverse ID? Or are there any other solutions? Am I wrong with my assumption that there may not be a space in the Fediverse ID?
Thanks for the great project, hope to hear from you soon

skariko:
Hello! Same problem here, did you find how to fix it?

mediaformat:
Technically you can search for a profile with a space in the username using [webfinger](https://webfinger.net/) and receive a valid response.
And you can fetch a profile using the links provided by webfinger.

The bug is with Mastodon, when pasting the author URL into Mastodon an error is returned:

> **Searching failed**
Server did not return an OK response code
(422: Validation failed: Username must contain only letters, numbers and underscores)

Searching Mastodon with webfinger seems to fail silently. Pleroma as well.

Zap finds the profile.

skariko:
> 

Yeah I see: https://webfinger.net/lookup/?resource=marco+calamari%40www.lealternative.net

I think that a workaround will be create a new author without space and give it all the posts :)

Issue #28, opened 2019-02-12
Make username (webfinger) customizable


merinleona:
I am not a fan of the fact that this used my loginname for wordpress as the user handle. I very much prefer Pterotypes approach to just use blog@. Now i deinstalled the plugin again, but i donÄt know how to delete the Account created on the Fediverse since I don't have access to it. How do I do that?
Thank you

pfefferle:
I have chosen the username, because this supports multiple users per blog. What are your concerns using the username?

You have no account in the „fediverse“ it all works decentral. If you reinstall the plugin, your account is gone.

pfefferle:
And, as far as I know pterotype also works this way...

merinleona:
i deleted both plugins. The Account created by pterotype (blog@my-domain) can't be find anymore. But the one created by Activitypub still shows up when I search for my domain on Mastadon. So it has to been entered into some registry or some part of the plugin does not deacitvate when you deinstall.

pfefferle:
Ah, I see... This is a problem with mastodons caching system... 😞

pfefferle:
Or it is, because pterotype also uses the username...

merinleona:
So it will disappear after a bit? I thought it had something to do with the plugin since I deinstalled both at the same time and the pterotype account vanished instantly.

merinleona:
I deinstalled pterotype as well

pfefferle:
You can try it on webfinger.net

If you provide me your blog URL, I will have a look.

merinleona:
I would not mind if it was my Display name btw. But I don't like it being my Admin login name.. it looks stupid^^ ( not your problem me picking a stupid name and being vane^^)

merinleona:
It's snowshaven.space

pfefferle:
Everything is deleted properly, at least as far as I can see (no more visible API endpoints). So it really seems a caching issue.

Suitbert:
I would like an input field, where I would enter a name of my choice. Don't like it, to be forced to use WP username. Would that be possible?

merinleona:
Thank you for your time :)

pfefferle:
@Suitbert because of any security reasons and/or other concerns or simply because of a "stupid" login name?

In the second case the easiest solution would be to change the login name: https://www.wpbeginner.com/wp-tutorials/how-to-change-your-wordpress-username/

pfefferle:
@Suitbert but to answer you question: Sure it is possible.

[...]

PR #915, opened 2024-09-30
Make username configurable

### LINKED ISSUE #141: Usernames vs. Nicknames
I would like to suggest that it is not a good idea to be publicly using the username as part of the URI for the ActivityPub plugin.  Using the username is creating a potential vulnerability on sites that wish to use this plugin.  In Wordpress usernames are the login ID's for user accounts.  This means that the URI is exposing the user ID of accounts publicly.

Instead it would be better to use Nickname, and handle the lookup to map them to the Username.  This way the account login ID is kept within the system and not made public.

pfefferle:
At least for the WebFinger ID

Fixes #141 #28 #71 

Screenshot:

Screenshot 2024-09-30 at 12 58 40

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -310,28 +310,9 @@ class Admin {
 				'show_in_rest'      => true,
 				'default'           => Blog::get_default_username(),
 				'sanitize_callback' => function ( $value ) {
-					// hack to allow dots in the username
-					$parts     = explode( '.', $value );
-					$sanitized = array();
+					$sanatized = self::sanatize_identifier( $value );
 
-					foreach ( $parts as $part ) {
-						$sanitized[] = \sanitize_title( $part );
-					}
-
-					$sanitized = implode( '.', $sanitized );
-
-					// check for login or nicename.
-					$user = new WP_User_Query(
-						array(
-							'search'         => $sanitized,
-							'search_columns' => array( 'user_login', 'user_nicename' ),
-							'number'         => 1,
-							'hide_empty'     => true,
-							'fields'         => 'ID',
-						)
-					);
-
-					if ( $user->results ) {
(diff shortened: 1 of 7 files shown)


includes/class-admin.php
```diff
@@ -758,4 +761,47 @@ public static function dashboard_glance_items( $items ) {
...
+				'fields'         => 'ID',
+			)
+		);
+
+		if ( $user->get_results() ) {
+			return new \WP_Error(
+				'identifier_exists',
+				\__( 'This identifier is already in use.', 'activitypub' )
+			);
+		}
```
akirk: This doesn't check whether the same identifier was already set for another username.

includes/collection/class-users.php
```diff
@@ -96,9 +98,14 @@ public static function get_by_username( $username ) {
 				'meta_query'  => array(
 					'relation' => 'OR',
 					array(
-						'key'     => 'activitypub_user_identifier',
+						'key'     => 'activitypub_identifier',
+						'value'   => $username,
+						'compare' => '=',
+					),
+					array(
```
akirk, line 105: It's probably not worth addressing but it could happen that you assign the identifier to a non-existent user at the time, and then register that username afterwards. In this scenario I believe the result here is undefined since you just pick the first result in the code below.

akirk (commented):
I am personally not sold on the need of changing the username but looks like there are many requests. It looks like this does the requested job.

mattwiebe:
This feels like unnecessary complexity? Can't these users just use the Blog user, which is already configurable, instead?

pfefferle:
@mattwiebe the renaming itself is pretty straight forward, but the collision detection is a bit complex. I think I will postpone it for now.

PR #917, opened 2024-10-01
Improve Auto-Post-Type-Detection

pfefferle:
Added checks:

* Use "Note" if "title" is empty
* Use "Note" if "content" is below 400 chars

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -30,6 +30,7 @@ require_once __DIR__ . '/includes/functions.php';
  */
 \defined( 'ACTIVITYPUB_REST_NAMESPACE' ) || \define( 'ACTIVITYPUB_REST_NAMESPACE', 'activitypub/1.0' );
 \defined( 'ACTIVITYPUB_EXCERPT_LENGTH' ) || \define( 'ACTIVITYPUB_EXCERPT_LENGTH', 400 );
+\defined( 'ACTIVITYPUB_NOTE_LENGTH' ) || \define( 'ACTIVITYPUB_NOTE_LENGTH', 400 );
 \defined( 'ACTIVITYPUB_SHOW_PLUGIN_RECOMMENDATIONS' ) || \define( 'ACTIVITYPUB_SHOW_PLUGIN_RECOMMENDATIONS', true );
 \defined( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS' ) || \define( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS', 3 );
 \defined( 'ACTIVITYPUB_HASHTAGS_REGEXP' ) || \define( 'ACTIVITYPUB_HASHTAGS_REGEXP', '(?:(?<=\s)|(?<=<p>)|(?<=<br>)|^)#([A-Za-z0-9_]+)(?:(?=\s|[[:punct:]]|$))' );
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -643,9 +643,16 @@ class Post extends Base {
 			return \ucfirst( $post_format_setting );
 		}
 
-		$has_title = post_type_supports( $this->wp_object->post_type, 'title' );
+		$has_title = \post_type_supports( $this->wp_object->post_type, 'title' );
+		$content   = \get_post_field( 'post_content', $this->wp_object->ID );
+		$content   = \wp_strip_all_tags( $content );
 
-		if ( ! $has_title ) {
+		// Check if the post has a title.
+		if (
+			! $has_title ||
+			! $this->wp_object->post_title ||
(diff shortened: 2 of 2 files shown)


includes/transformer/class-post.php
```diff
@@ -642,9 +642,16 @@ protected function get_type() {
...
+		$has_title = \post_type_supports( $this->wp_object->post_type, 'title' );
+		$content   = \get_post_field( 'post_content', $this->wp_object->ID );
+		$content   = \wp_strip_all_tags( $content );
+
+		// Check if the post has a title.
+		if (
+			! $has_title ||
+			! $this->wp_object->post_title ||
+			\strlen( $content ) <= ACTIVITYPUB_NOTE_LENGTH
+		) {
```
obenland, line 653: Would it make sense to add a check for whether the post has the `aside` post format to determine whether it's a note? (not sure what constitutes a note)

activitypub.php
```diff
@@ -29,6 +29,7 @@
  */
 \defined( 'ACTIVITYPUB_REST_NAMESPACE' ) || \define( 'ACTIVITYPUB_REST_NAMESPACE', 'activitypub/1.0' );
 \defined( 'ACTIVITYPUB_EXCERPT_LENGTH' ) || \define( 'ACTIVITYPUB_EXCERPT_LENGTH', 400 );
+\defined( 'ACTIVITYPUB_NOTE_LENGTH' ) || \define( 'ACTIVITYPUB_NOTE_LENGTH', 400 );
```
obenland, line 32: Can you help me understand the benefit of defining these as constants vs offering customization through a filter?

includes/transformer/class-post.php
pfefferle: We do that already: https://github.com/Automattic/wordpress-activitypub/blob/00a57c39a4e844949bc4f64a45714d1f21d9645d/includes/transformer/class-post.php#L667

`Article` is only used for the `Standard` Post-Format, every other will be "translated" to `Note`. This current change is only for cases, where users don't use custom-post-formats and every post defaults to `Article`.

includes/transformer/class-post.php
```diff
@@ -642,9 +642,16 @@ protected function get_type() {
 			return \ucfirst( $post_format_setting );
 		}
 
-		$has_title = post_type_supports( $this->wp_object->post_type, 'title' );
-
-		if ( ! $has_title ) {
+		$has_title = \post_type_supports( $this->wp_object->post_type, 'title' );
+		$content   = \get_post_field( 'post_content', $this->wp_object->ID );
+		$content   = \wp_strip_all_tags( $content );
```
```suggestion
		$content   = \wp_strip_all_tags( $this->wp_object->post_content );
```
Would this work?

pfefferle: sure!

activitypub.php
pfefferle: It's easier to configure it as a "hoster", but I am happy to add a filter or to change it to only use a filter.

obenland: I don't know much about how this plugin is used, so I'm just trying to understand

pfefferle: I tried to mimic the Mastodon server setup a bit, where you can customize the length of posts through an env var.

pfefferle: https://fouquet.me/2023/09/22/mastodon-char-limit

includes/transformer/class-post.php
```diff
@@ -642,9 +642,15 @@ protected function get_type() {
 			return \ucfirst( $post_format_setting );
 		}
 
-		$has_title = post_type_supports( $this->wp_object->post_type, 'title' );
-
-		if ( ! $has_title ) {
+		$has_title = \post_type_supports( $this->wp_object->post_type, 'title' );
```
mattwiebe, line 645: This will only check if the post_type supports having a title, not if there is one. Probably want something more like this?

```suggestion
		$has_title = \post_type_supports( $this->wp_object->post_type, 'title' ) && ! empty( get_the_title( $this->wp_object );
```

mattwiebe:
Do we have a screenshot or demo of what the `Article` type renders like in Mastodon these days?

includes/transformer/class-post.php
pfefferle: I check for a title some lines below (line 651).

pfefferle:
@mattwiebe here is a screenshot of an Article:

Screenshot 2024-10-10 at 20 21 23

mattwiebe:
Cool! Here's what it shows up as in Ivory (it does not seem to like the mp3)

Screenshot 2024-10-10 at 18 23 51

includes/transformer/class-post.php
mattwiebe: Ah! Should have kept reading 😅

mattwiebe (approved):
Nice to bring titles into the fediverse!

PR #920, opened 2024-10-02
Fix: Update() activity ID is not unique

### LINKED ISSUE #871: Update() activity ID is not unique
### Quick summary

WordPress instance sends Update(Actor) activity with ID in the form of `https://wordpress.example/author/dev/#update`

ActivityPub [requires](https://www.w3.org/TR/activitypub/#obj-id) IDs to be globally unique:

>all objects distributed by the ActivityPub protocol MUST have unique global identifiers, unless they are intentionally transient

### Steps to reproduce

Update profile info?

Activity ID to be unique

### What actually happened

Activity ID is always `https://wordpress.example/author/dev/#update`

### Impact

All

### Available workarounds?

There is no user impact

### Logs or notes

_No response_

pfefferle:
Fixes #871

### DIFF
--- a/includes/class-activity-dispatcher.php
+++ b/includes/class-activity-dispatcher.php
@@ -144,7 +144,6 @@ class Activity_Dispatcher {
 
 		// build the update
 		$activity = new Activity();
-		$activity->set_id( $user->get_url() . '#update' );
 		$activity->set_type( 'Update' );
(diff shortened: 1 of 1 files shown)


mattwiebe (approved):
Was not aware of this!

PR #922, opened 2024-10-02
Fix link regex

mattwiebe:
We are getting a bunch of errors like this on wpcom:

```
Warning: Undefined array key "host" in /wp-content/plugins/activitypub/dev/includes/class-link.php on line 84
```

This is because the previous regex was matching a string like `www foo` when it shouldn't be. I don't understand why the previous negated character class didn't work, but using the `\S` class is equivalent to `^\s`. I simplified the `https?` match a bit while I was in there.

## Proposed changes:
* improve `ACTIVITYPUB_URL_REGEXP` 

### Other information:

## Testing instructions:

* Go to '..'
*


includes/class-link.php
```diff
@@ -62,7 +62,7 @@ public static function replace_with_links( $result ) {
 			$result[0] = 'https://' . $result[0];
 		}
 		$parsed_url = \wp_parse_url( html_entity_decode( $result[0] ) );
-		if ( ! $parsed_url ) {
+		if ( ! $parsed_url || empty( $parsed_url['host'] ) ) {
 			return $result[0];
```
mattwiebe, line 66: This might not be needed but seemed like useful defensiveness.

pfefferle (changes requested):
@mattwiebe can you add the false positives to the test case?

mattwiebe:
> @mattwiebe can you add the false positives to the test case?

Will do! Having a little trouble with the tests (again) but I'll circle back with fresh energy tomorrow

pfefferle (approved):

PR #931, opened 2024-10-08
Add "visibility" feature

### LINKED ISSUE #923: Feature Request: Add filter to restrict ActivityPub not only per post_type but also per an individual post
### What

I would like to be able to offer other plugins the opportunity to disable ActivityPub for specific posts.

### Why

Let's take an event organizer with a WordPress site that displays a calendar (these cases are actually not fictional).

#### Use case 1

Sometimes the calendar serves two purposes: 
1. showing a plan when the main and only venue of the organizer is occupied.
2. informing the audience about public events.

Whereas **1.** is and may be visible publicly on the websites front-end events that are not part of **2.** should not be federated via the ActivityPub plugin.

#### Use case 2

Sometimes the event organizer has events in his venue he wants to advertise, but are not his own, these events even exist in the Fediverse. So he uses a plugin to import them, which actually means inserting a new post to the posts table (the post gets marked as e.g. 'from_federation', similar as to the way federated comments work).

These events are only treated as cache, querying them on the local URI would show the event on the local WordPress site. Using the the ActivityPub accept headers, a redirect to the original 'id' would be implemented.

The ActivityPub plugin should not create an internal copy of that, but it may announce the original event.

### How

I would offer providing a pull request that solves this.

My first idea was to add a boolean post_meta field `activitypub_disabled` to the post. Or use the `activitypub_status` and introduce a value `disabled`.

### Further thoughts

In general I would like to see more per post settings, for instance being added to the "Fediverse" column that currently handles the Content Warnings. For instance a settings could be to choose custom content template, having in mind many use the same post type for micro-blogging (Note) and macro-blogging (Article with summary/excerpt).

### LINKED ISSUE #929: Feature Request: Like and Dislike Interactions (Comments)
### What

This ticket is a request to add the ability for users on the fediverse to be able to engage with interactions made by other users on the fediverse to the wordpress post: to like and dislike those posts

### Why

The goal would be to allow for visitors to my wordpress blog to be able to upvote and downvote ActivityPub Posts made about my blog post, wihch appear as threaded comments below my blog post. This would allow for a typical UX like that of digg, reddit, lemmy, etc.

### How

* ActivityPub Like -> upvote
* ActivityPub Dislike -> downvote

### Example

Let's consider an example.  Earlier this year, I posted an article about Lemmy on my wordpress blog:

 * https://tech.michaelaltfield.net/2024/03/04/lemmy-fediverse-gdpr/

This article received a lot of dialog across the fediverse. Here's one post about it on one of the most popular lemmy instances, which has 267 comments (at the time of writing). In this example, I truncated it to only 9 comments for brevity.

 * https://lemmy.world/post/12727137

What I want is for each of these comments (which are ActivityPub Notes) to appear in the "Comments" section below my blog, and for my blog visitors to be able to upvote and downvote the comments directly on my site using their own federated account.

| [screenshot of a wordpress article with no comments](https://i.postimg.cc/R4yLqtqN/wordpress-activitypub-votes1.jpg) | [screenshot of a lemmy post with a lot of nested comments](https://i.postimg.cc/yVzhyzmK/wordpress-activitypub-votes2.jpg) | [screenshot of a wordpress article with a lot of nested comments after the article](https://i.postimg.cc/pPjYjdnk/wordpress-activitypub-votes3.jpg) |
|:--:|:--:|:--:|
| 1. My Blog Post (without ActivityPub comments) | 2. Comments on Lemmy | 3. My Blog Post (with ActivityPub comments) |

The screenshots above show:

1. (Left) The article on my wordpress blog with no comments
2. (Middle) A post on lemmy that links to my wordpress blog. There are a lot of nested lemmy comments. Each comment has an upvote and downvote button.
3. (Right) The same article on my wordpress blog as [1], but this time the ActivityPub comments in [2] are shown on my wordpress blog, just below the article. Every comment has an upvote and downvote button, so readers of my blog can like or dislike each federated ActivityPub Note directly. There is also set of filter buttons at the top, where users can sort the order that the comments appear.

pfefferle:
Fix #923

This PR allows users to change the visibility of a post.

Options:

* `Public` -> default
* `Quiet public` -> This behaves exactly like public, except the post will not appear in live feeds or hashtags, explore, or Mastodon search, even if you are opted-in account-wide.
* `Do not federate` -> Do not federate this post.

Screenshot 2024-10-08 at 12 30 31

Todo: Use constants


includes/functions.php
```diff
@@ -400,6 +400,31 @@ function is_activitypub_request() {
...
+/**
+ * Check if a post is disabled for ActivityPub.
+ *
+ * @param mixed $post The post object or ID.
+ *
+ * @return boolean True if the post is disabled, false otherwise.
+ */
+function is_post_disabled( $post ) {
+	$post       = \get_post( $post );
+	$return     = false;
```
mattwiebe: I would name this `$disabled` instead

includes/class-blocks.php
```diff
@@ -45,6 +45,16 @@ public static function register_postmeta() {
...
 				)
 			);
+			\register_post_meta(
+				$post_type,
+				'activitypub_content_visibility',
+				array(
+					'show_in_rest'      => true,
+					'single'            => true,
+					'type'              => 'string',
+					'sanitize_callback' => 'sanitize_text_field',
```
mattwiebe: We should probably define our own callback for this, checking to be sure it's one of our `ACTIVITYPUB_POST_VISIBILITY_LOCAL` etc constants. I'd hoped that `register_post_meta` would accept an `enum` type and went down a merry rabbit hole after an LLM hallucinated that it did 😅

src/editor-plugin/plugin.js
```diff
@@ -13,10 +14,26 @@ const EditorPlugin = () => {
...
+			<Icon icon={ icon } />
+			{text}
+		</Text>
+	);
+
 	return (
 		<PluginDocumentSettingPanel
 			name="activitypub"
-			title={ __( 'Fediverse', 'activitypub' ) }
+			title={ __( '⁂ Fediverse', 'activitypub' ) }
```
mattwiebe, line 36: 🙌🏻

mattwiebe (commented):
Poked around a bit, had a few thoughts, but the React looks like a seasoned pro!

pfefferle:
@mattwiebe 

> Poked around a bit, had a few thoughts, but the React looks like a seasoned pro!

😍

mattwiebe (commented):
A couple of questions:

1. Is this going to break Outbox pagination? Should we adjust the query there to exclude hidden posts?
2. Do we ever even want to save `public`? Should it be more of a default `null` case?

pfefferle:
@mattwiebe 

1. Mastodon does not really support the Outbox, so we could care about that later.
2. Good point. Will WordPress ignore the `null` value or will it persist the `null`?

pfefferle:
Updated icons! See: https://mastodon.social/@mattwiebe/113289210941629136

Screenshot 2024-10-11 at 21 56 57

Menrath:
@pfefferle I would like to add a feature here: could we add filters, so that other plugins might override the default visibility for a post? 

Concrete why: some event plugins have a option per event to mark it as "unlisted", which would fit perfectly here.

pfefferle:
@Menrath and @mattwiebe done!

mattwiebe (approved):
Had a run through, the "Public" type is no longer saved, everything looks in order! 🚢

PR #934, opened 2024-10-09
Only validate POST params and do not fall back

pfefferle:
Do not fall back to GET or other params if the POST-JSON is empty!

thanks @obenland

### DIFF
--- a/includes/rest/class-server.php
+++ b/includes/rest/class-server.php
@@ -8,6 +8,7 @@
 namespace Activitypub\Rest;
 
 use WP_Error;
+use WP_REST_Server;
 use WP_REST_Response;
 use Activitypub\Signature;
 use Activitypub\Model\Application;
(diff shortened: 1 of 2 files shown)


includes/rest/class-server.php
```diff
@@ -181,4 +183,32 @@ public static function validate_activitypub_requests( $response, $handler, $requ
...
+	 * @param string[]        $order   Array of types to check, in order of priority.
+	 * @param WP_REST_Request $request The request object.
+	 *
+	 * @return string[] The modified order of types to check.
+	 */
+	public static function request_parameter_order( $order, $request ) {
+		$route = $request->get_route();
+
+		// Check if it is an activitypub request and exclude webfinger and nodeinfo endpoints.
+		if ( ! \str_starts_with( $route, '/' . ACTIVITYPUB_REST_NAMESPACE ) ) {
```
obenland, line 199: `str_starts_with` was introduced in PHP 8, so we should update the min required PHP version for the plugin or replace it with `strpos`

pfefferle: we use a Polyfill for it: https://github.com/Automattic/wordpress-activitypub/blob/d532f8be06f0e3539f30ba1500dab95d391a99c6/includes/compat.php#L8

mattwiebe:
What was the bug here?

pfefferle:
@mattwiebe The current code falls back to `GET` params if there is no JSON param in the `POST` for validation and that might cause issues.

Example: Someone sends a `Create` with no `id` in the POST-JSON, but with an `?id={id}` as GET-Param. The code request is valid, because it falls back to GET, but because we further use `$request->get_json_params();`, the `id` is missing. This way, it is possible to bypass validation and that could cause issues later in the queue. With this PR, the validation follows the same rules as the following handling of the POST.

PR #940, opened 2024-10-14
Fixfor urls and hashtags in profiles are not converted

danielhuesken:
The removings for https://github.com/mastodon/mastodon/pull/28629 that should be only in post summery also removed it in profiles. This PR activates it only for Profiles again.


pfefferle (approved):

pfefferle:
🫣 oops

PR #944, opened 2024-10-18
Add custom Preview for "Fediverse"

pfefferle:
Screenshot 2024-10-18 at 10 52 23

* [X] Add menu-item
* [x] Add logic to open preview
* [x] Implement preview


pfefferle:
@mattwiebe I was in the mood for JavaScript ☺️

pfefferle:
Preview

Screenshot 2024-10-21 at 10 33 52

Menrath:
@pfefferle Wow! I love that! :heart_eyes: 

So often I got confronted with "but where do I see my Fediverse post now, what does it look like" (e.g., see this article https://event-federation.eu/2024/01/17/using-the-wordpress-activitypub-plugin-what-can-go-wrong/).

I wonder if I can adopt this for events as well. Could it be possible to hook into this in a clean way, to also add what it would look like on a typical Gancio/Mobilizon like platform? Or I would love inserting tabs at the top, because for events it's also important, what they will look like for instance on Mastodon.

I was just working on improving the underlying problem, but starting from the admin page. I was also just thinking of adding something with a JavaScript and modals as well in our Event Bridge plugin ([Screenshot](https://code.event-federation.eu/Event-Federation/wordpress-activitypub-event-bridge/attachments/300f138d-748d-415a-94da-35106fa095d7)).

pfefferle:
@Menrath nice idea! I will see how we can extend that a bit, to have a simple templating mechanism!?!

includes/class-activitypub.php
```diff
@@ -247,6 +252,7 @@ public static function template_redirect() {
 	 */
 	public static function add_query_vars( $vars ) {
 		$vars[] = 'activitypub';
+		$vars[] = 'preview';
```
mattwiebe, line 256: I don't know if we should use a core variable here, it already gets used for post previews in a `?p=1&preview=true` format. Are we leveraging something out of that? Or could we use our own query var here like `activitypub_preview`?

src/editor-plugin/plugin.js
```diff
@@ -60,4 +61,43 @@ const EditorPlugin = () => {
...
 
+function onActivityPubPreview() {
+	const previewLink = select( 'core/editor' ).getEditedPostPreviewLink();
+	const fediversePreviewLink = addQueryArgs( previewLink, { activitypub: 'true' } );
+
+	window.open( fediversePreviewLink, '_blank' );
+}
+
+const EditorPreview = () => {
+	const el = React.createElement;
```
mattwiebe: While directly calling `createElement` will work, it's not a typical way of working. If we look at the [devdocs page for @wordpress/primitives](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-primitives/), I see a different way. Although maybe the `text` node makes it hard to do that?

src/editor-plugin/plugin.js
```diff
@@ -60,4 +61,43 @@ const EditorPlugin = () => {
...
+	const svg = wp.primitives.SVG;
+	const text = el(
+		'text',
+		{ x: 7.5, y: 15, fill: '#000' },
```
mattwiebe: I think we'd want the `currentColor` keyword for `fill`

mattwiebe (commented):
Left a few comments, this is gonna be awesome 😁

src/editor-plugin/plugin.js
pfefferle: the question is, if this is really the right way... I experimented a bit, because I wanted to know if it is possible to load a text into an SVG container. Maybe it is better practice to load the SVG instead: https://symbol.fediverse.info/

includes/class-activitypub.php
pfefferle: I thought this might be the best way to go, because it already provides everything needed (loads the correct post or revision) and we only have to hijack the output: https://github.com/Automattic/wordpress-activitypub/pull/944/files#diff-661e7ae5b9a59de0559c8b57927dd88bac7bdfa00d0c1620112eff406a14a1daR65

src/editor-plugin/plugin.js
pfefferle: what about this instead:

Screenshot 2024-10-24 at 10 43 05

includes/class-activitypub.php
mattwiebe: Sounds good!

src/editor-plugin/plugin.js
mattwiebe: I like keeping it textual 👍

mattwiebe (changes requested):
When I have a draft and hit "Fediverse preview" I get our workaround for publish-then-set-to-draft:

Screenshot 2024-10-28 at 15 29 21

pfefferle:
Yes, this is still todo! I thought about a preview const!?!

pfefferle:
added `ACTIVITYPUB_PREVIEW` const!

mattwiebe (approved):
Tested again, worked with a draft as expected, I think this is good to ship and iterate on!

PR #949, opened 2024-10-23
Add Fediverse-Files video to the readme

pfefferle:
What about adding the fediverse files video to the readme?

I think it might help to understand the idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ The ActivityPub protocol is a decentralized social networking protocol based upo
 
 Enter the fediverse with **ActivityPub**, broadcasting your blog to a wider audience! Attract followers, deliver updates, and receive comments from a diverse user base of **ActivityPub**\-compliant platforms.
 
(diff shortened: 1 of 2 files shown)


davemart-in:
Totally optional, but if you want, you could embed a screenshot image of the video to draw more attention to it like this:

[image](https://www.youtube.com/watch?v=QzYozbNneVc)

We could even add the YouTube chrome at the bottom if you want.

Again, totally optional.

pfefferle:
@davemart-in WordPress.org will directly embed the video, so I think there is no need to add an additional screenshot!?!

pfefferle:
@davemart-in ah, you might mean the markdown file... maybe we add the image there to draw some attention to it... or maybe we find a way to embed it there too!!!

davemart-in:
Yeah, sorry. I wasn't specific in my comment. Not sure you can embed a video in the Readme file, but you can use an image screenshot.

PR #950, opened 2024-10-23
Provide contextual fallback for dynamic blocks

kraftbj:
When using blocks like Followers and Follow Me, previously, it reported that it would be empty in non-author contexts in the editor. In realty, this is only true if the blog-level profile is not enabled.

## Proposed changes:

* Conditionally include the empty disclaimer to only when the site profile is disabled.

### Other information:

## Testing instructions:

* With AP set to only author profiles, attempt to add a Follow Me block set to Dynamic User. See the fallback text that mentions it'll be empty.
* With AP set to have the blog and author profiles, do the same as above. See the fallback text without the empty disclaimer.


mattwiebe:
@kraftbj thanks so much for the PR! I updated it a bit to make proper use of i18n, because apparently adding a `/* translators */` comment wasn't enough 🤦

mattwiebe (approved):
🎉

PR #952, opened 2024-10-23
Add wp-env support

kraftbj:
It was helpful when testing my previous PR to test it out locally and wp-env was nice and simple. The AP plugin prefers a pretty permalink setting, so this adds wp-env as a "supported" local setup, including auto-setting the permalink when using `npm run env-start`

## Proposed changes:

* Adds .wp-env.json file.
* Adds script to package.json to start and stop the env.

### Other information:

## Testing instructions:

* `npm install && npm run env-start` does it work? Open the site (default user/pass: admin/password). Are permalinks already set?"
* `npm run env-stop` does it stop?


kraftbj:
We can probably chain things with `npm run dev` to auto-start the env too, but I defer to y'all who work on this daily on that one.

package.json
```diff
@@ -12,7 +12,9 @@
   "scripts": {
     "dev": "wp-scripts start",
     "build": "wp-scripts build",
-    "readme": "grunt wp_readme_to_markdown"
+    "readme": "grunt wp_readme_to_markdown",
+	"env-start": "wp-env start && wp-env run cli wp rewrite structure '/%year%/%monthnum%/%postname%/'",
```
pfefferle: @kraftbj can you use the same indents as the rest of the file (4 spaces)? I am also fine in switching to tabs, but it should be consistent.

.wp-env.json
```diff
@@ -0,0 +1,4 @@
+{
```
pfefferle, line 1: @kraftbj can you add `.wp-env.json` to the .distignore file, so that it will not be pushed to WordPress.org?

package.json
kraftbj: Will do. #BlameIDE

.wp-env.json
kraftbj: 940b1e0

package.json
kraftbj: 940b1e0

kraftbj:
I also alpha-sorted `.distignore` just for the sake of my own brain being able to parse the list faster.

pfefferle (approved):
thanks @kraftbj

PR #957, opened 2024-10-25
Fallback to blog-user if user is disabled

### LINKED ISSUE #948: Bug? Posts by Authors without ActivityPub privilege are not shared by Blog Profile
### Quick summary

(I'm not sure if this should be considered a bug or if it is working exactly how it is designed.)

My goal is that I have a WordPress Author that I'd prefer not to have an ActivityPub profile, but still want the Blog Profile to share posts by that Author. If I turn Disable ActivityPub for that WordPress Author, but have the Blog Profile turned on in the ActivityPub plugin's settings, the Blog Author does not share/post posts by that WordPress Author. 

### Steps to reproduce

1. In the ActivityPub plugin's settings: Enable Author-Profiles
2. In the ActivityPub plugin's settings: Enable Blog Profile
3. In Users panel in WordPress Admin, disable ActivityPub for a single user
4. Publish a post by that user

The Blog Profile would still share the post via ActivityPub.

### What actually happened

It did not.

### Impact

All

### Available workarounds?

### Logs or notes

Perhaps the user expectation is that this is working properly. If I disable an Author from ActivityPub, perhaps it is reasonable to conclude I don't want posts authored by the user to be shared at all. So perhaps this would need to be a new setting?

pfefferle:
Fallback to the blog-user if enabled and the specific user is disabled.

fix #948


tests/test-class-activitypub-activity-dispatcher.php
```diff
@@ -178,6 +178,65 @@ function ( $value, $type ) {
...
+		$followers = array( 'https://example.com/author/jon' );
+
+		add_filter(
+			'activitypub_is_user_type_disabled',
+			function ( $disabled, $type ) {
+				return false;
+			},
+			10,
+			2
+		);
```
obenland:
```suggestion
		add_filter( 'activitypub_is_user_type_disabled', '__return_false' );
```
WordPress has a few of these "frequently used" functions to make that easier.

pfefferle: Oh no, shame on me! I used them so many times in the other PR, but changed this function so many times that I didn't even notice how it ended up 😂

Thanks!

pfefferle:
@mattwiebe what do you think about this?

mattwiebe (approved):
Looks good!

cdevroe:
🙌

PR #959, opened 2024-10-28
Work around race condition when publishing local only posts

pfefferle:
It seems that the meta-information might not always be stored when saving a new post.

See discussion here: https://sueden.social/@freuwesen/113378039196820044 (it is in german but mastodon can translate).

## Proposed changes:

* clear audience fields, so that post will not be listed in any timeline when accidentally shared
* check a second time, right before federating it

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -409,10 +409,7 @@ function is_activitypub_request() {
  * @return boolean True if the post is disabled, false otherwise.
  */
 function is_post_disabled( $post ) {
-	$post = \get_post( $post );
-
-	\clean_post_cache( $post );
-
+	$post       = \get_post( $post );
 	$disabled   = false;
 	$visibility = \get_post_meta( $post->ID, 'activitypub_content_visibility', true );
 
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -409,7 +409,10 @@ function is_activitypub_request() {
  * @return boolean True if the post is disabled, false otherwise.
  */
 function is_post_disabled( $post ) {
-	$post       = \get_post( $post );
+	$post = \get_post( $post );
+
+	\clean_post_cache( $post );
+
 	$disabled   = false;
 	$visibility = \get_post_meta( $post->ID, 'activitypub_content_visibility', true );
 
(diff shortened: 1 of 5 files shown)


includes/functions.php
```diff
@@ -409,7 +409,10 @@ function is_activitypub_request() {
  * @return boolean True if the post is disabled, false otherwise.
  */
 function is_post_disabled( $post ) {
-	$post       = \get_post( $post );
+	$post = \get_post( $post );
+
+	\clean_post_cache( $post );
```
mattwiebe: I think that this will cause load problems more than it will solve postmeta cache problems? Is this actually doing anything useful?

Eg we could clean the cache up in `Dispatcher::send_post` above, before this check gets called. But even there I have my doubts that this is solving a problem the way it should be solved.

mattwiebe (commented):
Could we also maybe add a slight delay to the scheduler to give more time for the race condition to resolve?

pfefferle:
> Could we also maybe add a slight delay to the scheduler to give more time for the race condition to resolve?

with a "hacky" `sleep(2)`?

do have an idea if there is a better hook to use?

maybe https://github.com/Automattic/wordpress-activitypub/issues/668 ???

mattwiebe:
> > Could we also maybe add a slight delay to the scheduler to give more time for the race condition to resolve?
> 
> with a "hacky" `sleep(2)`?
> 
> do have an idea if there is a better hook to use?
> 
> maybe #668 ???

I'm thinking about where we [schedule the cron job here](https://github.com/Automattic/wordpress-activitypub/blob/ae6db6593d14acd5597da25310f343fd435ae8c4/includes/class-scheduler.php#L168). When the first arg is `time()` it's a near-instant cron job that is possibly running out ahead of the process that's trying to update postmeta to set private visibility. So we add 30 seconds:

```php
\wp_schedule_single_event( \time() + 30, $hook, $args );
```

Or even less, maybe 5-10, something like that. Just gives all the saving processes a little more time to settle before blasting out the post.

pfefferle:
I See! Makes sense!

pfefferle:
@mattwiebe added all changes!

PR #960, opened 2024-10-28
Improve `activity_handle_undo` action

pfefferle:
This PR improves the action handling and prevents PHP warnings.


includes/handler/class-undo.php
```diff
@@ -40,7 +40,9 @@ public static function handle_undo( $activity ) {
 			return;
 		}
 
-		$type = $activity['object']['type'];
+		$type    = $activity['object']['type'];
+		$state   = false;
+		$user_id = null;
```
obenland: Should this be 0? It's documented as `@param int   $user_id` below

pfefferle: I think I have to update the phpdoc then, because we use `0` for the blog-user (I'm no longer sure whether that was a good idea).

mattwiebe (dismissed):
Looks good!

includes/handler/class-undo.php
obenland: 0 means [there's no logged-in user](https://developer.wordpress.org/reference/functions/get_current_user_id/#return). So I guess it depends what's being done with it.

pfefferle:
@obenland @mattwiebe I changed the code to use the user that is passed by the inbox action.

includes/handler/class-undo.php
```diff
@@ -72,8 +76,15 @@ public static function handle_undo( $activity ) {
 			}
 
 			$state = wp_trash_comment( $comment );
-
-			do_action( 'activitypub_handled_undo', $activity, $user_id, isset( $state ) ? $state : null, null );
 		}
+
+		/*
```
obenland:
```suggestion
		/**
```

pfefferle: thanks!

obenland: Don't think I'd pass on a chance to pad my stats! 😄

mattwiebe (approved):

PR #962, opened 2024-10-29
Support `comment_previously_approved` setting

pfefferle:


mattwiebe (commented):
If I read this correctly, what we're doing here is seeing if a fediverse commenter has previously had an approved comment and if so, to approve future comments from them?

pfefferle:
@mattwiebe exactly!

mattwiebe (approved):
This is a nice improvement.

PR #966, opened 2024-11-03
Fix Fediverse preview for "blog actor only" mode.

Menrath:
The Fediverse preview added in https://github.com/Automattic/wordpress-activitypub/pull/944 does not work when the WordPress site is in single actor mode, that is only the blog-user is enabled.

**Fixes:**
- Get the user for the `post-preview.php` template via the ActivityPub objects `attributedTo` property.
- Additional bug that the current  \Activitypub\Collection\Users::get_by_resource( $uri ) did not return the blog user when the new `https://example.com/?author=0` blog actor ID is used.

This solution should work, but if anyone knows a cleaner solution I am happy to just have reported the issue.

### DIFF
--- a/templates/post-preview.php
+++ b/templates/post-preview.php
@@ -15,10 +15,8 @@ if ( \is_wp_error( $transformer ) ) {
 	);
 }
 
-$object        = $transformer->to_object();
-$attributed_to = $object->get_attributed_to();
-$user_uri      = is_array( $attributed_to ) ? reset( $attributed_to )[0] : $attributed_to;
-$user          = \Activitypub\Collection\Users::get_by_resource( $user_uri );
+$object = $transformer->to_object();
+$user   = $transformer->get_actor_object();
 
 ?>
 <DOCTYPE html>
--- a/templates/post-preview.php
+++ b/templates/post-preview.php
@@ -14,8 +14,12 @@ if ( \is_wp_error( $transformer ) ) {
 		404
 	);
 }
-$user   = \Activitypub\Collection\Users::get_by_id( $post->post_author );
-$object = $transformer->to_object();
+
+$object        = $transformer->to_object();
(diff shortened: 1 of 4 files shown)


templates/post-preview.php
```diff
@@ -14,8 +14,12 @@
 		404
 	);
 }
-$user   = \Activitypub\Collection\Users::get_by_id( $post->post_author );
-$object = $transformer->to_object();
+
+$object        = $transformer->to_object();
+$attributed_to = $object->get_attributed_to();
+$user_uri      = is_array( $attributed_to ) ? reset( $attributed_to )[0] : $attributed_to;
+$user          = \Activitypub\Collection\Users::get_by_resource( $user_uri );
```
pfefferle: what about making `$transformer->get_actor_object();` accessible? this feels like we do the same over and over!?!

Menrath: I favour the idea and would also suggest to investigate whether the whole logic of "which" actor is `owning` a post, and who is announcing or creating/updating could be more in one place. 
FYI: My background is that I wonder if it could be possible in the feature to also create new actors for event organizers, but currently this seems like a very distant goal which may also never be achieved due it's complexity, both here for us devs and for the users.

pfefferle: totally! I will start working on the outbox feature this week and try to do it as open as possible and to start painting the overall idea. I think this could make this a lot easier. At least I will try to keep your feedback about that in mind!

pfefferle: 🤔 The more I look into the code, the more I agree with your! I think we have to refactor this a bit within the Outbox task!

And to simplify the transformer, so that it will always return the correct user regardless of the post-author!

pfefferle: @Menrath I did some changes that should work for now. Let me know what you think!

pfefferle (approved):

PR #968, opened 2024-11-05
Rename `Users`-Collection to `Actors` to be more specific

pfefferle:
/cc @Menrath

### DIFF
--- a/includes/collection/class-actors.php
+++ b/includes/collection/class-actors.php
@@ -1,0 +1,289 @@
+<?php
+/**
+ * Actors collection file.
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\Collection;
+
+use WP_Error;
+use WP_User_Query;
+use Activitypub\Model\User;
+use Activitypub\Model\Blog;
+use Activitypub\Model\Application;
+
+use function Activitypub\object_to_uri;
+use function Activitypub\normalize_url;
+use function Activitypub\normalize_host;
+use function Activitypub\url_to_authorid;
+use function Activitypub\is_user_disabled;
+
+/**
+ * Actors collection.
+ */
+class Actors {
+	/**
+	 * The ID of the Blog Actor.
+	 *
+	 * @var int
+	 */
+	const BLOG_USER_ID = 0;
+
+	/**
+	 * The ID of the Application Actor.
+	 *
+	 * @var int
+	 */
+	const APPLICATION_USER_ID = -1;
+
+	/**
+	 * Get the Actor by ID.
+	 *
+	 * @param int $user_id The User-ID.
+	 *
+	 * @return User|Blog|Application|WP_Error The Actor or WP_Error if user not found.
+	 */
+	public static function get_by_id( $user_id ) {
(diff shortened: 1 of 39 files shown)


mattwiebe (dismissed):
"Oh no this is going to cause so many problems on wpcom, we use the `Users` collection in a few places"

"Why isn't the users class deleted?"

"Oh this is nice back compat work" 😅

Menrath:
Very nice! :) A consistent and specific language in ActivityPub is so important to improve collaboration and maintenance.

pfefferle:
> Idea: What about adding deprecation notices to the Users collection right away?

@Menrath what do you mean by that? I added deprecation messages to all functions in the Users Collection, is there an easier way?

Menrath:
Never mind! I didn't see the _deprecated... function, I was just looking for `https://developer.wordpress.org/reference/functions/_doing_it_wrong/` ;)

includes/collection/class-actors.php
```diff
@@ -0,0 +1,290 @@
...
+		switch ( $user_id ) {
+			case self::BLOG_USER_ID:
+				return new Blog();
+			case self::APPLICATION_USER_ID:
+				return new Application();
+			default:
+				return User::from_wp_user( $user_id );
+		}
+
+		return new WP_Error(
```
obenland: I don't know if this statement ever reached? The `default` in the `switch` statement should take precedent?

includes/collection/class-actors.php
```diff
@@ -0,0 +1,290 @@
...
+			'activitypub_user_not_found',
+			\__( 'Actor not found', 'activitypub' ),
+			array( 'status' => 404 )
+		);
+	}
+
+	/**
+	 * Get the Actor by username.
+	 *
+	 * @param string $username The Actor-Name.
```
```suggestion
	 * @param string $username Name of the Actor.
```
Considered `Actorname` but that is probably not a thing.

includes/collection/class-actors.php
```diff
@@ -0,0 +1,290 @@
...
+			'activitypub_user_not_found',
+			\__( 'Actor not found', 'activitypub' ),
+			array( 'status' => 404 )
+		);
+	}
+
+	/**
+	 * Get the Actor by resource.
+	 *
+	 * @param string $uri The Actor-Resource.
```
```suggestion
	 * @param string $uri The Actor resource.
```

includes/collection/class-actors.php
obenland:
```suggestion
	 * @param string $username The Actor name.
```
Or this?

includes/collection/class-actors.php
```diff
@@ -0,0 +1,290 @@
...
+					\__( 'Wrong scheme', 'activitypub' ),
+					array( 'status' => 404 )
+				);
+		}
+	}
+
+	/**
+	 * Get the Actor by resource.
+	 *
+	 * @param string $id The Actor-Resource.
```
```suggestion
	 * @param string $id The Actor resource.
```

includes/collection/class-actors.php
pfefferle: That was a PR and I really missed that! thanks 👍

PR #969, opened 2024-11-05
Hide sticky posts that are not public

pfefferle:
Fixes https://mastodon.social/@jollysea@chaos.social/113431319666591560

## Proposed changes:

*

### Other information:

- [x] Added Migration to delete old meta-fields

## Testing instructions:

* Go to '..'
*


pfefferle:
@mattwiebe added migration code and unittests

mattwiebe (approved):
Confirmed that the `null` postmeta is no longer set when switching a non-public post back to public. Looks good!

PR #974, opened 2024-11-08
Users: do not display submenus to users who cannot author posts

jeherve:
## Proposed changes:

Right now, subcribers, customers (when WooCommerce is installed) and any other registered users see the users's submenus created by the ActivityPub plugin. 

I believe only folks can author posts can be ActivityPub actors on the site, so I don't believe subscribers should have access to those menus.

### Other information:

## Testing instructions:

* Start with a site where you have 3 users: an admin, an author, and a subscriber.
* Activate the ActivityPub plugin
* Log in to the site as the admin
    * You should see the 2 extra menus displayed in the Users menu.
* Log in as the author
    * You should see those menus as well.
* Log in as the subscriber.
    * You should not see those submenus.

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -72,7 +72,7 @@ class Admin {
 			$followers_list_page = \add_users_page(
 				\__( '⁂ Followers', 'activitypub' ),
 				\__( '⁂ Followers', 'activitypub' ),
-				'edit_posts',
+				'activitypub',
 				'activitypub-followers-list',
 				array(
 					self::class,
@@ -88,7 +88,7 @@ class Admin {
 			\add_users_page(
 				\__( '⁂ Extra Fields', 'activitypub' ),
 				\__( '⁂ Extra Fields', 'activitypub' ),
-				'edit_posts',
+				'activitypub',
 				\esc_url( \admin_url( '/edit.php?post_type=ap_extrafield' ) )
 			);
 		}
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -72,7 +72,7 @@ class Admin {
 			$followers_list_page = \add_users_page(
 				\__( '⁂ Followers', 'activitypub' ),
 				\__( '⁂ Followers', 'activitypub' ),
-				'read',
+				'edit_posts',
 				'activitypub-followers-list',
 				array(
 					self::class,
@@ -88,7 +88,7 @@ class Admin {
 			\add_users_page(
 				\__( '⁂ Extra Fields', 'activitypub' ),
 				\__( '⁂ Extra Fields', 'activitypub' ),
-				'read',
+				'edit_posts',
 				\esc_url( \admin_url( '/edit.php?post_type=ap_extrafield' ) )
 			);
 		}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Dev
+
+### Fixed
+
+* Do not display ActivityPub's user sub-menus to users who do not have the capabilities of writing posts.
+
 ## 4.1.0 - 2024-11-08
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -152,6 +152,10 @@ For reasons of data protection, it is not possible to see the followers of other
 
 ## Changelog ##
 
+### Dev ###
+
+* Fixed: do not display ActivityPub's user sub-menus to users who do not have the capabilities of writing posts.
+
 ### 4.1.0 ###
 
 * Added: Add custom Preview for "Fediverse"


pfefferle:
Oh, thanks a lot for finding that and your PR @jeherve !

But can we use `activitypub` as cap instead of `edit_posts`? Would that work? https://github.com/Automattic/wordpress-activitypub/blob/6f140b4db11d2096468fcb9bd5d1c0ad9e27f8ae/includes/class-activitypub.php#L569

We added a capability some weeks/month ago, to allow admins to add/remove the capability to/from users!

jeherve:
> But can we use `activitypub` as cap instead of `edit_posts`? Would that work?

For this to work, I believe we would need to add some extra code, to clearly define that custom `activitypub` cap, and map it to an existing primitive capability defined by Core, i.e. `edit_posts` in my opinion.

To do that, we should probably add a new method in the main `Activitypub` class, hooked into the `map_meta_cap` filter, mapping the `activitypub` cap to `edit_posts`:
https://developer.wordpress.org/reference/hooks/map_meta_cap/

jeherve:
> To do that, we should probably add a new method in the main `Activitypub` class, hooked into the `map_meta_cap` filter, mapping the `activitypub` cap to `edit_posts`

To clarify, this is what I had in mind:

```diff
diff --git a/includes/class-activitypub.php b/includes/class-activitypub.php
index 26d6327..1bcf338 100644
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -40,6 +40,7 @@ class Activitypub {
 		\add_action( 'init', array( self::class, 'add_rewrite_rules' ), 11 );
 		\add_action( 'init', array( self::class, 'theme_compat' ), 11 );
 
+		\add_filter( 'map_meta_cap', array( self::class, 'map_activitypub_cap' ), 10, 4 );
 		\add_action( 'user_register', array( self::class, 'user_register' ) );
 
 		\add_action( 'in_plugin_update_message-' . ACTIVITYPUB_PLUGIN_BASENAME, array( self::class, 'plugin_update_message' ) );
@@ -570,6 +571,24 @@ class Activitypub {
 		}
 	}
 
+	/**
+	 * Map the custom 'activitypub' cap to the author role
+	 *
+	 * @param string[] $caps    Primitive capabilities required of the user.
+	 * @param string   $cap     Capability being checked.
+	 * @param int      $user_id The user ID.
+	 * @param array    $args    Adds context to the capability check, typically starting with an object ID.
+	 *
+	 * @return array
+	 */
+	public static function map_activitypub_cap( $caps, $cap, $user_id, $args ) { // phpcs:ignore Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed,VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+		if ( 'activitypub' === $cap ) {
+			$caps = array( 'edit_posts' );
+		}
+
+		return $caps;
+	}
+
 	/**
 	 * Delete `activitypub_content_visibility` when updated to an empty value.
 	 *
diff --git a/includes/class-admin.php b/includes/class-admin.php
index f959ce8..5a5cd81 100644
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -72,7 +72,7 @@ class Admin {
 			$followers_list_page = \add_users_page(
 				\__( '⁂ Followers', 'activitypub' ),
 				\__( '⁂ Followers', 'activitypub' ),
-				'edit_posts',
+				'activitypub',
 				'activitypub-followers-list',
 				array(
 					self::class,
@@ -88,7 +88,7 @@ class Admin {
 			\add_users_page(
 				\__( '⁂ Extra Fields', 'activitypub' ),
 				\__( '⁂ Extra Fields', 'activitypub' ),
-				'edit_posts',
+				'activitypub',
 				\esc_url( \admin_url( '/edit.php?post_type=ap_extrafield' ) )
 			);
 		}

```

I'm not familar with what else you did with that custom `activitypub` cap though, so I'll let you test whether this breaks some things :)

pfefferle:
Sorry if I was not precise enough: We already set the activitypub capability (but without a user-role) and I was not sure if this will work. I did some tests and it seems to work!

Btw. You can add/remove the capability on the users page in WordPress as an admin!

includes/class-admin.php
```diff
@@ -72,7 +72,7 @@ public static function admin_menu() {
 			$followers_list_page = \add_users_page(
 				\__( '⁂ Followers', 'activitypub' ),
 				\__( '⁂ Followers', 'activitypub' ),
-				'read',
+				'edit_posts',
```
pfefferle:
```suggestion
				'activitypub',
```

includes/class-admin.php
```diff
@@ -88,7 +88,7 @@ public static function admin_menu() {
 			\add_users_page(
 				\__( '⁂ Extra Fields', 'activitypub' ),
 				\__( '⁂ Extra Fields', 'activitypub' ),
-				'read',
+				'edit_posts',
```
pfefferle:
```suggestion
				'activitypub',
```

pfefferle (changes requested):

jeherve:
> I did some tests and it seems to work!

It doesn't work on my end, at least not until I map the `activitypub` cap to `edit_posts` as I've shown above. If I only change the cap as in your suggestion, subscribers can still see the menu.

pfefferle:
Does your subscriber maybe have the ActivityPub cap?

jeherve:
> Does your subscriber maybe have the ActivityPub cap?

As we discussed in slack, my test subscriber user was previously an editor (I changed its role), so the `activitypub` cap stayed active when the user was downgraded.

pfefferle (approved):

pfefferle:
thanks for the remote-pairing :)

PR #975, opened 2024-11-08
Embeds: only revert to URL if there is one

mattwiebe:
Saw a burst of warnings like:

```
[08-Nov-2024 16:50:11 UTC] Warning: Undefined array key "url" in wp-content/plugins/activitypub/includes/transformer/class-post.php on line 1074
```

Although I couldn't replicate on the site in question, it's probably just an embed block with no URL set. This should stop that.


pfefferle (approved):

pfefferle:
Makes sense!

PR #978, opened 2024-11-08
Add Mastodon Apps status provider

mattwiebe:
When used alongside Enable Mastodon Apps, we should prefer the ActivityPub representation as it would be sent out to Followers etc.

This work grew out of some other issues with fediverse comments being falsely attributed to the Blog or User.

## Proposed changes:
* All posts are the ActivityPub representation
* All comments received from the fediverse get attributed properly
* Bonus feature: local comments that leave a URL, if it turns out to be a valid actor, will be represented by that actor
* Local comments get their own handling like below

(Both images are screenshots in Phanpy)
Screenshot 2024-11-12 at 13 18 04
Screenshot 2024-11-12 at 13 18 48

## Testing instructions:
Activate Enable Mastodon Apps alongside ActivityPub with this PR. Is everything handled more or less properly?


integration/class-enable-mastodon-apps.php
```diff
@@ -324,6 +326,150 @@ function ( $field ) {
...
+	 */
+	private static function get_account_for_local_comment( $comment ) {
+		$maybe_actor = self::maybe_get_account_for_actor( $comment->comment_author_url );
+		if ( $maybe_actor ) {
+			return $maybe_actor;
+		}
+
+		// We will make a pretend local account for this comment.
+		$account                 = new Account();
+		$account->id             = 999999; // This is a fake ID.
```
pfefferle, line 413: this is fine because we do the lookup with the acct?

pfefferle (approved):

PR #984, opened 2024-11-12
Reuse constants once they're defined

obenland:
## Proposed changes:

* Updates plugin function usages with existing constants.

## Testing instructions:

* Check the plugin list and make sure the Settings action link is still there.

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -90,7 +90,7 @@ require_once __DIR__ . '/includes/functions.php';
 // Plugin related constants.
 \define( 'ACTIVITYPUB_PLUGIN_DIR', plugin_dir_path( __FILE__ ) );
 \define( 'ACTIVITYPUB_PLUGIN_BASENAME', plugin_basename( __FILE__ ) );
-\define( 'ACTIVITYPUB_PLUGIN_FILE', plugin_dir_path( __FILE__ ) . '/' . basename( __FILE__ ) );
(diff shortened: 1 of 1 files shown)


activitypub.php
```diff
@@ -90,7 +90,7 @@
 // Plugin related constants.
 \define( 'ACTIVITYPUB_PLUGIN_DIR', plugin_dir_path( __FILE__ ) );
 \define( 'ACTIVITYPUB_PLUGIN_BASENAME', plugin_basename( __FILE__ ) );
-\define( 'ACTIVITYPUB_PLUGIN_FILE', plugin_dir_path( __FILE__ ) . '/' . basename( __FILE__ ) );
+\define( 'ACTIVITYPUB_PLUGIN_FILE', ACTIVITYPUB_PLUGIN_DIR . basename( __FILE__ ) );
```
pfefferle, line 93: what about the `/` delimiter here?

pfefferle: @obenland you seem to never make any issues, I should have known that it simply works as you changed it 🙄

pfefferle (approved):

activitypub.php
obenland: `plugin_dir_path()` [returns the path with a trailing slash](https://developer.wordpress.org/reference/functions/plugin_dir_path/#source)

PR #985, opened 2024-11-12
Use version constant directly

obenland:
Since `ACTIVITYPUB_PLUGIN_VERSION` gets defined at the top of `activitypub.php`, `get_plugin_version()` will always return that and never fall back to the plugin meta information.

Not sure if 4.2.0 is the right version number in the deprecation documentation.
Not sure even if deprecating function is something this plugin does, or if it breaks back compat.

## Proposed changes:

* Deprecates `Activitypub\get_plugin_version()` and `Activitypub\Migration::get_target_version()` in favor of `ACTIVITYPUB_PLUGIN_VERSION`.
* Replaces all uses of the deprecated function with the constant.

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -255,15 +255,13 @@ function get_plugin_meta( $default_headers = array() ) {
 
 /**
  * Plugin Version Number used for caching.
+ *
+ * @deprecated 4.2.0 Use constant ACTIVITYPUB_PLUGIN_VERSION directly.
  */
 function get_plugin_version() {
-	if ( \defined( 'ACTIVITYPUB_PLUGIN_VERSION' ) ) {
-		return ACTIVITYPUB_PLUGIN_VERSION;
-	}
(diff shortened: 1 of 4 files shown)


pfefferle (dismissed):

obenland:
@pfefferle Do you have any guidance on this?

> Not sure if 4.2.0 is the right version number in the deprecation documentation.
> Not sure even if deprecating function is something this plugin does, or if it breaks back compat.

pfefferle:
@obenland Oh, haven't seen that, sorry!

We follow semantic versioning, so fully removing the functions would have required a 5.0.0 release due to the breaking change. Since you retained the functions and added a deprecation warning, opting for the next possible minor version makes total sense.

obenland:
Do you have a preference one way or the other?

pfefferle:
Exactly like you've done it: https://github.com/Automattic/wordpress-activitypub/pull/968/files#diff-2fb16a1f775688a0027ff486df52ce4bb27d1c754b878990408219982d3205b4 ☺️

Keeping the old files/classes/functions for a while, add deprecation message and remove them with the next major version.

pfefferle (approved):
makes sense!

PR #986, opened 2024-11-12
CLI: Remove meta command

obenland:
We can use `wp plugin get activitypub` instead.

See p1731432888774959-slack-C04TJ8P900J

## Proposed changes:

* Removes `meta` CLI command.
* Deprecates unused function `get_plugin_meta()`.


mattwiebe (approved):
My favorite color 🟥

activitypub.php
```diff
@@ -270,7 +274,7 @@ function get_plugin_version() {
 		'activitypub',
 		'\Activitypub\Cli',
 		array(
-			'shortdesc' => __( 'ActivityPub related commands: Meta-Infos, Delete and soon Self-Destruct.', 'activitypub' ),
+			'shortdesc' => 'ActivityPub related commands to manage plugin functionality and the federation of posts and comments.',
```
pfefferle, line 277: no translations for CLI commands?

obenland: I don't know! See #987. WP CLI [doesn't seem to translate their output](https://github.com/wp-cli/wp-cli/blob/e214df253932712ebe0669d50bb485276a9a5102/php/commands/src/CLI_Command.php#L243). And docs pulled from function comments are also not localized.

Do you use it in German?

pfefferle: The text seems to be translated...

Screenshot 2024-11-12 at 21 45 49

So the questions might be more like: do we want that? even more because it seems to be the only translated text!?!

pfefferle: let's switch to the other PR!

PR #987, opened 2024-11-12
CLI: Remove unneeded translation functions

obenland:
Could someone double-check me on my logic? 

I don't think translations get processed in the context of WP CLI (meaning docs or function messages get returned translated) but I could be wrong.

## Proposed changes:

* Removes translation functions from CLI commands.
* Updates parameter docs for `$args` params.


pfefferle:
No translations for CLI commands?

pfefferle:
Yoast-SEO seems to use translations, but the majority seems not to use them: https://github.com/search?q=WP_CLI%3A%3Asuccess&type=code

pfefferle:
The text seems to be translated (except the special-chars 🫣)...

Screenshot 2024-11-12 at 21 45 49

So the questions might be more like: do we want that? even more because it seems to be the only translated text!?!

obenland:
Interesting! Good to know

> So the questions might be more like: do we want that? even more because it seems to be the only translated text!?!

Up to you, I don't have a strong opinion

pfefferle:
I just wanted to write the same... I have no strong feelings on that one and am happy to remove them!

pfefferle (approved):

pfefferle:
you already removed the one from the description, so let's be consistent!

PR #988, opened 2024-11-12
Make change log entries mandatory with every PR

obenland:
It's a pretty barebones way to make sure changelogs get updated with every PR, instead of someone having to go through and doing it manually.
Adding the "Skip Changelog" to a PR will make the check pass, regardless of whether a changelog was updated or not.

## Proposed changes:

* Adds a Github action that checks CHANGELOG.md and readme.txt for updates.


obenland:
It looks like Gardening Bot adds the Docs label whenever Changelog gets updated :)

pfefferle:
yep ;)

pfefferle (dismissed):
I like it!

jeherve:
> It looks like Gardening Bot adds the Docs label whenever Changelog gets updated :)

That's a bug! :) It shouldn't be happening, we designed the action to ignore changes to `CHANGELOG.md`:
https://github.com/Automattic/jetpack/pull/30007#discussion_r1161900969

I'll look into why that's happening and get this fixed.

Edit: PR is here:
https://github.com/Automattic/jetpack/pull/40160

jeherve:
Maybe this is an opportunity to switch to using `automattic/jetpack-changelogger` to manage changelog entries and generation for this plugin?
https://packagist.org/packages/automattic/jetpack-changelogger

It is currently used in the Jetpack monorepo as well as in the WooCommerce monorepo, and can be used to generate single changelog entries for each PR, and then update a single changelog file on release.

Once the package is in use in the repo, it can be used from the command line when you're working on your PRs, or directly from the PR description, as you can see from the PR template in the WooCommerce repo:
https://github.com/woocommerce/woocommerce/blob/trunk/.github/PULL_REQUEST_TEMPLATE.md#changelog-entry

Is that something you'd be interested in? I'd be happy to set that up.

pfefferle (dismissed):

pfefferle:
@obenland do you think this is ready to merge?

obenland:
Looks likes it needs a new approval. Feel free to merge at your convenience

pfefferle (approved):

PR #989, opened 2024-11-13
Show '⁂ Fediverse Preview' link in Row-Actions

pfefferle:
Screenshot 2024-11-13 at 13 09 46


obenland:
I'm not sure I'd understand that that's meant to be an Activitypub Preview, until a few days ago I was unaware of the logo. Is this something other plugins offer as well in the list view?

pfefferle:
Good question... The icon is still pretty new... 🤔

I thought it might be nice to be as short as possible... do you think it should be `⁂ Fediverse Preview` instead? at least until people might recognize the icon...

obenland:
My preview looks like this, is that a configuration issue?

Screenshot 2024-11-14 at 10 13 59 AM

obenland:
It's not part of this PR but would be a good follow up, the preview link looks off with the Core preview:

image

pfefferle:
> It's not part of this PR but would be a good follow up, the preview link looks off with the Core preview

I saw this is already reported somewhere in the core-editor repo, but can't find it any more. The "Preview in new tab" seems to be not aligned with the top links and the fediverse links. I am not sure if we can do something there.

pfefferle:
> My preview looks like this, is that a configuration issue?

How have you done that!?! 😱

obenland:
I personally don't think a preview link is necessary in the list view. It makes more sense to me where it is now, in the Preview dropdown when editing a post.

pfefferle:
We had some feedbacks from people with the classic editor, that they can't use it now and because the list also includes the classic preview link...

mattwiebe (dismissed):
> We had some feedbacks from people with the classic editor, that they can't use it now and because the list also includes the classic preview link...

We could potentially only add that for `! site_supports_block()` then?

Otherwise I'm fine with this.

pfefferle:
> We could potentially only add that for ! site_supports_block() then?

That is a nice one!

includes/class-admin.php
```diff
@@ -811,7 +811,8 @@ public static function row_actions( $actions, $post ) {
 			! \post_type_supports( \get_post_type( $post ), 'activitypub' ) ||
 			! in_array( $post->post_status, array( 'pending', 'draft', 'future', 'publish' ), true ) ||
 			! \current_user_can( 'edit_post', $post->ID ) ||
-			ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL === get_content_visibility( $post )
+			ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL === get_content_visibility( $post ) ||
+			site_supports_blocks()
```
Menrath, line 815: @pfefferle Are you interested in having more fine control here? Sometimes plugins offering custom post types to also not use the block editor yet for their post types, so whether the button is shown could be depending on that as well.

pfefferle: @Menrath sure, can you create a PR for that?

PR #990, opened 2024-11-13
Better Post-Formats support

pfefferle:
This change uses the Post-Format also for Custom-Post-Type if they support them.

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -735,12 +735,8 @@ class Post extends Base {
 
 		if ( 'page' === $post_type ) {
 			$object_type = 'Page';
-		} else {
-			$post_format = \get_post_format( $this->wp_object );
-
-			if ( ! $post_format ) {
-				$object_type = 'Article';
-			}
+		} elseif ( ! \get_post_format( $this->wp_object ) ) {
+			$object_type = 'Article';
 		}
 
 		return $object_type;
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -735,7 +735,7 @@ class Post extends Base {
 
 		if ( 'page' === $post_type ) {
 			$object_type = 'Page';
-		} elseif ( 'post' === $post_type ) {
+		} else {
 			$post_format = \get_post_format( $this->wp_object );
 
 			if ( ! $post_format ) {
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -734,12 +734,11 @@ class Post extends Base {
 		$post_format = 'standard';
 
 		if ( \get_theme_support( 'post-formats' ) ) {
-			$post_format = \get_post_format( $this->wp_object );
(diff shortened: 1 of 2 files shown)


includes/transformer/class-post.php
```diff
@@ -750,12 +755,6 @@ protected function get_type() {
 						break;
 				}
 				break;
-			case 'page':
-				$object_type = 'Page';
-				break;
-			default:
-				$object_type = 'Article';
-				break;
 		}
```
obenland, line 740: I'm reading the logic as something like the following, starting at `// Default to Article.`:

```php
		// Default to Note.
		$object_type = 'Note';
		$post_type   = \get_post_type( $this->wp_object );

		if ( 'page' === $post_type ) {
			$object_type = 'Page';
		} else if ( 'post' === $post_type ) {
			$post_format = \get_post_format( $this->wp_object );

			if ( 'standard' === $post_format || '' === $post_format ) {
				$object_type = 'Article';
			}
		}
```

Is that about right?
What are circumstances where `$post_format` would be an empty string rather than `false`?

pfefferle: you are right!

pfefferle: changed

pfefferle: @obenland I think I mixed that up with comment-types somehow, because the default type for comments was an empty string before the migration to "comment". Anyways thanks for pointing this out!

obenland: How would you feel about replacing the nested switch statements with what I put together above?

pfefferle: @obenland sure, go for it! You can simply commit to my PR if you want!

pfefferle: FYI: the list was initially way bigger, that's why it is a switch/case: https://github.com/Automattic/wordpress-activitypub/commit/dde469c27c4e5ab6c09e7b284374b94d14beb9fa#diff-28c56747da3496a58aa702e51b60e1b36656e0263c4c2d97d0a013a6d2f07a08R64

obenland:
> This change uses the Post-Format also for Custom-Post-Type if they support them.

I think my suggested change didn't account for the goal of this PR. It's still gated to posts, I'm afraid

includes/transformer/class-post.php
```diff
@@ -729,33 +729,18 @@ protected function get_type() {
...
-					default:
-						$object_type = 'Note';
-						break;
-				}
-				break;
-			case 'page':
-				$object_type = 'Page';
-				break;
-			default:
+			if ( ! $post_format ) {
```
obenland: This is true for post types that don't support post formats. Is that the intention?
As opposed to post types that support post formats but the object doesn't have one assigned.

pfefferle: this is for both... either it does not support post-formats or the post format is "standard" (because standard is not in the list of post-formats https://developer.wordpress.org/reference/functions/get_post_format/#more-information)...

obenland (approved):

PR #991, opened 2024-11-13
Add baseline tests for Post::get_type()

obenland:
Allows us to make changes with more confidence in the future.

Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Dev
 
+### Added
+
+* Unit tests for the `ActivityPub\Transformer\Post` class
+
 ### Improved
 
 * Reuse constants once they're defined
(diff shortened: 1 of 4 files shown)


obenland:
@pfefferle I asked Claude.ai to create some unit tests for the `get_type()` method as it's written in trunk, so we can compare future changes with those outcomes.

It currently fails for posts without a post format (`test_get_type_handles_missing_post_format`), probably because of a wrong assumption? Posts with a title, more than 400 char content and no post format should be considered Articles, yes?

pfefferle:
@obenland I would say so, yes.

pfefferle (approved):
You are using a different naming convention for the new test. I have no strong feelings about how we should name them, but I would vote for keeping them consistent, so we should work on a follow up PR that renames the other tests.

obenland:
Yeah I got a phpcs error for the filename without the pre-pended `class-`. I'll do a follow up.

PR #1002, opened 2024-11-18
Have better checks if `audience` should be set or not

pfefferle:
Should help with issue: https://github.com/LemmyNet/lemmy/issues/5147


harcesz:
Could a blog be used a single user and still have an audience? Or to be more precise; would be possible to allow setting a single user blog so it shows up as a community in Lemmy, rather than a user? From my tests on the current version it does not seem to work.

pfefferle:
Not out of the box, but I can help with a code snippet that overrides the person type with group.

harcesz:
That would be much appreciated! Also such a setting would be great in future versions :)

pfefferle:
I got very controversial feedback because of the `Group` implementation and tweaked that a lot for that special usecase (Blog-User + Users enabled). I am not sure if I want to make it a setting, because it might break things if you do not exactly know what you are doing, especially on Friendi.ca!

harcesz:
Uh, yeah, who'd expect the social network protocol to have to deal with groups of people ;P But still - understood.   
  
I'd appreciate the mentioned snippet if you think generally I can use it without exploding anyones server.

PR #1005, opened 2024-11-20
Copy image, because otherwise it will be deleted.

pfefferle:
The sideload logic of WordPress deletes the source file and that might cause issues in local tests. This change generates a copy of the image to use that in tests.


akirk (approved):
Oh no, sorry!

pfefferle:
No problem :)

PR #1008, opened 2024-11-21
Add an early return filter for `get_upload_dir()`

mattwiebe:
Our use of `wp_upload_dir()` in `Activitypub\get_upload_baseurl()` was creating PHP warnings on dotcom due to its internal use of `wp_mkdir_p()` which fails there.

Fixes #

## Proposed changes:
* add a `pre_activitypub_get_upload_baseurl` filter to bypass the use of `wp_upload_dir()`


includes/functions.php
```diff
@@ -1478,12 +1478,22 @@ function get_attribution_domains() {
  * @return string The upload base URL.
  */
 function get_upload_baseurl() {
-	$upload_dir = \wp_get_upload_dir();
+	/**
+	 * Early filter to allow plugins to set the upload base URL.
+	 *
+	 * @param string|false $maybe_upload_dir The upload base URL or false if not set.
+	 */
+	$maybe_upload_dir = apply_filters( 'pre_activitypub_get_upload_baseurl', false );
```
pfefferle, line 1486: `pre_activitypub_get_upload_baseurl` or `activitypub_pre_get_upload_baseurl`???

includes/functions.php
```diff
@@ -1478,12 +1478,22 @@ function get_attribution_domains() {
...
+	if ( false !== $maybe_upload_dir ) {
+		return $maybe_upload_dir;
+	}
+
+	$upload_dir = \wp_upload_dir();
```
pfefferle: why not `wp_get_upload_dir()`?? to save a function call?

mattwiebe: Oh, TIL. Just switching to that would have fixed our issue too.

includes/functions.php
mattwiebe: Core always uses a `pre_` prefix to indicate a bypass filter so I used that here.

pfefferle:
@mattwiebe you also have to add the changelog entry to the readme.txt, then the tests are green!

includes/functions.php
```diff
@@ -1478,12 +1478,22 @@ function get_attribution_domains() {
...
+		return $maybe_upload_dir;
+	}
+
 	$upload_dir = \wp_get_upload_dir();
 
 	/**
 	 * Filters the upload base URL.
 	 *
-	 * @param string \wp_get_upload_dir()['baseurl'] The upload base URL.
+	 * @param string \wp_upload_dir()['baseurl'] The upload base URL.
```
pfefferle:
```suggestion
	 * @param string \wp_get_upload_dir()['baseurl'] The upload base URL.
```

if you change it back, we should also change the comment

mattwiebe: Ah, nice catch

mattwiebe:
> @mattwiebe you also have to add the changelog entry to the readme.txt, then the tests are green!

It does not like me 😆

mattwiebe:
We could also probably just scrap the regular filter entirely. I doubt anyone was using it, since it was just for us.

pfefferle (dismissed):

pfefferle (approved):

pfefferle (changes requested):
~~It seems that you accidentally bring back a lot of readme stuff!~~

I quickly fixed it

pfefferle (approved):

mattwiebe:
> I quickly fixed it

Thanks! The conflict resolution thing here was acting all kinds of weird for me

PR #1013, opened 2024-11-22
Actors: deprecate `as User_Collection` throughout

mattwiebe:
Deprecate `Activitypub\Collection\Actors as User_Collection` to simply use `Actors` throughout the codebase.

## Proposed changes:
* renamed `Users_Collection` to simply use `Actors`

### Other information:

- [x] Have you written new tests for your changes, if applicable? (not applicable, everything should be the same)

## Testing instructions:
Tests should pass, everything normal in function


pfefferle:
I think we should also update `Actors::APPLICATION_USER_ID` then!?!

mattwiebe:
> I think we should also update `Actors::APPLICATION_USER_ID` then!?!

Good idea!

pfefferle:
@mattwiebe and don't forget the changelog ;)

Menrath:
@mattwiebe @pfefferle I still have some doubts about this change, as the current `Actors::BLOG_USER_ID` and `Actors::APPLICATION_USER_ID` wordings show that the actors are indeed "mapped" to pseudo users. With a constant named `Actors::BLOG_ACTOR_ID` I would expect it more likely to be the full URI.

mattwiebe:
> @mattwiebe @pfefferle I still have some doubts about this change, as the current `Actors::BLOG_USER_ID` and `Actors::APPLICATION_USER_ID` wordings show that the actors are indeed "mapped" to pseudo users. With a constant named `Actors::BLOG_ACTOR_ID` I would expect it more likely to be the full URI.

Interesting! I guess we do have to show the seams of Actors <-> Users mapping somewhere, and this may be as good a place as any.

pfefferle (approved):

PR #1017, opened 2024-11-25
Added setting to enable/disable Authorized-Fetch

### LINKED ISSUE #726: Enable Authorized Fetch by default

pfefferle:
Fixes #726

Settings Screen:

Screenshot 2024-11-25 at 10 54 55


templates/settings.php
```diff
@@ -237,6 +237,33 @@
 			<?php \do_settings_fields( 'activitypub', 'general' ); ?>
 			<?php \do_settings_fields( 'activitypub', 'server' ); ?>
 		</div>
+		<div class="box">
```
obenland, line 240: should this be hidden when the constant is defined?

templates/settings.php
```diff
@@ -237,6 +237,33 @@
...
+							<p class="description">
+								<?php \esc_html_e( '⚠ Secure mode has its limitations, which is why it is not enabled by default. It is not fully supported by all software in the fediverse, and some features may break, especially when interacting with Mastodon servers older than version 3.0. Additionally, since it requires authentication for public content, caching is not possible, leading to higher computational costs.', 'activitypub' ); ?>
+							</p>
+							<p class="description">
+								<?php \esc_html_e( '⚠ Secure mode does not hide the HTML representations of public posts and profiles. While HTML is a less consistant format (that potentially changes often) compared to first-class ActivityPub representations or the REST API, it still poses a potential risk for content scraping.', 'activitypub' ); ?>
+							</p>
+						</td>
+					</tr>
+				</tbody>
+			</table>
```
obenland, line 266: Is there a reason we're not adding this as a settings field?

templates/settings.php
pfefferle: good point!

templates/settings.php
pfefferle: No, just legacy I would say ;)

pfefferle: I would recommend to not work on that as part of this PR, but as a refactor PR instead.

readme.txt
```diff
@@ -107,7 +107,7 @@ The plugin uses PHP Constants to enable, disable or change its default behaviour
 * `ACTIVITYPUB_USERNAME_REGEXP` - Change the default regex to detect @-replies in a text. Default: `(?:([A-Za-z0-9\._-]+)@((?:[A-Za-z0-9_-]+\.)+[A-Za-z]+))`.
 * `ACTIVITYPUB_URL_REGEXP` - Change the default regex to detect urls in a text. Default: `(www.|http:|https:)+[^\s]+[\w\/]`.
 * `ACTIVITYPUB_CUSTOM_POST_CONTENT` - Change the default template for Activities. Default: `<strong>[ap_title]</strong>\n\n[ap_content]\n\n[ap_hashtags]\n\n[ap_shortlink]`.
-* `ACTIVITYPUB_AUTHORIZED_FETCH` - Enable AUTHORIZED_FETCH. Default: `false`.
+* `ACTIVITYPUB_AUTHORIZED_FETCH` - Enable AUTHORIZED_FETCH.
```
obenland, line 110: Should this still be documented, now that it's removed? I suppose the preferred way to override would now be the filter?

templates/settings.php
```diff
@@ -237,6 +237,35 @@
...
+				<tbody>
+					<?php if ( ! defined( ACTIVITYPUB_AUTHORIZED_FETCH ) ) : ?>
+					<tr>
+						<th scope="row">
+							<?php \esc_html_e( 'Authorized-Fetch', 'activitypub' ); ?>
+						</th>
+						<td>
+							<p>
+								<label>
+									<input type="checkbox" name="activitypub_authorized_fetch" id="activitypub_authorized_fetch" value="1" <?php echo \checked( '1', \get_option( 'activitypub_authorized_fetch', '0' ) ); ?> />
```
obenland:
```suggestion
									<input type="checkbox" name="activitypub_authorized_fetch" id="activitypub_authorized_fetch" value="1" <?php \checked( '1', \get_option( 'activitypub_authorized_fetch', '0' ) ); ?> />
```
`checked()` echos automatically, unless the third parameter is set to `false`.

obenland (changes requested):
Sorry, I was ready to approve but found a couple more minor things 😬

readme.txt
pfefferle: I would not say that the filter is the preferred way. I think these are different use cases... As an admin, you want to use the const, as a plugin developer you might want to use the filter...

templates/settings.php
obenland: Agreed

templates/settings.php
```diff
@@ -237,6 +237,35 @@
 			<?php \do_settings_fields( 'activitypub', 'general' ); ?>
 			<?php \do_settings_fields( 'activitypub', 'server' ); ?>
 		</div>
+		<div class="box">
+			<h3><?php \esc_html_e( 'Security', 'activitypub' ); ?></h3>
+			<table class="form-table">
+				<tbody>
+					<?php if ( ! defined( ACTIVITYPUB_AUTHORIZED_FETCH ) ) : ?>
```
obenland:
```suggestion
					<?php if ( ! defined( 'ACTIVITYPUB_AUTHORIZED_FETCH' ) ) : ?>
```
This needs quotes to avoid triggering a fatal error.

pfefferle: 🫣

obenland (approved):

Issue #1020, opened 2024-11-25
Video attachments:


mattwiebe:
### Quick summary

Video attachments appear with two attachments, one of them broken

### Steps to reproduce

You can see it in my recent post with a video: https://mattwiebe.blog/2024/11/24/support-week-for-activitypub-for-wordpress/

What it looks like on Phanpy:

![Image](https://github.com/user-attachments/assets/5015e9a2-2234-42aa-b315-fe3ab09be252)

So when I get the ActivityPub version, the `attachment` field shows these two entries:

```json
[
  {
    "url": "https://videos.files.wordpress.com/9qROHgXQ/activitypub-office-hours-nov-25.mp4",
    "length": "114070319",
    "mediaType": "video/mp4"
  },
  {
    "type": "Document",
    "mediaType": "video/mp4",
    "url": "https://videos.files.wordpress.com/9qROHgXQ/activitypub-office-hours-nov-25_mp4_std.mp4",
    "width": "1920",
    "height": "1080",
    "name": "Video",
    "icon": {
      "type": "Image",
      "mediaType": "image/jpeg",
      "url": "https://videos.files.wordpress.com/9qROHgXQ/activitypub-office-hours-nov-25_mp4_hd_1080p.original.jpg"
    }
  }
]
```

Unsure if this is specifically a dotcom thing or not but it looks like the first one is broken and the second one works.

### What you expected to happen

One video

### What actually happened

Two videos, one broken.

### Impact

Some (< 50%)

### Available workarounds?

No but the platform is still usable

pfefferle:
WordPress.com seems to add the video as enclosure. From the feed:

```XML
<enclosure url="http://videos.files.wordpress.com/9qROHgXQ/activitypub-office-hours-nov-25.mp4" length="114070319" type="video/mp4"/>
```

I think we have to optimize the enclosure code a bit: https://github.com/Automattic/wordpress-activitypub/blob/129ea69ccf102612b4c9cbb52a8814c3e8f8d8a4/includes/transformer/class-post.php#L311

mattwiebe:
I have a fix for wpcom in D167257-code

obenland:
Fixed in [D167257-code](https://href.li/?https://code.a8c.com/D167257)

PR #1021, opened 2024-11-26
Fixed: Missing attachement-type for enclosures

pfefferle:
@mattwiebe this should fix the broken attachment, but not that it is added twice.

#1020


mattwiebe:
This did not fix my issue on wpcom. It seems like the problem only happens there, so I have a workaround.

This PR looks like good defensiveness though

mattwiebe (dismissed):
:shipit:

PR #1028, opened 2024-11-29
Add basic validation to enclosure function

pfefferle:
/cc @akirk

Fixes Fatal errors if the `enclosure` meta does not match the core-format.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1006,6 +1006,11 @@ function get_enclosures( $post_id ) {
 
 	$enclosures = array_map(
 		function ( $enclosure ) {
+			// Check if the enclosure is a string.
+			if ( ! $enclosure || ! is_string( $enclosure ) ) {
+				return false;
+			}
+
 			$attributes = explode( "\n", $enclosure );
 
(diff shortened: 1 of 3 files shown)


akirk (approved):
Thanks!

includes/functions.php
```diff
@@ -1006,6 +1006,11 @@ function get_enclosures( $post_id ) {
 
 	$enclosures = array_map(
 		function ( $enclosure ) {
+			// Check if the enclosure is a string.
+			if ( ! $enclosure || ! is_string( $enclosure ) ) {
+				return false;
```
akirk, line 1011: Should this return `null` to remove it from the array?

akirk (changes requested):
I think we should remove the invalid values using `array_filter()`

pfefferle:
There is an `array_filter` and it also filters `false`.

includes/functions.php
pfefferle: Both will be filtered by `array_filter` but I am also fine with `null`.

akirk (approved):
Oops, sorry. I missed that!

Issue #721, opened 2024-04-08
CSS uses hard-coded variables and doesn't support dark/light mode


edent:
### Quick summary

I'm unable to see the text box because it uses invalid colours:

```css
activitypub__dialog .activitypub-dialog__button-group input {
  background-color: var(--wp--preset--color--white);
  border: 1px solid var(--wp--preset--color--black);
  color: var(--wp--preset--color--black);
...
}
```

My theme doesn't have those `--wp--preset` variables. 

Additionally, the modal doesn't respect the dark / light mode of the theme:

![Pop up is light, background is dark.](https://github.com/Automattic/wordpress-activitypub/assets/837136/c1e5f729-3d37-4a36-bff4-92f7250de51c)

### Steps to reproduce

1. Use a theme which doesn't use Gutenberg / Blocks
2. Open the reply modal

### What you expected to happen

The pop-up should be themed as per the rest of the site.

### What actually happened

The hardcoded CSS doesn't respect dark/light mode - or many other theming options.

### Impact

One

### Available workarounds?

No but the platform is still usable

janboddez:
As mentioned elsewhere :-), the "hardcoded" values bug me, too.

I'm fairly sure a light-weight vanilla JS script could do all the things the current modal does, without most of the React stuff. The less (explicit) styles you define, the more it will inherit from the theme's overall CSS.

That said, the modal will need a background (and foreground) color. Ideally, we could use CSS variables (either the current ones, or something like `--wp--preset--color--base` and `--wp--preset--color--contrast`, which quite a few [?] block themes seem to use).

Unfortunately, classic themes have no standardized way of defining colors. (In fact, I'm not sure the "block theme" colors are all that standardized, either.)

Next best thing would be to allow site admins to easily set the colors (and fonts, etc.). It may be possible to leverage the (block) theme editor here (or the old customizer).

Next best thing ... just override the CSS in your theme, I guess? That's what I'll be doing, but I don't like it, because next thing you know, the class names have changed and you can start over.

janboddez:
One last, quick thought: since the whole thing is rendered in JS, it _should_ be possible to fetch the `body`'s background and foreground colors and somehow do something with those (like, figure out which one's lighter, and then maybe slightly alter them, and then use the outcome in the modal). But even that will likely fall short when it comes to dark mode.

edent:
I think there's an easier solution. Don't use a modal dialogue. 

[...]

Issue #425, opened 2023-09-14
Follow Me Block: address styling issues in dark themes


jeherve:
You can see the problem in action here:
https://notiz.blog/follow/#%23Fediverse

<img width="970" alt="image" src="https://github.com/Automattic/wordpress-activitypub/assets/426388/3a527586-36d8-407c-93ea-5dba938f3e2c">

This is happening because the modal background color is set to white:
https://github.com/WordPress/gutenberg/blob/f92f9e28ba4fb2e435cb89d5b327b707cb805ba7/packages/components/src/modal/style.scss#L21

All the while the text color is also a white color as defined by the theme.

I think it would be nice if the modal shipped with some styles to prevent that, maybe using core CSS vars provided by the theme.

pfefferle:
+ it breaks on mobile view.

mattwiebe:
I did fix this by overriding theme colors in the modal

PR #1034, opened 2024-12-02
Dialog: Add screenreader labels

### LINKED ISSUE #555: Follow overlay is missing input labels
### Quick summary

If I use the follow me block and open the overlay, I noticed that there are no labels available. As regular user, I’m fine with that, but at least for accessibility there should be invisible labels for screen readers.

### Steps to reproduce

1. Create a page with the "Follow me" block.
2. Open the page
3. Open the overlay
4. Look into the source code

A label for each input field is available, especially for screen readers.

### What actually happened

The label for the input fields are missing.

### Impact

Some (< 50%)

### Available workarounds?

### Logs or notes

_No response_

### LINKED ISSUE #861: "Follow me" element uses prohibited ARIA attributes
### Quick summary

`div class="activitypub-follow-me-block-wrapper wp-block-activitypub-follow-me" **aria-label**="Follow me on the Fediverse" [...]`

More info [https://dequeuniversity.com/rules/axe/4.9/aria-prohibited-attr](https://dequeuniversity.com/rules/axe/4.9/aria-prohibited-attr)

_Not all ARIA role-attribute combinations are valid. This Rule checks that noe of the attributes used with a particular role are listed as "prohibited" for that role in the latest version of WAI-ARIA.

The aria-label and aria-labelledby attributes are prohibited on presentation and none roles, as well as on text-like roles such as code, insertion, strong, etc._

### Steps to reproduce

1 - insert the "Follow me" block in a page/post
2 - run Google PageSpeed Insights
3 - check "Accessibility" section

N/A

### What actually happened

N/A

### Impact

Most (> 50%)

### Available workarounds?

### Logs or notes

_No response_

obenland:
Fixes #555.
Fixes #861.

## Proposed changes:

* Adds screenreader text for the dialog modal.

### After

https://github.com/user-attachments/assets/69796b77-a799-44c6-a31d-5778f8daff0c

## Testing instructions:

* Enable VoiceOver in Settings
* Create a post and insert the Follow Me block.
* Visit the frontend and tab through the modal.


obenland:
@mattwiebe What's the best way for me to test the comment dialog?

pfefferle:
@obenland I am not to deep into aria-labels, but FYI https://github.com/Automattic/wordpress-activitypub/issues/861

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
pfefferle, line 8: We should agree on one consistent WIP label. I am fine with `## [Unreleased]`, but have used `## Dev` for now.

obenland: Thanks for the heads-up! I didn't know and went with how https://keepachangelog.com/en/1.1.0/ has it in its example. Now I know! :)

pfefferle: Oh no, that is perfect! wasn't aware of that! then let's use `## [Unreleased]`! (always follow the spec!!!)

My point was only that we agree on something! :)

src/shared/lightbox.scss
```diff
@@ -10,7 +10,7 @@
 	}
 
 	.components-modal__header .components-button:hover {
-		color: var( --wp--preset--color--white );
+		color: #666;
```
pfefferle: this was added, because of day/night mode.

obenland: What would be the best way for me to test that?

pfefferle: a theme that supports light/dark mode and switch between both modes to see if it works in both.

obenland: Which one did you test with?

pfefferle: https://github.com/pfefferle/autonomie ☺️

but this may not be the best one... there might be better ones... 🫣

pfefferle: See https://github.com/Automattic/wordpress-activitypub/issues/721 and https://github.com/Automattic/wordpress-activitypub/issues/425

obenland: Fixed in [29db1be](https://github.com/Automattic/wordpress-activitypub/pull/1034/commits/29db1be3fad07740efd329f0183b05ecd8813f1f). Frankly, I'm a bit confused by it because the background and text colors of the modal don't change, but the hover for the button does.

Even with the changes it doesn't look great in dark mode, but at least it doesn't disappear now.

pfefferle (approved):
works fine, also with my theme ;)

PR #1037, opened 2024-12-03
Hashtag: Ignore tags within HTML attributes (like hex colors)

### LINKED ISSUE #955: css hex colors turned into tags
### Quick summary

Related ticket: 8924019-zd-a8c

Any css hex color codes in the post_content get treated as potential hashtags, and become tags on the post.

e.g.: if your post body includes this
```
<span style="color:#ffffff">text</span>
```

Then the resultant post will get a tag `ffffff` which cannot be removed, as it is reapplied every time the post is edited.

### Steps to reproduce

1. Create a new Post, use the code editor.
2. Enter an HTML tag with a css hex color code in its style. e.g.: `<span style="color:#ffffff">`
3. Save the post
4. Observe that the post now has a tag `ffffff`.
5. Try removing the tag - either in the "Edit Post" screen or via "Quick Edit"
6. The tag is re-added immediately upon removal.

Text that begins with a `#` inside an HTML attribute should be ignored when generating hashtags for a post.

### What actually happened

CSS color codes from the post content are turned into tags.

### Impact

One

### Available workarounds?

Yes, easy to implement

### Logs or notes

_No response_

obenland:
Fixes #955.

## Proposed changes:

* Strips tags before searching for hashtags to convert to tags.
* Adds unit test for various scenarios where hashtags should and shouldn't be converted.

### Other information:

## Testing instructions:

* Create a new Post, use the code editor.
* Enter an HTML tag with a css hex color code in its style. e.g.: `<span style="color: #ffffff">` (Important: preceding whitespace)
* Save the post
* Make sure the post doesn't have a tag called `ffffff`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+* Prevent hex color codes in HTML attributes from being added as post tags.
+
 ## [4.3.0] - 2024-12-02
 
 ### Added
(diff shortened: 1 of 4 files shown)


pfefferle (approved):
this is so simple and smart!

PR #1039, opened 2024-12-04
Shortcodes: Add support for only showing teaser in content shortcode

obenland:
~~The ticket is asking for tagging support in the excerpt, which doesn't really go with how WordPress thinks of the excerpt's use (just text, no links etc), but we have something similar built-in to the content.~~

Looks like [the premise of this PR is moot](https://github.com/Automattic/wordpress-activitypub/pull/1039#issuecomment-2517813478). Not sure if this should still be merged.

The More tag let's users define a teaser for a post and only shows that teaser on index and archive pages. We can use the same mechanism to allow Activitypub users to limit the amount of the content that's getting sent out with the `[ap_content]` shortcode.

This PR introduces a new, undocumented, shortcode attribute called `more_tag`, that will split out the content to only return the teaser, if a more tag is defined in the content. 

## Proposed changes:

* Adds a shortcode attribute to selectively only show the post's teaser.
* Updates shortcode docs to account for the new attribute.
* Adds unit tests

Docs entry after:
Screenshot 2024-12-04 at 9 16 44 AM

### Other information:

## Testing instructions:

* Go to Activitypub settings, select "Note" as Activity-Object-Type, and save changes.
* In post content, insert a `ap_content` shortcode with the `more_tag` attribute set to "yes", and save changes.
* Create a new post that includes a more tag and publish.
* Make sure that it only shows the teaser in Outbox.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Improved
 
 * Added screen reader text to the "Follow Me" block for improved accessibility
+* Added more tag support to content shortcode to control whether to show full post or teaser
 
 ### Fixed
 
--- a/includes/class-shortcodes.php
+++ b/includes/class-shortcodes.php
@@ -157,7 +157,7 @@ class Shortcodes {
 		remove_shortcode( 'ap_content' );
 
 		$atts = shortcode_atts(
-			array( 
+			array(
 				'apply_filters' => 'yes',
 				'more_tag'      => 'no',
 			),
--- a/includes/class-shortcodes.php
+++ b/includes/class-shortcodes.php
@@ -135,7 +135,12 @@ class Shortcodes {
 	/**
 	 * Generates output for the 'ap_content' Shortcode.
 	 *
-	 * @param array  $atts    The Shortcode attributes.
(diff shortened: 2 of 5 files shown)


pfefferle:
`[ap_excerpt]` does support @-replies and hashtags! The mention and hashtag parser will be also executed on the excerpt!

obenland:
Huh, I guess I took the bug report at face value instead of probing further. Let me do that

obenland:
Okay, it looks like the premise for this PR is moot. I don't have a strong preference for this still to be merged. Happy to just close it and write it off as a learning exercise for Cursor and shortcodes.

pfefferle:
Here’s my personal take on the shortcodes implementation in general: they were a creative early solution that offered a lot of flexibility, but I’d like to see them phased out over time. As the community focuses on [making long-form content a first-class citizen of the fediverse](https://socialwebfoundation.org/long-form-text-on-the-fediverse/), I believe we’ll eventually no longer need to rely on exceptions. Templating was essential for crossposting scenarios, but on the fediverse, the focus should be on federating original content as is. Platforms like Mastodon, for instance, don’t have a template feature to dictate how content appears on other servers.

obenland:
Thanks, that makes sense. Let's just close it :)

PR #1041, opened 2024-12-04
Prevent draft posts from being federated when deleted

### LINKED ISSUE #1030: Deleting a Post in status "Draft" sends out an ActivityPub Delete
### Quick summary

I was cleaning up my local drafts via bulk edit and it was taking ages. I checked my Stream logs and saw lots of `activitypub-outgoing` actions.

### Steps to reproduce

1. Create a post in draft.
2. Bulk delete that post.

The ActivityPub should do nothing.

### What actually happened

Deleting this draft is communicated via ActivityPub.

### Impact

All

### Available workarounds?

### Logs or notes

_No response_

obenland:
Fixes #1030.

## Proposed changes:

* Added status transition handling in `schedule_post_activity()` to prevent federation of draft posts.
* Added test coverage for post status transitions.

### Other information:

## Testing instructions:

* Create a draft post
* Create a published post
* Select both posts in the admin list view
* Bulk delete them
* Verify that:
  * The published post sends a Delete activity to followers
  * The draft post is deleted without federation

### DIFF
--- a/includes/class-scheduler.php
+++ b/includes/class-scheduler.php
@@ -138,24 +138,24 @@ class Scheduler {
 			return;
 		}
 
-		$type = false;
+		switch ( $new_status ) {
+			case 'publish':
+				$type = ( 'publish' === $old_status ) ? 'Update' : 'Create';
+				break;
 
-		if (
-			'publish' === $new_status &&
-			'publish' !== $old_status
-		) {
-			$type = 'Create';
-		} elseif (
-			'publish' === $new_status ||
-			// We want to send updates for posts that are published and then moved to draft.
-			( 'draft' === $new_status &&
-			'publish' === $old_status )
-		) {
-			$type = 'Update';
-		} elseif ( 'trash' === $new_status ) {
-			$type = 'Delete';
+			case 'draft':
+				$type = ( 'publish' === $old_status ) ? 'Update' : false;
+				break;
+
+			case 'trash':
+				$type = ( 'publish' === $old_status ) ? 'Delete' : false;
+				break;
+
+			default:
+				$type = false;
 		}
 
+		// No activity to schedule.
 		if ( empty( $type ) ) {
 			return;
 		}
--- a/includes/class-scheduler.php
+++ b/includes/class-scheduler.php
@@ -148,7 +148,7 @@ class Scheduler {
 				break;
 
 			case 'trash':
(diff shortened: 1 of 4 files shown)


pfefferle:
We do not really know the state, a draft had before the delete. A post could have been published, migrated to a draft and then deleted. In this case we have to send a `Delete`.

includes/class-scheduler.php
```diff
@@ -138,24 +138,24 @@ public static function schedule_post_activity( $new_status, $old_status, $post )
...
-		) {
-			$type = 'Update';
-		} elseif ( 'trash' === $new_status ) {
-			$type = 'Delete';
+			case 'draft':
+				$type = ( 'publish' === $old_status ) ? 'Update' : false;
+				break;
+
+			case 'trash':
+				$type = ( 'publish' === $old_status ) ? 'Delete' : false;
```
pfefferle: the check has to be if the post has the `activitypub_status` federated instead of checking the previous one: https://github.com/Automattic/wordpress-activitypub/blob/6c18433965d291459cc00ea885808ffc3bc8dea0/includes/class-activity-dispatcher.php#L194C3-L194C22

https://github.com/Automattic/wordpress-activitypub/blob/6c18433965d291459cc00ea885808ffc3bc8dea0/includes/functions.php#L898

obenland:
> migrated to a draft

Can you tell me more about that? Something that skips `transition_post_status()`?

pfefferle (changes requested):

pfefferle:
@obenland the scenario is:

1. A WordPress user publishes a new Blog-Post.
2. The plugin sends a `Create` Activity.
3. The user then decides to change the status of the Post to draft, because he added the wrong images.
4. The plugin sends an `Update` Activity, to hide the content on Mastodon et al., as long as the post is in draft mode, but it will NOT be deleted.
5. The user then decides that he does not want to have the Post republished and deletes it.
6. Your change will not send anything in that case, because the transition is from "draft" to "trashed".
7. The post will not be deleted from the fediverse

obenland:
That makes sense, thanks. Sounds like the switch to the federated check should cover that?

pfefferle:
it will set when the post/comment will be federated, so I would assume yes :)

pfefferle (approved):

Issue #1044, opened 2024-12-05
Images in the body get mangled in lemmy


db0:
### Quick summary

Example: https://lemmy.dbzer0.com/post/32758507

Image was embedded in the WP post. Got translated as html on lemmy side 

```
<img alt=“” src=“https://i0.wp.com/dbzer0.com/wp-content/uploads/2024/12/image.png?resize=620%2C413&ssl=1” height=“413” width=“620” />
```

### Steps to reproduce

1. post on WP site federated to lemmy.

### What you expected to happen

Images in the body should be converted to something usable by lemmy

### What actually happened

Images are put in as HTML

### Impact

All

### Available workarounds?

No but the platform is still usable

pfefferle:
@db0 this should be reported to lemmy I think!

pfefferle:
Some more context: The plugin does not strip any HTML, because some platforms like friendi.ca support inline images (see image).

![Image](https://github.com/user-attachments/assets/69f3b29a-2b5a-44f9-959d-9972532e85e6)

So I see this in the responsibility of the receiving platform, to remove the HTML that they do not support, instead of encoding it.

pfefferle:
I will close this issue for now, because the plugin seem to work as expected, please feel free to still add comments or re-open if you think it is still a problem.

db0:
One question: Does the plugin know when it's pushing to lemmy vs when it's pushing to friendica?

pfefferle:
we currently do not store this information. but I see what you try to propose :)

my 5 cents: It is way more complicated to have different representations for different platforms. this feels a lot like crossposting, back in the days. I prefer, that every platform should send as much and detailed informations as possible and the receiving should decide what they need and support.

db0:
I do seem to remember that you can declare a post type as text/html or text/markdown like [I do here in fediseer](https://github.com/Fediseer/fediseer/blob/f7370da78ceb1a7ccd6b37481aeaf1860e1b2647/fediseer/messaging.py#L69). Do you do that as well?

db0:
Anyway, the problem here is the usual that happens between two apub implementations. I see it constantly that the developers of each platform interaction which doesn't handle something correctly, insist that the problem is on the other side but ultimately a lot of edge incompatibility like this remains for the end-user.

pfefferle:
[...]

PR #1045, opened 2024-12-05
Add media type support

pfefferle:
improve content handling.

see #1044

### DIFF
--- a/includes/transformer/class-base.php
+++ b/includes/transformer/class-base.php
@@ -128,7 +128,7 @@ abstract class Base {
 	}
 
 	/**
-	 * Returns the default media type for ab Object.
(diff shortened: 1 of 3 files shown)


includes/transformer/class-base.php
```diff
@@ -120,11 +120,22 @@ abstract protected function get_id();
...
 	 * Get the replies Collection.
+	 *
+	 * @return array The replies collection.
 	 */
 	public function get_replies() {
 		return Replies::get_collection( $this->wp_object );
 	}
 
+	/**
+	 * Returns the default media type for ab Object.
```
```suggestion
	 * Returns the default media type for an Object.
```

obenland (approved):

PR #1049, opened 2024-12-06
Update tests to use plugin folder structure

obenland:
Makes it easier to identify file relationships between test and production files and makes filenames easier to read.
It does add a bit of overhead with the namespace being applied to all test files now.

## Proposed changes:

* Mimics folder structure in tests as they are in the plugin.
* Adds missing namespace declarations to test files that don't have them yet.
* Updates filenames to be in line with their production counterparts.
* Fixed a test that failed after the load order changed.

### Other information:

## Testing instructions:

Unit tests should suffice.

Before | After
--- | ---
Screenshot 2024-12-06 at 1 39 33 PM | Screenshot 2024-12-06 at 1 39 18 PM


tests/includes/activity/class-test-activity-dispatcher.php
```diff
@@ -5,12 +5,16 @@
  * @package Activitypub
  */
 
+namespace Activitypub\Tests\Activity;
```
pfefferle: this is (despite of the name) in the root includes folder. Just to keep it consistent, until it will be deleted by the inbox PR ;)

pfefferle (changes requested):

pfefferle (approved):

PR #1050, opened 2024-12-06
Admin: Move plugin settings link into Admin class

obenland:
## Proposed changes:

* Moves settings link callback into Admin class to consolidate it with other wp-admin changes.

## Testing instructions:

* Apply the diff
* Go to the Plugins screen and make sure there's still a "Settings" action link for Activitypub.

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -120,23 +120,6 @@ function plugin_init() {
 	}
 );
 
-/**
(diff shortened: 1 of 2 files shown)


pfefferle (approved):
Nice! And not only moved, but also improved :)

PR #1051, opened 2024-12-06
Use WP Autoloader class

obenland:
This is one of those changes I'm on the fence about as it doesn't strictly improve things, it just makes them "different".
The Autoloader class proposed here is based on [Dion's work for the WP.org Plugin Directory](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/plugin-directory/class-autoloader.php?rev=3083) and used in its proposed form in the `/lib/atomic-v2` library on Dotcom.

I was positively surprised that it seems to just be a drop-in replacement for the current Autoloader, which I didn't quite expect :)
It currently lacks the debug log for when a class is not readable or couldn't be loaded. Was that triggered at any point before?

## Proposed changes:

* Replaces inline Autoloader with a separate Autoloader class that can be re-used.
* Moves the integrations loader to be included with the other includes.

### Other information:

## Testing instructions:

* Make sure everything still works?

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -30,6 +30,10 @@ use WP_CLI;
 require_once __DIR__ . '/includes/compat.php';
 require_once __DIR__ . '/includes/functions.php';
 require_once __DIR__ . '/includes/constants.php';
+require_once __DIR__ . '/integration/load.php';
+
+require_once __DIR__ . '/includes/class-autoloader.php';
+Autoloader::register_path( __NAMESPACE__, __DIR__ . '/includes' );
 
 /**
  * Initialize REST routes.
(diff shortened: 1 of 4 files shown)


activitypub.php
```diff
@@ -161,8 +126,6 @@ function plugin_settings_link( $actions ) {
 	)
 );
 
-// Load integrations.
-require_once __DIR__ . '/integration/load.php';
```
pfefferle, line 148: Can we delete this? It is not the included folder!?!

obenland: It's moved to [line 34](https://github.com/Automattic/wordpress-activitypub/pull/1051/files#diff-d33bfd446d4ad05f88e9c8e907f31b2e4c06eca40c7f28cee7e58f3a2302958bR34), to group with the other includes.

pfefferle: Ah, haven't seen that! Thanks!

pfefferle (dismissed):
I like it and it looks way cleaner

includes/class-autoloader.php
```diff
@@ -0,0 +1,106 @@
...
+	/**
+	 * Registers Autoloader's autoload function.
+	 *
+	 * @throws \Exception When autoload_function cannot be registered.
+	 *
+	 * @param string $prefix Namespace prefix all classes have in common.
+	 * @param string $path   Path to the files to be loaded.
+	 */
+	public static function register_path( $prefix, $path ) {
+		$loader = new Autoloader( $prefix, $path );
```
pfefferle: Maybe use `self` if the file gets renamed sometime?

```suggestion
		$loader = new self( $prefix, $path );
```

pfefferle:
Maybe escaping the global functions!?!

activitypub.php
```diff
@@ -81,6 +85,22 @@ function plugin_init() {
 }
 \add_action( 'plugins_loaded', __NAMESPACE__ . '\plugin_init' );
 
+/**
+ * Add plugin settings link.
+ *
+ * @param array $actions The current actions.
+ */
+function plugin_settings_link( $actions ) {
```
pfefferle: haven't we moved that to the `class-admin.php`?

obenland:
> Maybe escaping the global functions!?!

What do you mean with that?

pfefferle:
Why have I used "escape" and I have no idea how it is called 🫣... I mean `\spl_autoload_register` instead of `spl_autoload_register`.

pfefferle (approved):

PR #1052, opened 2024-12-06
Integrations: Use Autoloader

obenland:
See #1051.

If we decide to move forward with #1051, we could also use it to autoload integrations files.

## Proposed changes:

* Creates a new Autoloader instance for integrations files.
* Remove file inclusions in favor of autoloading.

### Other information:

## Testing instructions:

* Install any of the plugins that we have integrations for and make sure they still work.


pfefferle (approved):
Maybe we need some kind of integration tests for the longer term... for now, I tested all of these by installing all the plugins for now...

PR #1053, opened 2024-12-09
This adds Icon support for Audio and Video attachments

### LINKED ISSUE #976: Feature Request: Add option to use the featured image of the post as profile image when publishing a podcast
### What

When [publishing](https://wikipediapodden.se/raderade-megafoner-287/) a podcast [in a way](https://mastodon.social/@pfefferle/113441876155981589) that Mastodon picks up the audio file and makes it playable directly on Mastodon, the image of the author is used as the image for the audio, see example below.

Image

The WordPress post has a featured image set, which is a cover art specifically for this episode. I would like an option to use the featured image instead of the author image.

### Why

It would be great for podcasts with unique artwork for each episode to draw more attention to the content of the podcast rather than who was the author of the post on WordPress. (Idea sparked from [this conversation](https://mastodon.social/@pfefferle/113448855634550530).)

### How

I imagine a simple toggle under the Fediverse dropdown on the post where this can be enabled (but it could also be a general setting rather than per post, I don't mind either).

pfefferle:
I changed all functions from static to the object context. Any reason why we used static functions that might cause issues after the change @mattwiebe ?

Fixes #976


obenland:
It would be great to add unit tests that cover the new functionality

pfefferle:
will wait for #1049 to be merged, to use the new structure

includes/transformer/class-post.php
```diff
@@ -208,7 +209,67 @@ protected function get_image() {
...
+	 *
+	 * @return array|null The Icon or null if no icon is available.
+	 */
+	protected function get_icon() {
+		$post_id = $this->wp_object->ID;
+		$id      = null;
+
+		// List post thumbnail first if this post has one.
+		if (
+			\function_exists( 'has_post_thumbnail' ) &&
```
obenland: What are circumstances where the function doesn't exist?

includes/transformer/class-post.php
```diff
@@ -208,7 +209,67 @@ protected function get_image() {
...
+	}
+
+	/**
+	 * Returns an Icon, based on the Featured Image with a fallback to the site-icon.
+	 *
+	 * @return array|null The Icon or null if no icon is available.
+	 */
+	protected function get_icon() {
+		$post_id = $this->wp_object->ID;
+		$id      = null;
```
obenland: `$id` gets never used, it's either the post thumbnail or the the option later on.

mattwiebe:
> I changed all functions from static to the object context. Any reason why we used static functions that might cause issues after the change @mattwiebe ?

Seems like the right thing to do!

includes/transformer/class-post.php
pfefferle: this was a copy and paste thing and I have no idea why I added it in the first place :(

obenland (approved):

PR #1054, opened 2024-12-09
Version bump 4.4.0

obenland:
## Proposed changes:

* Bumps versions to 4.4.0


pfefferle:
You have to add the links below and above the comment: https://github.com/Automattic/wordpress-activitypub/blob/d9a7c6677ca10313d4f089a81a1cff5bcaf77f1e/CHANGELOG.md?plain=1#L1129

obenland:
I have never seen this before! 😄 (As in, never scrolled down that file far enough)

pfefferle:
Because WordPress.org limits the size of the readme, you could also remove old changelog items. I would remove all `4.1.X` and below.

pfefferle (approved):
that's it!

PR #1065, opened 2024-12-09
Improve Interactions moderation

### LINKED ISSUE #1060: Improve Interactions moderation
Hide the Reply & History link
Add comment types to comment type filter

pfefferle:
Fixes #1060

Screenshot 2024-12-09 at 18 33 08

* [x] Check if Akismet code works like expected

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -595,19 +595,13 @@ class Admin {
 	 * @return array The modified actions.
 	 */
 	public static function comment_row_actions( $actions, $comment ) {
-		global $activitypub_comment_types;
-
 		if ( was_comment_received( $comment ) ) {
 			unset( $actions['edit'] );
 			unset( $actions['quickedit'] );
 		}
-
-		$comment_type = get_comment_type( $comment );
-
-		foreach ( $activitypub_comment_types as $type ) {
-			if ( $type['type'] === $comment_type ) {
-				unset( $actions['reply'] );
-			}
+		
+		if ( in_array( get_comment_type( $comment ), Comment::get_comment_type_names(), true ) ) {
+			unset( $actions['reply'] );
 		}
 
 		return $actions;
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -599,7 +599,7 @@ class Admin {
 			unset( $actions['edit'] );
 			unset( $actions['quickedit'] );
 		}
-		
+
 		if ( in_array( get_comment_type( $comment ), Comment::get_comment_type_names(), true ) ) {
 			unset( $actions['reply'] );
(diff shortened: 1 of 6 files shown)


includes/class-admin.php
```diff
@@ -594,11 +595,21 @@ function ( $views ) {
...
 			unset( $actions['quickedit'] );
 		}
 
+		$comment_type = get_comment_type( $comment );
+
+		foreach ( $activitypub_comment_types as $type ) {
+			if ( $type['type'] === $comment_type ) {
+				unset( $actions['reply'] );
+			}
+		}
```
obenland:
```suggestion
	public static function comment_row_actions( $actions, $comment ) {
		if ( was_comment_received( $comment ) ) {
			unset( $actions['edit'] );
			unset( $actions['quickedit'] );
		}
		
		if ( in_array( get_comment_type( $comment ), Comment::get_comment_type_names(), true ) ) {
			unset( $actions['reply'] );
		}
```

I forgot this was a global. We have an API for it, should we use that instead?

includes/class-admin.php
```diff
@@ -668,6 +679,27 @@ public static function manage_comments_custom_column( $column, $comment_id ) {
...
 
+	/**
+	 * Add the new ActivityPub comment types to the comment types dropdown.
+	 *
+	 * @param array $types The existing comment types.
+	 *
+	 * @return array The extended comment types.
+	 */
+	public static function comment_types_dropdown( $types ) {
+		global $activitypub_comment_types;
```
obenland: Same as above, should we use `Comment::get_comment_types()` instead of the global?

includes/class-admin.php
```diff
@@ -668,6 +679,27 @@ public static function manage_comments_custom_column( $column, $comment_id ) {
...
+
+		if ( ! is_array( $activitypub_comment_types ) ) {
```
obenland: What are circumstances where it wouldn't be an array?

includes/class-admin.php
pfefferle: good point!

includes/class-admin.php
pfefferle: Vorsicht ist die Mutter der Porzellankiste! ☺️

Because it is a global, it can be changed everywhere without any validation. But this might be the wrong place to check it. I think you are right with using the Comment functions and it might make sense to add the check there.

obenland: It's a global we set. If someone or something manipulates it to change its type, that should be so rare that we shouldn't have to account for it. Especially with good API functions for it where it wouldn't need to be manipulated directly.

obenland (approved):

PR #1070, opened 2024-12-10
Add a reaction block

### LINKED ISSUE #235: Show likes from fediverse

### LINKED ISSUE #1057: Template blocks for Likes and Reposts: display them on the frontend
Create a general reactions block that contains both likes and reblogs as a face pile.

See https://notiz.blog/2024/09/13/the-fediverse-files/

mattwiebe:
This will add a block to display Likes and Reposts in a single block. 

Fixes #1057 

## Proposed changes:
* Adds a Reactions block
* It automatically gets hooked to be placed after the post content block
* Enable reactions by default!

A (not quite up to date) video of it in action:

https://github.com/user-attachments/assets/e8d381b2-9627-489d-a528-b080b0ab498a

### Other information:

## Testing instructions:
With this PR active, view a post on the frontend that already has a number of Likes and Reposts. The reactions block should be inserted directly after the post content:

Screenshot 2024-12-12 at 16 04 57

The block is automatically added via a block hook. It should be opted in, and can be disabled in the Single Post template in the Site Editor: (also note the dummy content in the Editor)

Screenshot 2024-12-12 at 16 06 31

Every row of icons, if there are too many to fit the width, should limit the number of avatars shown so that none are cut off.

Screenshot 2024-12-12 at 16 15 35

Avatars and names are shown if you click on the **n reposts** (or likes) label to the right of the avatars:

Screenshot 2024-12-12 at 16 17 04

On a post with no likes or reposts, nothing will be shown, not even the title.

The title ("Fediverse reactions" by default) is editable.


pfefferle:
Awesome!! 🎉

Maybe one first little feedback: Maybe we should have a line break after the label (and use it more as a headline) so that the icon lists are all aligned?

mattwiebe:
> Maybe one first little feedback: Maybe we should have a line break after the label (and use it more as a headline) so that the icon lists are all aligned?

They should definitely be aligned! I'd prefer to keep them on a single line, maybe responsively moving to separate lines on smaller screens. Or I guess we could give some layout options, but that could be a future iteration. 🤔

jeherve:
Nice!

I believe this would fix #235 as well?

pfefferle:
@mattwiebe what about icons at the beginning of each line https://wordpress.github.io/gutenberg/?path=/docs/icons-icon--docs ???

* `reusableBlock` for `reposts`
* and the heart you are already using for `like`

We could use the icon attribute for that. Either store the `@wordpress/icons` key or an SVG? https://github.com/Automattic/wordpress-activitypub/blob/e7fa73c84669d8564050490cdd0ed6062431ef93/includes/class-comment.php#L616

includes/rest/class-interaction.php
```diff
@@ -43,6 +43,24 @@ public static function register_routes() {
...
+				'callback'            => array( self::class, 'get_reactions' ),
+				'permission_callback' => '__return_true',
+				'args'                => array(
+					'id' => array(
+						'required'          => true,
+						'validate_callback' => function ( $val ) {
+							return is_numeric( $val );
+						},
+					),
+				),
```
obenland: We could just copy [what Core does for the post id](https://github.com/WordPress/wordpress-develop/blob/af1682d4d327655781bc0b18db0f5f31f06d9b2b/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L114-L121). Setting the type to integer triggers the built-in validation and sanitization calls

mattwiebe: Even better!

pfefferle:
Is it possible to use the comment type registry instead of hardcoding to the two types? So that it is easier to add new types, either in the plugin or an external one?

pfefferle:
Does not have to be in the first iteration!

mattwiebe:
> Is it possible to use the comment type registry instead of hardcoding to the two types? So that it is easier to add new types, either in the plugin or an external one?

Done in https://github.com/Automattic/wordpress-activitypub/pull/1070/commits/fb55c1f99f23c9128ff968f086e616f940190fc9 ! (props Windsurf lol)

pfefferle:
Can you ask Windsurf to also update the react code to support them? 🥸

pfefferle:
I ask for a friend... 😅

mattwiebe:
> Can you ask Windsurf to also update the react code to support them? 🥸
> I ask for a friend... 😅

We had another good collaboration https://github.com/Automattic/wordpress-activitypub/pull/1070/commits/c37516344ea74164ca837074d30396fbfddd98e3 - this is the type of stuff I would wave off to keep focused on shipping but refactors like this go surprisingly well.

includes/rest/class-interaction.php
```diff
@@ -43,6 +44,22 @@ public static function register_routes() {
 				),
 			)
 		);
+
+		\register_rest_route(
+			ACTIVITYPUB_REST_NAMESPACE,
+			'/reactions/(?P<id>\d+)',
```
pfefferle: This reads like the ID is the ID of a reaction, but these are the reactions for this given ID, so maybe we should change it to something like `/posts/(?P<id>\d+)/reactions` instead.

That would be also more consistent with the comments and other posts endpoints:

* https://github.com/Automattic/wordpress-activitypub/blob/253cffff12291ae2ae40538ae16fefc44b69e9a3/includes/rest/class-comment.php#L38
* https://github.com/Automattic/wordpress-activitypub/blob/253cffff12291ae2ae40538ae16fefc44b69e9a3/includes/rest/class-collection.php#L81

In this context, perhaps we can also create a dedicated `class-post.php` file, to encapsulate the post endpoints.

pfefferle:
> The block is automatically added via a block hook. It should be opted in, and can be disabled in the Single Post template in the Site Editor: (also note the dummy content in the Editor)

I do not really understand if and how it is possible to have that deactivated and place it in the template. It seems that every time I place the block somewhere else, the checkbox is activated again. But this maybe me, simply not understanding FSE properly enough :( 

reactions

Besides of that, the preview got lost while trying this. And if I place the block in the comment section, it shows up twice and if I deactivate the hook again, it removes both.

Screenshot 2024-12-13 at 10 03 56

I tested with "Twenty Twenty-Five"

pfefferle:
On "Twenty Twenty-Five" the icons do overlap the text a bit:

Screenshot 2024-12-13 at 10 01 52

It seems to be a problem with the length of the line.

Tested with latest Firefox version on the Mac.

pfefferle:
I also tested adding another comment-type and it worked like a charm!!! 🎉🎉🎉🎉

Screenshot 2024-12-13 at 10 12 58

mattwiebe:
@pfefferle the editor component appears to have broken. That's why nothing was showing for you, there. Should be fixed by https://github.com/Automattic/wordpress-activitypub/pull/1070/commits/c4a87a2e7c34096b1db5fef3fcbc9f9279f5f5e0

includes/class-federated-reactions-settings.php
```diff
@@ -0,0 +1,105 @@
+<?php
+/**
+ * Federated Reactions Settings file.
+ *
+ * @package ActivityPub
+ */
+
+namespace Activitypub;
```
pfefferle: I would put the file in a subfolder and use the namespace `Activitypub\Settings` or similar. Then we could migrate the rest of the settings over time.

pfefferle: and do we have to use `federated` in the filename?

pfefferle:
@mattwiebe nice! I will re-test on Monday!

pfefferle:
@mattwiebe the FSE editing experience seems to work fine now, but I can't see any likes in the frontend!

includes/class-blocks.php
```diff
@@ -21,12 +21,15 @@ public static function init() {
...
 
-		\add_action( 'wp_enqueue_scripts', array( self::class, 'add_data' ) );
-		\add_action( 'enqueue_block_editor_assets', array( self::class, 'add_data' ) );
+		\add_action( 'wp_enqueue_scripts', array( self::class, 'inject_activitypub_options' ) );
+		\add_action( 'admin_print_scripts', array( self::class, 'inject_activitypub_options' ) );
 		\add_action( 'load-post-new.php', array( self::class, 'handle_in_reply_to_get_param' ) );
 		// Add editor plugin.
 		\add_action( 'enqueue_block_editor_assets', array( self::class, 'enqueue_editor_assets' ) );
 		\add_action( 'init', array( self::class, 'register_postmeta' ), 11 );
+		if ( ! \wp_is_block_theme() ) {
```
pfefferle: @obenland FYI https://github.com/Automattic/wordpress-activitypub/blob/8770b5b3b8c3a696ca8e0f919e9b1cf2acf82a93/includes/functions.php#L571

includes/class-federated-reactions-settings.php
```diff
@@ -0,0 +1,105 @@
...
+			);
+		}
+	}
+
+	/**
+	 * Register settings.
+	 */
+	public static function register_settings() {
+		register_setting(
+			'discussion',
```
pfefferle: @obenland I like the integration on the discussion page, but I would also add the setting to the ActivityPub options. I would assume that users will first search in the plugins settings.

includes/class-blocks.php
pfefferle: It is a bit misleading if you have the setting even on block (FSE) themes. Why would you add the block to your frontend if you then deactivate it through the settings?

includes/class-blocks.php
```diff
@@ -21,12 +21,15 @@ public static function init() {
...
+			\add_action( 'the_content', array( self::class, 'maybe_add_fediverse_reactions' ) );
```
pfefferle: @obenland I would prefer using one of the comment hooks! The `the_content` hook is always tricky in combination with for example sharing plugins, because it will always be part of the content! It also produces ugly artefacts in combination with Webmentions!

What about `comment_form_before` and `comment_form_comments_closed` instead?


https://github.com/pfefferle/wordpress-webmention/blob/c0312741a98dcf99ad7c21cf3c7ffae89035fb64/includes/class-comment-walker.php#L23

pfefferle:
@mattwiebe the animation takes a bit of getting used to 😳

includes/class-blocks.php
pfefferle: what about setting the complete init of the class behind `if ( ! \wp_is_block_theme() ) {`???

includes/class-blocks.php
```diff
@@ -358,4 +400,24 @@ public static function render_follower( $follower ) {
...
+		if ( ! is_singular() ) {
+			return $content;
+		}
+
+		if ( ! Federated_Reactions_Settings::is_reactions_enabled() ) {
+			return $content;
+		}
+
+		// Create a block instance for proper rendering.
+		$block_content = do_blocks( '<!-- wp:activitypub/reactions {"title":"' . esc_attr__( 'Fediverse Reactions', 'activitypub' ) . '"} /-->' );
```
pfefferle, line 420: what about if the site does not support blocks at all?

includes/class-blocks.php
pfefferle: (maybe not high priority, because it only adds the empty container `<div>` ant the rest is JS)

includes/class-blocks.php
pfefferle: Oh, you only can add the block to your template if the settings is enabled!?!

This still feels a bit inconsistent! For the classic editor, the setting will directly add the block to the output and for block themes it "only" allows you to place the block somewhere.

pfefferle: An other example: I am working on a hybrid theme, that supports blocks in the post and in the template using `block_template_part`. `wp_is_block_theme` is false for this theme, so I only have the option to show the reactions as part of the content, even if it would be possible to place it somewhere else on the site. Because if I deactivate the setting, I will not be able to place it somewhere else, and if it is activated, it will directly added to the content.

includes/class-federated-reactions-settings.php
```diff
@@ -0,0 +1,105 @@
...
+	public static function render_reactions_enabled_field() {
+		?>
+		<fieldset>
+			<legend class="screen-reader-text"><?php esc_html_e( 'Federated Reactions', 'activitypub' ); ?></legend>
+			<label>
+				<input type="checkbox" name="activitypub_reactions_enabled" value="1" <?php checked( '1', get_option( 'activitypub_reactions_enabled', '1' ) ); ?> />
+				<?php esc_html_e( 'Show federated reactions on posts', 'activitypub' ); ?>
+			</label>
+			<p class="description">
+				<?php esc_html_e( 'This can be overridden on individual posts.', 'activitypub' ); ?>
```
pfefferle: You can not, if you use the classic editor for example.

includes/class-blocks.php
obenland: yeah, no strong opinion here. @mattwiebe suggested `the_content`, happy to go either direction.

includes/class-federated-reactions-settings.php
obenland: Fixed in [2b5c696](https://github.com/Automattic/wordpress-activitypub/pull/1070/commits/2b5c69694a4ff539416220729013b05fbad55763)

includes/class-federated-reactions-settings.php
obenland: Fixed the filename in [c5d6551](https://github.com/Automattic/wordpress-activitypub/pull/1070/commits/c5d65516b52b0684472f71d6de8701967e4c41b2).

obenland: > I would put the file in a subfolder and use the namespace `Activitypub\Settings` or similar. Then we could migrate the rest of the settings over time.

Agreed! I'd even go a step further and create an admin subfolder and have Settings as part of that, to bundle all admin-related functionality. Let's do that as a follow-up?

includes/class-blocks.php
obenland: > Because if I deactivate the setting, I will not be able to place it somewhere else

Why not? Should it not behave like any other block?

pfefferle: Let me ask it differently: How can I add the block somewhere else in the hybrid theme, without having it added to the post?

pfefferle: Because the setting has a different meaning in the different modes, I can not add the block if the setting is disabled. But if I enable the setting, the code adds the block to the `the_content` hook, because it "thinks" it is a classic theme!

Screenshot 2024-12-16 at 16 04 01

pfefferle: I would keep the block-template (FSE) behaviour completely out of that. Enabling the setting should only be to add/remove it from the `the_content` hook... like a `legacy` mode...

The per post setting on the other hand can work for both.

includes/class-federated-reactions-settings.php
obenland: Fixed in 3129aea7ffed1fc8e028378c294acff72b3e0fd8

obenland:
### Remaining Tasks

- [x] Limit Reaction Setting to non-block themes.
- [x] Per-post setting should be for both, global setting only for non-block themes.
- [x] Add a link to Site Editor for block themes in Settings.
- [x] Tone down hover animation.

mattwiebe:
@Automattic/fediverse Ready to test some more, I think everything should be addressed!

includes/class-reactions-settings.php
```diff
@@ -0,0 +1,216 @@
...
+	/**
+	 * Check if reactions are enabled for a post.
+	 *
+	 * @param int $post_id The post ID.
+	 * @return bool
+	 */
+	public static function is_reactions_enabled( $post_id = 0 ) {
+		// Check global setting first - if disabled globally, return false.
+		if ( ! self::is_reactions_globally_enabled() ) {
+			return false;
```
pfefferle: This will still prevent, that the block will load outside of the post.

pfefferle (dismissed):
tests are failing

obenland:
It looks like aab821d4 broke the unit test that checked whether reactions are enabled for an individual post while they are deactivated globally.
Is that something we don't want to support?

pfefferle:
If we want to have a setting that deactivates the feature globally then we need to have a different setting to add/remove the reactions to the post, these two should not be implicit!

mattwiebe:
We've removed support for auto-appending the Reactions block to classic themes. Those users can either insert the block manually, or edit their PHP templates to output the block like:

```php
do_blocks( '<!-- wp:activitypub/reactions {"title":"Fediverse Reactions"} /-->' );
```

We may revisit this, but it was getting too complex and blocking shipping

obenland (approved):

PR #1071, opened 2024-12-11
Remove asset updater

pfefferle:
This is meant to work perfect with dev branches but not how we use GitHub.

Besides of that, the plugin is so actively developed, so that it makes no sense either.

---

Context:

This action updates the assets, like images and the readme, but only if there is no other change in the repo.


obenland:
What's the harm in keeping it?

pfefferle:
It always runs on a PR and shows an error :)

It will maybe work once in 1000 PRs.

obenland:
Ah, yes, showing the error is something I don't understand about that action either.

obenland (approved):

PR #1072, opened 2024-12-11
Improved comment-type handling

pfefferle:
It now uses the comment-type as index and keeps a list of mapped activity-types.

Also added a function to get the comment-type by the activity-type.

`Comment::get_comment_type_by_activity_type `

props @obenland

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -541,36 +541,25 @@ class Comment {
 	/**
 	 * Is this a registered comment type.
 	 *
-	 * @param string $slug The name of the type.
+	 * @param string $name The name of the type.
+	 *
 	 * @return boolean True if registered.
 	 */
-	public static function is_registered_comment_type( $slug ) {
-		$slug = \strtolower( $slug );
-		$slug = \sanitize_key( $slug );
+	public static function is_registered_comment_type( $name ) {
+		$name = \strtolower( $name );
+		$name = \sanitize_key( $name );
 
 		$registered_comment_types = self::get_comment_types();
 
-		return isset( $registered_comment_types[ $slug ] );
+		return isset( $registered_comment_types[ $name ] );
 	}
 
 	/**
 	 * Return the registered custom comment types names.
 	 *
-	 * @deprecated 4.5.0 Use `self::get_comment_type_slugs` instead.
-	 *
 	 * @return array The registered custom comment type names.
 	 */
 	public static function get_comment_type_names() {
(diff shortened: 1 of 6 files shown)


includes/class-comment.php
```diff
@@ -523,8 +545,8 @@ public static function get_comment_types() {
 	 * @return boolean True if registered.
 	 */
 	public static function is_registered_comment_type( $slug ) {
-		$slug = strtolower( $slug );
-		$slug = sanitize_key( $slug );
+		$slug = \strtolower( $slug );
+		$slug = \sanitize_key( $slug );
 
 		return in_array( $slug, array_keys( self::get_comment_types() ), true );
```
```suggestion
		$registered_comment_types = self::get_comment_types();

		return isset( $registered_comment_types[ $slug ] );
```
Would this work? I wonder if that's faster.

obenland:
With this change, `get_comment_type_names()` could be deprecated in favor of  `get_comment_type_slugs()` and just return `array_keys( self::get_comment_types() )`

includes/class-comment.php
```diff
@@ -520,14 +542,17 @@ public static function get_comment_types() {
 	/**
 	 * Is this a registered comment type.
 	 *
-	 * @param string $slug The name of the type.
+	 * @param string $name The name of the type.
```
obenland: I'd leave it as slug, as that's what it is (hence my proposal to rename `get_comment_type_names()` to `get_comment_type_slugs()`). The name would be the translated "Repost" and "Like".

pfefferle: so reverting this commit?

obenland: oh, I didn't realized you had changed it and reverted again. Yes, I think so, if you don't have a strong preference. [WP uses slug for these things](https://developer.wordpress.org/?s=slug).

includes/collection/class-interactions.php
```diff
@@ -111,14 +111,13 @@ public static function add_reaction( $activity ) {
 			return false;
 		}
 
-		$type = $activity['type'];
+		$comment_type = Comment::get_comment_type_by_activity_type( $activity['type'] );
```
obenland, line 114: Could there ever be multiple comment types for an activity type? Is that something we have to worry about?

obenland (approved):

includes/collection/class-interactions.php
pfefferle: I don't think so... there is the possibility that we might store items in different places (direct messages for example), but that is a different discussion...

PR #1073, opened 2024-12-11
Comments: Account for custom comment types when calculating comment counts

### LINKED ISSUE #1069: Fix comment count
On posts that receive Likes and/or Reposts, the comment count is wrong. We should exclude likes and reposts from the count.

Eg: this post: https://mattwiebe.blog/2024/11/24/support-week-for-activitypub-for-wordpress/

Image

Shows "34 responses" in the comments section. It should only show 5 responses, after the 29 combined likes and reposts are removed from it.

List of possible hooks:

* [ ] [`get_comments_number`](https://developer.wordpress.org/reference/functions/get_comments_number/)
* [ ] [`wp_count_comments`](https://developer.wordpress.org/reference/functions/wp_count_comments/)
* [X] [`pre_wp_update_comment_count_now`](https://developer.wordpress.org/reference/functions/wp_update_comment_count_now/)
* [ ] [`get_page_of_comment_query_args`](https://developer.wordpress.org/reference/functions/get_page_of_comment/)?
* [ ] [`parse_comment_query`](https://developer.wordpress.org/reference/hooks/parse_comment_query/)?

obenland:
Fixes #1069.

The benefit of this approach is that it doesn't add a DB query to front end requests.
The downsides are that if another plugin uses the same filter, there's a potential for conflict, and comment counts won't change until WordPress finds a reason to update the count, like approving/unapproving a comment.

One more consideration:
We might want to add an upgrade routine so we can grab all posts with a like or reblog and update their counts, after the plugin was upgraded ([example](https://github.com/WordPress/wordpress-develop/commit/d5a7a3a45dc18ad1554e185c6611ec54ced9b40c)).

There is almost certainly more work needed to make this work correctly.

## Proposed changes:

* Adds callback that queries comment counts without Activitypub custom comment types.

### Other information:

## Testing instructions:

* Apply the diff.
* Go to the comments list on your development site.
* Approve/unapprove a like or repost.
* Make sure the new comment count reflects that number of actual comments (and pings).

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -671,10 +671,6 @@ class Comment {
 			return;
 		}
 
-		if ( isset( $query->query_vars['count'] ) && true === $query->query_vars['count'] ) {
-			return;
-		}
-
 		// Exclude likes and reposts by the ActivityPub plugin.
 		$query->query_vars['type__not_in'] = self::get_comment_type_names();
 	}
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -30,6 +30,7 @@ class Comment {
 		\add_action( 'pre_get_comments', array( static::class, 'comment_query' ) );
 		\add_filter( 'pre_comment_approved', array( static::class, 'pre_comment_approved' ), 10, 2 );
 		\add_filter( 'get_avatar_comment_types', array( static::class, 'get_avatar_comment_types' ), 99 );
+		\add_filter( 'pre_wp_update_comment_count_now', array( static::class, 'pre_wp_update_comment_count_now' ), 10, 3 );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


includes/class-comment.php
```diff
@@ -715,4 +716,27 @@ public static function pre_comment_approved( $approved, $commentdata ) {
...
+	 * @return int|null The updated comment count, or null to use the default query.
+	 */
+	public static function pre_wp_update_comment_count_now( $new_count, $old_count, $post_id ) {
+		if ( null === $new_count ) {
+			global $wpdb;
+
+			$excluded_types = self::get_comment_type_names();
+
+			// phpcs:ignore WordPress.DB
+			$new_count = (int) $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->comments WHERE comment_post_ID = %d AND comment_approved = '1' AND comment_type NOT IN ('" . implode( "','", $excluded_types ) . "')", $post_id ) );
```
pfefferle, line 732: why not count all "comment" instead?

obenland: Pingbacks and trackbacks are part of the count, too. And we don't know what other custom comment types might exist

pfefferle: good time to fix that too ;)

pfefferle:
I've done some testing, how these changes behave with different settings and it seems that if you enable "paging" for comments, it shows no comments at all. Will see how this happens.

This is not directly because of this change, but maybe we could use this PR to also fix that.

pfefferle:
I fixed paging for classic themes, and it seems to also work fine for FSE themes.

pfefferle (approved):

PR #1075, opened 2024-12-12
Add a CLI command to generate demo reactions (comments, likes, reposts)

pfefferle:
This helps with testing new features!

```bash
$ wp activitypub generate comments --count=100 --post_id=123
```

### DIFF
--- a/includes/class-cli.php
+++ b/includes/class-cli.php
@@ -9,6 +9,7 @@ namespace Activitypub;
 
 use WP_CLI;
 use WP_CLI_Command;
+use WP_CLI\Utils;
 
 /**
  * WP-CLI commands.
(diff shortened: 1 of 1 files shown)


obenland:
Maybe this could just live in a branch, to be used when needed?

pfefferle:
@obenland What are you afraid of?

obenland:
It serves a very specific need, doesn't seem to be needed outside of development and possible even outside of this one feature. [Theme Unit Test](https://make.wordpress.org/themes/handbook/review/theme-unit-test/) covers a lot of the need for test data, too.

pfefferle:
But the frontend for example is hard to test automatically.

The WP CLI provides some similar features btw.

* https://developer.wordpress.org/cli/commands/post/generate/
* https://developer.wordpress.org/cli/commands/comment/generate/

obenland:
Given the existence of [developer.wordpress.org/cli/commands/comment/generate](https://developer.wordpress.org/cli/commands/comment/generate/) I think there's even less of a need for a custom one here.

I don't feel overly strongly about this, so please feel free to move forward regardless of my feedback. To me, CLI commands that ship with the plugin should make plugin functionality more accessible or easier to use for developers, which this doesn't seem to fit.

Issue #1077, opened 2024-12-13
Separate API endpoints more strictly/use different varification method


pfefferle:
The current codebase does not strictly separate the endpoints we need "locally", like for example the endpoints needed by the blocks, from the ActivityPub APIs.

[Update] An alternate solution #1078 

This could cause some issues in the future, or at lest bring some more complexity, when it comes to signature verification for example. Request to the ActivityPub endpoints should be verified, local requests might require a session or are open in general.

This is already a problem when you turn on `Authorized-Fetch`!!!

akirk:
In #1078 you show that technically it's not necessary, more of a question of adding the right permisson_callbacks. I don't disagree with separating it out but it might be just more efficient for now to be very dilligent about the callbacks.

obenland:
A couple of options come to mind for me. We could group Editor related endpoints in a subdirectory of `/includes/rest` and give them their own namespace. We could also have all endpoint classes extend the same class that has the signature verification as one of its methods, so they all inherit it.

pfefferle:
I think we can discuss other ways to optimize the structure, but #1078 should fix the issue for now.

PR #1078, opened 2024-12-13
Use a more explicit signature verification

pfefferle:
This is a proposal to use the `permission_callback`, instead of a general hook, to verify signatures.

The advantage is, that it is easier to enable/disable verification for specific endpoints this way.

See #1077

### DIFF
--- a/includes/rest/class-server.php
+++ b/includes/rest/class-server.php
@@ -36,7 +36,7 @@ class Server {
 	 */
 	public static function add_hooks() {
 		\add_filter( 'rest_request_before_callbacks', array( self::class, 'validate_activitypub_requests' ), 9, 3 );
-		//\add_filter( 'rest_request_before_callbacks', array( self::class, 'authorize_activitypub_requests' ), 10, 3 );
+		// \add_filter( 'rest_request_before_callbacks', array( self::class, 'authorize_activitypub_requests' ), 10, 3 );
 		\add_filter( 'rest_request_parameter_order', array( self::class, 'request_parameter_order' ), 10, 2 );
 	}
 
--- a/includes/rest/class-server.php
+++ b/includes/rest/class-server.php
@@ -35,7 +35,7 @@ class Server {
 	 * Add sever hooks.
 	 */
 	public static function add_hooks() {
-		\add_filter( 'rest_request_before_callbacks', array( self::class, 'validate_activitypub_requests' ), 9, 3 );
+		\add_filter( 'rest_request_before_callbacks', array( self::class, 'validate_requests' ), 9, 3 );
 		\add_filter( 'rest_request_parameter_order', array( self::class, 'request_parameter_order' ), 10, 2 );
 	}
 
@@ -139,7 +139,7 @@ class Server {
 	 *
 	 * @return mixed|WP_Error The response, error, or modified response.
 	 */
-	public static function validate_activitypub_requests( $response, $handler, $request ) {
+	public static function validate_requests( $response, $handler, $request ) {
 		if ( 'HEAD' === $request->get_method() ) {
 			return $response;
 		}
--- a/includes/rest/class-server.php
+++ b/includes/rest/class-server.php
@@ -78,7 +78,7 @@ class Server {
 	 * The function is meant to be used as a permission callback for the rest api.
 	 *
 	 * It verifies the signature of POST, PUT, PATCH and DELETE requests and also the GET requests in secure mode.
-	 * You can use the filter 'activitypub_defer_signature_verification' to defer the signature verification.
+	 * You can use the filter 'activitypub_defer_verify_signature' to defer the signature verification.
 	 * The HEAD request is always bypassed.
 	 *
 	 * @see https://www.w3.org/wiki/SocialCG/ActivityPub/Primer/Authentication_Authorization#Authorized_fetch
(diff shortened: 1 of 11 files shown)


pfefferle:
@mediaformat any idea why we have not used it in the first place?

includes/rest/class-server.php
```diff
@@ -145,6 +146,55 @@ public static function authorize_activitypub_requests( $response, $handler, $req
...
+	 * @see WP_REST_Request
+	 *
+	 * @see https://www.w3.org/wiki/SocialCG/ActivityPub/Primer/Authentication_Authorization#Authorized_fetch
+	 * @see https://swicg.github.io/activitypub-http-signature/#authorized-fetch
+	 *
+	 * @param \WP_REST_Request $request  Request used to generate the response.
+	 *
+	 * @return bool Whether the request is authorized.
+	 */
+	public static function authorize_activitypub_permissions( $request ) {
```
jeherve: Do you think it would make sense to have a more specific name for that method, to indicate what it does exactly (check for `HEAD` and verify signature), and to separate for any future permission callback you may have for other endpoints, that will require different permissions?

includes/rest/class-server.php
```diff
@@ -145,6 +146,55 @@ public static function authorize_activitypub_requests( $response, $handler, $req
...
+		 */
+		$defer = \apply_filters( 'activitypub_defer_signature_verification', false, $request );
+
+		if ( $defer ) {
+			return true;
+		}
+
+		if (
+			// POST-Requests are always signed.
+			'GET' !== $request->get_method() ||
```
jeherve: Would it make sense to specifically check for `POST` here, or to `\WP_REST_Server::EDITABLE`?

pfefferle: I wanted to be on the safe side, so I thought I would love to everything except `GET` :)

includes/rest/class-server.php
pfefferle: Totally, I just wanted to have something out, to start a discussion about if this even makes sense or if I miss something.

We had a discussion about that long time ago and we decided to use what we currently use and I can't remember the background of the decision 🫣

includes/rest/class-server.php
pfefferle: We had an issue, because initially I only checked for `POST`, that's why I was a bit paranoid with this change :)

includes/rest/class-server.php
```diff
@@ -74,39 +73,24 @@ public static function application_actor() {
...
-	 *                                                                    Usually a WP_REST_Response or WP_Error.
-	 * @param array                                             $handler  Route handler used for the request.
-	 * @param \WP_REST_Request                                  $request  Request used to generate the response.
+	 * @param \WP_REST_Request $request The request object.
 	 *
-	 * @return mixed|WP_Error The response, error, or modified response.
+	 * @return bool|\WP_Error True if the request is authorized, WP_Error if not.
 	 */
-	public static function authorize_activitypub_requests( $response, $handler, $request ) {
+	public static function signature_verification( $request ) {
```
obenland: Could this be `verify_signature` to use the same nomenclature as other methods in this class?

mediaformat:
> @mediaformat any idea why we have not used it in the first place?

The callbacks use is more granular but I think that makes sense!

pfefferle:
@Automattic/fediverse we really should merge this ASAP. I forgot that I enabled "Authorized-Fetch" in my test system, so it took me some time, to figure out, why the likes/repost widgets won't be displayed!

includes/rest/class-server.php
```diff
@@ -74,39 +73,24 @@ public static function application_actor() {
...
 	/**
-	 * Callback function to authorize each api requests
+	 * Callback function to authorize an api request.
 	 *
-	 * @see WP_REST_Request
+	 * The function is meant to be used as a permission callback for the rest api.
+	 *
+	 * It verifies the signature of POST, PUT, PATCH and DELETE requests and also the GET requests in secure mode.
+	 * You can use the filter 'activitypub_defer_verify_signature' to defer the signature verification.
+	 * The HEAD request is always bypassed.
```
obenland:
```suggestion
	 * The function is meant to be used as part of permission callbacks for rest api endpoints.
	 *
	 * It verifies the signature of POST, PUT, PATCH, and DELETE requests, as well as GET requests in secure mode.
	 * You can use the filter 'activitypub_defer_verify_signature' to defer the signature verification.
	 * HEAD requests are always bypassed.
```

includes/rest/class-server.php
```diff
@@ -120,29 +104,29 @@ public static function authorize_activitypub_requests( $response, $handler, $req
...
+		$defer = \apply_filters( 'activitypub_defer_verify_signature', false, $request );
 
 		if ( $defer ) {
-			return $response;
+			return true;
 		}
 
 		if (
-			// POST-Requests are always signed.
+			// POST-Requests have to be always signed.
```
obenland:
```suggestion
			// POST-Requests always have to be signed.
```

obenland:
I think [1cffea7](https://github.com/Automattic/wordpress-activitypub/pull/1078/commits/1cffea7f3dc66c83ce56872b188d26d64d40e8d7) went a bit too far. [My feedback](https://github.com/Automattic/wordpress-activitypub/pull/1078#discussion_r1883963934) was specifically about the `signature_verification` method, the other hooks and function names were fine as they were.

pfefferle:
oops, I think I have not checked the search/replace properly enough! 🫣

obenland:
Do existing unit tests cover the change or should there be new ones added?

pfefferle:
@obenland there were test for all parts in the stack, but I added one that integrates all of them.

obenland (approved):
💖

PR #1080, opened 2024-12-13
Add comment count upgrade routine 

obenland:
Untested yet but ready for initial feedback.

## Proposed changes:

* Updates `lock()` to not overwrite existing locks.
* Adds update routine for comment counts on posts with activitypub activity.
* Adds unit tests to cover the new functionality.

### Other information:

## Testing instructions:

* Create posts with likes and reposts from the Fediverse.
* View one of those posts in the front-end and see the incorrect comment number.
* Delete the `activitypub_db_version` option or change it to a low version number.
* Trigger the upgrade routine by loading any page on your test blog.
* Reload the post with with likes and reposts from the Fediverse and make sure the comment counts are updated.


pfefferle:
We already have a migration class that cares about the versioning and also supports locking, so that migrations will not run in parallel. You can simply add your migration logic there.

Let me know if I can support!

obenland:
Migrate! 🙈

includes/class-migration.php
```diff
@@ -146,6 +147,9 @@ public static function maybe_migrate() {
 		if ( \version_compare( $version_from_db, '4.1.0', '<' ) ) {
 			self::migrate_to_4_1_0();
 		}
+		if ( \version_compare( $version_from_db, '4.5.0', '<' ) ) {
```
pfefferle, line 161: If you move this block to the `async_migration` function, it will be handled async!

pfefferle (changes requested):

includes/class-migration.php
obenland: I wasn't clear on how to best solve passing the batch and offset variables in that approach

pfefferle: you don't have to!

pfefferle: ah, these are batches...

pfefferle: I always love to have something that works for all implementations... should be add a third param for various data, so that all async jobs can add additional data?

obenland: I knew you would say that 😄
Let me take another look at it

obenland: With the batching I'm not sure how to make it work in a reasonable way. We'd have to carry forward the old version, re-triggering any upgrade routines in `async_migration`, depending on which version the plugin upgraded from.

includes/class-migration.php
```diff
@@ -53,9 +53,20 @@ public static function get_version() {
...
 	 * Locks the database migration process to prevent simultaneous migrations.
+	 *
+	 * @return bool|int True if the lock was successful, timestamp of existing lock otherwise.
 	 */
 	public static function lock() {
-		\update_option( 'activitypub_migration_lock', \time() );
+		global $wpdb;
+
+		// Try to lock.
+		$lock_result = $wpdb->query( $wpdb->prepare( "INSERT IGNORE INTO `$wpdb->options` ( `option_name`, `option_value`, `autoload` ) VALUES (%s, %s, 'no') /* LOCK */", 'activitypub_migration_lock', \time() ) ); // phpcs:ignore WordPress.DB
```
pfefferle: what is wrong with `update_option`?

obenland: `update_option` ignores any existing lock and updates it

pfefferle (approved):

PR #1086, opened 2024-12-17
Include @mentions in the JSON representation of the reply

### LINKED ISSUE #844: Comment replies via Federation: include @mentions in the JSON representation of the reply
### What

When clicking reply to reply to a comment added via Federation, it would be nice if the handle of the person you're replying to, and maybe even the handles of the all the people @-mentioned in the comment you're replying to, were added to the JSON representation of the reply.

### Why

When on a WordPress site, one does not expect to see every reply to a comment start with `@person`. Comment threading (when it is enabled) makes it very clear that the reply is in response to another comment.
Outside of the WordPress context, however, folks have gotten used to seeing `@person` in front of the replies to a post. It makes it easy to understand who is involved in a conversation.

Here is an example of a typical conversation on the Fediverse. 

- I posted an update (via a WP site)
- Donncha replied via the Fediverse, thus prepending `@jeremy` to his reply.
- Matthias replied to Donncha on the Fediverse, thus prepending both `@jeremy` and `@donncha` to his reply.
- At that point, I replied to Matthias from my WordPress site since all the replies above had been federated. I think it would be fair to expect my reply to follow the same syntax. Those `@mentions` do not need to appear on the site itself, but I think it would be nice for them to appear on the Fediverse.

Screenshot 2024-08-09 at 10 09 40

### How

Related conversation: p1723189722720219-slack-C04TJ8P900J

pfefferle:
Fix #844

> When clicking reply to reply to a comment added via Federation, it would be nice if the handle of the person you're replying to, and maybe even the handles of the all the people @-mentioned in the comment you're replying to, were added to the JSON representation of the reply.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


includes/transformer/class-comment.php
```diff
@@ -7,6 +7,7 @@
 
 namespace Activitypub\Transformer;
 
+use WP_HTML_Tag_Processor;
```
obenland:
```suggestion
```

It doesn't look like that's used anywhere?

includes/transformer/class-comment.php
```diff
@@ -53,31 +54,30 @@ public function change_wp_user_id( $user_id ) {
...
+
+		$content       = $this->get_content();
+		$at_replies    = '';
+		$reply_context = $this->extract_reply_context( array() );
+
+		foreach ( $reply_context as $acct => $url ) {
+			$at_replies .= sprintf(
+				'<a class="u-mention mention" href="%s">%s</a> ',
+				$url,
+				$acct
```
obenland:
```suggestion
		foreach ( $reply_context as $account => $url ) {
			$at_replies .= sprintf(
				'<a class="u-mention mention" href="%s">%s</a> ',
				$url,
				$account
```
Should `$url` and `$account` be escaped? Does this string end up being displayed?

Multiple mentions will be glued together without a space, not sure if that's on purpose?

tests/includes/transformer/class-test-comment.php
```diff
@@ -0,0 +1,168 @@
...
+		$transformer = new Comment( $comment );
+		$object      = $transformer->to_object();
+
+		// Get the content.
+		$content = $object->get_content();
+
+		// Test that reply context is added.
+		$this->assertStringContainsString( 'Reply comment', $content );
+		$this->assertStringContainsString( '<a class="u-mention mention" href="https://remote.example/@author">@author@remote.example</a>', $content );
+
```
obenland, line 93: It might be helpful to add a test with multiple mentions that also compares the entire output to make sure we're happy with whitespace.

includes/transformer/class-comment.php
pfefferle: oops... I experimented with it ;)

tests/includes/transformer/class-test-comment.php
pfefferle: it already adds multiple, but I will extend the test, to check for all of them.

includes/transformer/class-comment.php
pfefferle: We have to escape them! good point, but I would love to stick with `$acct`, because it is the WebFinger that uses `acct:` as protocol, or we could use `$webfinger` instead.

obenland: Makes sense!

obenland (dismissed):

PR #1087, opened 2024-12-17
PHPUnit: Fix slow tests

obenland:
Takes test execution from 7.8s to ~3s.

[`class-test-followers.php`](https://github.com/Automattic/wordpress-activitypub/compare/fix/slow-tests?expand=1#diff-f4a1cd985da53a0d11e13f79cd35f252fcdce604e0d61148cd3ba68a7daeef05) needs a keen eye to see if the AI has inadvertently removed an implicit test that we should keep.

Once we drop support for PHP 7.0, the listener can be enabled by default.
 
## Proposed changes:

* Adds a timing listener to tests
* Adds http request mocks and breaks up tests to make them faster.

### Other information:

### DIFF
--- a/tests/includes/class-test-mention.php
+++ b/tests/includes/class-test-mention.php
@@ -30,6 +30,24 @@ class Test_Mention extends \WP_UnitTestCase {
 	);
 
 	/**
+	 * Set up the test case.
+	 */
+	public function set_up() {
+		parent::set_up();
+		add_filter( 'pre_get_remote_metadata_by_actor', array( get_called_class(), 'pre_get_remote_metadata_by_actor' ), 10, 2 );
+		add_filter( 'pre_http_request', array( $this, 'pre_http_request' ), 10, 3 );
+	}
+
+	/**
+	 * Tear down the test case.
+	 */
+	public function tear_down() {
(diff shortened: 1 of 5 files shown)


pfefferle:
I am Fine with catching up with WordPress and require 7.2!

tests/includes/class-test-mention.php
```diff
@@ -77,17 +91,74 @@ public function the_content_provider() {
...
+		// Handle remote users.
+		if ( 'pfefferle@notiz.blog' === $actor ) {
+			return array(
+				'id'   => 'https://notiz.blog/author/matthias-pfefferle/',
+				'url'  => 'https://notiz.blog/author/matthias-pfefferle/',
+				'name' => 'pfefferle',
+			);
+		}
+
+		if ( 'pfefferle@lemmy.ml' === $actor ) {
```
pfefferle: my initial idea was to have an integration test and directly check if the changes still work with lemmy and WordPress. I understand that this makes the tests slow, so what do you think about copying their profile JSON to use them locally and to maybe have a regular update of the files?

pfefferle (dismissed):
Maybe we can use JSON files in a future PR... something like `sanitize_title( $host ).json`. It would be easier to add new platforms or to update the existing ones, this way.

PR #1089, opened 2024-12-17
Workflow: Install SVN before trying to deploy

obenland:
## Proposed changes:

* Install SVN before using it in the action.


mattwiebe (approved):
let's try it!

PR #1090, opened 2024-12-17
Post meta: Update meta keys to be private

### LINKED ISSUE #943: Feature Request: Underscore prefix for meta keys
### What

I'd like to see underscore prefixes for the meta keys like `activitypub_user_id` (to be `_activitypub_user_id` instead).

### Why

Meta keys without underscore prefixes show up in the Custom Fields UI, which clogs up the select/dropdown menu.

Image

### How

_No response_

obenland:
WordPress convention states that meta keys prefixed with an underscore are considered private and hidden from the custom fields UI. Since our meta keys are for internal plugin use only, they should follow this convention.

This change helps prevent accidental modification of ActivityPub meta data through WordPress's custom fields interface while maintaining the expected pattern for plugin developers.

See: https://developer.wordpress.org/plugins/metadata/managing-post-metadata/#hidden-custom-fields

Fixes #943

## Proposed changes:

* Added a migration routine in class-migration.php to prefix meta keys with underscores
* Updated `class-test-followers.php` to use the new prefixed meta keys (_activitypub_*)
* Updated `class-test-migration.php` to test the migration directly instead of through `maybe_migrate()`
* Fixed all meta key references across the codebase to use the new prefixed versions (e.g., `activitypub_user_id` → `_activitypub_user_id`)

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+* Added a filter to make custom comment types manageable in WP.com Calypso
+
 ### Changed
 
 * Hide ActivityPub post meta keys from the custom Fields UI
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+* Hide ActivityPub post meta keys from the custom Fields UI
+
 ### Fixed
 
 * Undefined array key warnings in various places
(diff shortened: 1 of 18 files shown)


obenland:
@pfefferle I remember you mentioning certain post meta keys that should be private, which ones were those?

pfefferle:
It was more about if there are some that should not be private. Like for example the visibility one. So that classic editor users can add that through the meta forms for example.

pfefferle (dismissed):
Nice!

pfefferle (dismissed):

mattwiebe (dismissed):
This looks good to me. We've got a couple of places that directly call `activitypub_user_id` in the wpcom codebase that will need to be updated before/when this lands.

obenland:
Thanks for the heads up!

PR #1092, opened 2024-12-18
Don't send direct message notification for received public activities

akirk:
With 4.5.0, I started receiving direct message e-mails for things that are not direct messages:

Screenshot 2024-12-18 at 08 53 45

I added a unit test without a reply and fixed the if. I am not sure why it was tested to be both a reply and public and not just public.


pfefferle (approved):
thanks a lot @akirk

pfefferle (changes requested):
I will add some more tests!

pfefferle (approved):
should be good to go without further changes!

jeherve:
Thanks for the quick fix!

PR #1098, opened 2024-12-18
Editor: Make Fediverse preview template filterable

### LINKED ISSUE #967: Feature: Add filter to allow custom fediverse previews
The current implementation of the Fediverse preview (#994), has a generic name but is still quite limited. Therefore I propose as suggested by @pfefferle to make it more flexible.

Two options come in to my mind:

1. **Easy and straight forward:** Apply a filter that lets change the preview template complety (based on the post-type).

2. **Additional preview tabs within the fediverse preview:** Within the preview, multiple preview tabs could also be possible, but it has a drawback that would add quite some complexity: e.g. an `as:Event` is rendered differently in Mastodon than a `Note`, same like `Article`. So the template would have to take Converted object types into account. So maybe here the preview tabs could be concrete Fediverse applications. 

Reference:

             @Menrath nice idea! I will see how we can extend that a bit, to have a simple templating mechanism!?!

_Originally posted by @pfefferle in https://github.com/Automattic/wordpress-activitypub/issues/944#issuecomment-2426941679_

obenland:
Fixes #967.

## Proposed changes:

* Adds a filter for the preview template path.
* Adds unit tests for the filter.

### Other information:

## Testing instructions:

* Go to the Editor to create or edit a post.
* Click Preview > Fediverse Preview.
* Make sure the preview still works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+* Add a filter to allow modifying the ActivityPub preview template
+
 ## [4.5.1] - 2024-12-18
 
 ### Improved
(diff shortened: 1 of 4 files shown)


pfefferle:
I think this is a nice first step, but I would keep the ticket open, to discuss a template manager, that allows you to view your post in different templates. Or should we file a different issue for that?

/cc @Menrath

pfefferle (dismissed):

pfefferle (approved):

PR #1099, opened 2024-12-18
Docs: Add missing hook documentation

obenland:
Adds action/filter docs for hooks that are missing them and updates existing docs to include all parameters and more unified language with Core.

## Proposed changes:

* Added and update hook documentation.


pfefferle (approved):
Nice!

PR #1100, opened 2024-12-18
Reply Block: Embed referenced post when possible

### LINKED ISSUE #1027: Federated Reply block: embed original post on frontend
### What

The Federated reply block is very useful for a site to become a real actor on the Fediverse. Once we publish a post using that block, our reply looks good on the Fediverse. I was wondering if we could embed the original post instead of "only" rendering a link to that post:

https://github.com/Automattic/wordpress-activitypub/blob/9beea625fd715f77bf0682c2dc26ae4277e88c3f/includes/class-blocks.php#L301-L326

### Why

This would allow readers that consume the site content from the site, and not from the Fediverse, to read the original post right away, without having to click, and thus have all the context of the conversation in one place, just like folks on the Fediverse experience.

### How

I know that Fediverse embeds are a bit tricky, but maybe we could start with some support for Mastodon, based off the learnings from [this trac ticket](https://core.trac.wordpress.org/ticket/58120).

obenland:
Fixes #1027.

In this first pass the Mastodon embed provider only gets registered when the Reply block gets rendered. I'm open to making it generally available but initially it didn't feel like functionality that this plugin should be responsible for globally.

There's also an action to register additional Fediverse oembed providers beyond Mastodon.social. Not sure if there are others we should or need to consider.

The current implementation supports embedding posts from all supported oembed providers. I wasn't sure whether it should be limited to Mastodon or whether it should be all—please let me know if you have an opinion on that.

Finally, this maintains the anchor, that [IndieWeb seems to be looking for](https://indieweb.org/in-reply-to).

## Proposed changes:
* Adds oEmbed support to the reply block, automatically embedding content from supported platforms.
* Registers Mastodon as an oEmbed provider for better integration with Mastodon posts.

### Other information:
- [x] Have you written new tests for your changes?

#### Testing Reply Block with Mastodon Posts:
1. Create a new post.
2. Add a Reply block.
3. Enter a Mastodon post URL (e.g., https://mastodon.social/@obenland/113561729606054851).
4. Verify that:
   - The post is embedded above the reply link.
   - The reply link is still visible and functional.
   - The URL is properly formatted in the link text.

#### Testing with Non-Embeddable URLs:
1. Create a new post.
2. Add a Reply block.
3. Enter a URL that doesn't support oEmbed.
4. Verify that:
   - Only the reply link is shown (no embed).
   - The link is properly formatted.

#### Testing with Other oEmbed Providers:
1. Create a new post.
2. Add a Reply block.
3. Try URLs from other oEmbed-enabled platforms.
4. Verify that embeds work as expected while maintaining the reply link functionality.


includes/class-blocks.php
```diff
@@ -339,7 +339,19 @@ public static function render_reply_block( $attrs ) {
 		$html = '';
 
 		if ( ! empty( $attrs['url'] ) ) {
-			$html = sprintf(
+			wp_oembed_add_provider( '#https?://mastodon\.social/(@.+)/([0-9]+)#i', 'https://mastodon.social/api/oembed', true );
```
pfefferle: isn't this too specific, because it only works for mastodon.social? what about using the ActivityPub JSON instead?

pfefferle: yes, Mastodon.social might be the biggest instance, but we should work on a generic solution!

jeherve: > what about using the ActivityPub JSON instead?

Oh, that's not something I had considered when I created the issue, sorry @obenland! It's true that pulling data from the ActivityPub representation instead of trying to embed may be even cleaner, I like the idea!

pfefferle: @obenland @jeherve I refactored the code a bit, so that it is easier to add ActivityPub support with oEmbed fallback and added some of the biggest mastodon instances.

Let me know what you think!

obenland: First pass in [3db22d2](https://github.com/Automattic/wordpress-activitypub/pull/1100/commits/3db22d2c9a46e02cef23685fd9c0d50c2a284cbd).
This could use some heavy testing with posts from various different platforms.

~Should we try oembed first?~ Tests say Yes :)

includes/functions.php
```diff
@@ -1593,3 +1593,143 @@ function is_self_ping( $id ) {
...
+	if ( $embed ) {
+		return $embed;
+	}
+
+	// Create unique transient keys for both HTTP calls.
+	$main_transient_key = '_activitypub_embed_' . md5( $url );
+	$embed              = \get_transient( $main_transient_key );
+
+	if ( false === $embed ) {
+		$response = Http::get( $url );
```
pfefferle: There is already a function that provides all the needed headers and has caching built in: https://github.com/Automattic/wordpress-activitypub/blob/b2e14e9e00350d2cadfef1f08e0241a55f637610/includes/class-http.php#L224C25-L224C42

includes/functions.php
```diff
@@ -1593,3 +1593,143 @@ function is_self_ping( $id ) {
...
+				if ( isset( $attachment['type'] ) && 'Document' === $attachment['type'] ) {
+					$image = $attachment['url'];
+					break;
+				}
+			}
+		}
+
+		ob_start();
+		?>
+		<div class="activitypub-embed">
```
pfefferle: I would prefer a template and `load_template` here. (I am not a big fan of inline HTML)

jeherve (commented):
This is working really well, nice work!

I've tested it with replies to posts on `mastodon.social` as well as smaller instances not mentioned in this PR, and non-Mastodon instances (in which case you can get a 401 on the embed request) and it's all displayed nicely.

I had a question about the design. 
On Mastodon, loading a reply clearly displays the reply in focus, and the original post above it in a darker color. In WordPress, the original post naturally gets displayed above the reply so it has more emphasis than the reply you're writing. 
I wonder if it would be helpful to clearly indicate "in response to" above the embed, or something like that, so it's clearer that the embed is the original post you're responding to?

pfefferle:
@obenland can you maybe add some example screenshots to the PR?

jeherve:
@pfefferle You can check some examples on my own site right now:

- Reply to a `mastodon.top` message: https://herve.bzh/20250102-15328/
- Reply to a `mastodon.social` (added as embed source in this PR) message: https://herve.bzh/20240820-1963/
- Reply to a non-mastodon post: https://herve.bzh/20250108-15536/

pfefferle:
Thanks a lot @jeherve !

Should we make this configurable? This is a very big block that we add to each post!?!

pfefferle:
@obenland I added some microformats to match the https://indieweb.org/reply-context

jeherve:
> Should we make this configurable? This is a very big block that we add to each post!?!

That's a good question. I personally think it makes sense to provide all that context, but I could imagine some folks wanting a more consise display. How about adding a filter so folks could choose between an expanded and a compact view? Later on and if that proves necessary, we could make that a setting in the Reply block itself.

includes/class-activitypub.php
```diff
@@ -738,4 +740,16 @@ public static function updated_postmeta( $meta_id, $object_id, $meta_key, $meta_
...
+	 * Register some Mastodon oEmbed providers.
+	 */
+	public static function register_oembed_providers() {
+		\wp_oembed_add_provider( '#https?://mastodon\.social/(@.+)/([0-9]+)#i', 'https://mastodon.social/api/oembed', true );
+		\wp_oembed_add_provider( '#https?://mastodon\.online/(@.+)/([0-9]+)#i', 'https://mastodon.online/api/oembed', true );
+		\wp_oembed_add_provider( '#https?://mastodon\.cloud/(@.+)/([0-9]+)#i', 'https://mastodon.cloud/api/oembed', true );
+		\wp_oembed_add_provider( '#https?://mstdn\.social/(@.+)/([0-9]+)#i', 'https://mstdn.social/api/oembed', true );
+		\wp_oembed_add_provider( '#https?://mastodon\.world/(@.+)/([0-9]+)#i', 'https://mastodon.world/api/oembed', true );
+		\wp_oembed_add_provider( '#https?://mas\.to/(@.+)/([0-9]+)#i', 'https://mas.to/api/oembed', true );
+	}
```
mattwiebe, line 790: Would it be possible to just add these dynamically based on what posts the user tries to embed? Assume that they use this oembed pattern? 

Otherwise anybody trying to embed a post that's not in the list won't work.

mattwiebe (commented):
I am so happy these are happening - I wanted this from the start

mattwiebe:
Another thought I'm having: can we actually use the embed block in the Editor so that it gives a truer representation of what will be on the frontend?

pfefferle:
@obenland ~if you change the setting afterwards, it will not be saved.~ the embed flag is not stored in my test env.

includes/class-activitypub.php
pfefferle: The code falls back to Activity-JSON parsing if the embed does not return a result.

obenland:
Yeah, the toggle is currently not working right. Turns out that it's next to impossible to have different defaults for a new block attribute for existing blocks vs newly inserted blocks.

Right now the easiest path forward seems to be to default it to showing embeds, even for existing blocks. Or we make this opt-in, which would be a bummer.

pfefferle:
I am fine with either way. let's decide what default status we prefer and go with it!

One thing: We need to get sure that this part won't be federated and remove it the same way we remove the link!

https://github.com/Automattic/wordpress-activitypub/blob/e2908e8b55bd3bb9bfaa35e852d623e78471462a/includes/transformer/class-post.php#L503C1-L503C88

obenland:
> Another thought I'm having: can we actually use the embed block in the Editor so that it gives a truer representation of what will be on the frontend?

@mattwiebe I'd love that! Would you be up for taking a stab at it?

includes/class-activitypub.php
mattwiebe: Oh! The Activity JSON fallback isn't working in my local environment so I didn't know it was there. Will test in a different env 👍🏻

mattwiebe:
> > Another thought I'm having: can we actually use the embed block in the Editor so that it gives a truer representation of what will be on the frontend?
> 
> @mattwiebe I'd love that! Would you be up for taking a stab at it?

Sure thing, I'll dig in

obenland:
Could we use Core's embed block to display the embed?

If that's not possible, there's [an existing oembed API endpoint](https://github.com/WordPress/wordpress-develop/blob/95c00d0f68a662cf96066ee518e70250b9917ba4/src/wp-includes/class-wp-oembed-controller.php#L65) to retrieve the oembed response, could that be re-used? It doesn't contain our fallback html, I wonder if that could be added through filters.

obenland:
Do we need register a separate embed preview in Gutenberg for the oembed providers we added? Or a block variation or something. Just pasting a mastodon URL doesn't show a preview in the editor, just a blank square
image

mattwiebe:
> Could we use Core's embed block to display the embed?
> 
> If that's not possible, there's [an existing oembed API endpoint](https://github.com/WordPress/wordpress-develop/blob/95c00d0f68a662cf96066ee518e70250b9917ba4/src/wp-includes/class-wp-oembed-controller.php#L65) to retrieve the oembed response, could that be re-used? It doesn't contain our fallback html, I wonder if that could be added through filters.

Yeah, I was halfway through all of this when I wondered the same. It would be a lot cleaner if possible, but the control flow logic as I understand it in both `WP_oEmbed` and the `oembed/1.0/proxy` endpoint is such that it will only work if you use the `pre_oembed_result` filter, which would require us to add a preflight ActivityPub representation request to every oEmbed request in order to make it work, AND that would have to supercede any registered oEmber handler. 

Although maybe I could just hook in at a later point on the REST Request itself to avoid that… let me investigate that and see if I can remove complexity

mattwiebe:
> Do we need register a separate embed preview in Gutenberg for the oembed providers we added? Or a block variation or something. Just pasting a mastodon URL doesn't show a preview in the editor, just a blank square 

It _should_ show up there fine for any registered provider. I might be able to circle back to actually just using the core block for display if the filter approach works

mattwiebe:
Ok! Looks like I've messed up some tests here but the new, **More Like Core** approach is working. This comes with some downsides, especially the UX in the Editor isn't necessarily better with leveraging the `core/embed` block. It's not updating properly _and_ you can edit that embed's URL independently of the main reply block. Some more work to be done, there.

mattwiebe:
This should be ready to go!

obenland:
@jeherve @pfefferle Could you take this for another spin?

pfefferle:
I have issues with other WordPress sites:

Screenshot 2025-02-21 at 15 13 22

pfefferle:
The block takes a lot of space even if there is no preview available

Screenshot 2025-02-21 at 15 20 46

even more when un-hover

Screenshot 2025-02-21 at 15 24 20

obenland:
@mattwiebe When `embedPost` is set to false in edit.js, let's not attempt to embed the post when the URL is added/updated/changed.

pfefferle:
~@obenland I talk about enabled `embedPost` but without result!~

It seems to be the same result, never mind!

But it is a bit irritating that the setting is checked:

Screenshot 2025-02-21 at 15 33 06

I would assume that it is unchecked if it is not supported?

Maybe that is the also the spacing issue, because if I use a link with preview and uncheck the embed, the spacing seems fine.

Screenshot 2025-02-21 at 15 35 13

obenland:
Oh, my comment was not in reference to something you said, but general feedback

pfefferle:
But apart from the feedback and the small glitches! This is great and I love the improvements the PR has gone through!

mattwiebe:
Ok I think this should be pretty much there UX-wise. Addressed big margins when the embed is disabled. The Loading state sometimes disappears entirely?

> When embedPost is set to false in edit.js, let's not attempt to embed the post when the URL is added/updated/changed.

@obenland The trick with what I'm doing as of [7809d3a](https://github.com/Automattic/wordpress-activitypub/pull/1100/commits/7809d3a03e2721eaab63499f243d9f4f996cee74) is that I'm always setting to true automatically when a valid embed is found, and always setting to false automatically when an invalid embed is found, aka embed not found. 

The user is always able to disable `embedPost` from a perfectly valid embed state, and that will save with `embedPost=false` in the block's saved state. The frontend won't try to render the embed even though it theoretically could. But otherwise, just like the `core/embed` block won't let you embed a block for a URL that fails (and this is what I needed to override that state and only show the block on success), we should default to "they want to embed if they can."

# BUT

I've now talked myself into a bit more scope creep on this PR that has dragged itself across this week with all my "help" lol

There's room for us to take this a step further and test that we can get a valid ActivityPub response from the URL. We're probably being far too permissive here. In this case we would need to change some more of our approach. Because right now we'll embed anything, even something that isn't an ActivityPub URL. As long is it returns a good core oembed response. And that's not ideal. Want to embed something that isn't ActivityPub? Use the core embed. It's not really ultimately appropriate to embed other URLs here is it? Isn't this a great chance for a preflight check? This URL can't be contacted. Or can. Inform the user, set expectations.

mattwiebe:
Not that I want to back out the oEmbed enhancement. That gives us instant embeds for an ActivityPub site that isn't explicitly registered as a WP oEmbed provider.

But this would also resolve the issue for weird-looking WP embeds: we don't want those. Or would we?

obenland:
Yes, let's stay ActivityPub specific for this block and leave anything else to Core to handle

mattwiebe:
It's all coming back to me now, why I didn't put embeds in the first time around! 😅

mattwiebe:
This is ready for another round of testing. We now only target for ActivityPub embeds, and we always render those ActivityPub embeds with our own embed template rather than even trying for an oEmber representation.

We will not allow the user to set `embedPost=true` for any URL that we fail to find an ActivityPub representation for, effectively gated by `Http::get_remote_object()`. We now have good insight whether the URL we're trying to reply to is even any good to us.

I've refined the behaviour around defaulting to `embedPost=true` whenever a successful embed is found. I want to do this optimistically, to opt users into having an embed when they will in most cases want one, but also not causing undesired edits simply by loading a post with the old `null` value or one where we've set `embedPost=false` explicitly.

Now the behaviour tracks to a new `optimisticEmbed` state. When true, we will automatically set `embedPost=true` whenever a successful embed is found under three situations:

1. The block first run was with `embedPost=true` (logic: loading a block previously saved as true)
2. This is a newly-created block (logic: first run has an empty `url` prop)
3. Whenever the user explicitly toggle the sidebar ToggleControl to true for a block

This covers the scenarios where optimistic embedding seems called for, but otherwise we don't want to interfere with user preference for no embedding, or updating state on old saved blocks.

jeherve (commented):
This tested well for me. I have a few questions / remarks:

1. Could we display the user's handle here instead of this long URL?
image

2. The label text is really about 2 different things: 
    - This is a post from a Fediverse software.
    - You can embed that original post if you'd like.

Maybe the first point should appear above the toggle?

And maybe the toggle text shouldn't say "below", since it's not below?

3. I wonder if we should do away with the technical terms ("supports", "ActivityPub replies"). Could we be more generic somehow, and say that you can reply to this post on the Fediverse?

4. The "non-supported" message reads like an error message, even for posts that I know are from ActivityPub platforms like Mastodon:

Screenshot 2025-02-24 at 15 46 20

It seems to want me to input a different URL, but this one is not wrong.

jeherve:
> This tested well for me

Scratch that, the embed is included in the ActivityPub representation of the post:

Screenshot 2025-02-24 at 17 15 14
Screenshot 2025-02-24 at 17 14 59

mattwiebe:
Great test finds @jeherve ! I'll go brew up some changes to this, we're almost there right? 😅

mattwiebe:
@jeherve I haven't been able to replicate the block itself showing up in the outbox content, but I've 1) moved the embeds to a webfinger style `@user@domain.com` and 2) improved the labels and loading states.

jeherve:
> @jeherve I haven't been able to replicate the block itself showing up in the outbox content, but I've 1) moved the embeds to a webfinger style `@user@domain.com` and 2) improved the labels and loading states.

I'm afraid I'm still running into the issue:

Here is an example (I'm sharing screenshots because I'll be switching back to `trunk`):

image
image

Even if I don't enable the large embed, the Federated reply block contents are shown in the ActivityPub context:

image
image

You can see the post in `trunk` here:
https://herve.bzh/20250225-58394/activitypub/
image

mattwiebe:
Thanks for braving the branch in prod @jeherve 🫡🫡🫡

Got to bottom of the problem in https://github.com/Automattic/wordpress-activitypub/pull/1100/commits/22be38908ada2797e0ad657a1084114d387943d7 and decided to scratch one last itch that's been bugging me with the previous behaviour of "show nothing from the block when federating" - that's not how most platforms work, they show a `@reply` link. Why don't we do that too? https://github.com/Automattic/wordpress-activitypub/pull/1100/commits/08bbd8ea2db062251909d69862df0e833d4b9efa took care of that.

We should be good, with that, now. Let me know how things are on the UX front, cc @Automattic/fediverse as well, I think that this is really finally actually done (ducks)

obenland:
Are we back to preferring the fallback embed handler over the mastodon embeds?

image

obenland:
Embeds in the front-end don't get expanded anymore, I tested mastodon and WordPress embeds.

image

includes/functions.php
```diff
@@ -1633,3 +1712,91 @@ function is_actor( $data ) {
...
+
+/**
+ * Get an ActivityPub embed HTML for a URL.
+ *
+ * @param string  $url        The URL to get the embed for.
+ * @param boolean $inline_css Whether to inline CSS. Default true.
+ *
+ * @return string|false The embed HTML or false if not found.
+ */
+function get_embed_html( $url, $inline_css = true ) {
```
obenland, line 1599: With this function it looks like we can remove `embed_get()`?
I liked the attempt in `embed_get()` to get an oembed response first, should we do that here, too?

includes/rest/class-url-validator-controller.php
```diff
@@ -0,0 +1,120 @@
...
+			'/' . $this->rest_base,
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'validate' ),
+					'args'                => array(
+						'url' => array(
+							'required'          => true,
+							'type'              => 'string',
+							'sanitize_callback' => 'esc_url_raw',
```
obenland:
```suggestion
							'format' => 'uri',
							'required' => true,
```
With these changes most checks in `validate()` should be possible to remove

includes/rest/class-url-validator-controller.php
```diff
@@ -0,0 +1,120 @@
...
+	 * Register routes.
+	 */
+	public function register_routes() {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'validate' ),
```
obenland:
```suggestion
					'callback'            => array( $this, 'get_items' ),
```
Let's reuse the format in WP_REST_Controller?

includes/rest/class-url-validator-controller.php
```diff
@@ -0,0 +1,120 @@
...
+					'args'                => array(
+						'url' => array(
+							'required'          => true,
+							'type'              => 'string',
+							'sanitize_callback' => 'esc_url_raw',
+						),
+					),
+					'permission_callback' => array( $this, 'validate_url_permissions_check' ),
```
obenland:
```suggestion
					'permission_callback' => array( $this, 'get_items_permissions_check' ),
```

obenland (commented):
There's no way around the custom API endpoint?

includes/functions.php
mattwiebe: Yes we can almost certainly get rid of `oembed_get` but I agree that I like its approach of serving a real `oEmbed` where we can get one. It's just that the lifecycle is very tricky because we can only use  a `pre_oembed_get` filter, we can't filter on a `not found` attempt except in the way we're doing it by filtering the endpoint itself. I think there's a way to juggle all of this but it's precarious. But I would like to manage this compexity on the dev side in favour of user simplicity.

mattwiebe:
> There's no way around the custom API endpoint?

We absolutely need it to some degree, it's the only way to give a reliable `is_activitypub` check. But maybe we can leverage more Core stuff.

includes/rest/class-url-validator-controller.php
mattwiebe: Excellent. Nice to have a REST API sleuth on the job to compensate for my Windsurf slop ;)

mattwiebe:
Demo'd this internally today. Two notes:

1. The default to optimistically show the embed is not working, at least not in the bookmarklet flow.
2. Our non-oEmbed, ActivityPub embeds (using `core/embed`) is not currently working in the Editor.

mattwiebe:
I am going to add some more unit tests but I've fixed the bugs from yesterday's demo and this should be good for a hopefully-final round of testing!

What has improved:

1. Unregistered oEmbed providers will now render, using our ActivityPub template, even in the Core Embed block!
2. The bookmarklet will set `embedPost=true` so that there's a nice smooth flow there.
3. We stopped using the `core/embed` block for rendering oEmbeds because it was doubling and even tripling the embed (only inside our block, which means I was probably Doing It Wrong but I couldn't figure it out).

jeherve (commented):
I gave it another try, and would have a few notes:

*********

When you choose to embed the post, we now display both the embed and the link in the editor and on the frontend. I expected to see only the embed, or the link if I choose not to embed. I did not expect to see both at once:

Screenshot 2025-03-11 at 18 23 14

*********

I see we now display the full username you are replying to in the ActivityPub representation, like we do for federated replies via comments. That's a nice improvement, although it surprised me when I edited an old post to add the embed. I ended up with 2 mentions, since I had been manually adding the mentions until now.

I think it's fine to add it, I doubt many people will:

1. Have manually added the mention in their existing posts like I used to do.
2. Will go back and edit those old posts.

That said, I'm curious why one of the mentions is in full (`@username@domain.com`) while the other is shortened (`@username`). You can check my reply under this post for an example:
https://social.lol/@macmanx/114113978996512965

The first mention is added by the ActivityPub plugin, while the other is the one I had added to the post content manually.

Here is how things look like in the editor:

image

And in the ActivityPub representation of the post:

image

https://herve.bzh/20250306-65760/activitypub/

It would seem like our rules to display a mention in the ActivityPub representation are different whether it's done within the post content or added alongside the Federated reply block.

************

When I click to embed a post in the editor, I see 2 small layout issues, outlined in this video:

- The message to check whether the post can be embedded seems to have a lot of padding, so it causes the page to jump a bit.
- The embed can include a vertical scrollbar, but it cannot actually be scrolled (noting that on the frontend the embed doesn't include a scrollbar, it's just taller to adjust to the height).


https://github.com/user-attachments/assets/691e45d5-7964-49ce-b74a-358c0024dddb

pfefferle:
If you try to use a non-oembed link, it does lock the preview button.

mattwiebe:
> If you try to use a non-oembed link, it does lock the preview button.

@pfefferle do you mean when you're trying to embed a non-ActivityPub oEmbed?

pfefferle:
@mattwiebe if I use a mastodon instance that is not in the allow-list for example.

As far as I understood the PR, the plugin should fall back to use ActivityPub to generate a preview based on that data, but that part does not work properly.

mattwiebe:
> @mattwiebe if I use a mastodon instance that is not in the allow-list for example.
> 
> As far as I understood the PR, the plugin should fall back to use ActivityPub to generate a preview based on that data, but that part does not work properly.

In this case it was `mastodon.art` that was causing the trouble. It fails when fetching an ActivityPub representation with a `401 Unauthorized` error and the message `{"error":"Request not signed"}`. So it appears that it requires `AUTHORIZED_FETCH` to be enabled.

In cases like this, when it is not enabled, it would be nice to do a retry with authorized fetch. I will restrain myself from bloating this PR further heh.

pfefferle:
🤔

I thought we always send the signature... also for GET... I will check that and also check the preview with a different (not whitelisted) instance...

pfefferle:
Ah, I see... This is because of localhost! We send the signature, but it can't be verified, because the remote server can not access localhost!

With a bonn.social site, it works like expected!

pfefferle (dismissed):

jeherve (commented):
I gave this PR another try, and currently run into this issue when I try to activate embedding for an existing post (with an embed from mastodon.social)

image

The same error appears on the frontend when I save my changes:
https://herve.bzh/sorvagsvatn-un-lac-aux-iles-feroe/

Interestingly, that error seems to be in response to an embed request using a Friends endpoint:

https://herve.bzh/wp-json/friends/v1/embed?url=https://mastodon.social/@STPo/114182842753572276#?secret=xxxx

This may indicate a conflict with the Friends plugin (cc @akirk )

It doesn't happen with all replies, though. Another one is displayed properly for example (although I still run into the scrollbar issue I mentioned earlier).

The issue with duplicate mentions remains, though:

image

image

mattwiebe:
I think we're done here. I haven't been able to replicate @jeherve's embed failures, even alongside Friends. I moved to a [shorter handle in replies](https://github.com/Automattic/wordpress-activitypub/pull/1100/commits/5f9bc6c156b1b3ce64aca10e3c7d9f68ced8c56e) to match #1510 . 

We already had an approval from @pfefferle and @obenland would you mind one final test?

pfefferle:
It looks a bit strange if you try to use the URL of a Blog that has ActivityPub enabled:

Screenshot 2025-03-28 at 10 56 16

Neither the Embed-Code nor the ActivityPub representation seems to be used 🤔

mattwiebe:
> Neither the Embed-Code nor the ActivityPub representation seems to be used

Yeah that's not good. The neverending PR continues to never end. :/

mattwiebe:
Ok, the WP embed has been solved.

Unfortunately `notiz.blog` will not be embeddable because of `x-frame-options`:

```
curl -I https://notiz.blog/2025/03/24/blogtastisch-2-blogs-und-das-fediverse/embed/ 
HTTP/2 200
date: Mon, 31 Mar 2025 23:55:53 GMT
content-type: text/html; charset=UTF-8
vary: Accept-Encoding
server: nginx
x-cache: miss
x-pingback: https://notiz.blog/xmlrpc.php
link: <https://notiz.blog/wp-api/webmention/1.0/endpoint>; rel="webmention", <https://notiz.blog/wp-api/webmention/1.0/endpoint>; rel="http://webmention.org/", <https://notiz.blog/wp-api/>; rel="https://api.w.org/", <https://notiz.blog/wp-api/wp/v2/posts/53144>; rel="alternate"; title="JSON"; type="application/json", <https://notiz.blog/b/Ekj>; rel=shortlink
server-timing: wp-total;dur=63.49
cache-control: max-age=0, public
expires: Mon, 31 Mar 2025 23:55:53 GMT
x-xss-protection: 1; mode=block
x-content-type-options: nosniff
strict-transport-security: max-age=31536000
x-frame-options: SAMEORIGIN
referrer-policy: strict-origin-when-cross-origin
x-clacks-overhead: GNU Terry Pratchett
```

A perfect solution would then fall back to the ActivityPub representation, but I'm not doing that. :)

includes/transformer/class-post.php
```diff
@@ -535,12 +536,70 @@ protected function get_content() {
...
+			$webfinger = '@' . $author['preferredUsername'] . '@' . $domain;
+		}
+
+		if ( ! $webfinger ) {
+			return '';
+		}
+
+		// Generate HTML @ link.
+		return \sprintf(
+			'<p class="ap-reply-mention"><a rel="mention ugc" href="%1$s" title="%2$s">%3$s</a></p>',
```
jeherve, line 596: Do we need the wrapper to be a paragraph? Could it be a `span`? I'm mostly asking because it feels off to have the mention displayed on its own line before the reply. That's not what we typically see in other Fediverse software ; the @mention is usually immediately followed by the content.

jeherve (commented):
I'm still running into the same embed issues with Friends, but since that seems specific to my setup I'll let others test!

includes/transformer/class-post.php
obenland: It doesn't look like removing the `p` makes much of a difference when using the block editor, even if we removed `wpautop()`, too. The following paragraph will be wrapped in its own paragraph tag and displayed on a new line.

obenland:
@pfefferle @jeherve I think we've spent enough time on this PR, it's decision-time. Let's either merge it or close it. I'm happy for us to create follow-up PRs once it's in, but at some point we have to call it on this one, and I think this is it.

pfefferle (approved):

PR #1102, opened 2024-12-19
Ensure that direct messages are only sent when the user is the recipient

akirk:
I was still receiving e-mails when someone's message was a reply to only followers. This adds a check to see if the user is a direct recipient.

### DIFF
--- a/includes/class-mailer.php
+++ b/includes/class-mailer.php
@@ -152,13 +152,13 @@ class Mailer {
 
 		if (
 			is_activity_public( $activity ) ||
-			! in_array( \get_author_posts_url( $user_id ), $recipients )
+			! in_array( \get_author_posts_url( $user_id ), $recipients, true )
 		) {
 			return;
 		}
 
 		// Only accept messages that have the user in the "to" field.
-		if ( ! in_array( \get_author_posts_url( $user_id ), $activity['to'] ) ) {
+		if ( ! in_array( \get_author_posts_url( $user_id ), $activity['to'], true ) ) {
 			return;
 		}
 
--- a/includes/class-mailer.php
+++ b/includes/class-mailer.php
@@ -157,6 +157,11 @@ class Mailer {
 			return;
 		}
 
+		// Only accept messages that have the user in the "to" field.
+		if ( ! in_array( \get_author_posts_url( $user_id ), $activity['to'] ) ) {
+			return;
+		}
+
 		$actor = get_remote_metadata_by_actor( $activity['actor'] );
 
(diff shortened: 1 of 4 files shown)


includes/class-mailer.php
```diff
@@ -148,8 +148,12 @@ public static function new_follower( $notification ) {
 	 * @param int   $user_id  The id of the local blog-user.
 	 */
 	public static function direct_message( $activity, $user_id ) {
-		// Check if Activity is public or not.
-		if ( is_activity_public( $activity ) ) {
+		if (
+			is_activity_public( $activity ) ||
+			// Only accept messages that have the user in the "to" field.
+			empty( $activity['to'] ) ||
+			! in_array( \get_author_posts_url( $user_id ), (array) $activity['to'], true )
```
pfefferle: it can also be the `?author=<id>` URL I think we should use `User->get_id()` instead

includes/class-mailer.php
```diff
@@ -148,8 +148,12 @@ public static function new_follower( $notification ) {
 	 * @param int   $user_id  The id of the local blog-user.
 	 */
 	public static function direct_message( $activity, $user_id ) {
-		// Check if Activity is public or not.
-		if ( is_activity_public( $activity ) ) {
+		if (
+			is_activity_public( $activity ) ||
+			// Only accept messages that have the user in the "to" field.
+			empty( $activity['to'] ) ||
```
pfefferle, line 155: should we support empty `to`? is there a usecase for that?

akirk: This is to avoid a warning in case there is no array key `to`.

includes/class-mailer.php
akirk: Ok, I'll change it!

includes/class-mailer.php
pfefferle: 🤦‍♂️

includes/class-mailer.php
```diff
@@ -148,8 +149,12 @@ public static function new_follower( $notification ) {
...
+			! in_array( ( new User( $user_id ) )->get_id(), (array) $activity['to'], true )
```
pfefferle: how does this even work, the User class has no constructor that sets the ID 😳

I think this should be `Actors::get_by_id( $user_id )->get_id()` instead.

akirk: Oops, thanks for catching it. Likely it worked because it used a 0-id in the test.

pfefferle (approved):
thanks a lot

PR #1104, opened 2024-12-19
Jetpack: Support comment management in Calypso

obenland:
Fixes peLCyI-uB-p2#comment-492

## Proposed changes:

* Adds callbacks to expose custom comment types to the comment API endpoints that power the Calypso comment screen.

## Testing instructions:

* Apply these changes to a WoA test site or any Jetpack site.
* Go to https://wordpress.com/comments/all/SITE
* Make sure you see any like and repost comments and the comment counts are accurate.

image


jeherve (commented):
This doesn't work in my tests on a self-hosted site ; I must be missing something in setting up my site for that review:

image
image

obenland:
@jeherve Is the comment count correct and the comments don't show up? Or does none of it work?

pfefferle (dismissed):

pfefferle (dismissed):

jeherve:
> Is the comment count correct and the comments don't show up? Or does none of it work?

The comment count is off by one for some reason, and the reactions are not displayed at all, only the comments:

**Calypso**

image

**WP-Admin**

Screenshot 2025-01-06 at 14 47 56

obenland:
@jeherve Are there other comment types from other plugins by any chance? 

To display the comments it requires [the `jetpack_json_api_comment_types` filter](https://github.com/Automattic/jetpack/blob/a4e72c9b07e1e575fcb733a505ade99f57db7870/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-comment-endpoint.php#L74-L83) that comes with Jetpack 14.2.

jeherve:
> To display the comments it requires [the `jetpack_json_api_comment_types` filter](https://github.com/Automattic/jetpack/blob/a4e72c9b07e1e575fcb733a505ade99f57db7870/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-comment-endpoint.php#L74-L83) that comes with Jetpack 14.2.

🤦! This is what was missing on my site, and as an idiot, I thought that had already shipped! My bad. It works well now.

jeherve (dismissed):

PR #1134, opened 2025-01-07
Clean up options on uninstall

obenland:
## Proposed changes:

* Adds removals for registered user and post meta, as well as options to the uninstall hook.

Is there more data that we add, that should be removed here?


includes/class-activitypub.php
```diff
@@ -77,6 +77,46 @@ public static function deactivate() {
 	 */
 	public static function uninstall() {
 		Scheduler::deregister_schedules();
+
+		delete_metadata( 'user', 0, '_activitypub_user_identifier', '', true );
```
pfefferle: oh nice, wasn't aware of ID `0` is deleting everything!

pfefferle:
I like to cleanup in general, but this is very destructive and might cause some data loss by accident, even more because it was different before!

What about adding a consent layer when the user clicks the remove link or adding a "remove and delete all data" link?

obenland:
> might cause some data loss by accident, even more because it was different before!

This only runs when the plugin gets deleted, not when it's deactivated.
With it being different than before, do you mean that it now deletes data where it didn't use to?

> What about adding a consent layer when the user clicks the remove link or adding a "remove and delete all data" link?

Core has a built-in "Are You Sure" modal when you try to delete a plugin.

image

includes/class-activitypub.php
```diff
@@ -86,6 +86,46 @@ public static function uninstall() {
...
+		delete_metadata( 'user', 0, 'magic_sig_public_key', '', true );
+		delete_metadata( 'user', 0, 'magic_sig_private_key', '', true );
+
+		delete_metadata( 'post', 0, '_activitypub_actor_json', '', true );
+		delete_metadata( 'post', 0, '_activitypub_canonical_url', '', true );
+		delete_metadata( 'post', 0, '_activitypub_errors', '', true );
+		delete_metadata( 'post', 0, '_activitypub_inbox', '', true );
+		delete_metadata( 'post', 0, '_activitypub_user_id', '', true );
+		delete_metadata( 'post', 0, 'activitypub_content_warning', '', true );
+		delete_metadata( 'post', 0, 'activitypub_content_visibility', '', true );
```
pfefferle: As discussed: let's split this PR and start with cleaning up the options. I would love to have an export of the followers, before we delete them permanantly.

What do you think @Automattic/fediverse

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Comment counts get updated when the plugin is activated/deactivated/deleted
 * Added a filter to make custom comment types manageable in WP.com Calypso
+* Cleanup of option values when the plugin is uninstalled
```
pfefferle: this is in the wrong version now

readme.txt
```diff
@@ -155,6 +155,7 @@ For reasons of data protection, it is not possible to see the followers of other
 
 * Added: Comment counts get updated when the plugin is activated/deactivated/deleted
 * Added: A filter to make custom comment types manageable in WP.com Calypso
+* Added: Cleanup of option values when the plugin is uninstalled
```
pfefferle: wrong version!

pfefferle (dismissed):

PR #1144, opened 2025-01-09
NodeInfo: Use Rest Controller structure

obenland:
Follow-up to #1146.

## Proposed changes:

* Updates NodeInfo endpoint to extend `WP_REST_Controller`.
* Adds unit tests for NodeInfo endpoints.
* Removes built-in support for nodeinfo2.

### Other information:


pfefferle:
@obenland can you please use `git mv` for the following "Use Rest Controller structure" restructurings? This way it is easier to track the changes of the PR.

includes/rest/class-nodeinfo-controller.php
```diff
@@ -0,0 +1,215 @@
...
+
+		\register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/(?P<version>\d\.\d)',
+			array(
+				'args' => array(
+					'version' => array(
+						'description' => 'The version of the NodeInfo schema.',
+						'type'        => 'string',
+						'enum'        => array( '2.0' ),
```
pfefferle: The idea of the `activitypub_rest_nodeinfo_pre` filter is to allow other plugins to add their own NodeInfo versions/endpoints. with this limitation to 2.0 this is no longer possible.

includes/rest/class-nodeinfo-controller.php
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	/**
+	 * Retrieves the NodeInfo discovery profile.
+	 *
+	 * @param \WP_REST_Request $request The request object.
+	 *
+	 * @return \WP_REST_Response Response object.
+	 */
+	public function get_items( $request ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
```
pfefferle, line 79: wasn't it possible to simple remove `$request` param?

what about a more descriptive function name? like `get_discovery` or something?

includes/rest/class-nodeinfo-controller.php
```diff
@@ -0,0 +1,215 @@
...
+		return \rest_ensure_response( $response );
+	}
+
+	/**
+	 * Retrieves the NodeInfo profile.
+	 *
+	 * @param \WP_REST_Request $request The request object.
+	 * @return \WP_REST_Response Response object.
+	 */
+	public function get_item( $request ) {
```
pfefferle, line 102: This is more consistent with the `get_nodeinfo2` function

```suggestion
	public function get_nodeinfo( $request ) {
```

includes/rest/class-nodeinfo-controller.php
obenland: I didn't do a terribly good job documenting the intention of these changes, so I updated #1146 to make up for that:

> With the ambition of this plugin becoming a candidate for Core inclusion, the endpoints should be as close as possible to what Core would expect them to be. This PR is a first step towards that state, with endpoint classes extending the base `WP_REST_Controller` class and defining a schema of the output.

`WP_REST_Controller` creates a structure that can be followed, with [get_items](https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/rest-api/endpoints/class-wp-rest-controller.php#L75-L90) being one of the methods it provides. They're not abstract methods, so they don't _have_ to be implemented, and in cases like this one, where the endpoint doesn't manage a collection of items, it can probably be argued that we don't really need to follow that convention, but it's the closest I can think of.

The `$request` param is part of the function signature in the parent class and so we'll have to follow it to avoid a PHP warning/error.

includes/rest/class-nodeinfo-controller.php
obenland: With the context of how `WP_REST_Controller ` is implemented, `get_nodeinfo2` is the inconsistent part, I'm afraid. I actually thought about suggesting to remove it, however I think it's used to populate https://the-federation.info?

includes/rest/class-nodeinfo-controller.php
obenland: Fixed in 6a0fed3.

includes/rest/class-nodeinfo-controller.php
pfefferle: ah I see! ...this is a very "creative" way of using `get_items` and `get_item` in this context ☺️

the-federation.info also supports NodeInfo, so let's remove it then... there is a NodeInfo plugin that will bring back nodeinfo2 support if someone needs it... Keep it simple, stupid!

includes/rest/class-nodeinfo-controller.php
pfefferle: oh, "becoming a candidate for Core inclusion" is a nice north star ⭐️!

includes/rest/class-nodeinfo-controller.php
```diff
@@ -0,0 +1,218 @@
...
+
+		\register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/(?P<version>\d\.\d)',
+			array(
+				'args' => array(
+					'version' => array(
+						'description' => 'The version of the NodeInfo schema.',
+						'type'        => 'string',
+						'enum'        => array( '1.0', '1.1', '2.0', '2.1' ),
```
pfefferle: Why not just allow any version number in the format x.x? Otherwise, we'd need to manually add each new version, regardless of whether we support it!

(there is already a 2.2 btw. https://github.com/jhass/nodeinfo/tree/main/schemas/2.2 ☺️)

obenland: Yes, the regex is probably enough

obenland: Fixed in 2178c16.

includes/rest/class-nodeinfo-controller.php
obenland: Great! Removed in 2c304a3.

readme.txt
```diff
@@ -142,6 +142,7 @@ For reasons of data protection, it is not possible to see the followers of other
 * Fixed: @-mentions in federated comments being displayed with a line break
 * Fixed: Fetching replies from the same instance for Enable Mastodon Apps
 * Fixed: Image captions not being included in the ActivityPub representation when the image is attached to the post
+* Removed:  Built-in support for nodeinfo2. Use the [NodeInfo plugin](https://wordpress.org/plugins/nodeinfo/) instead.
```
pfefferle:
```suggestion
* Removed: Built-in support for nodeinfo2. Use the [NodeInfo plugin](https://wordpress.org/plugins/nodeinfo/) instead. 
```

includes/rest/class-nodeinfo-controller.php
```diff
@@ -0,0 +1,162 @@
...
+	 * @param \WP_REST_Request $request The request object.
+	 *
+	 * @return \WP_REST_Response Response object.
+	 */
+	public function get_items( $request ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+		$response = array(
+			'links' => array(
+				array(
+					'rel'  => 'https://nodeinfo.diaspora.software/ns/schema/2.0',
+					'href' => get_rest_url_by_path( '/nodeinfo' ),
```
pfefferle: The version is missing

```suggestion
					'href' => get_rest_url_by_path( '/nodeinfo/2.0' ),
```

pfefferle (changes requested):

includes/rest/class-nodeinfo-controller.php
obenland: Nice catch! Fixed in 893a128.

readme.txt
```diff
@@ -138,11 +138,12 @@ For reasons of data protection, it is not possible to see the followers of other
 * Added: A filter to make custom comment types manageable in WP.com Calypso
 * Changed: Hide ActivityPub post meta keys from the custom Fields UI
 * Changed: Bumped minimum required PHP version to 7.2
+* Changed: Print `_activityPubOptions` in the `wp_footer` action on the frontend.
 * Fixed: Undefined array key warnings in various places
 * Fixed: @-mentions in federated comments being displayed with a line break
 * Fixed: Fetching replies from the same instance for Enable Mastodon Apps
 * Fixed: Image captions not being included in the ActivityPub representation when the image is attached to the post
-* Changed: Print `_activityPubOptions` in the `wp_footer` action on the frontend.
+* Removed: Built-in support for nodeinfo2. Use the [NodeInfo plugin](https://wordpress.org/plugins/nodeinfo/) instead.
```
pfefferle: should be under "Unreleased" now.

pfefferle (approved):

PR #1145, opened 2025-01-09
Application: Use Rest Controller structure

obenland:
Follow-up to #1146.

Separates concerns with the `Server` class and makes it a more general REST API plugin.
Future steps could include creating a signature-specific controller that contains `verify_signature()` and endpoints can inherit from.

## Proposed changes:

* Creates new Application controller including schema.
* Adds unit tests for the new controller

### Other information:


pfefferle:
There seem to be a unittest error

pfefferle:
Wrong title?

includes/rest/class-application-controller.php
```diff
@@ -0,0 +1,169 @@
...
+					'format' => 'uri',
+				),
+				'type'                      => array(
+					'type' => 'string',
+					'enum' => array( 'Application' ),
+				),
+				'name'                      => array(
+					'type' => 'string',
+				),
+				'icon'                      => array(
```
pfefferle, line 95: I like the idea, but sadly this is not that easy to verify. Per spec an Icon (and many other elements) can be a `Link` or `Image` object or simply the `URI` as `string` or an array of these: https://www.w3.org/TR/activitystreams-vocabulary/#dfn-icon

obenland: Thanks! That's exactly the type pf feedback I was looking for. Fixed in 75015c3.

pfefferle (approved):

PR #1148, opened 2025-01-10
Introduce Query Singleton, to store/query all ActivityPub related informations based on the $wp_query

pfefferle:
This came up in a discussion with @menrath to better support custom endpoints/uris with content negotiation and authorized fetch.

This PR brings a new Query-Object that generates the ActivityPub object based on the queried object. This simplifies the code and fully relies on the Transformer system to provide custom JSON representations.

props @Menrath 

## Proposed changes:

* Add Query Singleton
* Handle and simplify `is_activitypub_request`
* Get ActivityPub Object(-IDs) based on the queried_object

### Other information:

- [x] Add check if Users, single Posts and whole Post-Types are supported by ActivityPub (to the transformer?)

## Testing instructions:

* Go to an author profile or a blog post
* Add `?activitypub` at the end of the URL or call it with the `application/activity+json` header
* Check if the page returns JSON
* Change the capabilities of the user or the Content-Visibility of a post
* Re-check that the page does no longer return JSON


pfefferle:
I would love to hear your feedback on that @Menrath !

Menrath:
Thanks for drafting this! For completeness and documentation, so that other interested persons can follow this discussion better I will write some more introduction lines, and review/propose improvements in a second step.

This review took me quite some hours. And it's not complete yet :/

**Use cases:**  
1. Redirect of cached remote ActivityPub objects. The internal use case is redirecting cached remote `Note`s [1] which are stored/used as WP_Comments.
2. Dealing with ActivityPub objects that are stored within WordPress but may not be referenced by the it's post URL.
3. Allow ActivityStreams representations of non-public post types (e.g., an Event plugin might have a custom post type for locations: these are non-public in the sense there is no public archive for humans, but I would like them to be fetchable by their ID via ActivityPub, and also be able to send updates of those, rather than having to send Updates off all Events using that location).
4. Additional and new – I wanted to mention it here: Maybe also be able to negotiate to ActivityPub transformers for objects other than _Posts_, _Comments_ or _Users_/Actors, e.g., _Terms_.

**Goals:**
- Readability
- Performance (e.g., not being forced to call the function `is_activitypub_request` again).

**My concrete examples:**

1. Dealing with cached remote ActivityPub objects.
[The Event Bridge For ActivityPub](https://wordpress.org/plugins/event-bridge-for-activitypub/) is inserting remote ActivityPub objects as WP_Posts, they are treated as cache. Therefore they should not be federated by the current [Dispatcher](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/class-activity-dispatcher.php)/[Scheduler](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/class-activity-scheduler.php) class.
This is prevented by adding a filter to 'activitypub_is_post_disabled' which is applied in the [ActivityPub\is_post_disabled](https://github.com/Automattic/wordpress-activitypub/blob/81001534a2ce9cd1ea0d16c2e53083194d8b8ea8/includes/functions.php#L420-L442) function. These posts are displayed locally when queried without ActivityPub request headers. But if it is an ActivityPub requ

    Redirect of cached remote ActivityPubobjects. The internal use case is redirecting cached remote Notes [1] which are stored/used as WP_Comments.est one must be redirected to the origin. These can be achieved in the [same manner as it happens with the comments](https://github.com/Automattic/wordpress-activitypub/blob/81001534a2ce9cd1ea0d16c2e53083194d8b8ea8/includes/class-activitypub.php#L230-L261). 
Maybe this template_redirect function could be more generic and cover all cases.

2. This touched both use cases 2 and 3. I was writing an ActivityPub transformer for [EventPrime](https://de.wordpress.org/plugins/eventprime-event-calendar-management/) events. I needed to Negotiate content that is referenced by a shortcode, the shortcode is referencing a post. The post has the custom post type where EventPrime stores its events. However, this post-type is non-public (which is kind of odd, but not up to my choice – and therefore not processed by the ActivityPub plugin, which is a sane default).

The filter `activitypub_uri_supports_conneg` only brings me one step further in my case.

Because I cannot get passed https://github.com/Automattic/wordpress-activitypub/blob/change/improve-conneg-extensibility/includes/functions.php#L390 as I have no guarantee that the post type `page` is "enabled" for ActivityPub, even though some `page` posts might contain the actual reference to what I want to show in the ActivityPub world.

**Some general ideas:**
I would propose separating concerns a bit more:

* Having a function that checks if the request is an ActivityPub request, in the sense of it contains an activitystreams/json-ld header or the activitypub query parameter.
* Doing the actual negotiation: What is the content being queried, does it have a ActivityPub representation enabled.

pfefferle:
@Menrath I think you are right, especially because we already check for the allowed object types (`is_author`, `is_post`, ...) in several places. So I would also vote for simplifying the `is_activitypub_request` and have the specific checks in the template loader!

Thank you for taking the time to thoroughly review the code!!!!

pfefferle:
what about a `queried_activitypub_object` that is generated with the transformer and stored in the `$wp_query` ???

includes/class-activitypub.php
```diff
@@ -50,11 +50,10 @@ public static function init() {
 		}
 
 		\add_filter( 'activitypub_get_actor_extra_fields', array( Extra_Fields::class, 'default_actor_extra_fields' ), 10, 2 );
+		\add_filter( 'pre_render_activitypub_template', array( self::class, 'add_headers' ) );
```
Menrath: Isn't this actually the other way round?
Shouldn't it be, note the **non**, of course this action does not exist yet.
```php
		\add_filter( 'pre_render_non_activitypub_template', array( self::class, 'add_headers' ) );
```

Menrath:
Actually I like the idea about getting the Transformer once and storing it "globally". This would make it possible to refactor and simplify a lot of functions and checks.

However, I wonder if the Transformer is the right place to pack the functions to determine **if and which ActivityPub actor owns an object**.

> what about a queried_activitypub_object that is generated with the transformer and stored in the $wp_query ???

Also, I am unsure about adding stuff to WordPress globals. I have another idea in mind that might work: Create a new singleton class like ActivityPub_Request, ActivityPub_Negotiation or put it in the ActivityPub class. This would allow storing certain results for later use.

On construct (GET, non-REST requests) this would do some checks like:
* is_activitypub_request
* Negotiate the object the request is targeting (possibly remote or local)
* The visibility of the object (there may be access restrictions)
* which actor owns the object 
* Which transformer will take care of the transformation 
* Register other action hooks and filters

I haven't thought this through in detail, but maybe it will help.

pfefferle:
@Menrath I like the idea a lot and prepared something. That would simplify so much and would allow you to simply use the transformer!

pfefferle:
Would love to also get your feedback on this @Automattic/fediverse, @jeherve and @akirk !

Is this the right way to go?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Hide ActivityPub post meta keys from the custom Fields UI
 * Bumped minimum required PHP version to 7.2
+* Improved conneg and AUTHORIZED_FETCH support for custom extensions
```
obenland:
```suggestion
* Improved content negotiation and AUTHORIZED_FETCH support for third-party plugins
```

This would make it a bit clearer for me as a relatively new reader

includes/class-activitypub.php
```diff
@@ -100,17 +101,16 @@ public static function render_activitypub_template( $template ) {
 			return $template;
 		}
 
+		self::add_headers();
```
obenland, line 104: Should this be called after the `is_activitypub_request()` check?

includes/class-query.php
```diff
@@ -0,0 +1,193 @@
...
+	 * @var object
+	 */
+	private $activitypub_object;
+
+	/**
+	 * Whether the current request is an ActivityPub request.
+	 *
+	 * @var bool
+	 */
+	private $is_activitypub_request;
```
obenland, line 48: This seems to be unused

includes/class-query.php
```diff
@@ -0,0 +1,193 @@
...
+
+		return $this->activitypub_object->get_id();
+	}
+
+	/**
+	 * Get the queried object.
+	 *
+	 * This adds support for Comments by `?c=123` IDs and Users by `?author=123` and `@username` IDs.
+	 *
+	 * @return WP_Term|WP_Post_Type|WP_Post|WP_User|WP_Comment|null The queried object.
```
obenland:
```suggestion
	 * @return \WP_Term|\WP_Post_Type|\WP_Post|\WP_User|\WP_Comment|null The queried object.
```

includes/functions.php
```diff
@@ -365,51 +365,29 @@ function ( $matches ) {
...
 		return true;
 	}
 
 	/*
-	 * The other (more common) option to make an ActivityPub request
-	 * is to send an Accept header.
-	 */
+	* The other (more common) option to make an ActivityPub request
+	* is to send an Accept header.
+	*/
```
```suggestion
	 * The other (more common) option to make an ActivityPub request
	 * is to send an Accept header.
	 */
```
Was this an IDE setting? They should stay aligned

includes/functions.php
```diff
@@ -365,51 +365,29 @@ function ( $matches ) {
...
-		 * - application/ld+json
-		 * - application/json
-		 */
+		* $accept can be a single value, or a comma separated list of values.
+		* We want to support both scenarios,
+		* and return true when the header includes at least one of the following:
+		* - application/activity+json
+		* - application/ld+json
+		* - application/json
+		*/
```
```suggestion
		 * $accept can be a single value, or a comma separated list of values.
		 * We want to support both scenarios,
		 * and return true when the header includes at least one of the following:
		 * - application/activity+json
		 * - application/ld+json
		 * - application/json
		 */
```

includes/functions.php
```diff
@@ -420,17 +398,26 @@ function is_activitypub_request() {
...
+	// Check if the post type supports ActivityPub.
+	if ( ! \post_type_supports( $post->post_type, 'activitypub' ) ) {
 		$disabled = true;
+	} else {
+		$post       = \get_post( $post );
+		$visibility = \get_post_meta( $post->ID, 'activitypub_content_visibility', true );
+
+		if ( ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL === $visibility ) {
+			$disabled = true;
+		}
```
obenland: I'd leave the function as is and just add the `post_type_supports()` check to the if clause

obenland (commented):
Testing steps in the PR would be helpful for me to better understand the changes.
Let's add more unit tests to make sure things work as expected and don't break in the future.

includes/class-activitypub.php
pfefferle: this is the header that should be added to show apps where to find the JSON version, so there is no problem with adding it to all headers

includes/class-query.php
```diff
@@ -0,0 +1,243 @@
...
+		if ( ! $wp_object ) {
+			// If the object is not a valid ActivityPub object, try to get a virtual object.
+			$this->activitypub_object = $this->maybe_get_virtual_object();
+			return;
+		}
+
+		$transformer = Factory::get_transformer( $wp_object );
+
+		if ( $transformer && ! is_wp_error( $transformer ) ) {
+			$this->activitypub_object = $transformer->to_object();
```
Menrath: I see a small performance issue here, that in cases 

* where access to the ActivityPub resource might be denied (imagine a follower-only post in the future)
* a redirect occurs to the origin (remote received items)

the transformation will still be done.

Of course changing the logic here, would lead to a lot of changes: having to change the logic for `has_activitypub_object` etc. too.

As you probably have use cases in mind where you want to use `has_activitypub_object` and `get_activitypub_object`. As these are not used yet and an example would be good to understand why these new functions exist.

includes/class-query.php
```diff
@@ -0,0 +1,243 @@
...
+	 * Get the queried object.
+	 *
+	 * This adds support for Comments by `?c=123` IDs and Users by `?author=123` and `@username` IDs.
+	 *
+	 * @return \WP_Term|\WP_Post_Type|\WP_Post|\WP_User|\WP_Comment|null The queried object.
+	 */
+	public function get_queried_object() {
+		$queried_object = \get_queried_object();
+
+		if ( $queried_object ) {
```
Menrath, line 147: Here a filter could be useful to actually change the `$queried_object`.

Examples of use cases:

**Multilang plugins (see https://github.com/Automattic/wordpress-activitypub/pull/1166):**.
Most multilang plugins use multiple posts to translate a single post, which are then linked together. In the ActivityPub world, there could only be one ActivityPub object with a `contentMap`/`nameMap` containing all the translations. So when querying the pretty permalink of a translation, the ActivityPub-ID should probably be linked to the corresponding post in the main language of the site.

**Content negotiated via shortcodes:**.
The EventPrime event plugin uses `page` posts that only contain shortcodes to link to posts of the `em_event` post type which has set `public => false` (probably some non WordPress-codex legacy, I do not support things like this...) . It would be nice to be able to change the target queried object used for the ActivityPub transformer here.

includes/class-query.php
```diff
@@ -0,0 +1,243 @@
...
+	 */
+	private function __construct() {
+		$wp_object = $this->get_queried_object();
+
+		if ( ! $wp_object ) {
+			// If the object is not a valid ActivityPub object, try to get a virtual object.
+			$this->activitypub_object = $this->maybe_get_virtual_object();
+			return;
+		}
+
```
Menrath: Maybe the redirects for "received" posts/comments etc should be done here? E.g. via an filter `activitypub_was_item_received`/`activitypub_item_is_remote`.

Then the `template_redirect` filter from the `Activitypub` class could be dropped and rewritten as a applied filter for that one. This would also allow other third-party plugins to cleanly have other remote ActivityPub items cached and redirected.

Or maybe even better an action like `maybe_redirect_activitypub_request_to_origin`.

includes/class-query.php
```diff
@@ -0,0 +1,243 @@
...
+	public function get_queried_object_id() {
+		if ( ! $this->get_queried_object() ) {
+			return null;
+		}
+
+		return $this->get_queried_object()->ID;
+	}
+
+	/**
+	 * Get the virtual object.
```
Menrath, line 168: I think it would be extremely helpful to have more comments here, especially for non-insiders.

Virtual Objects currently are things like the Application Actor, or the Blog Actor? Or in other words things that do not have their information stored in a certain way in the database.

includes/transformer/class-comment.php
```diff
@@ -53,6 +54,11 @@ public function change_wp_user_id( $user_id ) {
 	 * @return string The User-URL.
 	 */
 	protected function get_attributed_to() {
+		// If the comment was received via ActivityPub, return the author URL.
```
Menrath, line 57: This fix does not seem to be related to this PR?

includes/transformer/class-factory.php
```diff
@@ -72,12 +75,22 @@ public static function get_transformer( $data ) {
...
+				if ( 'attachment' === $data->post_type && ! is_post_disabled( $data ) ) {
 					return new Attachment( $data );
+				} elseif ( ! is_post_disabled( $data ) ) {
+					return new Post( $data );
 				}
-				return new Post( $data );
+				// fall-through.
 			case 'WP_Comment':
-				return new Comment( $data );
+				if ( ! is_local_comment( $data ) ) {
```
Menrath, line 85: I had a hard time understanding this change as the wording `is_local_comment` is ambiguous.
I see now that it refers to comments that have been made locally but are not federated via ActivityPub, because there are normal received WordPress commends via the default form etc.

`is_non_activitypub_comment` would be clearer. Or an inverse function called `is_federated_comment`.

includes/class-query.php
pfefferle: > the transformation will still be done.

the latest changes to the transformer will no longer run transformations for posts, comments or actors that are not enabled or public and the result is stored as query var, so this should be very effective and fast.

> As you probably have use cases in mind where you want to use has_activitypub_object and get_activitypub_object. As these are not used yet and an example would be good to understand why these new functions exist.

the `get_activitypub_object` and `has_activitypub_object` functions are already in use!

See the `activitypub-json.php` template and the `Query::get_activitypub_object_id` function.

includes/class-query.php
pfefferle: I am not sure if I like the query to automatically redirect! you voted for a separating of concerns and I like the query object to only check and return things but not redirecting. If hooking the object to filters I fear that it might result in coincidents and maybe race conditions.

includes/class-query.php
pfefferle: do we really need a filter here? if there is a queried object, it will be transformed, so it uses the functionality you already implemented for EventPrime and others!?!

includes/transformer/class-comment.php
pfefferle: kind of... because this plugin is now responsible for conneg, I found this bug of the current code...

includes/class-query.php
Menrath: You are totally right. I wrote this in the concern of the above, to maybe prevent a non-needed full transformation of an item.

includes/class-query.php
Menrath: I think it would be the right way to go. But I would love to hear what the others think.

E.g., the solutions I came up for EventPrime (only this event plugin) with are kind of dirty:
1. Change the ActivityPub template to a custom one which changes the object and calls the transformer factory again in the last moment.
2. Checking if it's an ActivityPub request and change the global $post variable with the new target. (pobably won't work any longer with this changes, unless I call the `__destructor` method, hmm).

includes/class-query.php
Menrath: > the latest changes to the transformer will no longer run transformations for posts, comments or actors that are not enabled or public and the result is stored as query var, so this should be very effective and fast.

I could not find the appropriate code that does this?! 🧐 (For example for posts: https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/transformer/class-post.php#L74).
And while it is possible that the transformer could deal with this, I think it should not be the transformers task.
For example, take comments: if queried directly, received remote ActivityPub comments should be redirected and not transformed. But if a replies collection containing such a comment is queried, it probably should be transformed.

Menrath: I was wondering if the singleton class `Activitypub\Query` truly needs to manage anything beyond identifying 'what transformer', 'whether it is an ActivityPub request' and 'what the ActivityPub ID' is.

Maybe the transformer should have a public get_activitypub_id function which just wraps it's protected get_id.

Menrath:
@pfefferle I have formulated some perhaps petty questions in my review, but on the whole I think these changes are fantastic!

includes/class-query.php
pfefferle: but for conneg I need that anyways!

includes/class-query.php
pfefferle: I will see how I might change it to a later binding method. to not call it on the constructor. but I like the idea of a queried activitypub object a lot. even more if it is cached then and can be used at several states in one request.

includes/class-query.php
pfefferle: ah, now I understand what you mean with `redirecting`! I am curious what Mastodon is doing in this case!

includes/class-query.php
pfefferle: As I wrote below, I will see if it is possible to have a late binding on that, instead of calling it always in the constructor!

pfefferle:
thanks for your review @Menrath ! I will try to optimize the code to use late bindings instead of the constructor.

includes/class-query.php
Menrath: Mastodon:

1. ActivityPub request: It always redirects to the origin. E.g., https://graz.social/@ORF_Steiermark@orfodon.org/113833276827247699
2. Non-ActivityPub request: 
   * Logged-In: No-Redirect if the post is already cached.
   * Not-Logged-In: Redirect but with a human warning that that tells the user that the page is left.

```shell
curl --request GET \                                                                                                                                                                    
  --url https://graz.social/@ORF_Steiermark@orfodon.org/113833276827247699 \
  --header 'accept: application/activity+json'  -I -L -o /dev/null -w "%{url_effective}"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0  4433    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
https://orfodon.org/users/ORF_Steiermark/statuses/113833276742240716
```

```shell
curl --request GET \                                                                                                                                                                      
  --url https://graz.social/@ORF_Steiermark@orfodon.org/113833276827247699 \
  -I -L -o /dev/null -w "%{url_effective}" 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0 13909    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
https://graz.social/redirect/statuses/113833276827247699
```

includes/class-query.php
Menrath: > but for conneg I need that anyways!

Would't the ActivityPub ID be enough?

pfefferle: I like to have both... it helps also for the preview... but I try to disconnect them from each other: https://github.com/Automattic/wordpress-activitypub/pull/1148/commits/4b954f35510013b103769bdd133f72eb720d7440

pfefferle: I think a `to_id` transformation is not bad at all, because it also helps for announces and likes.

pfefferle:
Let me know what you think about the latest changes @Menrath !?!

includes/class-query.php
```diff
@@ -0,0 +1,278 @@
...
+	 * Get the queried object ID.
+	 *
+	 * @return int The queried object ID.
+	 */
+	public function get_queried_object_id() {
+		if ( ! $this->get_queried_object() ) {
+			return null;
+		}
+
+		return $this->get_queried_object()->ID;
```
Menrath: The property ID is not available on WP_Term and WP_Post_Type.

includes/class-query.php
```diff
@@ -0,0 +1,278 @@
...
+	 * Get the ActivityPub object ID.
+	 *
+	 * @return int The ActivityPub object ID.
+	 */
+	public function get_activitypub_object_id() {
+		if ( $this->activitypub_object_id ) {
+			return $this->activitypub_object_id;
+		}
+
+		$wp_object                   = $this->get_queried_object();
```
Menrath: These lines of code are a bit duplicated. Also saving the transformer could make this a bit cleaner, it also prevents the same transformer from being initialized multiple times. But apart from that it's maybe just a personal preference.

```php
		if ( ! isset( $this->transformer ) ) {
			$this->set_transformer();
		}

		if ( $this->transformer && ! is_wp_error( $this->transformer ) ) {
			$this->activitypub_object_id = $this->transformer->to_id();
		}
```

includes/class-query.php
```diff
@@ -0,0 +1,278 @@
...
+	 * Whether the current request is an ActivityPub request.
+	 *
+	 * @var bool
+	 */
+	private $is_activitypub_request;
+
+	/**
+	 * The constructor.
+	 *
+	 * Transform the queried object to an ActivityPub object.
```
Menrath: This comment is outdated

includes/class-query.php
```diff
@@ -0,0 +1,278 @@
...
+	 */
+	private function __construct() {
+		// Do nothing.
+	}
+
+	/**
+	 * The destructor.
+	 */
+	public function __destruct() {
```
Menrath, line 60: I don't know the reason why a destructor could/is needed here. But that could be my lack of years of coding experience.

Menrath (commented):
I did not find time to have a detailed view, on the outbox PR yet, as I get more and more the feeling that those two are quite related, so there may be more to say. But atm the last changes look good to me!

includes/class-query.php
Menrath: @pfefferle So you're not in favor of adding a filter? I genuinely believe that the example above of multilang-plugins where multiple WordPress posts lead to a single ActivityPub Query, is a valid use case. Allowing for a filter here would provide a clean and readable solution. That said, I'll also explore whether such implementation would work together with the new outbox scheduling.

includes/class-query.php
pfefferle: this is mostly to be able to manually destroy the object for the unittests

pfefferle:
I think this is a good improvement so far... what do you think about merging it and then start improving it even more!?!

Menrath:
@pfefferle Merging as it, I'd be totally happy! It makes it easy to create separate issues/PRs for the different points of potential improvement! And it's easier to follow to others as well.

pfefferle:
@Automattic/fediverse what do you think?

obenland (approved):

PR #1149, opened 2025-01-10
Interaction: Use Rest Controller structure

obenland:
Follow-up to #1146.

## Proposed changes:

* Updates Interactions endpoint to extend `WP_REST_Controller`.
* Adds unit tests for Interactions endpoint.

### Other information:


includes/rest/class-interaction-controller.php
```diff
@@ -46,28 +51,19 @@ public static function register_routes() {
...
 	 *
-	 * @return WP_REST_Response Redirect to the editor or die.
+	 * @return \WP_REST_Response|\WP_Error Response object on success, or WP_Error object on failure.
 	 */
-	public static function get( $request ) {
+	public function get_item( $request ) {
 		$uri          = $request->get_param( 'uri' );
 		$redirect_url = null;
-		$object       = Http::get_remote_object( $uri );
+		$object       = \Activitypub\Http::get_remote_object( $uri );
```
pfefferle: let's stick with using `use` for Namespaced classes. this is easier to read and consistent with the rest of the code.

obenland: Fixed in fbbd9d5.

includes/rest/class-interaction-controller.php
```diff
@@ -46,28 +53,19 @@ public static function register_routes() {
...
+	public function get_item( $request ) {
 		$uri          = $request->get_param( 'uri' );
 		$redirect_url = null;
 		$object       = Http::get_remote_object( $uri );
 
-		if (
-			\is_wp_error( $object ) ||
-			! isset( $object['type'] )
-		) {
-			\wp_die(
```
pfefferle, line 64: this was a wp_die, because it might be user facing.

pfefferle: what about adding a switch based on the accept header? if `text/html` return `wp_die` otherwise a `WP_Error`?

obenland: I wondered about that! What's the scenario for when it's user-facing? How can I test that?

includes/rest/class-interaction-controller.php
```diff
@@ -104,31 +102,22 @@ public static function get( $request ) {
...
 		 * @param string $redirect_url The URL to redirect to.
 		 * @param string $uri          The URI of the object.
-		 * @param array  $object       The object.
+		 * @param array  $object       The object being interacted with.
 		 */
 		$redirect_url = \apply_filters( 'activitypub_interactions_url', $redirect_url, $uri, $object );
 
 		// Check if hook is implemented.
 		if ( ! $redirect_url ) {
-			\wp_die(
```
pfefferle, line 117: same as above

includes/rest/class-interaction-controller.php
pfefferle: go to a user on a mastodon instance, where you have no account on (for example https://mastodon.art/@pfefferle), click the follow button, then use your blogs WebFinger-ID in the follow dialogue.

pfefferle: better build a custom URI instead of using an API endpoint?

pfefferle (approved):

PR #1159, opened 2025-01-13
Add release script

obenland:
Adds a release script that replaces version numbers and opens a PR automatically. See #1158.

## Testing instructions:

* `npm run release`


obenland:
@Automattic/fediverse Do you think this would be helpful?

pfefferle (approved):
totally misunderstood the usecase and thought it was for the .com deployment 🫣

this is awesome!

PR #1161, opened 2025-01-14
Fix missing migration for v4.7.0

pfefferle:
The renaming of the meta-tags #943 was planned for version 4.6.0, so the migration was set to this version, but we postponed the release to 4.7.0 without updating the migration, that is why it never ran. This results in empty follower lists.

This PR updates the version number for the migration, so that it will run after we release 4.7.1

Fixes https://wordpress.org/support/topic/did-4-7-0-break-things/

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


akirk (approved):
Nice! I think it would be worth to add a high priority label.

PR #1162, opened 2025-01-14
Actions: Check migration versions before release

obenland:
## Proposed changes:

* A GitHub action that makes sure newly added upgrade routines run on the same version number as the tag that's being released.
* Also runs on PRs that add a new upgrade routine to make sure the version number is higher than the current version.


obenland:
I asked Windsurf for help with this. I'm not sure the PR check is hugely helpful.
For new tags we'd have to make sure that it prevents syncing to WP.org. Can we abort tag creation with an action?

PR #1165, opened 2025-01-14
Transformers: Migrate from wp_object to items

obenland:
Non-functional changes that split out the property updates around wp_object => item from #593.

### DIFF
--- a/includes/transformer/class-base.php
+++ b/includes/transformer/class-base.php
@@ -28,28 +28,28 @@ abstract class Base {
 	 *
 	 * @var WP_Post|WP_Comment
 	 */
-	protected $wp_object;
+	protected $item;
 
 	/**
 	 * Static function to Transform a WordPress Object.
 	 *
 	 * This helps to chain the output of the Transformer.
 	 *
-	 * @param WP_Post|WP_Comment $wp_object The WordPress object.
+	 * @param WP_Post|WP_Comment $item The WordPress object.
 	 *
 	 * @return Base
 	 */
-	public static function transform( $wp_object ) {
-		return new static( $wp_object );
+	public static function transform( $item ) {
+		return new static( $item );
 	}
 
 	/**
 	 * Base constructor.
 	 *
-	 * @param WP_Post|WP_Comment $wp_object The WordPress object.
+	 * @param WP_Post|WP_Comment $item The WordPress object.
 	 */
-	public function __construct( $wp_object ) {
-		$this->wp_object = $wp_object;
+	public function __construct( $item ) {
+		$this->item = $item;
 	}
 
 	/**
@@ -124,7 +124,7 @@ abstract class Base {
 	 * @return array The replies collection.
 	 */
 	public function get_replies() {
-		return Replies::get_collection( $this->wp_object );
+		return Replies::get_collection( $this->item );
 	}
 
 	/**
--- a/includes/transformer/class-base.php
+++ b/includes/transformer/class-base.php
@@ -31,6 +31,15 @@ abstract class Base {
 	protected $item;
 
 	/**
+	 * The WP_Post or WP_Comment object.
+	 *
+	 * @deprecated 4.8.0 Use $item instead.
+	 *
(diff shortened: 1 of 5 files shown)


obenland:
@pfefferle Can you share the why behind these changes? 

I remember that there were some prohibited variable names we had to deal with when we did all the PHPCS stuff, is this an extension of that? I think it took offense with `$object`, `$wp_object` should be fine?

pfefferle:
This is because with the new transformers it does not have to be a WordPress object any more.

includes/transformer/class-base.php
```diff
@@ -28,28 +28,28 @@ abstract class Base {
...
 	/**
 	 * Base constructor.
 	 *
-	 * @param WP_Post|WP_Comment $wp_object The WordPress object.
+	 * @param WP_Post|WP_Comment $item The WordPress object.
 	 */
-	public function __construct( $wp_object ) {
-		$this->wp_object = $wp_object;
+	public function __construct( $item ) {
+		$this->item = $item;
```
Menrath: Please also add a fallback to also set wp_object to allow third party transformers to adjust this.

obenland: Oh, I see, a back-compat layer for third-party transformers relying on `wp_object` to be there?

obenland: Fixed in 5ad4eee.

Menrath: Thanks! Can be marked as solved.

pfefferle (approved):
If @Menrath is fine, I am too!

Menrath:
I have some doubts about this change and try to explain why, but in order to do so I have give a more general answer, also targeting PR #593. So this is kind of a mixed review, I hope that is O.K.

I wonder whether `item` is the best term (in terms of readability and technical specificity), not just here, but especially in the context of [the PR adding the outbox collection](https://github.com/Automattic/wordpress-activitypub/pull/593).
The [ActivityPub specification](https://www.w3.org/TR/activitypub/#outbox) describes the outbox stream as containing **activities** that the user has published. Consequently, I would suggest using the name `activity` instead of `item` within the outbox collection, as it aligns better with the specification and improves clarity.

For the transformer, I believe `wp_object` is a quite fitting name. It reflects that the entity represents a WordPress construct, such as a WP_Post, WP_Comment, WP_Term, or potentially other types. `item`, on the other hand, does not convey this relationship as clearly. Perhaps one could also update the PHPDoc comment for the transformer to include `mixed`, as this would make the expected (?) flexibility of the wp_object type more explicit.

So when I put this together: The transformer classes take care of the `object` which may be set in an `Activity` (outbox collection item).

The wording `item` in the `Shortcodes` class is fine and clear enough for me. It’s not wrong per se, as it is only used locally within that function, and the PHPDoc provides sufficient

I will take some time tomorrow to review the current status of the outbox PR in detail. So far, I’ve only taken a quick look at it.  


Off topic: I want to point out the naming of the namespace `Activitypub\Activity.` When I first started working with this plugin, it confused me quite a bit. The namespace contains all kinds of object types, while `Activity` is a specific subtype of the generic ActivityPub object (e.g., `Activitypub\Activitypub`, `Activitypub\Object`, `Activitypub\Object_Types\`). That said, I understand this might be a legacy decision, and I think it’s something I can live with. :)

obenland:
Thank you for your thoughtful feedback!

We can't use `object` since it's [a reserved word in PHP](https://www.php.net/manual/en/reserved.other-reserved-words.php) and `wp_object` [is no longer an accurate description](https://github.com/Automattic/wordpress-activitypub/pull/1165#issuecomment-2591708644) of the content. `activity_object`, which came up previously, is ambiguous with Activities being Objects themselves. I'm kind of at a loss for a more semantic name for it. :(

Building on your PHPDoc suggestion, is this problem maybe something that could be "solved" with more verbose docs for these properties?

pfefferle:
Let's close this, because we are about to release the outbox PR this, latest next week!

PR #1166, opened 2025-01-15
Add support for WPML post locale

thomas-pike:
Adds WPML compatibility

## Proposed changes:

* Gets the post locale from WPML Multilingual CMS plugin if installed

### Other information:

## Testing instructions:

* Install the WPML Multilingual CMS plugin
* Publish a post or a translation of an existing post in a language other than the default site language.
* Check that the activity object is sent out with the expected language.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+* Support for WPML post locale
+
 ### Removed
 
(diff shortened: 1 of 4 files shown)


Menrath:
This is a good first step. Wouldn't it be better to actually have one ActivityPub object, that contains all the languages in the `contentMap`? The current development in https://github.com/Automattic/wordpress-activitypub/pull/1148 might be able to make this possible without much overhead.

mattwiebe:
Thanks for the PR @thomas-pike !

pfefferle (approved):
I think this is a nice start and we can improve on this.

PR #1131, opened 2025-01-07
Frontend scripts: print options later

mattwiebe:
I noticed that hooking to `wp_enqueue_scripts` on the frontend was printing our `_activitypubOptions` var nearly at the top. Seemed bad for SEO among other things.

## Proposed changes:

Before:
Screenshot 2025-01-07 at 12 27 46

After (it's much lower, still there, just not way up top)
Screenshot 2025-01-07 at 12 27 18

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

All frontend blocks that rely on `_activitypubOptions` still works: Follow Me, Followers, Reactions

PR #1168, opened 2025-01-15
Use `_activityPubOptions` at runtime

mattwiebe:
It turns out that #1131 breaks stuff on dotcom. But even printing in `wp_head` wasn't good enough, because the scripts were demanding the option to be present at instantiation time. But with a more flexible approach, we can assume that the option will be there at runtime, which is good enough. So I introduced a `getOptions` helper to do that.

Note that we could get around this by adding proper inline script dependencies to each script in turn, but that was giving problems with the script handles registered through `block.json` via `register_block_type_from_metadata`. This is why I moved to just outputting it.

## Testing

All blocks, in editor and the frontend, should keep working properly


pfefferle (dismissed):

src/shared/get-options.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Get options from window._activityPubOptions.
+ *
+ * @param {string|undefined} key - Optional key to retrieve specific option.
+ * @return {*|boolean} Option value if found, false if not found, or entire options object if no key provided.
+ */
+const getOptions = ( key ) => {
```
obenland: How would you feel about `getOption` to reflect only the one option that's being passed?

mattwiebe: I think that would have been a good idea, I didn't like `getOptions` which is why I switch instead to destructuring from `useOptions` with https://github.com/Automattic/wordpress-activitypub/pull/1168/commits/6277ac6b8f2c3a719eefbddc372eba5420cdd42f

obenland (dismissed):

PR #1170, opened 2025-01-15
Outbox: Update API Endpoint

obenland:
## Proposed changes:

* Updates `Outbox::add()` to return any WP_Errors it receives so we can return them in the API.
* Updates Outbox endpoint to extending `WP_REST_Controller`.
* Adds `per_page` collection parameter.
* Moves `user_id` validation into its own callback.
* Switches Outbox data from posts to `ap_outbox` items.
* Adds `prepare_item_for_response()` callback to convert posts to Activity objects.
* Adds collection schema information.
* Updates Json transformer to define its own `get_cc()` callback.
* Adds unit tests for Outbox endpoint.

### Other information:

## Testing instructions:

* Download [the ZIP file of this branch](https://github.com/Automattic/wordpress-activitypub/archive/refs/heads/update/outbox-endpoint.zip) and upload it to a test site.
* Deactivate prod ActivityPub plugin and activate this one.
* Create a few posts and comments, maybe edit and delete some of them.
* View your site/account on your favorite instance and make sure it shows these items.


includes/rest/class-outbox-controller.php
```diff
@@ -0,0 +1,284 @@
...
+	 * @return array Response object on success, or WP_Error object on failure.
+	 */
+	public function prepare_item_for_response( $item, $request ) {
+		$transformer = Factory::get_transformer( $item->post_content );
+
+		$type  = 'Object';
+		$terms = wp_get_object_terms( $item->ID, 'ap_activity_type' );
+		if ( isset( $terms[0]->name ) ) {
+			$type = ucfirst( $terms[0]->name );
+		}
```
obenland: These lines don't feel very reassuring. Is there a way to store the activity type with the outbox item itself somehow?

includes/transformer/class-json.php
```diff
@@ -30,4 +30,15 @@ public function __construct( $item ) {
...
 	}
+
+	/**
+	 * Returns the public secondary audience of this object
+	 *
+	 * @see https://www.w3.org/TR/activitystreams-vocabulary/#dfn-cc
+	 *
+	 * @return array The secondary audience of this object.
+	 */
+	protected function get_cc() {
```
obenland, line 41: I don't quite understand why I had to add this here to keep the cc in the activity object. If I didn't the cc entry would just be empty. @pfefferle Do you know what's missing here? The cc is available in the `$item`, but it doesn't get carried over into the Activity object in `prepare_item_for_response()` ^

tests/includes/collection/class-test-outbox.php
```diff
@@ -53,7 +53,7 @@ public function activity_object_provider() {
 				),
 				'Create',
 				1,
-				'{"@context":["https://www.w3.org/ns/activitystreams",{"Hashtag":"as:Hashtag","sensitive":"as:sensitive"}],"id":"https://example.com/1","type":"Note","content":"This is a note","contentMap":{"en":"This is a note"},"tag":[],"to":["https://www.w3.org/ns/activitystreams#Public"],"cc":[],"mediaType":"text/html","sensitive":false}',
+				'{"@context":["https://www.w3.org/ns/activitystreams",{"Hashtag":"as:Hashtag","sensitive":"as:sensitive"}],"id":"https://example.com/1","type":"Note","content":"This is a note","contentMap":{"en":"This is a note"},"tag":[],"to":["https://www.w3.org/ns/activitystreams#Public"],"mediaType":"text/html","sensitive":false}',
```
obenland, line 59: Same as above. Which of my changes caused the `cc` to disappear?

obenland:
@Automattic/fediverse  Can you think of a good way to get real world examples for post requests to the outbox endpoint?

includes/rest/class-outbox-controller.php
pfefferle: I planned the `ap_activity_type` taxonomy for that: https://github.com/Automattic/wordpress-activitypub/pull/593/files#diff-480b2542905cc428144536831a25aa56f1c4307d2667a40f719353b6ed91bf8dR574

includes/transformer/class-json.php
pfefferle: This is one thing we should re-work. currently every implementer has to know how to add the `to` and `cc` to have a private, public, ... Activity. We should use the content-visibility flags, to set the audience for an Activity and the dispatcher and/or collection should set the correct `cc` and `to` based on that informations.

includes/rest/class-outbox-controller.php
```diff
@@ -0,0 +1,317 @@
...
+		$outbox_query = new \WP_Query();
+		$query_result = $outbox_query->query( $query_args );
+
+		$response = array(
+			'@context'     => array( 'https://www.w3.org/ns/activitystreams' ),
+			'id'           => \get_rest_url( null, \sprintf( 'actors/%d/outbox', $user_id ) ),
+			'generator'    => 'https://wordpress.org/?v=' . \get_bloginfo( 'version' ),
+			'actor'        => $user->get_id(),
+			'type'         => 'OrderedCollectionPage',
+			'partOf'       => \get_rest_url( null, \sprintf( 'actors/%d/outbox', $user_id ) ),
```
pfefferle: this seem to ignore the namespace (`http://localhost:8076/wp-json/actors/1/outbox` instead of `http://localhost:8076/wp-json/activitypub/1.0/actors/1/outbox`). besides of that, we should use `get_rest_url_by_path` so that it will be compatible with WordPress.com

includes/rest/class-outbox-controller.php
```diff
@@ -0,0 +1,317 @@
...
+			'totalItems'   => $outbox_query->found_posts,
```
pfefferle, line 170: this also seems to be broken. on my test system this returns `13` but I can only see two results.

includes/rest/class-outbox-controller.php
obenland: Thanks, fixed in 5d53ef8.

includes/rest/class-outbox-controller.php
obenland: With the understanding that we'll only look at Outbox items and not any other posts, this can be simplified. Fixed in 3f7a27d.

includes/rest/class-outbox-controller.php
```diff
@@ -0,0 +1,302 @@
...
+
+		$query_args = array(
+			'posts_per_page' => $request->get_param( 'per_page' ),
+			'author'         => $user_id > 0 ? $user_id : null,
+			'paged'          => $page,
+			'post_type'      => Outbox::POST_TYPE,
+			'post_status'    => 'draft',
+		);
+
+		$outbox_query = new \WP_Query();
```
pfefferle, line 160: what about moving this to the outbox collection class? this way you can use it in different contexts?

obenland: Let's wait until there's a need for it?

includes/rest/class-outbox-controller.php
```diff
@@ -0,0 +1,294 @@
...
+							'default'     => 10,
+							'minimum'     => 1,
+							'maximum'     => 100,
+						),
+					),
+				),
+				array(
+					'methods'             => \WP_REST_Server::CREATABLE,
+					'callback'            => array( $this, 'create_item' ),
+					'permission_callback' => array( 'Activitypub\Rest\Server', 'verify_signature' ),
```
pfefferle: posting should not be depending on the signature, but on an authentication method!

pfefferle:
Now that we have a persistent outbox, we can use the ID of the outbox-item as ID for the Activity, but no idea how to set that properly!

includes/rest/class-outbox-controller.php
pfefferle:
```suggestion
					'permission_callback' => function () {
						return current_user_can( 'activitypub' );
					},
```

obenland: Fixed in 230f859.

includes/rest/class-outbox-controller.php
```diff
@@ -0,0 +1,296 @@
...
+				),
+			),
+		);
+
+		if ( ! current_user_can( 'activitypub' ) ) {
+			$args['meta_query'][] = array(
+				'key'     => 'activitypub_content_visibility',
+				'value'   => array( ACTIVITYPUB_CONTENT_VISIBILITY_QUIET_PUBLIC, ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL ),
+				'compare' => 'NOT IN',
+			);
```
pfefferle: I think the current check requires the meta and for the public posts it is not, so they will not be included.

Shouldn't it be instead:

```suggestion
			$args['meta_query'][] = array(
				'key'     => 'activitypub_content_visibility',
				'compare' => 'NOT EXISTS',
			);
```

includes/rest/class-outbox-controller.php
```diff
@@ -0,0 +1,296 @@
...
+
+		$args = array(
+			'posts_per_page' => $request->get_param( 'per_page' ),
+			'author'         => $user_id > 0 ? $user_id : null,
+			'paged'          => $page,
+			'post_type'      => Outbox::POST_TYPE,
+			'post_status'    => 'draft',
+
+			// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+			'meta_query'     => array(
```
pfefferle, line 126: should we add the limitation also only to the public queries?

obenland: Sure, works for me

pfefferle:
What about some integration tests, to test the query limitation in the logged in and logged out state?

obenland:
`test_get_items_content_visibility()` is currently testing [for logged in vs logged out](https://github.com/Automattic/wordpress-activitypub/pull/1170/files#diff-fa19e16ffd687c0cb509463423251180ff751eb6fcc63e0d9baff78b6c1b3803R433-R450), is that what you're looking for?

pfefferle:
I was just curious why the changed result was not erroring ;)

pfefferle (approved):
I am working on the broken tests on the `add/outbox-collection` branch... this has nothing to do with this PR, so please merge it...

PR #1178, opened 2025-01-17
Outbox: Generalize functionality of the transformers

pfefferle:
To better support

* extraction of mentions for arbitrary data
* generating the audience based on the content visibility
* ...


includes/transformer/class-activity-object.php
```diff
@@ -25,7 +25,7 @@ public function to_object() {
 	 *
 	 * @return array The list of @-Mentions.
 	 */
-	protected function get_mentions() {
```
obenland: Do we need to keep this around for third party plugins?

pfefferle: 🤔 good point

pfefferle: ok, I think it is not worth changing it then! I will revert it!

obenland: You could just deprecate it

includes/transformer/class-base.php
```diff
@@ -116,8 +125,74 @@ protected function transform_object_properties( $activity_object ) {
...
+	 * @return Base_Object The ActivityPub Object.
+	 */
+	protected function set_audience( $activity_object ) {
+		$public = 'https://www.w3.org/ns/activitystreams#Public';
+		$actor  = Actors::get_by_resource( $this->get_attributed_to() );
+		if ( ! $actor || is_wp_error( $actor ) ) {
+			$followers = array();
+		} else {
+			$followers = $actor->get_followers();
+		}
```
obenland, line 177:
```suggestion
		$followers = array();
		if ( method_exists( $actor, 'get_followers' ) ) {
			$followers = $actor->get_followers();
		}
```
Not sure if that's much better?

includes/transformer/class-base.php
```diff
@@ -243,4 +305,69 @@ protected function get_summary_map() {
...
+		if ( $mentions ) {
+			foreach ( $mentions as $mention => $url ) {
+				$tag    = array(
+					'type' => 'Mention',
+					'href' => \esc_url( $url ),
+					'name' => \esc_html( $mention ),
+				);
+				$tags[] = $tag;
+			}
+		}
```
```suggestion
		foreach ( $mentions as $mention => $url ) {
			$tags[] = array(
				'type' => 'Mention',
				'href' => \esc_url( $url ),
				'name' => \esc_html( $mention ),
			);
		}
```
Could be simplified a little bit. If `$mentions` is empty it'll just not iterate

obenland (approved):

obenland:
YOLO

PR #1181, opened 2025-01-17
Migration: flush post caches

mattwiebe:
The migrations from #1161 did not flush post caches, which can lead to problems on sites with a persistent object cache.

Aka `get_post_meta( ID, '_activitypub_actor_json' )` would return empty even though the data was in the DB.

## Proposed changes:
* Add a migration to flush caches

### Other information:

## Testing instructions:
On a site with a persistent object cache:
* Start a site on ActivityPub 4.6.0. 
* Receive a follower
* Upgrade to this version

Everything should work normally. Without this we were seeing Followers not display correctly, eg in the Followers block or in wp-admin Followers screen


obenland (approved):
Code review

PR #1186, opened 2025-01-20
Update Dispatcher to use Outbox

pfefferle:
Simple dispatcher based on the Outbox-Collection.

This is a simple rewrite of the current dispatcher system, to use the Outbox instead of the Scheduler.

This is a first draft and will be improved over time, to better handle:

* Re-tries
* Errors
* Logging
* Batch processing


mattwiebe:
> This is a first draft and will be improved over time, to better handle:
> 
> * Re-tries
> * Errors
> * Logging
> * Batch processing

I'm going to take on some or all of those, let's see how far I get.

mattwiebe:
@pfefferle I assume we need to get these tests passed? I'd really like this merged into #593 before I bring in my error/logging stuff

includes/class-dispatcher.php
```diff
@@ -0,0 +1,206 @@
...
+ * @author Matthias Pfefferle
+ *
+ * @see https://www.w3.org/TR/activitypub/
+ */
+class Dispatcher {
+	/**
+	 * Initialize the class, registering WordPress hooks.
+	 */
+	public static function init() {
+		\add_action( 'activitypub_process_outbox', array( self::class, 'process_outbox' ), 10, 1 );
```
obenland:
```suggestion
		\add_action( 'activitypub_process_outbox', array( self::class, 'process_outbox' ) );
```
Priority 10 and 1 accepted arg are [defaults](https://developer.wordpress.org/reference/functions/add_action/).

includes/class-dispatcher.php
```diff
@@ -0,0 +1,206 @@
...
+		}
+
+		$json = $activity->to_json();
+
+		foreach ( $inboxes as $inbox ) {
+			safe_remote_post( $inbox, $json, $actor_id );
+		}
+
+		$outbox_item->post_status = 'publish';
+		\wp_update_post( $outbox_item );
```
obenland:
```suggestion
		\wp_publish_post( $outbox_item );
```
I don't know if one is "better" than the other but I wanted to offer it

includes/rest/class-outbox-controller.php
```diff
@@ -120,8 +120,7 @@ public function get_items( $request ) {
 			'author'         => $user_id > 0 ? $user_id : null,
 			'paged'          => $page,
 			'post_type'      => Outbox::POST_TYPE,
-			'post_status'    => 'draft',
-
+			'post_status'    => 'any',
```
```suggestion
			'post_status'    => 'any',

```
You don't want to keep the new line?

mattwiebe:
#1197 adds logging, and can form the basis of batching and retries as well

includes/functions.php
```diff
@@ -1554,43 +1555,34 @@ function is_self_ping( $id ) {
...
+	$outbox_activity_id = Outbox::add( $activity_object, $activity_type, $user_id, $content_visibility );
 
-	if ( ! $id ) {
+	if ( ! $outbox_activity_id ) {
 		return false;
 	}
 
-	$hook = 'activitypub_process_outbox';
-	$args = array( $id );
+	\do_action( 'post_activitypub_add_to_outbox', $outbox_activity_id, $activity_object, $user_id, $content_visibility );
```
obenland, line 1593: This would benefit from hook docs

includes/transformer/class-base.php
```diff
@@ -146,7 +146,7 @@ public function to_activity( $type ) {
 		$activity->set_object( $object );
 
 		// Use simple Object (only ID-URI) for Like and Announce.
-		if ( in_array( $type, array( 'Like', 'Announce' ), true ) ) {
+		if ( in_array( $type, array( 'Like' ), true ) ) {
```
```suggestion
		if ( 'Like' === $type ) {
```

includes/scheduler/class-post.php
```diff
@@ -104,4 +110,41 @@ public static function schedule_post_activity( $new_status, $old_status, $post )
...
+
+		// Only if the activity is a Create, Update or Delete.
+		if ( ! in_array( $activity_type, array( 'Create', 'Update', 'Delete' ), true ) ) {
+			return;
+		}
+
+		$transformer = Factory::get_transformer( $activity_object );
+		$activity    = $transformer->to_activity( $activity_type );
+
+		Outbox::add( $activity, 'Announce', Actors::BLOG_USER_ID, ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC );
```
obenland: Further up we bail if `$actor_id` is the blog user id, here we set it to the blog user id, could you help me understand the difference?

pfefferle: If the `$user_id` of the original Activity is the one of the blog user, it should not `Announce` its own Activities.

includes/scheduler/class-post.php
```diff
@@ -45,6 +49,8 @@ public static function transition_attachment_status( $post_id ) {
 				self::schedule_post_activity( 'trash', '', $post_id );
 				break;
 		}
+
+		\add_action( 'post_activitypub_add_to_outbox', array( self::class, 'send_announces' ), 10, 4 );
```
obenland: @pfefferle Was this meant to be added here, rather than in `init()`? Or did I mess up a rebase?

pfefferle: this was meant to be in the init!

pfefferle: you messed it up ;)

obenland: haha let me fix it!

mattwiebe: Oh, I messed that up in a merge somewhere, ha

obenland (approved):

obenland (approved):

PR #1188, opened 2025-01-20
Outbox: Use callbacks that can be unhooked.

obenland:
## Proposed changes:

* Moves closures into methods that can be unhooked in tests.

### DIFF
--- a/includes/scheduler/class-comment.php
+++ b/includes/scheduler/class-comment.php
@@ -25,18 +25,7 @@ class Comment {
 
 		// Comment transitions.
 		\add_action( 'transition_comment_status', array( self::class, 'schedule_comment_activity' ), 20, 3 );
-		\add_action(
-			'edit_comment',
-			function ( $comment_id ) {
-				self::schedule_comment_activity( 'approved', 'approved', $comment_id );
-			}
-		);
-		\add_action(
-			'wp_insert_comment',
-			function ( $comment_id ) {
-				self::schedule_comment_activity( 'approved', '', $comment_id );
-			}
-		);
+		\add_action( 'wp_insert_comment', array( self::class, 'schedule_comment_activity_on_insert' ), 10, 2 );
 	}
 
 	/**
@@ -83,5 +72,17 @@ class Comment {
 		}
 
 		add_to_outbox( $comment, $type, $comment->user_id );
+	}
+
+	/**
+	 * Schedule Comment Activities on insert.
+	 *
+	 * @param int         $comment_id Comment ID.
+	 * @param \WP_Comment $comment    Comment object.
+	 */
+	public static function schedule_comment_activity_on_insert( $comment_id, $comment ) {
+		if ( 1 === (int) $comment->comment_approved ) {
+			self::schedule_comment_activity( 'approved', '', $comment );
+		}
 	}
 }
--- a/tests/includes/scheduler/class-test-comment.php
+++ b/tests/includes/scheduler/class-test-comment.php
@@ -38,6 +38,9 @@ class Test_Comment extends \WP_UnitTestCase {
 	public static function wpSetUpBeforeClass( $factory ) {
 		self::$user_id = $factory->user->create( array( 'role' => 'author' ) );
 		self::$post_id = $factory->post->create( array( 'post_author' => self::$user_id ) );
+
+		// Add activitypub capability to the user.
+		get_user_by( 'id', self::$user_id )->add_cap( 'activitypub' );
 	}
 
 	/**
--- a/tests/includes/scheduler/class-test-comment.php
+++ b/tests/includes/scheduler/class-test-comment.php
@@ -1,0 +1,168 @@
+<?php
+/**
+ * Test Comment scheduler class.
+ *
+ * @package Activitypub\Tests\Scheduler
+ */
+
+namespace Activitypub\Tests\Scheduler;
+
+use Activitypub\Collection\Outbox;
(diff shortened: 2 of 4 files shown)


includes/scheduler/class-comment.php
```diff
@@ -25,18 +25,7 @@ public static function init() {
 
 		// Comment transitions.
 		\add_action( 'transition_comment_status', array( self::class, 'schedule_comment_activity' ), 20, 3 );
-		\add_action(
-			'edit_comment',
-			function ( $comment_id ) {
-				self::schedule_comment_activity( 'approved', 'approved', $comment_id );
-			}
-		);
```
obenland, line 33: @pfefferle `edit_comment` has a call to `wp_transition_comment_status()` following [a few lines later](https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/comment.php#L2654-L2654). Do we need a separate callback for that?

includes/scheduler/class-comment.php
```diff
@@ -25,18 +25,7 @@ public static function init() {
 
 		// Comment transitions.
 		\add_action( 'transition_comment_status', array( self::class, 'schedule_comment_activity' ), 20, 3 );
```
obenland, line 27: @pfefferle What's the reason for going with priority `20` rather than `10`?

includes/scheduler/class-comment.php
```diff
@@ -84,4 +73,16 @@ public static function schedule_comment_activity( $new_status, $old_status, $com
...
+	/**
+	 * Schedule Comment Activities on insert.
+	 *
+	 * @param int         $comment_id Comment ID.
+	 * @param \WP_Comment $comment    Comment object.
+	 */
+	public static function schedule_comment_activity_on_insert( $comment_id, $comment ) {
+		if ( 1 === (int) $comment->comment_approved ) {
+			self::schedule_comment_activity( 'approved', '', $comment );
+		}
```
obenland, line 86: Should a comment Create activity only be added when the comment was approved?

includes/scheduler/class-comment.php
pfefferle: I wanted to get sure that all other hooks where called before we schedule the federation process.

includes/scheduler/class-comment.php
pfefferle: Sure!

I think this was not checked yet, because we only federate comments of registered blog users and I think they will be auto-approved!?!

But better to have it checked explicitly 👍

includes/scheduler/class-comment.php
pfefferle: 🤔 I don't think so...

pfefferle: 🤔 seems to be obsolete then!

sadly I can't remember why I initially added it!

maybe we should check if there is a unit-test that checks for `create` and `update`???

pfefferle (approved):
I like your changes, maybe re-check if all cases are handled by unit-tests...

tests/includes/scheduler/class-test-comment.php
```diff
@@ -0,0 +1,169 @@
...
+				'user_id'          => self::$user_id,
+				'comment_approved' => 0,
+			)
+		);
+		$activitpub_id = \Activitypub\Comment::generate_id( $comment_id );
+
+		wp_set_comment_status( $comment_id, 'approve' );
+
+		$post = $this->get_leatest_outbox_item( $activitpub_id );
+		$this->assertSame( $activitpub_id, $post->post_title );
```
obenland, line 84: @pfefferle This test fails because the above is considered [a "local" comment ](https://github.com/Automattic/wordpress-activitypub/blob/update/comment-outbox/includes/transformer/class-factory.php#L87)and [the Transformer called in add_to_outbox](https://github.com/Automattic/wordpress-activitypub/blob/update/comment-outbox/includes/functions.php#L1564) returns an error.

Is there more nuance to what kind of comment we expect an Outbox item for?

tests/includes/scheduler/class-test-comment.php
```diff
@@ -0,0 +1,169 @@
...
+			array(
+				'comment_post_ID'  => self::$post_id,
+				'user_id'          => self::$user_id,
+				'comment_approved' => 1,
+			)
+		);
+		$activitpub_id = \Activitypub\Comment::generate_id( $comment_id );
+
+		$post = $this->get_leatest_outbox_item( $activitpub_id );
+		$this->assertSame( $activitpub_id, $post->post_title );
```
obenland, line 103: Same as above ^

includes/scheduler/class-comment.php
obenland: Fixed in [cbe3451](https://github.com/Automattic/wordpress-activitypub/pull/1188/commits/cbe345110361da8a489ba840cecfc36996dc4974).

tests/includes/scheduler/class-test-comment.php
pfefferle: only comments by registered users with the `activitypub` cap.

obenland: Adding the cap doesn't seem to fix it. It looks like it expects the comment to already having been federated via ActivityPub or received via ActivityPub.

obenland: Looks like you changed it in #1186. Adopting that fixes the tests.

pfefferle (approved):

PR #1191, opened 2025-01-21
Fix rewrite rules for NodeInfo

obenland:
## Proposed changes:

* Add an upgrade routine that flushes rewrite rules to account for changes in 4.7.2

### Other information:

## Testing instructions:

* On a test site install [4.7.1](https://downloads.wordpress.org/plugin/activitypub.4.7.1.zip) and update to 4.7.2.

### DIFF
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -167,6 +167,9 @@ class Migration {
 		if ( \version_compare( $version_from_db, '4.7.2', '<' ) ) {
 			self::migrate_to_4_7_2();
 		}
+		if ( \version_compare( $version_from_db, 'unreleased', '<' ) ) {
+			add_action( 'init', 'flush_rewrite_rules', 20 );
+		}
 
 		/*
 		 * Add new update routines above this comment. ^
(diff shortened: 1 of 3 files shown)


includes/class-migration.php
```diff
@@ -167,6 +167,9 @@ public static function maybe_migrate() {
 		if ( \version_compare( $version_from_db, '4.7.2', '<' ) ) {
 			self::migrate_to_4_7_2();
 		}
+		if ( \version_compare( $version_from_db, 'unreleased', '<' ) ) {
```
mattwiebe, line 170: Why `unreleased`? Are you just trying to ensure that we always flush rewrites in any and every upgrade routine?

obenland: The release script will replace it with the next version that gets released. It's supposed to help avoid what happened in 4.7.0 where an upgrade script was merged after the targeted release had already gone out (#1161).

mattwiebe: Aha! Very nice. Should be good then.

mattwiebe (dismissed):

pfefferle (approved):

PR #1194, opened 2025-01-21
Releases: Replace versions in upgrade scripts

obenland:
It's hard to test without a pending upgrade routine in trunk, I'm afraid.

## Proposed changes:

* Changes version regex to hopefully replace upgrade routine versions correctly in the future


obenland:
This is not the right fix it looks like. I'll circle back later.

PR #1196, opened 2025-01-21
Outbox: Tests and fixes for Post Scheduler

obenland:
## Proposed changes:

* Adds tests for Post and Actor Schedulers.
* Fixes user updates.
* Moves extrafields handling into Actor class.
* Updates Transformer to account for inheritance in Activity objects.

### Other information:

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1561,16 +1561,20 @@ function is_self_ping( $id ) {
  * @return boolean|int The ID of the outbox item or false on failure.
  */
 function add_to_outbox( $data, $type = 'Create', $user_id = 0, $content_visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
-	$transformer = Transformer_Factory::get_transformer( $data );
+	if ( $data instanceof \Activitypub\Activity\Base_Object ) {
+		$activity = $data;
+	} else {
+		$transformer = Transformer_Factory::get_transformer( $data );
 
-	if ( ! $transformer || is_wp_error( $transformer ) ) {
-		return false;
-	}
+		if ( ! $transformer || is_wp_error( $transformer ) ) {
+			return false;
+		}
 
-	$activity = $transformer->to_object();
+		$activity = $transformer->to_object();
 
-	if ( ! $activity || is_wp_error( $activity ) ) {
-		return false;
+		if ( ! $activity || is_wp_error( $activity ) ) {
+			return false;
+		}
 	}
 
 	set_wp_object_state( $data, 'federate' );
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1561,13 +1561,24 @@ function is_self_ping( $id ) {
  * @return boolean|int The ID of the outbox item or false on failure.
  */
 function add_to_outbox( $data, $type = 'Create', $user_id = 0, $content_visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
+	$trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+	$debug = false;
+	foreach ($trace as $call) {
+		if (isset($call['class']) && isset($call['function'])) {
+			if ($call['class'] === 'Activitypub\Tests\Scheduler\Test_Actor' && $call['function'] === 'test_user_update') {
+				// Method was called
+				$debug = true;
+			}
+		}
+	}
 	$transformer = Transformer_Factory::get_transformer( $data );
-
+$debug ? var_dump($data,$transformer) : null;
 	if ( ! $transformer || is_wp_error( $transformer ) ) {
 		return false;
 	}
 
 	$activity = $transformer->to_object();
+	$debug ? var_dump($activity) : null;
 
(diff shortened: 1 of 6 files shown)


includes/functions.php
```diff
@@ -1561,16 +1561,20 @@ function is_self_ping( $id ) {
...
+		}
 
-	$activity = $transformer->to_object();
+		$activity = $transformer->to_object();
 
-	if ( ! $activity || is_wp_error( $activity ) ) {
-		return false;
+		if ( ! $activity || is_wp_error( $activity ) ) {
+			return false;
+		}
```
obenland: @Automattic/fediverse Heads up! ^ This accounts for `Actors` already being "Activities" I guess and failing transformation.

includes/scheduler/class-post.php
```diff
@@ -51,28 +51,26 @@ function ( $post_id ) {
 	public static function schedule_post_activity( $new_status, $old_status, $post ) {
 		$post = get_post( $post );
 
-		if ( ! $post || is_post_disabled( $post ) ) {
+		if ( ! $post ) {
```
obenland: `is_post_disabled()` needs to be moved down so `ap_extrafield` and `ap_extrafield_blog` posts are not considered disabled and can be processed.

includes/scheduler/class-post.php
```diff
@@ -51,28 +51,26 @@ function ( $post_id ) {
...
 
-		// Do not send activities if post is password protected.
-		if ( \post_password_required( $post ) ) {
+		if ( is_post_disabled( $post ) ) {
 			return;
 		}
 
-		// Check if post-type supports ActivityPub.
-		$post_types = \get_post_types_by_support( 'activitypub' );
-		if ( ! \in_array( $post->post_type, $post_types, true ) ) {
```
obenland, line 75: Checking for post type support is now part of `is_post_disabled()`

includes/scheduler/class-post.php
pfefferle: It was pragmatic to add the `ap_extrafield` to the hook, but technically it should live in the actor scheduler... should we keep it here for now?

obenland: I think I'd prefer to move it to the Actor scheduler

includes/functions.php
pfefferle: I fixed that in my Dispatcher PR, so that this should work properly... I would keep the transformation, because of possible additions to the `cc` and `to` fields: https://github.com/Automattic/wordpress-activitypub/pull/1186/commits/a90e03a6608975aff518318c8162753944e12663

pfefferle: https://github.com/Automattic/wordpress-activitypub/pull/1186/commits/a90e03a6608975aff518318c8162753944e12663

obenland: That doesn't seem to work. The transformer is looking at [the class itself](https://github.com/Automattic/wordpress-activitypub/blob/a90e03a6608975aff518318c8162753944e12663/includes/transformer/class-factory.php#L31) and not the instance, so it doesn't account for inheritance.

pfefferle: but we should fix it in the transformer, to stay consistent!

obenland: Fixed in 904848b1e95efcc13c067aa1bad2aff284d5d21a

includes/transformer/class-factory.php
```diff
@@ -7,6 +7,7 @@
 
 namespace Activitypub\Transformer;
 
+use Activitypub\Activity\Base_Object;
```
pfefferle: Is that possible? I thought you have to always use the full path in the check!?!

obenland: I think only if we were to pass it as a string. But I'm happy to move it inline to be safe.

pfefferle: no no, I am just curious!!!

obenland:
I'm at a loss as to why tests are failing in 7.2. They succeed for me locally :(

obenland:
> I'm at a loss as to why tests are failing in 7.2. They succeed for me locally :(

Turns out we were using an action that was only introduced in WP 6.3, while our min supported version is 5.5.

includes/scheduler/class-actor.php
```diff
@@ -89,6 +92,21 @@ public static function blog_user_update( $value = null ) {
...
 	}
 
+	/**
+	 * Schedule Activities.
+	 *
+	 * @param string   $new_status New post status.
+	 * @param string   $old_status Old post status.
+	 * @param \WP_Post $post       Post object.
+	 */
+	public static function schedule_post_activity( $new_status, $old_status, $post ) {
```
mattwiebe, line 102: If I named this function, I would call it `maybe_schedule_profile_update`. You can take it or leave it :)

I do like splitting it out from the rest of the post updates.

mattwiebe (approved):
Looks good to me!

includes/scheduler/class-post.php
pfefferle: I can't find this check in the `is_post_disabled` function!?!

pfefferle (approved):

pfefferle:
Nice one!

PR #1200, opened 2025-01-22
Outbox: Fix Query

pfefferle:
If the `author` field is null, the `WP_Query` will return every activity regardless of the Actor.

To have a proper result for the blog and application Actor, we have to also add the `_activitypub_activity_actor` to the query.

### DIFF
--- a/includes/rest/class-outbox-controller.php
+++ b/includes/rest/class-outbox-controller.php
@@ -137,8 +137,8 @@ class Outbox_Controller extends \WP_REST_Controller {
 			// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
 			'meta_query'     => array(
 				array(
-					'key'     => '_activitypub_activity_actor',
-					'value'   => $actor_type,
+					'key'   => '_activitypub_activity_actor',
+					'value' => $actor_type,
 				),
 			),
 		);
--- a/includes/rest/class-outbox-controller.php
+++ b/includes/rest/class-outbox-controller.php
@@ -115,15 +115,32 @@ class Outbox_Controller extends \WP_REST_Controller {
 		 */
 		$activity_types = apply_filters( 'rest_activitypub_outbox_activity_types', array( 'Announce', 'Create', 'Like', 'Update' ) );
 
+		switch ( $user_id ) {
+			case -1:
+				$actor_type = 'application';
+				break;
+			case 0:
+				$actor_type = 'blog';
+				break;
+			default:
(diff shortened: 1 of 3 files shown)


includes/rest/class-outbox-controller.php
```diff
@@ -115,15 +115,32 @@ public function get_items( $request ) {
...
 			'post_type'      => Outbox::POST_TYPE,
-			'post_status'    => 'draft',
+			'post_status'    => 'any',
 
 			// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
 			'meta_query'     => array(
+				array(
+					'key'     => '_activitypub_activity_actor',
+					'value'   => $actor_type,
+					'compare' => '=',
```
```suggestion
```
We should be able to remove that, [it's the default](https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/class-wp-meta-query.php#L544).

obenland: A few lines down we're removing all meta queries for users with the `activitypub` capability—is this one that should be maintained?

obenland (commented):
It would be great to add tests to cover this change

includes/rest/class-outbox-controller.php
pfefferle: yes, you should never see results of a different user in your endpoint!

pfefferle: Good point!

pfefferle:
Besides of that, the check seems to be wrong. We would have to check if the user that is logged in is the same user that owns the endpoint!

obenland:
@pfefferle Would you want to add tests that check for the change you made for who gets to see who's data? And the actor type meta?

obenland (approved):

tests/includes/rest/class-test-outbox-controller.php
```diff
@@ -491,6 +502,183 @@ public function test_get_items_content_visibility( $visibility, $public_visible,
...
+					'_activitypub_activity_actor'    => 'user',
+					'activitypub_content_visibility' => ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC,
+				),
+			)
+		);
+
+		// Create a post with blog actor type.
+		$blog_post_id = self::factory()->post->create(
+			array(
+				'post_author'  => $user_id,
```
pfefferle: Shouldn't this `0` then?

```suggestion
				'post_author'  => 0,
```

pfefferle:
@obenland thanks much for adding tests!!!

PR #1205, opened 2025-01-23
Accept Follower: Migrate to outbox

pfefferle:
This PR adds:

* Support for adding Activities to the Outbox
* Use Outbox to send `Accept` responses to `Follow` requests

This should also fix https://community.nodebb.org/topic/18518/wordpress-follow-failure

/cc @julianlam

### Other information:


includes/functions.php
```diff
@@ -1600,3 +1601,62 @@ function add_to_outbox( $data, $activity_type = 'Create', $user_id = 0, $content
...
+			'TentativeAccept',
+			'TentativeReject',
+			'Travel',
+			'Undo',
+			'Update',
+			'View',
+		)
+	);
+
+	if ( is_array( $data ) && isset( $data['type'] ) ) {
```
obenland, line 1653:
```suggestion
	if ( isset( $data['type'] ) ) {
```
If `$data` is not an array, the isset() will return false

includes/functions.php
```diff
@@ -1600,3 +1601,62 @@ function add_to_outbox( $data, $activity_type = 'Create', $user_id = 0, $content
...
+		return in_array( $data['type'], $types, true );
+	}
+
+	if ( is_object( $data ) && $data instanceof Base_Object ) {
```
obenland, line 1657:
```suggestion
	if ( $data instanceof Base_Object ) {
```
If `$data` is not an object, the instanceof check will fail

tests/includes/class-test-functions.php
```diff
@@ -216,4 +216,111 @@ public function object_to_uri_provider() {
...
+			array( 'type' => null ),
+			array(),
+		);
+
+		foreach ( $invalid_activities as $activity ) {
+			$this->assertFalse(
+				\Activitypub\is_activity( $activity ),
+				sprintf( 'Activity type %s should be invalid', isset( $activity['type'] ) ? $activity['type'] : 'null' )
+			);
+		}
```
obenland: Let's break these out in two separate tests with `dataProvider`s for each Activity type

tests/includes/class-test-functions.php
```diff
@@ -216,4 +216,111 @@ public function object_to_uri_provider() {
...
+			new \stdClass(),
+		);
+
+		foreach ( $invalid_inputs as $input ) {
+			$this->assertFalse(
+				\Activitypub\is_activity( $input ),
+				sprintf( 'Input of type %s should be invalid', gettype( $input ) )
+			);
+		}
+	}
```
obenland, line 301: This test can be part of the "invalid Activities" test

includes/handler/class-follow.php
```diff
@@ -102,21 +104,9 @@ public static function send_follow_response( $actor, $activity_object, $user_id,
...
-		$activity->set_actor( $user->get_id() );
-		$activity->set_to( $actor );
-		$activity->set_id( $user->get_id() . '#follow-' . \preg_replace( '~^https?://~', '', $actor ) . '-' . \time() );
-
-		$activity = $activity->to_json();
+		// Send response only to the Follower.
+		$activity_object['to'] = $actor;
 
-		Http::post( $inbox, $activity, $user_id );
+		add_to_outbox( $activity_object, 'Accept', $user_id, ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE );
```
obenland, line 110: it looks like `send_follow_response()` no longer sends a follow response?

obenland (changes requested):

includes/functions.php
pfefferle: I have to re-add the check, otherwise the code tries to access an access key also on objects and that fails!

obenland (approved):

PR #1208, opened 2025-01-23
Outbox: Fall back to blog author for non-authorized authors

obenland:
I noticed Outbox items being created when I imported content, including from authors that were not part of the site, and did not have the `activtipub` cap.

## Proposed changes:

* Skip adding Outbox items for post authors that don't have the cap.
* Adds unit test to cover that case.

### Other information:

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1580,9 +1580,13 @@ function add_to_outbox( $data, $activity_type = 'Create', $user_id = 0, $content
 		return false;
 	}
 
-	// If the user is disabled, use the blog user.
-	if ( is_user_disabled( $user_id ) && ! is_user_disabled( Actors::BLOG_USER_ID ) ) {
-		$user_id = Actors::BLOG_USER_ID;
+	// If the user is disabled, fall back to the blog user when available.
+	if ( is_user_disabled( $user_id ) ) {
+		if ( is_user_disabled( Actors::BLOG_USER_ID ) ) {
+			return false;
+		} else {
+			$user_id = Actors::BLOG_USER_ID;
+		}
 	}
 
 	set_wp_object_state( $data, 'federate' );
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1580,6 +1580,11 @@ function add_to_outbox( $data, $activity_type = 'Create', $user_id = 0, $content
 		return false;
 	}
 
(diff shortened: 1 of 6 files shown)


includes/scheduler/class-post.php
```diff
@@ -71,6 +71,11 @@ public static function schedule_post_activity( $new_status, $old_status, $post )
 			return;
 		}
 
+		// Skip if the post author is not allowed to publish activities.
+		if ( $post->post_author > 0 && ! \user_can( $post->post_author, 'activitypub' ) ) {
```
pfefferle: What about `is_user_disabled`?

obenland: Ah, yes!

obenland: Fixed in 6e6b973.

includes/scheduler/class-post.php
```diff
@@ -71,6 +72,11 @@ public static function schedule_post_activity( $new_status, $old_status, $post )
 			return;
 		}
 
+		// Skip if the post author is not allowed to publish activities.
+		if ( is_user_disabled( $post->post_author ) ) {
```
pfefferle: Shouldn't it be?

```suggestion
		if ( is_user_disabled( $post->post_author ) && is_user_disabled( Actors::BLOG_USER_ID ) ) {
```

obenland: Fixed in febccc5.

includes/functions.php
```diff
@@ -1574,6 +1574,11 @@ function add_to_outbox( $data, $activity_type = 'Create', $user_id = 0, $content
 		return false;
 	}
 
+	// If the user is disabled, use the blog user.
+	if ( is_user_disabled( $user_id ) && ! is_user_disabled( Actors::BLOG_USER_ID ) ) {
+		$user_id = Actors::BLOG_USER_ID;
+	}
```
pfefferle, line 1590: we also need the check the case `if ( is_user_disabled( $user_id ) && is_user_disabled( Actors::BLOG_USER_ID ) )` to bail out then!?!

obenland: Thanks! Fixed in 25b1e73.

pfefferle (approved):

PR #1210, opened 2025-01-23
Admin: Don't require publish_post cap to get activitypub cap

obenland:
Gives context to why a user couldn't be given ActivityPub access, rather than failing silently.

## Proposed changes:

* Adds a callback to display an error message 

image

## Testing instructions:

* Create a new user with a Subscriber role.
* In the Users list, select the Subscriber and enable Activitypub for them.
* See the error message.

### DIFF
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -650,15 +650,12 @@ class Activitypub {
 	}
 
 	/**
-	 * Add the 'activitypub' capability to users who can publish posts.
+	 * Add the 'activitypub' capability to new users.
 	 *
 	 * @param int $user_id User ID.
 	 */
 	public static function user_register( $user_id ) {
-		if ( \user_can( $user_id, 'publish_posts' ) ) {
-			$user = \get_user_by( 'id', $user_id );
-			$user->add_cap( 'activitypub' );
-		}
+		\get_user_by( 'id', $user_id )->add_cap( 'activitypub' );
 	}
 
 	/**
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -650,12 +650,15 @@ class Activitypub {
 	}
 
 	/**
-	 * Add the 'activitypub' capability to new users.
+	 * Add the 'activitypub' capability to users who can publish posts.
 	 *
 	 * @param int $user_id User ID.
 	 */
 	public static function user_register( $user_id ) {
-		\get_user_by( 'id', $user_id )->add_cap( 'activitypub' );
+		if ( \user_can( $user_id, 'publish_posts' ) ) {
+			$user = \get_user_by( 'id', $user_id );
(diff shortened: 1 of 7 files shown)


pfefferle:
This was on purpose. The `publish_post` cap is not required, because big news companies might want to post in the name of guest authors for example.

obenland:
Hm, not quite sure I understand. It currently only adds the `activtipub` cap if the user already has `publish_posts`.

The PR doesn't change that, fwiw, it just surfaces a message now when that condition gets hit, to no longer fail silently.

pfefferle:
No, this is only the default we set when you install the plugin, but you can add the cap to every user regardless of the `publish_posts` cap.

includes/class-admin.php
```diff
@@ -781,11 +782,12 @@ public static function handle_bulk_request( $sendback, $action, $users ) {
 
 		foreach ( $users as $user_id ) {
 			$user = new \WP_User( $user_id );
-			if (
-				'add_activitypub_cap' === $action &&
-				user_can( $user_id, 'publish_posts' )
```
obenland, line 786: @pfefferle Are we talking about the same thing? This is checking for publish_posts.

pfefferle: then this is a bug and should check the `activitypub` cap

includes/class-activitypub.php
```diff
@@ -555,15 +555,12 @@ private static function register_post_types() {
...
+	 * Add the 'activitypub' capability to new users.
 	 *
 	 * @param int $user_id User ID.
 	 */
 	public static function user_register( $user_id ) {
-		if ( \user_can( $user_id, 'publish_posts' ) ) {
-			$user = \get_user_by( 'id', $user_id );
-			$user->add_cap( 'activitypub' );
-		}
+		\get_user_by( 'id', $user_id )->add_cap( 'activitypub' );
```
pfefferle: By default we only want to add the "activitypub" capability to users that can edit posts. The admin can change this afterwards.

includes/class-migration.php
```diff
@@ -514,15 +514,8 @@ public static function add_default_settings() {
...
-		$users = \get_users(
-			array(
-				'capability__in' => array( 'publish_posts' ),
-			)
-		);
-
-		// Add ActivityPub capability to all users that can publish posts.
-		foreach ( $users as $user ) {
+		// Add ActivityPub capability to all users.
+		foreach ( \get_users() as $user ) {
```
pfefferle: Same here: Only users with the `publish_post` capability should be migrated, the rest has to be done by hand.

includes/class-activitypub.php
obenland: Oh, I misunderstood your comment then

pfefferle (approved):

PR #1214, opened 2025-01-24
Outbox: Fix race condition with Post-Metas

pfefferle:
The post meta data seems not yet saved when `transition_post_status` is triggered.

### DIFF
--- a/includes/scheduler/class-post.php
+++ b/includes/scheduler/class-post.php
@@ -160,10 +160,10 @@ class Post {
 	/**
 	 * Filter the post data before it is inserted via the REST API.
 	 *
-	 * @param \WP_Post         $post     The post object.
+	 * @param \stdClass        $post     An object representing a single post prepared for inserting or updating the database.
 	 * @param \WP_REST_Request $request  The request object.
 	 *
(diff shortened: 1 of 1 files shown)


includes/scheduler/class-post.php
```diff
@@ -149,4 +156,35 @@ public static function send_announces( $outbox_activity_id, $activity_object, $a
...
 		Scheduler::schedule_outbox_activity_for_federation( $outbox_activity_id );
 	}
+
+	/**
+	 * Filter the post data before it is inserted via the REST API.
+	 *
+	 * @param \WP_Post         $post     The post object.
+	 * @param \WP_REST_Request $request  The request object.
+	 *
+	 * @return \WP_Post The post object.
```
obenland:
```suggestion
	 * @param \stdClass        $post     An object representing a single post prepared for inserting or updating the database.
	 * @param \WP_REST_Request $request  The request object.
	 *
	 * @return \stdClass The prepared post.
```
Not sure [if it matters](https://developer.wordpress.org/reference/hooks/rest_pre_insert_this-post_type/#parameters)

obenland (approved):

PR #1222, opened 2025-01-28
Check response body for tombstone type (#1209)

### LINKED ISSUE #1209: Deletion not being respected if tombstone remains
### Quick summary

The plugin checks if the object is "really" a tombstone before deleting the object.  But this check is only that the response code is 404 or 410, which matches Mastodon's behaviour.  Friendica's behaviour is to return the object as a tombstone.  Since this is a valid response, the plugin concludes that this is not a tombstone after all, and leaves the comment in place.

The plugin should actually parse the response, if there is one, and check if the object type is tombstone.  This should also result in `is_tombstone` returning true.

### Steps to reproduce

1. Write an article on Wordpress
2. Load article URL in Friendica
3. Make a comment on the article in Friendica
4. Approve the article in Wordpress, confirm that it is visible
5. Delete the comment in Friendica

### Site owner impact

### Severity

Moderate

No revenue impact

_No response_

mexon:
Allows remote servers to delete messages if they return valid `Tombstone` responses, rather than requiring them to reply with 404 or 410 response codes.  The response codes are recommended by the spec but not required.  Friendica, at least, replies with a `Tombstone` and a 200 response code.

Fixes #1209

## Proposed changes:

* In case response code is neither 404 or 410, check if body is an object with type `Tombstone`, and returns true if so.

### Other information:

## Testing instructions:

* On Wordpress, make a new post
* On Friendica, load the new post
* Make a reply in Friendica
* Wait until reply is visible on Wordpress, approving if necessary
* On Friendica, delete the reply
* Verify that reply is deleted in Wordpress


pfefferle (dismissed):
Thanks a lot! This looks good to me, I only simplified the tests a bit.

pfefferle (approved):

PR #1223, opened 2025-01-28
Outbox: Basic rescheduling

pfefferle:
If the initial scheduler fails, the object will never be federated. This recurring scheduler re-adds schedules for pending activities and re-adds a scheduler for these items.


includes/collection/class-outbox.php
```diff
@@ -74,4 +74,23 @@ public static function add( $activity_object, $activity_type, $user_id, $content
...
+	 * @param int $limit The number of activities to get.
+	 *
+	 * @return WP_Post[] The pending activities.
+	 */
+	public static function get_pending( $limit = 10 ) {
+		$activites = \get_posts(
+			array(
+				'post_type'      => self::POST_TYPE,
+				'post_status'    => 'pending',
+				'posts_per_page' => $limit,
```
obenland:
```suggestion
				'posts_per_page' => $limit,
				'fields'         => 'ids',
```
Since we only need IDs in `reprocess_outbox`, we can make this query faster by not asking for the whole object.

includes/collection/class-outbox.php
```diff
@@ -74,4 +74,23 @@ public static function add( $activity_object, $activity_type, $user_id, $content
 
 		return $id;
 	}
+
+	/**
+	 * Get pending activities.
+	 *
+	 * @param int $limit The number of activities to get.
+	 *
+	 * @return WP_Post[] The pending activities.
```
obenland:
```suggestion
	 * @return \WP_Post[] The pending activities.
```
Makes my IDE happy

includes/collection/class-outbox.php
```diff
@@ -74,4 +74,23 @@ public static function add( $activity_object, $activity_type, $user_id, $content
...
+	 */
+	public static function get_pending( $limit = 10 ) {
+		$activites = \get_posts(
```
obenland: We could avoid an unnecessary local variable and return the results right away

includes/class-scheduler.php
```diff
@@ -59,6 +60,10 @@ public static function register_schedules() {
 		if ( ! \wp_next_scheduled( 'activitypub_cleanup_followers' ) ) {
 			\wp_schedule_event( time(), 'daily', 'activitypub_cleanup_followers' );
 		}
+
+		if ( ! \wp_next_scheduled( 'activitypub_reprocess_outbox' ) ) {
+			\wp_schedule_event( time(), 'hourly', 'activitypub_reprocess_outbox' );
```
obenland, line 67: Is this still missing a hook to connect the `activitypub_reprocess_outbox` event to the `reprocess_outbox` callback? Or did I miss that?

obenland (approved):

PR #1226, opened 2025-01-28
Outbox: Fix `to`, `cc` and `id`

pfefferle:
Use Outbox `guid` as `id` and `array_unique` for `to` and `cc`.

props @obenland


obenland (approved):
Love the unit tests!

PR #1230, opened 2025-01-28
Outbox: Account for invalid json

obenland:
Not quite sure this is the best way to go. I'd be fine with restoring `init_from_json()` to working with an empty array, too.

See p1738085487855409/1738079794.569609-slack-C04TJ8P900J

```
PHP Fatal error:  Uncaught Error: Call to undefined method WP_Error::set__id() in /wp-content/plugins/activitypub/includes/model/class-follower.php:339
Stack trace:
#0 /wp-content/plugins/activitypub/includes/collection/class-followers.php(202): Activitypub\Model\Follower::init_from_cpt(Object(WP_Post))
```

## Proposed changes:

* Adds WP Error checks to all calls of `init_from_json()`.
* Adds test to cover the possibility of invalid json.

### Other information:

### DIFF
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -68,12 +68,6 @@ class Dispatcher {
 		$activity->set_object( \json_decode( $outbox_item->post_content, true ) );
 		$activity->set_actor( Actors::get_by_id( $outbox_item->post_author )->get_id() );
 
-		// Pre-fill the Activity with data (for example cc and to).
-		$activity_object = Activity::init_from_json( $outbox_item->post_content );
-		if ( ! is_wp_error( $activity_object ) ) {
-			$activity->set_object( $activity_object );
-		}
-
 		// Use simple Object (only ID-URI) for Like and Announce.
 		if ( in_array( $type, array( 'Like', 'Delete' ), true ) ) {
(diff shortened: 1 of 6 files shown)


jeherve (approved):
That fixes it, thank you!

includes/class-dispatcher.php
```diff
@@ -64,10 +64,14 @@ public static function process_outbox( $id ) {
...
 		$activity->set_type( $type );
 		$activity->set_id( $outbox_item->guid );
-		// Pre-fill the Activity with data (for example cc and to).
-		$activity->set_object( Activity::init_from_json( $outbox_item->post_content ) );
 		$activity->set_actor( Actors::get_by_id( $outbox_item->post_author )->get_id() );
 
+		// Pre-fill the Activity with data (for example cc and to).
+		$activity_object = Activity::init_from_json( $outbox_item->post_content );
+		if ( ! is_wp_error( $activity_object ) ) {
+			$activity->set_object( $activity_object );
```
pfefferle: Should't we fail here instead of working around it and nulling the object?

obenland: What would failing look like in this case? Publish the outbox item so it doesn't get retried and then return?

pfefferle (approved):

PR #1233, opened 2025-01-29
Outbox: Fix profile updates

pfefferle:
Better check casting to the correct Activity-Object for Actors

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1614,7 +1614,7 @@ function add_to_outbox( $data, $activity_type = 'Create', $user_id = 0, $content
 /**
  * Check if an object is an Activity.
  *
- * @param array|object $data The object to check.
+ * @param array|object|string $data The object to check.
  *
  * @see https://www.w3.org/ns/activitystreams#activities
  *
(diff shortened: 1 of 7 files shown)


includes/transformer/class-user.php
```diff
@@ -21,7 +21,15 @@ class User extends Base {
 	 * @return \Activitypub\Activity\Base_Object The ActivityPub Object
```
obenland: This will need an update for the new return type. Any downstream uses of `to_object()` that now need wp_error handling?

includes/functions.php
```diff
@@ -1614,7 +1614,7 @@ function add_to_outbox( $data, $activity_type = 'Create', $user_id = 0, $content
 /**
  * Check if an object is an Activity.
```
obenland: This, and the parameter comment should be updated as well

includes/functions.php
```diff
@@ -1659,6 +1659,51 @@ function is_activity( $data ) {
...
+}
+
+/**
+ * Check if an object is an Actor.
+ *
+ * @param array|object|string $data The object to check.
+ *
+ * @see https://www.w3.org/ns/activitystreams#actor
+ *
+ * @return boolean True if the object is an Activity, false otherwise.
```
```suggestion
 * @see https://www.w3.org/ns/activitystreams#actor
 *
 * @param array|object|string $data The thing to check.
 *
 * @return boolean True if the passed argument is an Actor, false otherwise.
```

includes/rest/class-outbox-controller.php
```diff
@@ -233,10 +233,14 @@ public function get_items( $request ) {
 	 * @return array Response object on success, or WP_Error object on failure.
 	 */
 	public function prepare_item_for_response( $item, $request ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
```
```suggestion
	public function prepare_item_for_response( $item, $request ) {
```
Looks like `$request` is now used

includes/transformer/class-base.php
```diff
@@ -271,6 +271,10 @@ public function get_replies() {
 	 * @return array The content map for the post.
 	 */
 	protected function get_content_map() {
+		if ( ! \method_exists( $this, 'get_content' ) || ! $this->get_content() ) {
+			return null;
```
obenland, line 275: Return type for this method is array, should we return an empty array here instead? Or update the docs?

pfefferle: Update the docs. With an empty array, it will be added to the output, with `null`, it will be ignored.

includes/rest/class-outbox-controller.php
```diff
@@ -232,11 +232,15 @@ public function get_items( $request ) {
...
-		$activity    = $transformer->to_activity( $type );
+	public function prepare_item_for_response( $item, $request ) {
+		$type = \get_post_meta( $item->ID, '_activitypub_activity_type', true );
+
+		$activity = new Activity();
+		$activity->set_type( $type );
 		$activity->set_id( $item->guid );
+		// Pre-fill the Activity with data (for example cc and to).
+		$activity->set_object( \json_decode( $item->post_content, true ) );
+		$activity->set_actor( Actors::get_by_various( $request->get_param( 'user_id' ) )->get_id() );
```
obenland, line 243: Looks like we don't need set_actor? It doesn't change the content of the actor property.
Is there a scenario where we grab posts that are not by the passed user and need to update the actor to be the passed user?

pfefferle: On `Announces` the Activity-Actor is the Blog-Actor and the Object is attributed to the Author.

obenland (approved):

PR #1234, opened 2025-01-29
Fix: Importing tags from the Fediverse

### LINKED ISSUE #1167: WP blog picks up tags from around the fediverse
### Quick summary

Approximately two weeks after publishing a single post on my new Wordpress blog, the list of tags contains multiple entries that I did not use in the post. I recognize some tags from the description of my own Mastodon accounts that follow the blog, but others are completely random (and some of them all kinds of racist/NSFW/...). The single post has not been interacted that much, so I'm not sure if all of them are from other fediverse accounts known to the plugin, or if there's some other mechanism at play here.

I feel that this should be considered a bug rather than intended behaviour, because it leads to a very polluted tag list. Also, it changes the resulting page when visiting, say, `example.blog/tag/cannabis` or `example.blog/tag/some_racial_slur` from the standard 404 page to a "nothing was found" search result page.

### Steps to reproduce

Assuming that nothing is special about my WP/plugin installation, just visit the tags admin page at `/wp-admin/edit-tags.php?taxonomy=post_tag` and compare the tags listed there to your blog's actual content.

Plugin version is currently 4.7.1.

### Site owner impact

### Severity

Moderate

No revenue impact

_No response_

pfefferle:
This PR adds a check, so that only supported posts will be parsed for tags. This will prevent, that followers and other custom post types will be parsed for Hashtags.

Fixes #1167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Handle deletes from remote servers that leave behind an accessible Tombstone object.
+* No longer parses tags for post types that don't support Activitypub.
 
 ## [4.7.3] - 2025-01-21
 
--- a/includes/class-hashtag.php
+++ b/includes/class-hashtag.php
@@ -53,6 +53,11 @@ class Hashtag {
 	 * @param \WP_Post $post    Post object.
 	 */
 	public static function insert_post( $post_id, $post ) {
+		// Check if the post supports ActivityPub.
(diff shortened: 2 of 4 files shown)


obenland:
Can you help me understand how follower's tags get into a post's tags list?

pfefferle:
@obenland we parse the post content/excerpt for hashtags and add them as WordPress tags and because followers will be saved as a custom post type, their content/excerpt will be also parsed. Guess where we store the profile description!?! ;)

obenland (dismissed):

obenland (approved):

PR #1235, opened 2025-01-29
WIP: Allow federating arbitrary data

pfefferle:
This PR uses the data you pass to the Activity via `set_object` to keep all the custom attributes.


jeherve (dismissed):
It worked in my tests, thank you!

PR #1239, opened 2025-01-31
Change usage of fediverse symbol

pfefferle:
Move "⁂" to the end of a line, otherwise the links look a bit disorganized


obenland:
Is it worth invalidating all translations for these strings for that change? I'm good with it if the answer is Yes.

obenland (approved):

PR #1240, opened 2025-02-01
Stream: Only surface errors in Outbox processing

obenland:
Also adds support for comment and user types.

## Proposed changes:

* Adds support for user types.
* Only shows Inbox results for errors.

## Testing instructions:

* Make a change to your user profile.
* Add a comment reply to a federated comment.
* If developing with `wp-env` run `npm run env -- run cli wp cron event run --all ` to process cron.
* Double-check the information in Stream.


pfefferle:
This still fails if you have more then around 4 or 5 errors in one federation process.

pfefferle (changes requested):
This still fails if you have more then around 4 or 5 errors in one federation process. I think we should one dedicated entry for every error.

pfefferle (approved):

PR #1242, opened 2025-02-02
Fix/Change some return types in PHPDocs

Menrath:
* Make replies (collection) nullable: So if we support no comments at all, one can just not send this field.
* Fix return type for ActivityPub object ID

props to PHPStan :)


pfefferle (dismissed):

includes/transformer/class-base.php
```diff
@@ -259,7 +259,7 @@ public function get_media_type() {
 	/**
 	 * Get the replies Collection.
 	 *
-	 * @return array The replies collection.
+	 * @return ?array The replies collection.
```
obenland: Looks like `Replies::get_collection()` returns an array or null, let's reflect that here as well. Would you mind updating it for `Replies::get_collection()` as well?

Menrath: @obenland Completed.

obenland: Apologies, I should have been more clear. Since we know it’ll be either an array or null, let’s just say that. So `array|null` rather than `?array`.

includes/collection/class-replies.php
```diff
@@ -83,7 +83,7 @@ private static function get_id( $wp_object ) {
 	 *
 	 * @param WP_Post|WP_Comment $wp_object The post or comment to fetch replies for.
 	 *
-	 * @return array An associative array containing the replies collection without JSON-LD context.
+	 * @return ?array An associative array containing the replies collection without JSON-LD context.
```
```suggestion
	 * @return array|null An associative array containing the replies collection without JSON-LD context.
```

includes/transformer/class-base.php
```diff
@@ -260,7 +260,7 @@ public function get_media_type() {
...
```
```suggestion
	 * @return array|null The replies collection on success or null on failure.
```

obenland (approved):

PR #1243, opened 2025-02-02
Fix negotiation of ActivityPub requests for custom post types

Menrath:
Please try to reproduce this first. It seems the WordPress function `\get_queried_object() `used in the `\ActivityPub\Query->get_queried_object()` does not work with custom post types, when pretty permalinks are used and the query URL is `?p=<post_id>`.


Menrath:
This is dangerous and can not be merged as is. I will explain later.

Menrath:
I think now that https://github.com/Automattic/wordpress-activitypub/pull/1246 is merged this should no longer be risky, because whether the post later will be transformed and therefore shown is handled by the transformer factory anyway.

pfefferle (approved):

PR #1256, opened 2025-02-03
Settings: Make tabs filterable

### LINKED ISSUE #1250: Make Settings tabs filterable
### What

Let's make it easier for third-party plugins to add their own tabs to the ActivityPub settings page.
As a separate task, we should also move the hard-coded settings over to using the Settings API, so it's easier for third-party plugins to add their own settings sections and settings fields.

### Why

So it's easier for third-party plugins to integrate with the ActivityPub plugin.

### How

_No response_

obenland:
Fixes #1250.

## Proposed changes:

* Adds a filter to add custom settings tabs.
* Updates template loader and tabs template to account for custom tabs.
* Adds developer docs.

## Testing instructions:

* Go to the settings page under Settings > ActivityPub and make sure it still works as expected.
* Switching between tabs should show the correct active status.
* Try extending the tabs and see that it works.

### DIFF
--- a/templates/admin-header.php
+++ b/templates/admin-header.php
@@ -5,16 +5,10 @@
  * @package Activitypub
  */
 
+global $settings_tabs;
+
 /* @var array $args Template arguments. */
-$args = wp_parse_args(
-	$args,
-	array(
-		'welcome'      => '',
-		'settings'     => '',
-		'blog-profile' => '',
-		'followers'    => '',
-	)
-);
+$args = wp_parse_args( $args, array_fill_keys( array_keys( $settings_tabs ), '' ) );
 ?>
 <div class="activitypub-settings-header">
 	<div class="activitypub-settings-title-section">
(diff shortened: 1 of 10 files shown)


obenland:
Needs docs with examples

templates/admin-header.php
```diff
@@ -5,42 +5,32 @@
  * @package Activitypub
  */
 
+global $settings_tabs;
```
pfefferle: why do we have to make it global? and if so, shouldn't we prefix it, to avoid collisions?

obenland: Fixed in [8b74a14](https://github.com/Automattic/wordpress-activitypub/pull/1256/commits/8b74a14babdf7d501a211fb44998efb5393671a5)

pfefferle (approved):
nice one!!

Menrath:
Nice!

Menrath:
@obenland @pfefferle Looks great :)
image

PR #1258, opened 2025-02-03
Add support for a reply "context"

pfefferle:
This PR adds a reply `context` to WordPress Posts and Comments.

The idea of the context is, that relying parties will be able to "import" the whole conversation, even if they will not be pinged in some of the cases. The context provides a list of all ids of every reply, including the main topic.

Some context:

* https://codeberg.org/fediverse/fep/src/branch/main/fep/7888/fep-7888.md
* https://spectra.video/w/xwCSYfZh1mJY64zJ9GngbE

## Proposed changes:

* Added a context API endpoint
* The endpoint lists all comments ordered by the publish date

## Testing instructions:

* Checkout the branch
* Load a post with comments with the ActivityPub `Accept` header
* Check for the `context` attribute
* Load the referenced endpoint
* Check if all comments will be listed

/cc @julianlam

props @julianlam (thanks for taking your time to test that with me)


julianlam:
NodeBB checks for context when receiving a post object, but if WordPress also supports pulling in an AP object by URL then I recommend you also support if someone pastes a NodeBB topic URL in.

Otherwise, nice work! Can't wait to see it live.

pfefferle:
@julianlam Let me know when you are back from vacation. Would love to try it with nodebb before I release it ☺️

pfefferle:
@Automattic/fediverse I justed tested it with @julianlam and it imported the post and all comments nicely into nodeBB!

So this should be ready to review!

CHANGELOG.md
```diff
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* A reply `context` for Posts and Comments to allow relying parties to discover the whole conversation of a thread.
 * Setting to adjust the number of days Outbox items are kept before being purged.
+* Undo API for Outbox items.
```
obenland, line 14: Is this a rebase remnant?

includes/collection/class-replies.php
```diff
@@ -141,20 +144,56 @@ public static function get_collection_page( $wp_object, $page, $part_of = null )
...
+
 	/**
 	 * Get the ActivityPub ID's from a list of comments.
 	 *
 	 * It takes only federated/non-local comments into account, others also do not have an
 	 * ActivityPub ID available.
 	 *
-	 * @param WP_Comment[] $comments The comments to retrieve the ActivityPub ids from.
+	 * @param WP_Comment[] $comments              The comments to retrieve the ActivityPub ids from.
+	 * @param boolean      $include_blog_comments Include blog comments in the returned array.
```
```suggestion
	 * @param boolean      $include_blog_comments Optional. Include blog comments in the returned array. Default false.
```

obenland (commented):
Nice, found only minor stuff

CHANGELOG.md
pfefferle: I would say yes! But is that wrong?

pfefferle: ah, missed to remove it from fixed!

pfefferle: This is wrong in trunk!

obenland (approved):

julianlam:
:clap: :rocket:

PR #1262, opened 2025-02-04
Process Follower Inboxes in Batches

### LINKED ISSUE #564: Feature Request: Improve scheduling?
### What

I was simply wondering; if you had a quite a few followers on different servers, that means a lot of POST requests are made for every new post.

These are made in a single WP-Cron call:

https://github.com/Automattic/wordpress-activitypub/blob/efd98acd0bf3f2c42f0ffc3d8f10920e74cb8512/includes/class-activity-dispatcher.php#L82-L84

Now I recently learned WP-Cron runs asynchronously even when WP_DISABLE_CRON is not set, and is normally triggered from a cron job when it _is_ set.

 The latter hopefully also means a longer (or even infinite) time-out limit.

Still, I wonder if it would be possible to either "spread these out" in time a bit more, make them ... less demanding.

### Why

I imagine it can be quite taxing  on smaller servers to send out a few dozen HTTP requests.

Note that I've only ever experienced real issues with _incoming_ requests, after moving to an unmanaged VPS at least. (My previous (shared) web host seemed to have trouble even with outgoing requests.)

But I don't have very many followers either. :-P

### How

It might be possible to either "slice" the `$inboxes` array and schedule only, e.g., 5 requests at a time, moving the remaining requests forward by a minute or so until they've all run. This might require keeping track of which inboxes were already "processed"---which would be a nice feature anyway, to allow for retries? Or the processed ones could simply be dropped from the array and the remainder could be passed onto the callback as an additional argument. Or something like that. :-D

Or allowing a true "job queue" or something to process them. (I guess a custom plugin could already replace the `activitypub_send_activity` callback with something "more efficient," or rather, less demanding. Based on Redis or whatever.)

Note that this likely creates race conditions too, i.e., the callback, when it eventually runs, should again verify that the post wasn't deleted or made "private," etc., in the meantime.

But: such an additional verification step _might_ at one point anyhow become a necessity. In fact, the plugin seems to already be running one such check: https://github.com/Automattic/wordpress-activitypub/blob/efd98acd0bf3f2c42f0ffc3d8f10920e74cb8512/includes/class-activity-dispatcher.php#L64C8-L64C49. (I'd say even more such checks are in order. But that's a different matter.)

Come to think of it, if I had WP_DISABLE_CRON set right now and only ran cron once every 10 minutes or so, I'd have the same potential issue (that the post's visibility changed).

obenland:
Fixes #564

## Proposed changes:

* Separated Follower inboxes and processes them in batches.

### Other information:

## Testing instructions:

* Enable WP_DEBUG and WP_DEBUG_LOG.
* Install and activate the Stream plugin.
* Create or update a post.
* Observe in Stream that the outbox item was processed.
* Check in error logs for errors.

### DIFF
--- a/includes/class-debug.php
+++ b/includes/class-debug.php
@@ -64,9 +64,6 @@ class Debug {
 		if ( \is_wp_error( $result ) ) {
 			// phpcs:ignore WordPress.PHP.DevelopmentFunctions
 			\error_log( "[DISPATCHER] Failed Request to: {$inbox} with Result: " . \print_r( $result, true ) );
-		} else {
-			// phpcs:ignore WordPress.PHP.DevelopmentFunctions
-			\error_log( "[DISPATCHER] Success Request to: {$inbox} with Result: " . \print_r( $result, true ) );
 		}
 	}
 
--- a/includes/class-debug.php
+++ b/includes/class-debug.php
@@ -7,9 +7,6 @@
 
 namespace Activitypub;
 
-use WP_DEBUG;
-use WP_DEBUG_LOG;
-
 /**
  * ActivityPub Debug Class.
  *
@@ -20,9 +17,11 @@ class Debug {
 	 * Initialize the class, registering WordPress hooks.
 	 */
 	public static function init() {
-		if ( WP_DEBUG_LOG ) {
+		if ( \WP_DEBUG && \WP_DEBUG_LOG ) {
 			\add_action( 'activitypub_safe_remote_post_response', array( self::class, 'log_remote_post_responses' ), 10, 2 );
 			\add_action( 'activitypub_inbox', array( self::class, 'log_inbox' ), 10, 3 );
+
+			\add_action( 'activitypub_sent_to_inbox', array( self::class, 'log_sent_to_inbox' ), 10, 2 );
 		}
 	}
 
@@ -52,6 +51,22 @@ class Debug {
 			$url   = object_to_uri( $actor );
 			// phpcs:ignore WordPress.PHP.DevelopmentFunctions
 			\error_log( "[INBOX] Request From: {$url} with Activity: " . \print_r( $data, true ) );
+		}
+	}
+
+	/**
+	 * Log the sent to follower action.
+	 *
+	 * @param array  $result The result of the remote post request.
+	 * @param string $inbox  The inbox URL.
+	 */
+	public static function log_sent_to_inbox( $result, $inbox ) {
+		if ( \is_wp_error( $result ) ) {
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions
+			\error_log( "[DISPATCHER] Failed Request to: {$inbox} with Result: " . \print_r( $result, true ) );
+		} else {
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions
+			\error_log( "[DISPATCHER] Success Request to: {$inbox} with Result: " . \print_r( $result, true ) );
 		}
 	}
 
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -44,11 +44,19 @@ class Dispatcher {
 		\add_filter( 'activitypub_interactees_inboxes', array( self::class, 'add_inboxes_by_mentioned_actors' ), 10, 3 );
 		\add_filter( 'activitypub_interactees_inboxes', array( self::class, 'add_inboxes_of_replied_urls' ), 10, 3 );
 
-		// Fallback to followers if no other inboxes are found.
-		// @deprecated Use `activitypub_interactees_inboxes` instead.
+		// Fallback for `activitypub_send_to_inboxes` filter.
 		\add_filter(
 			'activitypub_send_to_inboxes',
 			function ( $inboxes, $actor_id, $activity ) {
+				/**
+				 * Filters the list of interactees inboxes to send the Activity to.
+				 *
+				 * @param array    $inboxes  The list of inboxes to send to.
+				 * @param int      $actor_id The actor ID.
+				 * @param Activity $activity The ActivityPub Activity.
+				 *
+				 * @deprecated Unreleased Use `activitypub_interactees_inboxes` instead.
+				 */
 				return \apply_filters_deprecated( 'activitypub_send_to_inboxes', array( $inboxes, $actor_id, $activity ), 'Unreleased', 'activitypub_interactees_inboxes' );
 			},
 			10,
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -91,7 +91,19 @@ class Dispatcher {
 		}
 
 		self::send_activity_to_followers( $activity, $actor_id, $outbox_item );
-		self::async_batch_processor( $activity, $actor_id, $outbox_item );
+
+		// TODO: Should we call Scheduler::async_batch directly here?
+		\wp_schedule_single_event(
+			\time(),
+			'activitypub_async_batch',
+			array(
+				array( self::class, 'send_to_followers' ),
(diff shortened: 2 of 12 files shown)


includes/class-dispatcher.php
```diff
@@ -73,7 +79,161 @@ public static function process_outbox( $id ) {
...
+			$result = safe_remote_post( $inbox, $json, $actor_id );
+
+			// @TODO Handle errors more elegantly.
+			if ( wp_remote_retrieve_response_code( $result ) >= 400 ) {
+				$attempt = get_post_meta( $follower_id, '_activitypub_outbox_attempt_' . $outbox_item->ID, true );
+				$attempt = $attempt ? $attempt + 1 : 1;
+
+				if ( $attempt <= 3 ) {
+					// Log attempt and move on.
+					update_post_meta( $follower_id, '_activitypub_outbox_attempt_' . $outbox_item->ID, $attempt );
```
pfefferle: is that a good idea? in the worst case this would add tens of thousands of metas.

obenland: But only until they're processed. We should remove them once that happened. Even if "finished processing" means we tried 3 times unsuccessfully.

pfefferle: Can't that kill shared hosting dbs? 1000++ db inserts in a few milli seconds?

obenland: It's batches of 50 currently and this only runs on failed posts. How did you get to thousands?

pfefferle: oh, I may have commented on the wrong line... but some lines later, the code also stores the processed ones!?!

obenland: Correct, [they get deleted](https://github.com/Automattic/wordpress-activitypub/pull/1262/files#diff-f53a4c6f8c1e61852fbbdd1132034377c601fdabc3bebb9dd2ac9c834c41a9ddR180) once an outbox item finished processing.

obenland:
@pfefferle This should be in a test-able state now, hooked into the Outbox processing

includes/class-scheduler.php
```diff
@@ -184,4 +185,96 @@ public static function reprocess_outbox() {
...
+		self::lock( $key );
+
+		$callback = array_shift( $args ); // Remove $callback from arguments.
+		$next     = \call_user_func_array( $callback, $args );
+
+		self::unlock( $key );
+
+		if ( ! empty( $next ) ) {
+			// Schedule the next run, adding the result to the arguments.
+			\wp_schedule_single_event(
```
pfefferle, line 233: shouldn't we always add the next schedule before we start the federation? otherwise the batch processing might break if one of the batches does not work properly?

includes/class-dispatcher.php
```diff
@@ -73,7 +79,55 @@ public static function process_outbox( $id ) {
...
+	public static function send_to_followers( $activity, $actor_id, $outbox_item, $batch_size = 50, $offset = 0 ) {
+		$inboxes = Followers::get_inboxes_for_activity( $actor_id, $activity, $batch_size, $offset );
+		$json    = $activity->to_json();
+
+		foreach ( $inboxes as $inbox ) {
+			$result = safe_remote_post( $inbox, $json, $actor_id );
+
+			// @TODO Handle errors more elegantly.
+			if ( wp_remote_retrieve_response_code( $result ) >= 400 ) {
+				\error_log( 'Error sending to follower: ' . $inbox ); // phpcs:ignore
```
pfefferle: no more storage of errors?

obenland: See `// @TODO Handle errors more elegantly.` in line 119 :) 

But frankly, I'm not sure how we'll do that for Stream. Any ideas?

includes/class-scheduler.php
obenland: We could keep track of when the first batch gets started and clear it after the last batch ran, then have `reprocess_outbox` looks for those and pick them back up?

includes/class-dispatcher.php
pfefferle: should we simply call an action? this way we can use that for error processing.

* error log if WP_Debug
* stream if installed
* post-meta
* ...

obenland: Sure! For each call individually or per batch?

pfefferle: for each error?

includes/collection/class-followers.php
```diff
@@ -326,6 +327,93 @@ public static function get_inboxes( $user_id ) {
...
+				),
+				array(
+					'key'     => '_activitypub_inbox',
+					'value'   => '',
+					'compare' => '!=',
+				),
+			),
+		);
+
+		if ( self::maybe_add_inboxes_of_blog_user( $json, $actor_id ) ) {
```
pfefferle, line 362: the `maybe_add_inboxes_of_blog_user` does no longer return anything useful

includes/class-dispatcher.php
```diff
@@ -19,17 +19,30 @@
  * @see https://www.w3.org/TR/activitypub/
  */
 class Dispatcher {
+
+	/**
+	 * Batch size.
+	 *
+	 * @var int
+	 */
+	public static $batch_size = 50;
```
pfefferle: I would start with 100.

```suggestion
	public static $batch_size = 100;
```

Should we use a const for it, so that it is easier to replace?

includes/class-dispatcher.php
```diff
@@ -73,7 +86,86 @@ public static function process_outbox( $id ) {
 			$activity->set_object( $activity->get_object()->get_id() );
 		}
 
-		self::send_activity_to_followers( $activity, $actor_id, $outbox_item );
+		if ( ! self::should_send_to_followers( $activity, $actor_id, $outbox_item ) ) {
+			return;
+		}
+
+		// Send to mentioned and replied-to users. Everyone other than followers.
+		self::send_to_interactees( $activity, $actor_id, $outbox_item );
```
pfefferle: `send_to_interactees` should be called even if `should_send_to_followers` is false, because even private messages should be sent to the mentioned users.

obenland: Nice catch! Fixed in [9fe47dd](https://github.com/Automattic/wordpress-activitypub/pull/1262/commits/9fe47dd29a93f895fe380a7be05ec60f3c4351e9).

includes/class-dispatcher.php
obenland: Fixed in [350f4ab](https://github.com/Automattic/wordpress-activitypub/pull/1262/commits/350f4abae80f42207e056a9a370db8c65c39bb5b)

includes/class-debug.php
```diff
@@ -55,6 +54,22 @@ public static function log_inbox( $data, $user_id, $type ) {
...
+	 * @param array  $result The result of the remote post request.
+	 * @param string $inbox  The inbox URL.
+	 */
+	public static function log_sent_to_inbox( $result, $inbox ) {
+		if ( \is_wp_error( $result ) ) {
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions
+			\error_log( "[DISPATCHER] Failed Request to: {$inbox} with Result: " . \print_r( $result, true ) );
+		} else {
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions
+			\error_log( "[DISPATCHER] Success Request to: {$inbox} with Result: " . \print_r( $result, true ) );
```
obenland: @pfefferle Is success of interest?

pfefferle: spam the logs!!!

pfefferle: 🙌

pfefferle (approved):

pfefferle:
@obenland good job!

PR #1266, opened 2025-02-04
Update Followers endpoint to extend WP_REST_Controller

obenland:
Follow-up to #1146.

## Proposed changes:

* Updates Followers endpoint to extend `WP_REST_Controller`.
* Adds unit tests and schema information for Followers endpoint.

### Other information:


includes/rest/class-followers-controller.php
```diff
@@ -0,0 +1,290 @@
...
+		if ( $page && ( ( \ceil( $response['totalItems'] / $per_page ) ) > $page ) ) {
+			$response['next'] = \add_query_arg( 'page', $page + 1, $response['partOf'] );
+		}
+
+		if ( $page && ( $page > 1 ) ) {
+			$response['prev'] = \add_query_arg( 'page', $page - 1, $response['partOf'] );
+		}
+
+		$response['first'] = \add_query_arg( 'page', 1, $response['partOf'] );
+		$response['last']  = \add_query_arg( 'page', \ceil( $response['totalItems'] / $per_page ), $response['partOf'] );
```
pfefferle: by default (no followers) it shows `0` and loading that page throws an error then. either this should link to min `1` or only show up if $data has more than one page.

pfefferle (changes requested):

pfefferle:
Now the pages `2` and above will show no errors. Is that fine?

pfefferle (dismissed):

obenland:
Could you describe the scenario where you'd expect an error? I'm not quite following

pfefferle:
oh sorry! if you have only one page and change the page number in the URL to a way higher page. should we show an error then or simply show an empty items array?

obenland:
Oh! Core's Posts controller accounts for that, I opted not to bother in the Outbox controller, but happy to reconsider!

pfefferle:
I am fine with both... just wanted to bring up the question...

pfefferle (approved):

Issue #1269, opened 2025-02-04
Race conditions e.g. on post creation: Post not "ready" when added to outbox.


Menrath:
### Quick summary

Sometimes there is a race condition: 

* Post creation/update
* The Post Scheduler gets called immediately
* In the Scheduler `add_to_outbox` gets called immediately 
* In `add_to_outbox` the transformer gets called directly: **sometimes even before the object that should get transformed is fully ready (e.g. post-meta is not set yet)** 

### Steps to reproduce

Use the plugin "The Events Calendar". Create a new event post with a venue. The post scheduler gets called which calls `add_to_outbox`. At this stage the venue is not yet linked to the event post: `\tribe_has_venue( $post_id )` returns false. As are most or all other meta values.

### If a workaround is available, please outline it here.

Use scheduling within the `add_to_outbox` function.

obenland:
We had to [add a callback to make post meta available earlier](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/scheduler/class-post.php#L162-L195) for posts that get saved through the Rest API. Is that  what you're running into here? Or is it something else?

Menrath:
@obenland This could indeed by the same problem. But maybe there are further issues, e.g. when on the request where a post is created also other related posts or terms are created (venue post/term of an event post that is created on the fly).

obenland:
For related posts, my assumption would be that they get added to the outbox through existing hooks. Is terms support something you're working on? Would a callback like the one we used help in that case?

Menrath:
Not all post types are also scheduled. 

But lets take the **"The Events Calendar" plugin as an example**, it is easier to understand that way and that plugin is highly popular:

* Venues are stored in a custom post type (not-public). Events are too, of course (public).
* Venues can be created independently but I'd say most of the time they are created in the same fly on the event edit page.
* Venues are transformed to `as:Place` within the `as:location` of an `as:Event`. The venue information is also used for the `as:Event`s `as:summary`.
* Currently in the wild there is currently (might change) need to send/schedule venues independently. Their current purpose is federated caching, and to be able to reliably map them.
* With the current status quo they can be queried by ID/URL though, because the custom transformer that is applied and has **it's own** `is_post_disabled` check.

However, regarding "The Events Calendar" the problem could be reduced to the post_meta race condition because it seems that `wp_insert_post` for the venue post is getting called before the post gets inserted/updated. I have to double check that again.

**Other example:**

[...]

PR #1270, opened 2025-02-05
Fix race-condition on post creation/adding to outbox

Menrath:
This is a draft for a Hotfix for #1269.

Is there any better solution? This PR would change the behaviour a lot and many unit tests would have to be adopted.


pfefferle:
I think we fixed this, can you verify it with the event integration @Menrath ?

Issue #1263, opened 2025-02-04
Feature Request: True Unique ID for transformed WP_Term


Menrath:
### What

Currently when writing a third party transformer for WP_Term objects no true unique ID can be used that can be resolved.

### Why

E.g. Transformer for venues of event plugins stored in terms, which are transformer to `as:Place`.

See, the current best approach I could find (note that the PHPdoc is faulty...):

https://codeberg.org/Event-Federation/wordpress-event-bridge-for-activitypub/src/commit/ba329f34a0de952d0b9abcaeef1450f7352f028f/includes/activitypub/transformer/place/class-base-term-place.php#L57

It is not that severe, as an ID for a place is not that crucial (at least at the moment) and the slug of a term changes rarely, I would say!

### How

Make `\ActivityPub\Query->get_queried_object` even more flexible, though I do not like seeing it getting bloated more and more :/

pfefferle:
@Menrath I am not really sure what the actual request is? that the plugin provides unique URLs for terms?

Menrath:
It is not a direct request, It is more like a question if the plugin should provide something.

The question is, should there by unique IDs for WP_Term objects by which the ActivityPub object can by queried with the ActivityPub Accept headers?

If the WP core function `get_queried_object()` can not identify the Term (it normally only can do so by the terms slug) the `ActivityPub\Query` will not find it. @pfefferle Remember I was proposing an filter for the queried object, this is related. Or maybe a plugin developer should write global redirects of some way?

pfefferle:
🤔

this is a hard and very theoretical question. I am not sure if the arbitrary data, the event plugins provide, have to be queryable through conneg.

I just saw that I have an issue in the current code, because what I wanted the transformer to be always called, even with an empty queried object: https://github.com/Automattic/wordpress-activitypub/blob/84c6cbf9a6401df792c0cddada35be8299e1c032/includes/class-query.php#L95

This way you would not need a filter, because the transformer would be called always!?!

pfefferle:
Nonetheless, I am happy to add support for terms. Do you know the ugly url WordPress is using for them, if you have disabled pretty permalinks?

pfefferle:
@Menrath does this help: https://github.com/Automattic/wordpress-activitypub/pull/1272 ???

Menrath:
> Do you know the ugly url WordPress is using for them, if you have disabled pretty permalinks?

[...]

PR #1272, opened 2025-02-05
ActivityPub Query: Improve third party plugin support

pfefferle:
* Add filter to change the queried object
* Always call transformer

See #1263


Menrath:
Either this or we could also introduce a query like `?t=<term_id>` for ActivityPub requests. This would kind of unify behaviour for third party plugins and maybe reduce the possibility of errors. 


But the approach here is indeed useful for some cases: I have already real life examples in mind:

Some event plugins have "events" and "event occurances". Usually the occurances are stored in a custom table, because the posts table is just not adequate for such special cases. This is also what "The Events Calendar" and "GatherPress" do.

With that filter one could easily set an "occurance" as the targeted query, which would not be possible (in an easy and straight-forward way) without the filter I would guess.

**Edit:** Maybe that example is not good, because the custom transformer might be able to negotiate that, if the related WP_Post object gets "identified" as the queried object by WordPress.

obenland:
What would be a good way for me to test this change? Go to an author archive page?

pfefferle:
this handles the content negotiation for authors, posts, comments and all other post-types. so you can check this by calling one or all of these sites with the `application/json+activity` Accept header.

pfefferle:
Ah and adding the query bar was only to get sure that we have access to it!

obenland (approved):

PR #1273, opened 2025-02-05
Add: Show ActivityPub preview in row actions, when block editor enabled but is not used for post type.

Menrath:
Another small PR containing a hopefully small improvement from me these days.


pfefferle (approved):
@Menrath come on! you are the embodiment of improvement! :)

PR #1278, opened 2025-02-06
Add/sent to inboxes actions

pfefferle:
Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-debug.php
+++ b/includes/class-debug.php
@@ -7,9 +7,6 @@
 
 namespace Activitypub;
 
-use WP_DEBUG;
-use WP_DEBUG_LOG;
-
 /**
  * ActivityPub Debug Class.
  *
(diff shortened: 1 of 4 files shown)


includes/class-debug.php
```diff
@@ -55,6 +54,24 @@ public static function log_inbox( $data, $user_id, $type ) {
...
+	/**
+	 * Log the sent to follower action.
+	 *
+	 * @param array  $result The result of the remote post request.
+	 * @param string $inbox  The inbox URL.
+	 * @param int    $actor_id The ID of the actor.
+	 * @param array  $activity The activity.
+	 */
+	public static function log_sent_to_follower( $result, $inbox, $actor_id, $activity ) {
+		if ( \is_wp_error( $result ) ) {
```
obenland, line 64: `$result` will be a `WP_REST_Response` object, not a `WP_Error`. It does have[ a `as_error()` method](https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/rest-api/class-wp-rest-response.php#L225-L246) though, that helps with error handling.
So you could convert it first and then check it with `is_wp_error()`

pfefferle: this is the custom function and it returns an error 🫣

https://github.com/Automattic/wordpress-activitypub/blob/8c9f6a842e02d89e573073bedd2bf3aaae9f9f20/includes/class-http.php#L68

obenland: Oh! We could consider updating that to return the actual error that's being returned

includes/class-debug.php
```diff
@@ -20,9 +17,11 @@ class Debug {
 	 * Initialize the class, registering WordPress hooks.
 	 */
 	public static function init() {
-		if ( WP_DEBUG_LOG ) {
+		if ( \WP_DEBUG && \WP_DEBUG_LOG ) {
```
obenland, line 20: I'd keep it at only checking for WP_DEBUG_LOG. WP_DEBUG, to me, generally suggests that there'll be front-end facing debug information. Logs could be collected without that

PR #1279, opened 2025-02-06
Add support for deprecated version replacements

obenland:
In preparations for our next release that contains a deprecated method.


pfefferle:
Do we have a chance to add JS tests?

obenland:
Do we have to? 🫣
With this being an internal tool I wasn't planning on it

pfefferle:
it is more a general question, because we currently have no infrastructure for JS tests and we do not test the blocks yet.

pfefferle (approved):

PR #1281, opened 2025-02-06
Outbox: Allow querying of outbox posts

obenland:
Keeps UI and front-end querying disabled, but allows REST API requests for the post type.

Useful for debugging.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to wp-admin and make sure the admin UI for Outbox items doesn't show up.
* Query the rest api with `/wp-json/wp/v2/ap_outbox` and make sure it works and has author and meta information.
* Pick a `guid` and try to access it from the front-end (it should fail)


obenland:
Should we do that with all our custom post types?

pfefferle (approved):

PR #1283, opened 2025-02-06
Reply Block: add broader text

kraftbj:
When using this block, I paused to try to think through if this was restricted to only Mastodon APIs or any federated post.

Including both Fediverse and Mastodon that should help ease that wonder in others.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Add a federated reply block.
* See the new text under the URL field.


obenland (approved):

obenland:
Thank you!

PR #1284, opened 2025-02-07
Improve the wording of the Reply-Block

pfefferle:
The text change that comes with #1283 can be misinterpreted, that the Fediverse and Mastodon are the same. This change should make it clear, that Mastodon is only an example.

Aside from that, it regenerates the JS.


obenland (approved):

kraftbj (commented):
Belated reply from me, but good change. Thanks!

PR #1285, opened 2025-02-07
Outbox Batch: Only pass outbox id to jobs

obenland:
## Proposed changes:

* Moves author and activity creation into separate methods
* Updates inbox processor to generate the activity based on the passed oubtox item id
* Updates `process_outbox()` to reuse the new utility methods.

### Other information:

### DIFF
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -10,6 +10,9 @@ namespace Activitypub;
 use Activitypub\Activity\Activity;
 use Activitypub\Collection\Actors;
 use Activitypub\Collection\Followers;
+use Activitypub\Model\Application;
+use Activitypub\Model\Blog;
+use Activitypub\Model\User;
 
 /**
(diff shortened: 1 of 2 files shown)


includes/class-dispatcher.php
```diff
@@ -10,6 +10,9 @@
 use Activitypub\Activity\Activity;
 use Activitypub\Collection\Actors;
 use Activitypub\Collection\Followers;
+use Activitypub\Model\Application;
```
pfefferle: Do we really need them? Is this only for the PHPDoc?

obenland: yeah its not great :/

pfefferle:
should we change the static class to use an object instead?

obenland:
> should we change the static class to use an object instead?

No strong opinions here. What's the benefit?

pfefferle:
If you see no advantage, then probably not ;)

pfefferle (approved):

PR #1286, opened 2025-02-07
Outbox: Improve batch handling

pfefferle:
After a new item is added to the Outbox, every pending item of the same type will be "invalidated". This way, we ensure that older `Update`s won't be published if there is already a newer one.

Aside from that, this PR allows us to filter by ID and to better debug by title.


includes/collection/class-outbox.php
```diff
@@ -72,6 +79,52 @@ public static function add( $activity_object, $activity_type, $user_id, $content
...
+				'post_status' => 'pending',
+				'exclude'     => array( $current_id ),
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+				'meta_query'  => $meta_query,
+				'fields'      => 'ids',
+			)
+		);
+
+		if ( $existing_items ) {
+			foreach ( $existing_items as $existing_item_id ) {
```
obenland: I think `get_posts()` always returns an array, even if it's empty. If that's the case the if check shouldn't be necessary.

obenland: At this point we should probably also look for any in-progress jobs and remove them, and also delete the outbox_offset post meta.

pfefferle: how can I kill in-progress jobs?

obenland: I guess we can't kill in-progress batches, but if this happens to fall in-between two batches we can remove the scheduled job. [This](https://github.com/Automattic/wordpress-activitypub/blob/trunk/tests/includes/class-test-dispatcher.php#L97-L106) is how we can check if there's a timestamp and then pass it to `wp_unschedule_event()`.

pfefferle: Do we need the actor ID in the schedule? this is part of the outbox item already!

obenland: We don't, the only reason I kept it in was that we already had it in `process_outbox` and I didn't need to get it again with `self::get_author()`. Feel free to change that

pfefferle: 😩

includes/class-dispatcher.php
```diff
@@ -104,26 +103,27 @@ public static function process_outbox( $id ) {
 		} else {
 			// No followers to process for this update. We're done.
 			\wp_publish_post( $outbox_item );
+			\delete_post_meta( $outbox_item->ID, '_activitypub_outbox_offset' );
```
obenland, line 106: This shouldn't be needed since we don't process followers when we get here.

pfefferle: I thought it might be nice to clean up!?

includes/collection/class-outbox.php
```diff
@@ -72,6 +79,51 @@ public static function add( $activity_object, $activity_type, $user_id, $content
...
+				'post_type'   => self::POST_TYPE,
+				'post_status' => 'pending',
+				'exclude'     => array( $current_id ),
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+				'meta_query'  => $meta_query,
+				'fields'      => 'ids',
+			)
+		);
+
+		foreach ( $existing_items as $existing_item_id ) {
```
obenland, line 136:
```suggestion
		foreach ( $existing_items as $existing_item_id ) {
			$event_args = array(
				Dispatcher::$callback,
				$existing_item_id,
				Dispatcher::$batch_size,
				\get_post_meta( $outbox_item->ID, '_activitypub_outbox_offset', true ) ?: 0, // phpcs:ignore
			);
			$timestamp = wp_next_scheduled( 'activitypub_async_batch', $event_args );
			wp_unschedule_event( $timestamp, 'activitypub_async_batch', $event_args );
```

obenland: Something like this?

pfefferle: nice! I was just sitting on that :)

includes/class-dispatcher.php
obenland: Hm. I guess it's _possible_ when a reprocess job gets here and the result of `should_send_to_followers` changed since the original run? Feels so unlikely

includes/collection/class-outbox.php
```diff
@@ -72,6 +79,59 @@ public static function add( $activity_object, $activity_type, $user_id, $content
...
+			$event_args = array(
+				Dispatcher::$callback,
+				$existing_item_id,
+				Dispatcher::$batch_size,
+				\get_post_meta( $outbox_item->ID, '_activitypub_outbox_offset', true ) ?: 0, // phpcs:ignore
+			);
+			$timestamp = wp_next_scheduled( 'activitypub_async_batch', $event_args );
+			wp_unschedule_event( $timestamp, 'activitypub_async_batch', $event_args );
```
obenland:
```suggestion
			wp_unschedule_event( $timestamp, 'activitypub_async_batch', $event_args );

```

includes/collection/class-outbox.php
obenland: @pfefferle It's untested! I just wrote it down from memory

includes/class-dispatcher.php
pfefferle: I have no strong feeling here, happy to remove it...

obenland:
We currently don't seem to account for the blog user's reposts 
image

pfefferle:
Oh, this might be missing in the Stream integration.

pfefferle:
@obenland what do you mean? all the `Announces` are the reposts.

obenland:
Right, I would have expected those to also be canceled and the Outbox item published

pfefferle:
Ah, that is what you meant! makes sense!

includes/collection/class-outbox.php
```diff
@@ -72,6 +86,62 @@ public static function add( $activity_object, $activity_type, $user_id, $content
...
+	 * @param string $activity_type The type of the activity.
+	 * @param int    $current_id    The ID of the current outbox item to exclude.
+	 *
+	 * @return void
+	 */
+	private static function invalidate_existing_items( $object_id, $activity_type, $current_id ) {
+		$meta_query = array(
+			array(
+				'key'   => '_activitypub_object_id',
+				'value' => $object_id,
```
obenland, line 113: For Blog users, the object_id contains a timestamp (`http://localhost:8888/?p=1834#activity-update-2025-02-11T21:02:54Z`), so there won't be a shared object_id between activities. Should we remove the hash for this purpose?

pfefferle: we should either use the guid of the original Outbox-Item or the `get_id` of the original `$wp_object`.

obenland: I like the original `$wp_object`, which it already saves currently, just with the added hash. we could strip the hash in Outbox::add()

pfefferle: This feels hacky! the hash was a workaround because we had no persistant IDs. I would like to fix that, so that we do no longer have the hashes anywhere.

obenland: Sure! Could I leave that with you? I still don't feel super confident in that area

pfefferle: Should work now!

includes/collection/class-outbox.php
```diff
@@ -38,17 +42,32 @@ public static function add( $activity_object, $activity_type, $user_id, $content
...
 			'post_author'  => \max( $user_id, 0 ),
 			'post_status'  => 'pending',
 			'meta_input'   => array(
-				'_activitypub_activity_type'     => $activity_type,
-				'_activitypub_activity_actor'    => $actor_type,
-				'activitypub_content_visibility' => $content_visibility,
+				'_activitypub_object_id'         => \esc_url_raw( $activitypub_object_id ),
+				'_activitypub_activity_type'     => \esc_attr( $activity_type ),
+				'_activitypub_activity_actor'    => \esc_attr( $actor_type ),
+				'activitypub_content_visibility' => \esc_attr( $content_visibility ),
```
obenland: What makes it necessary to escape these three?

pfefferle: shouldn't we escape anything to go into the db?

pfefferle: Happy to remove it

pfefferle: Ah, we already do that in the register_post_meta call! 🫣

obenland: Yeah, the naming of esc_url_raw() compared with the rest of the `esc_*()` family of functions is confusing. [This VIP doc](https://docs.wpvip.com/security/validating-sanitizing-and-escaping/) is a good primer on the difference between sanitization and escaping in WP and links to [a handy list of available sanitization functions](https://developer.wordpress.org/apis/security/sanitizing/#sanitization-functions).

`esc_url_raw()` calls `sanitize_url()` which then calls `esc_url()` with the `db` context.

obenland: Let's also remove them from `invalidate_existing_items()`

pfefferle: done

obenland (approved):

PR #1287, opened 2025-02-07
Bump Min WP Version to 6.4

obenland:
Let's bump the min version to 6.4. The last three versions is what the Plugin Directory expects plugins to be tested with and it gets us guaranteed `WP_HTML_Tag_Processor` support. 6.4 is also the first version that supports PHP 8.3.

Based on https://fedi.wrm.sr/software/wordpress/versions, 126 installs are below 6.4, or around 1.47% of all ActivityPub installs.

## Proposed changes:

* Bumps min WP version to 6.4.
* Removes now unnecessary class_exists checks and fallbacks.

### Other information:


includes/functions.php
```diff
@@ -553,17 +553,6 @@ function is_single_user() {
  * @return boolean True if the site supports the block editor, false otherwise.
  */
 function site_supports_blocks() {
-	if ( \version_compare( \get_bloginfo( 'version' ), '5.9', '<' ) ) {
-		return false;
-	}
-
-	if (
-		! \function_exists( 'register_block_type_from_metadata' ) ||
```
pfefferle, line 479: let's keep that, so that it will still work with classicpress or the classic editor enabled.

includes/transformer/class-post.php
```diff
@@ -883,43 +879,6 @@ protected function get_classic_editor_image_embeds( $max_images ) {
...
-	/**
-	 * Get image attachments from the classic editor.
-	 * This is imperfect as the contained images aren't necessarily the
-	 * same as the attachments.
-	 *
-	 * @param int $max_images The maximum number of images to return.
-	 *
-	 * @return array The attachment IDs.
-	 */
-	protected function get_classic_editor_image_attachments( $max_images ) {
```
pfefferle, line 902: let's keep that, for people that still use the classic editor

pfefferle: Ah, this is only about the attachments 🤔

includes/functions.php
obenland: The functions exist with the classic editor enabled. Should ClassicPress be a consideration?

includes/transformer/class-post.php
obenland: Good to remove?

mattwiebe: I think we should leave this in. This is what will allow people who don't use blocks via the Classic Editor plugin, or who post using the REST API, to still have their attachments show up in the Activity.

obenland: That doesn't seem to be accounted for in lines 799-803, it only looks for `WP_HTML_Tag_Processor` and uses `get_classic_editor_image_embeds()` when it's available

mattwiebe: Oh, I see. I thought we were ditching all classic editor support with this, didn't read closely enough!

I think we should be good to get rid of it, then.

mattwiebe (dismissed):
:shipit:

includes/transformer/class-post.php
```diff
@@ -796,11 +796,7 @@ protected function get_classic_editor_images( $media, $max_images ) {
 		}
 
 		if ( \count( $media['image'] ) <= $max_images ) {
-			if ( \class_exists( '\WP_HTML_Tag_Processor' ) ) {
```
pfefferle, line 803: no need to stick with the `get_classic_editor_images`, because we can move `get_classic_editor_image_embeds` up one level and move the max images check to the `get_classic_editor_image_embeds` function.

pfefferle (changes requested):

includes/transformer/class-post.php
obenland: Could this be done in a separate PR?

pfefferle: sure

includes/collection/class-extra-fields.php
```diff
@@ -114,8 +114,7 @@ public static function fields_to_attachments( $fields ) {
 			if (
 				\stripos( $link_content, '<a' ) === 0 &&
 				\stripos( $link_content, '<a', 3 ) === false &&
-				\stripos( $link_content, '</a>', \strlen( $link_content ) - 4 ) !== false &&
-				\class_exists( '\WP_HTML_Tag_Processor' )
```
pfefferle: can we keep that one for a little longer?

obenland: Reverted in [e54d59e](https://github.com/Automattic/wordpress-activitypub/pull/1287/commits/e54d59e9ee0a53c898529773e6030c5c05292e49)

pfefferle (approved):

PR #1288, opened 2025-02-07
Purge Outbox Items on Schedule

obenland:
Helps preventing tons of Outbox items from polluting the wp_posts table. This should be followed up with a PR that makes the time offset customizable.

## Proposed changes:

* Adds a recurring event to purge Outbox items.
* Adds a callback that purges outbox items when there's at least 20 items.
* Adds unit tests.

### Other information:

## Testing instructions:

* Have a ton of Outbox items, at least 20.
* Maybe manually adjust the days threshold in `purge_outbox()` so it catches a few posts.
* Schedule the purging job: ` wp cron event schedule activitypub_outbox_purge`
* Run the job: `wp cron event run --due-now`

### DIFF
--- a/includes/class-scheduler.php
+++ b/includes/class-scheduler.php
@@ -209,6 +209,11 @@ class Scheduler {
 	 * Purge outbox items based on a schedule.
 	 */
 	public static function purge_outbox() {
+		$total_posts = (int) wp_count_posts( Outbox::POST_TYPE )->publish;
+		if ( $total_posts <= 20 ) {
+			return;
+		}
+
 		$days     = 180; // TODO: Replace with a setting.
 		$timezone = new \DateTimeZone( 'UTC' );
 		$date     = new \DateTime( 'now', $timezone );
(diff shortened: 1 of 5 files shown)


obenland:
On a blog with low activity, this could result in an empty Outbox. Is that okay?

pfefferle:
what about two factors... keep at least x items and delete every item older than x... ?

obenland:
yeah, what would be a good number?

pfefferle:
something between 20 and 50?

includes/class-scheduler.php
```diff
@@ -199,6 +205,39 @@ public static function reprocess_outbox() {
 		}
 	}
 
+	/**
+	 * Purge outbox items based on a schedule.
+	 */
+	public static function purge_outbox() {
+		$total_posts = (int) wp_count_posts( Outbox::POST_TYPE )->publish;
+		if ( $total_posts <= 20 ) {
```
pfefferle, line 213: should we use a const for the number?

pfefferle (approved):

includes/class-scheduler.php
obenland: Let's keep it simple for now and see if we get feedback about it?

PR #1291, opened 2025-02-07
Add basic relay support!

pfefferle:
Support for sending Activities to ActivityPub Relays, to improve discoverability of public content.

Some context:

* https://relaylist.com/
* https://joinfediverse.wiki/Relays
* https://fedi.tips/using-relays-to-quickly-expand-a-servers-view-of-the-fediverse/

Screenshot 2025-02-28 at 13 30 26

## Proposed changes:

* Add Form to add relays
* Add some context for users to understand the feature
* Send Activities to relays

### Other information:

## Testing instructions:

* See unittests


Menrath:
I would love to read the story behind this :)

pfefferle:
There is no big story behind it, the idea is mainly to make a blog a bit more visible by publishing posts to a relay. Very similar to WordPress' classic ping service.

includes/class-dispatcher.php
```diff
@@ -97,6 +97,9 @@ public static function process_outbox( $id ) {
 		// Send to mentioned and replied-to users. Everyone other than followers.
 		self::send_to_interactees( $activity, $actor->get__id(), $outbox_item );
 
+		// Send to relays.
+		self::send_to_relays( $activity, $actor, $outbox_item );
```
obenland: maybe `send_to_interactees()` was too specific a naming? I worry that we'll keep adding groups of recipients that are not followers, which don't really need to be broken up. Like, we should be able to use a filter like `activitypub_send_to_inboxes` to add any non-follower inboxes?

pfefferle: I initially used the filter, but as you said, the name is a bit too specific

pfefferle: I used a more generic filter name and rewrote the relay implementation to use that filter. Happy to use a different name, curious what you think about the change!

includes/class-dispatcher.php
```diff
@@ -49,12 +49,13 @@ public static function init() {
 		\add_action( 'activitypub_process_outbox', array( self::class, 'process_outbox' ) );
 
 		// Default filters to add Inboxes to sent to.
-		\add_filter( 'activitypub_interactees_inboxes', array( self::class, 'add_inboxes_by_mentioned_actors' ), 10, 3 );
-		\add_filter( 'activitypub_interactees_inboxes', array( self::class, 'add_inboxes_of_replied_urls' ), 10, 3 );
+		\add_filter( 'activitypub_custom_inboxes', array( self::class, 'add_inboxes_by_mentioned_actors' ), 10, 3 );
+		\add_filter( 'activitypub_custom_inboxes', array( self::class, 'add_inboxes_of_replied_urls' ), 10, 3 );
+		\add_filter( 'activitypub_custom_inboxes', array( self::class, 'add_inboxes_of_relays' ), 10, 3 );
```
obenland: How would you feel about `activitypub_additional_inboxes` for the hook?

includes/class-dispatcher.php
```diff
@@ -95,7 +100,7 @@ public static function process_outbox( $id ) {
 		$activity = Outbox::get_activity( $outbox_item );
 
 		// Send to mentioned and replied-to users. Everyone other than followers.
-		self::send_to_interactees( $activity, $actor->get__id(), $outbox_item );
+		self::send_to_custom_inboxes( $activity, $actor->get__id(), $outbox_item );
```
obenland: How would you feel about `activitypub_initial_inboxes()` or `activitypub_primary_inboxes()` instead of custom?

includes/class-dispatcher.php
```diff
@@ -405,4 +412,37 @@ protected static function should_send_to_followers( $activity, $actor, $outbox_i
...
+		}
+
+		$relays = \get_option( 'activitypub_relays', array() );
+
+		if ( empty( $relays ) ) {
+			return $inboxes;
+		}
+
+		$inboxes = array_merge( $inboxes, $relays );
+		$inboxes = array_unique( $inboxes );
```
obenland: I think we can avoid calling this here, the function unique's it after the filter.

includes/wp-admin/class-settings-fields.php
```diff
@@ -466,4 +482,27 @@ public static function render_authorized_fetch_field() {
...
+		</p>
+		<textarea
+			id="activitypub_relays"
+			name="activitypub_relays"
+			class="large-text"
+			cols="50"
+			rows="5"
+		><?php echo esc_textarea( implode( PHP_EOL, $value ) ); ?></textarea>
+		<p class="description">
+			<?php echo wp_kses( __( 'Enter the <strong>Inbox-URLs</strong> (e.g. <code>https://relay.example.com/inbox</code>) of the relays you want to use, one per line.', 'activitypub' ), array( 'strong' => array() ) ); ?>
```
obenland: The allowlist will need `code`, too, unless you want to add it through a placeholder

includes/wp-admin/class-settings-fields.php
```diff
@@ -466,4 +482,27 @@ public static function render_authorized_fetch_field() {
...
+
+	/**
+	 * Render relays field.
+	 */
+	public static function render_relays_field() {
+		$value = get_option( 'activitypub_relays', array() );
+		?>
+		<p>
+			<?php echo wp_kses( __( 'A <strong>Fediverse-Relay</strong> distributes content across instances, expanding reach, engagement, and discoverability, especially for smaller instances.', 'activitypub' ), 'default' ); ?>
+		</p>
```
obenland: How would you feel about moving this below the textarea? In previous PRs I noticed there's no margins between this text and the input box.

includes/class-dispatcher.php
pfefferle: is this really "initial" or "primary"?

includes/wp-admin/class-settings-fields.php
pfefferle: I thought that this might be a relevant info you should read before filling out stuff!?!

includes/class-dispatcher.php
obenland: I guess it's initial in the way that it gets processed before followers 🤷 
`send_to_non_follower_inboxes`?

includes/wp-admin/class-settings-fields.php
obenland: I do hope users read descriptions regardless of where they're placed

includes/class-dispatcher.php
pfefferle: I am not a fan of negations... what about something like `activitypub_special_inboxes`?

I thought about `custom`, because they are indirectly customized by the user through adding relays or replying to someone.

obenland: Maybe we could go with `send_to_additional_inboxes()` to mirror the filter?

obenland (approved):

PR #1292, opened 2025-02-07
Add URI transformer

### LINKED ISSUE #1290: Feature Request: Add the ability to use the outbox to Announce and Like any URL
### What

Use case from the Friends plugin: there you can like or announce any URL, this is currently not possible with the outbox.

### Why

Because ActivityPub supports announcing just URLs, not whole objects.

### How

This would need a URL transformer.

pfefferle:
Allow Activities on URLs instead of requiring Activity-Objects. This is useful especially for sending Announces and Likes.

Fixes #1290

## Proposed changes:

*

### Other information:


tests/includes/transformer/class-test-factory.php
```diff
@@ -214,4 +214,53 @@ function () {
...
+		// WP_Error für fehlgeschlagene Anfrage erstellen
+		$fake_request = function () {
+			return new \WP_Error( 'fetch_error', 'Failed to fetch remote object' );
+		};
+
+		add_filter( 'pre_http_request', $fake_request, 10 );
+
+		$uri_transformer = Factory::get_transformer( 'https://example.com/invalid' );
+
+		$this->assertInstanceOf( \WP_Error::class, $uri_transformer );
```
obenland: No need to change this, but you recently mentioned how much you enjoy learning new unit test functions: WordPress created [`assertWPError()`](https://github.com/WordPress/wordpress-develop/blob/381731aafc815c399b13a524c519fc1c3fe00018/tests/phpunit/includes/abstract-testcase.php#L837) as a wrapper for the `assertInstanceOf()` check :)

includes/transformer/class-factory.php
```diff
@@ -96,6 +106,8 @@ public static function get_transformer( $data ) {
 				break;
 			case 'json':
 				return new Json( $data );
+			case 'uri':
+				return new Uri( $data );
```
obenland: Is this PR missing a file? `Uri` is undefined

obenland: Looks like e5a94ab68b339f12c104677f748ad26755ae06b0 removed it

obenland (dismissed):

obenland (approved):

PR #1299, opened 2025-02-10
Enforce: 200 status header

### LINKED ISSUE #1296: Maybe enfore Status Code 200 when reaching ActivityPub template
When the queried object in the `\ActivityPub\Query` class is not found by the core `\get_queried_object()`, WordPress already sets the status code to **404**. Maybe when reaching the ActivityPub template a Status O.K. should be enforced.

Meanwhile I applied a fix via the filter in the template on my side: https://codeberg.org/Event-Federation/wordpress-event-bridge-for-activitypub/pulls/130/files

pfefferle:
Fix #1296

props @Menrath

### DIFF
--- a/tests/includes/class-test-activitypub.php
+++ b/tests/includes/class-test-activitypub.php
@@ -15,6 +15,25 @@ use Activitypub\Collection\Outbox;
  * @coversDefaultClass \Activitypub\Activitypub
  */
 class Test_Activitypub extends \WP_UnitTestCase {
+	/**
+	 * Test user ID.
+	 *
+	 * @var int
+	 */
+	protected static $user_id;
+
+	/**
+	 * Create fake data before tests run.
+	 *
+	 * @param WP_UnitTest_Factory $factory Helper that creates fake data.
+	 */
+	public static function wpSetUpBeforeClass( $factory ) {
+		self::$user_id = $factory->user->create(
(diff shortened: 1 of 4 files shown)


obenland:
Let’s add a unit test, recreating the original bug

pfefferle:
I have not had the chance to reproduce it yet 🫣

pfefferle:
So now I could reproduce it and added some tests!

Any feedback @obenland and @Menrath ???

tests/includes/class-test-activitypub.php
```diff
@@ -115,4 +134,89 @@ public function activity_meta_sanitization_provider() {
...
+		unset( $_SERVER['HTTP_ACCEPT'] );
+		_unregister_post_type( 'test_cpt' );
+	}
+
+	/**
+	 * Test that ActivityPub requests for custom post types return 404 even with post type support.
+	 *
+	 * @covers ::render_activitypub_template
+	 */
+	public function test_custom_post_type_with_support_returns_404() {
```
Menrath: This function name is misleading now :)

pfefferle: ooops, that's because I first tried it without the 200 fix to reproduce it! damn AI :)

obenland (approved):

PR #1301, opened 2025-02-10
Add Undo functionality

### LINKED ISSUE #1289: Feature Request: Add the ability to send out an "Undo" for a previous activity
### What

In ActivityPub, certain actions can be undone, for example announces and likes. For this to work, we need to wrap the `Announce` or `Like` activity in an object of type `Undo`.

### Why

It would be convenient to have a function in the outbox to achieve this because you need to reconstruct the inner object which already was in the outbox earlier.

### How

Add an `undo()` method to the Outbox object.

obenland:
Fixes #1289

## Proposed changes:

* Moves `get_activity()` and `get_author()` methods to `Outbox` class, where they feel more appropriate. These methods take an outbox id.
* Adds `undo()` method to `Outbox` class, that accepts an outbox item and creates an Undo activity based on it. It accounts for `Create` and `Add` activities requiring their own "undo" type.
* Updates `Base::transform_object_properties()` and `Base_Object::to_array()` to not only discard `null` values but also empty values unless they're `false`.
* Adds unit tests for the new functionality.

### Other information:

## Testing instructions:

* Test by running it in CLI: `wp activitypub undo 123` where 123 is an Outbox post id.
* Test by running it in CLI: `wp activitypub undo "https://example.com/?post_type=ap_outbox&p=123"` where the URL is an Outbox post guid.
* The changes to `Base::transform_object_properties()` and `Base_Object::to_array()` are probably what has the highest potential for unintended consequences. Anything standing you there?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+* Undo API for Outbox items.
+
 ## [5.2.0] - 2025-02-13
 
 ### Added
@@ -13,7 +19,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Outbox processed events get logged in Stream and show any errors returned from inboxes.
 * Outbox items older than 6 months will be purged to avoid performance issues.
 * REST API endpoints for likes and shares.
-* Undo API for Outbox items.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Outbox processed events get logged in Stream and show any errors returned from inboxes.
 * Outbox items older than 6 months will be purged to avoid performance issues.
 * REST API endpoints for likes and shares.
+* Undo API for Outbox items.
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


obenland:
Tests fail because the activity saved in Outbox items has `tag` and `cc` properties, and the Activity object doesn't?

obenland:
@pfefferle Setting [the default for `cc` to `null`](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/transformer/class-base.php#L174) actually results in it showing up in objects as `cc:[null]` since [there are no more checks after `set_audience()` runs](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/transformer/class-base.php#L134-L136).

Not sure how you feel about [defaulting to an empty array](https://github.com/Automattic/wordpress-activitypub/pull/1301/files#diff-40461128132f656da7bc0f129679b45bfff64d9131f39c4eef9fa1ff729afd7cR246-R248) for it in `Activity`?

obenland:
New approach: Being [more](https://github.com/Automattic/wordpress-activitypub/pull/1301/files#diff-9f46d66a9baa398c489c736ffd8882a6387b810c1c200d718183509a0d5f2f9b) [prescriptive](https://github.com/Automattic/wordpress-activitypub/pull/1301/files#diff-e234ab9bc159f49c9fbd132cea9d72fbfb70a6e9f161d703a27608a2bb1b8159) than checking for `isset()` also removes these empty properties.

mattwiebe (commented):
This seems good to me. 

Where, if anywhere, could we also expose this in the UI? Should we add a `wp-cli` command?

obenland:
I like that, yes!

obenland:
Added a cli command in [f5b2b1f](https://github.com/Automattic/wordpress-activitypub/pull/1301/commits/f5b2b1fd568196979c3d200b05ea3053e0828e02)

pfefferle (approved):

PR #1303, opened 2025-02-10
Break up Collections endpoint and extend WP_REST_Controller

obenland:
Follow-up to https://github.com/Automattic/wordpress-activitypub/pull/1146.

## Proposed changes:

* Creates separate controllers for Replies and Moderators collections.
* Rewrites all three endpoints to build on WP_REST_Controller.
* Adds unit tests.

### Other information:


includes/rest/class-replies-controller.php
```diff
@@ -0,0 +1,171 @@
...
+		} else {
+			$response = Replies::get_collection( $wp_object );
+		}
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		// Add ActivityPub Context.
+		$response['@context'] = Base_Object::JSON_LD_CONTEXT;
```
pfefferle: it does not really make a difference, but can we make `@context` the first item of the array, because of readability.

pfefferle (dismissed):

PR #1304, opened 2025-02-11
Fix: duplicate content check

pfefferle:
If a comment is sent twice and, due to a race condition, both copies are stored in the database, the current faulty check will cause all new comments to be added as separate entries instead of extending the first one!

---

There are scenarios where comments might be sent twice simultaneously, and since WordPress doesn’t use transactions, this could result in duplicate database entries for the same Activity. The current code checks if multiple comments share the same ID and, if so, treats it as if no comment exists.

This means that if a comment is duplicated in the database (for whatever reason), any new Activity related to that comment would be added as a separate entry.

The PR fixes this behavior, but we still need a solution to prevent the initial duplication.


obenland:
What's the race condition?

pfefferle:
there are possibilities were comments might be sent twice at the same time and because WordPress does not use transactions, it might result in two db entries for the same Activity. The current code checks if there are more than one comment with the same id, if so, it was handled like there is no comment.

So if you have a comment twice in the db (however this might happen), the code would add any new Activity on that comment as a new one.

The PR fixes this.

We still need a fix that prevents the initial duplicate content.

includes/class-comment.php
```diff
@@ -264,17 +264,15 @@ public static function object_id_to_comment( $id ) {
...
+				'orderby'    => 'comment_date',
+				'order'      => 'DESC',
 			)
 		);
 
 		if ( ! $comment_query->comments ) {
 			return false;
 		}
 
-		if ( count( $comment_query->comments ) > 1 ) {
```
mattwiebe, line 274: We could probably just combine with previous check to be

```php
if ( count( $comment_query->comments ) !== 1 ) {
	return false
}
```

pfefferle: the thing is: it does not make sense to check for the number, because if there is at least one (or more), it should return the first one.

obenland:
Have you found a way to reproduce this?

pfefferle:
@obenland don't get me wrong, this does not prevent the case that there might be two Activities at the same time, that would cause duplicate DB entries. To prevent that, we might have to think about db transactions or maybe work on a persistent inbox.

this PR fixes the follow up issue, that it does not become more than these two entries!

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Enforce 200 status header for valid ActivityPub requests.
+* Duplicate comment check.
```
obenland: Let's expand on this? What was fixed, how?

obenland (changes requested):
I really appreciate everything you've done so far in terms of improving the PR description and giving more context, and I want to apologize for maybe coming across as pedantic, but I still don't feel like I have a good idea of the problem that needs fixing and how this PR fixes it.

How should this be tested? Are there steps you could share to reproduce the bug, and testing instructions for the fix?

akirk:
As someone affected, currently I receive comments twice:
Screenshot 2025-02-12 at 15 36 41

Granted, Mastodon is posting this twice:
```
[12/Feb/2025:15:36:30 +0100] "POST /wp-json/activitypub/1.0/actors/1/inbox HTTP/1.1" 202 12 "-" "http.rb/5.1.1 (Mastodon/4.2.13; +https://mastodon.nl/)"
[12/Feb/2025:15:36:30 +0100] "POST /wp-json/activitypub/1.0/actors/0/inbox HTTP/1.1" 202 12 "-" "http.rb/5.1.1 (Mastodon/4.2.13; +https://mastodon.nl/)"
```

But there is certainly [only one reply that I sent](https://mastodon.nl/@akirktest/113991417183901082).

I saw other reports about double-posting of Likes, too.

pfefferle:
How it should work:

* Add a new comments with a `source_id`
* Try to add a new comment with the same `source_id`

In this case, the second comment should update the first comment.

The problem:

* Add two comments to the db that have the same `source_id`
* Try to add a third comment with the same `source_id`. The new comment will be added as a third comment now.

The problem is, that the code currently falsely checks if there is more than one item with the same `source_if` and if so, it falsely returns `false`, what means that there is no comment with that `source_id`.

The result: Every update of that reply will create a new comment!

obenland:
Thanks, that is helpful. How come the duplicate comment check in wp_allow_comment() doesn't catch these?

pfefferle:
@obenland I think we remove it.

obenland:
Looks like [only in tests](https://github.com/search?q=repo%3AAutomattic%2Fwordpress-activitypub%20duplicate_comment_id&type=code)?

pfefferle:
Isn't it part if the `check_comment_flood` hook?

https://github.com/Automattic/wordpress-activitypub/blob/f96a3a6839cc3678b10a15f02b1e81df053d4c5f/includes/collection/class-interactions.php#L292

pfefferle:
@obenland I see that there is lots of possibility to improve the code, but I think we should write isses/tasks for that. This simple PR prevents to get spammed with comments, that is why I recommend to merge it as soon as possible.

obenland (approved):
I don't know, it feels to me like we're fixing the symptom and not the cause. Maybe I still don't fully understand what's going on here. We should investigate why the duplicate comment check doesn't catch these.

Approving because I trust your judgement and I don't want perfection to stand in the way of progress.

PR #1307, opened 2025-02-11
Magic method docs

obenland:
## Proposed changes:

* Adds docs for magic methods where I could find them.

Can be amended in the future if there are any I missed. Just did a quick visual search in my IDE.

### DIFF
--- a/includes/activity/class-base-object.php
+++ b/includes/activity/class-base-object.php
@@ -191,7 +191,7 @@ class Base_Object {
 	 *
 	 * @see https://www.w3.org/TR/activitystreams-vocabulary/#dfn-icon
 	 *
-	 * @var string
+	 * @var string|array
 	 *    | Image
 	 *    | Link
 	 *    | array<Image>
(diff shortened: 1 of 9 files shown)


includes/activity/class-base-object.php
```diff
@@ -191,7 +191,7 @@ class Base_Object {
 	 *
 	 * @see https://www.w3.org/TR/activitystreams-vocabulary/#dfn-icon
 	 *
-	 * @var string
+	 * @var string|array
 	 *    | Image
```
pfefferle: I think we have to either remove the following then, or put them also in line

pfefferle:
There is a conflict

includes/activity/class-base-object.php
obenland: I can put them inline. Are you okay with removing ActivityPub spec types like Link or ObjectType in favor of PHP types?

pfefferle: Sure!

pfefferle (approved):

PR #1308, opened 2025-02-12
Add rest controller for likes and shares

pfefferle:
This PR adds endpoints for likes and shares and fixes some issues in the replies collection. The likes and shares endpoints list only the absolute number without any items.

A WordPress post on Mastodon (or any other Fediverse-Platform) does not properly show the likes and shares count, because of the missing endpoints. This PR should fix this issue.

Added:

* Shares endpoint, that lists a counter how much an entry was shared
* Likes endpoint, that lists a counter how much an entry was liked

Fixed issues:

* Properly nest collections page in Collection
* Set first page to "1" instead of "0"
* Fix unittests

### Other information:

## Testing instructions:

* Load `/wp-json/activitypub/1.0/posts/<id>/likes` or `/wp-json/activitypub/1.0/posts/<id>/shares` on your test system

### DIFF
--- a/includes/rest/class-replies-controller.php
+++ b/includes/rest/class-replies-controller.php
@@ -161,10 +161,10 @@ class Replies_Controller extends \WP_REST_Controller {
 	 * @return array Response collection of likes.
 	 */
 	public function get_likes( $request, $wp_object ) {
-		if ( ! $wp_object instanceof \WP_Post ) {
-			$likes = 0;
-		} else {
+		if ( $wp_object instanceof \WP_Post ) {
 			$likes = Interactions::count_by_type( $wp_object->ID, 'like' );
+		} else {
+			$likes = 0;
 		}
 
 		$response = array(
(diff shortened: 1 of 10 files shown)


includes/rest/class-replies-controller.php
```diff
@@ -90,29 +100,104 @@ public function get_items( $request ) {
...
+	 *
+	 * @return array Response collection of likes.
+	 */
+	public function get_likes( $request, $wp_object ) {
+		if ( ! $wp_object instanceof \WP_Post ) {
+			$likes = 0;
 		} else {
-			$response = Replies::get_collection( $wp_object );
+			$likes = Interactions::count_by_type( $wp_object->ID, 'like' );
 		}
```
obenland: In cases like this, where we have something positive to test for, let's do that in the if case and fall back to the default in the else statement.
Alternatively you could define the default and then just do the if case, to change it if the check succeeds.

includes/rest/class-replies-controller.php
```diff
@@ -90,29 +100,104 @@ public function get_items( $request ) {
...
+	 * @param \WP_Post|\WP_Comment $wp_object The WordPress object.
+	 *
+	 * @return array Response collection of shares.
+	 */
+	public function get_shares( $request, $wp_object ) {
+		if ( ! $wp_object instanceof \WP_Post ) {
+			$shares = 0;
+		} else {
+			$shares = Interactions::count_by_type( $wp_object->ID, 'repost' );
 		}
```
obenland, line 192: Same as above

obenland (approved):
I noticed we don't send a `totalItems` count in replies responses, should we? 
It's not a regression, this is good to go!

pfefferle:
I would always send a totalItems... some platforms only consume that number!

PR #1311, opened 2025-02-12
Use Settings API for future proofing

obenland:
Using the Settings API will make it easier to move with any future wp-admin redesign and enables third-party plugins to add settings to sections we register. 

## Proposed changes:

* Moves hardcoded settings into Settings API calls.

### Before
Screenshot 2025-02-12 at 2 07 00 PM

### After
Screenshot 2025-02-12 at 2 17 02 PM

### Other information:

## Testing instructions:

* Go to Settings > ActivityPub.
* Change a setting and save.
* Make sure it persists and is actually updated.


pfefferle:
Can we use templates instead of directly returning HTML?

pfefferle:
What about the welcome screen?

obenland:
I was considering moving the Welcome screen over as well. I think most of it could be displayed using the settings api, maybe with the exception of the accordion at the bottom?

At the end of the day they're not settings, though. The Settings API is not meant as a tool to structure admin pages, so I think I'd want to keep it purpose-aligned.

obenland:
> Can we use templates instead of directly returning HTML?

Do you mean for the display callbacks that render the setting? I don't know, I've not considered it. Templates would be one more level of abstraction where it might not be needed? Currently everything is encapsulated in that one file and it's easy to read and jump back and forth. 

What would be reasons for having it in templates?

Menrath:
@obenland Maybe Maintainability and Readability.

You might want to be interested in what GatherPress did, they have a very generic and modular approach: https://github.com/GatherPress/gatherpress/blob/b5edb98c1bcf889edb553e4f8f7fb7373985fd16/includes/core/classes/class-settings.php

pfefferle:
@obenland "separation of concerns", "MVC", ...

includes/class-settings-fields.php
```diff
@@ -0,0 +1,384 @@
+<?php
```
pfefferle, line 1: should we put these under the admin namespace and in an admin subfolder? like `Activitypub\Admin\Settings_Fields`??

obenland: Yes! There are a few other classes that would fit in there, and I'd like to combine it with hooking them into `admin_init` rather than `init`, so I thought that could be a good follow-up PR

pfefferle (approved):

PR #1324, opened 2025-02-13
Settings: Remove boxes from Welcome page

obenland:
## Proposed changes:

* Removes the custom boxes layout and reverts to a more wp-admin look.

### Before
Screenshot 2025-02-13 at 3 12 35 PM

### After
Screenshot 2025-02-13 at 3 12 00 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


obenland:
It's definitely harder to differentiate individual sections than it was before.
Maybe it would help to give the Blog Profile and Author Profile sections a Settings treatment and display them in two columns?

mattwiebe:
I think it looks better before :)

obenland:
Yeah it's not great. I'll try the settings lock for blog and author profiles

obenland:
Not really better 
image

pfefferle:
What about removing all the boxes except the profile-blocks?

pfefferle:
the profiles do not look very good that way 🤔

obenland:
Yeah. Let's leave it as-is for now?

obenland:
Closing for now, until we can think of a better way forward.

PR #1331, opened 2025-02-17
Add: Preview support for Articles

pfefferle:
Mastodon does not support `Articles` yet, so we need a way to provide subsets of the `Article`, so that Microblogging services can show a summary.

The `summary` filed of an `Article` is not reliable enough, because it is (mis)used for the content warning.

This PR adds a `Note` as `preview`, so that Microblogging services have something to fall back when they do not support `Article`s

Some context:

* https://codeberg.org/fediverse/fep/src/branch/main/fep/b2b8/fep-b2b8.md#preview
* https://codeberg.org/evanp/fep/issues/21

props @Menrath 

## Proposed changes:

* Check if Object-Type is `Article` and add summary as preview `Note` object

### Other information:

## Testing instructions:

* Create an Article
* Load the permalink
* Add `?activitypub`
* Check JSON

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Undo API for Outbox items.
 * Allow Activities on URLs instead of requiring Activity-Objects. This is useful especially for sending Announces and Likes.
 
-
 ## [5.2.0] - 2025-02-13
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* A `preview` property to the `Article` object.
+* A fallback `Note` for `Article` objects to improve previews on services that don't support Articles yet.
 * Undo API for Outbox items.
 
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+* A `preview` property to the `Article` object.
```
obenland: This is not overly helpful, I'm afraid. Why did we add a preview property to the object? What's the underlying reason for this change?

obenland (changes requested):
> the summary is used for content warnings and therefore not reliable enough.

Not reliable enough for what? Could you add more context to the PR description and changelog? It's not obvious to me why this change is necessary.

CHANGELOG.md
obenland:
```suggestion
* A fallback `Note` for `Article` objects to improve previews on services that don't support Articles yet.
```

obenland: Something like that?

obenland (dismissed):
Thank you!

obenland (approved):

PR #1332, opened 2025-02-17
Add: Resolvable Outbox-Activities

pfefferle:
This PR makes the Outbox Activities resolvable through their ID URLs (post-guid).

## Important!

Aside from that, it fixes an error in the Actor-validation in several REST-Controllers.

## Proposed changes:

* Fix Actor validation for REST-API queries
* Makes Outbox Activities resolvable

### Other information:

## Testing instructions:

* Enable `WP_DEBUG`
* Go to `/wp-admin/edit.php?post_type=ap_outbox`
* Copy a `guid`
* Load `guid` URL with ActivityPub `Accept`header


includes/collection/class-outbox.php
```diff
@@ -7,8 +7,8 @@
 
 namespace Activitypub\Collection;
 
-use Activitypub\Activity\Activity;
 use Activitypub\Dispatcher;
+use Activitypub\Activity\Activity;
```
obenland, line 11: Don't like ordering them alphabetically?

includes/rest/class-outbox-controller.php
```diff
@@ -84,7 +84,7 @@ public function register_routes() {
 	public function validate_user_id( $user_id ) {
 		$user = Actors::get_by_various( $user_id );
 		if ( \is_wp_error( $user ) ) {
-			return $user;
+			return false;
```
obenland: Why can't we return the WP_Error?

templates/outbox-json.php
```diff
@@ -0,0 +1,27 @@
...
+
+// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+echo $activity->to_json();
+
+/**
+ * Fires after an ActivityPub blog profile has been generated and sent to the client.
+ *
+ * @param int $user_id The ID of the WordPress blog user whose profile was generated.
+ */
+\do_action( 'activitypub_json_post', $activity );
```
obenland: This file can be deleted now, right?

includes/rest/class-outbox-controller.php
pfefferle: The doc says boolean and because it wasn't working properly I tried to follow the spec as best as possible.

includes/collection/class-outbox.php
pfefferle: By length 🫣

includes/rest/class-outbox-controller.php
obenland: [wp_errors should work](https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/rest-api/class-wp-rest-request.php#L913-L922)—what did you experience?

obenland (approved):

PR #1335, opened 2025-02-18
Move: Support incoming Move Activities

pfefferle:
This is the first PR to fully support Mastodons `Move` feature.

A `Move` Activity is normally used if a user changes instances. It forces the instances of all followers to update the user and point to the new instance.

## Proposed changes:

* Improve profile update activites
* Support incoming `Move` activites and update the follower

### Other information:

## Testing instructions:

This is a bit hard to test and I would recommend to check the code and the unit tests, but you can give it a try with the following prerequisites.

* You need a Follower that started a migration
* You have to send a `Move` Activity with the `Actor` as source `Person` and the `Object` as target `Person`.
* You can check your followers list if it was done properly


includes/handler/class-move.php
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\Handler;
+
+use Activitypub\Http;
+use Activitypub\Collection\Followers;
+
+use function Activitypub\is_actor;
```
obenland:
```suggestion
```
Looks like it's unused

includes/handler/class-move.php
```diff
@@ -0,0 +1,192 @@
...
+		$origin_object = Http::get_remote_object( $origin );
+
+		$verified = self::verify_move( $target_object, $origin_object );
+
+		if ( ! $verified ) {
+			return;
+		}
+
+		$target_follower = Followers::get_follower_by_actor( $target, false );
+		$origin_follower = Followers::get_follower_by_actor( $origin, false );
```
obenland:
```suggestion
		$target_follower = Followers::get_follower_by_actor( $target );
		$origin_follower = Followers::get_follower_by_actor( $origin );
```
`get_follower_by_actor()` doesn't have a second arg.

includes/handler/class-move.php
```diff
@@ -0,0 +1,192 @@
...
+			// Clear the cache.
+			wp_cache_delete( $origin_id, 'posts' );
+			return;
+		}
+
+		// If the new target is followed, but the origin is not,
+		// everything is fine, so we can return.
+		if ( $target_follower && ! $origin_follower ) {
+			return;
+		}
```
obenland: Should we check this first and bail early?

includes/handler/class-move.php
```diff
@@ -0,0 +1,192 @@
...
+
+			// Get all user ids from $origin_users that are not in $target_users.
+			$users = \array_diff( $origin_users, $target_users );
+
+			foreach ( $users as $user_id ) {
+				\add_post_meta( $target_follower->get__id(), '_activitypub_user_id', $user_id );
+			}
+
+			$origin_follower->delete();
+			return;
```
obenland:
```suggestion
```
Not needed

includes/handler/class-move.php
```diff
@@ -0,0 +1,192 @@
...
+		}
+	}
+
+	/**
+	 * Extract the target from the activity.
+	 *
+	 * The ActivityStreams spec devine the `target` attribute as the
```
obenland:
```suggestion
	 * The ActivityStreams spec define the `target` attribute as the
```

includes/handler/class-move.php
```diff
@@ -0,0 +1,192 @@
...
+			return object_to_uri( $activity['object'] );
+		}
+
+		return null;
+	}
+
+	/**
+	 * Extract the origin from the activity.
+	 *
+	 * The ActivityStreams spec devine the `origin` attribute as source
```
obenland:
```suggestion
	 * The ActivityStreams spec define the `origin` attribute as source
```

includes/handler/class-move.php
```diff
@@ -0,0 +1,192 @@
...
+
+		if ( ! $verified ) {
+			return;
+		}
+
+		$target_follower = Followers::get_follower_by_actor( $target, false );
+		$origin_follower = Followers::get_follower_by_actor( $origin, false );
+
+		// If the new target is not followed, but the origin is,
+		// update the origin follower to the new target.
```
obenland: Let's update multiline comments to follow [WP docs guidelines](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-inline-comments).

includes/handler/class-update.php
```diff
@@ -7,6 +7,8 @@
 
 namespace Activitypub\Handler;
 
+use Activitypub\Model\Follower;
```
```suggestion
```
Appears unused

phpcs.xml
```diff
@@ -1,8 +1,9 @@
 <?xml version="1.0"?>
 <ruleset name="WordPress ActivityPub">
 	<description>WordPress ActivityPub Standards</description>
-	<file>.</file>
 
+	<file>.</file>
+	<exclude-pattern>.(git|github|vscode|idea|wordpress-org)</exclude-pattern>
 	<exclude-pattern>*\.(inc|css|js|svg)</exclude-pattern>
```
obenland, line 7: Could these be merged?

tests/includes/handler/class-test-move.php
```diff
@@ -0,0 +1,257 @@
...
+		// Mock the HTTP request to return an error.
+		add_filter(
+			'pre_http_request',
+			function () {
+				return new \WP_Error( 'http_request_failed', 'Invalid request' );
+			}
+		);
+
+		// Should return without error.
+		$this->assertNull( Move::handle_move( $activity ) );
```
obenland: `handle_move()` never returns anything, so this might not be the best thing to check. Maybe verifying that it didn't change the original followers guid instead?

tests/includes/handler/class-test-move.php
```diff
@@ -0,0 +1,257 @@
...
+
+	/**
+	 * Test the handle_move method without a target or origin.
+	 */
+	public function test_handle_move_without_target_or_origin() {
+		$activity = array(
+			'type' => 'Move',
+		);
+
+		$this->assertNull( Move::handle_move( $activity ) );
```
obenland: same as above

tests/includes/handler/class-test-move.php
```diff
@@ -0,0 +1,257 @@
...
+		$origin_follower->set_url( $origin );
+		$id = $origin_follower->upsert();
+
+		// Add the user ID meta value.
+		add_post_meta( $id, '_activitypub_user_id', $this->user_id );
+
+		// Mock the HTTP request.
+		add_filter(
+			'pre_http_request',
+			function ( $preempt, $args, $url ) use ( $target, $target_object, $origin, $origin_object ) {
```
obenland: in `class-test-update.php` these `pre_http_request` callbacks get removed, let's do the same here

obenland (changes requested):

phpcs.xml
pfefferle: I am not sure that this is a good idea. The first line ignores dotfiles, the second one file extensions.

tests/includes/handler/class-test-move.php
pfefferle: ~I agree that this test makes no sense, but checking existing followers is quite complex. I think I will test the verify method instead.~ AI FTW!

obenland (approved):

obenland:
Nice work!

PR #1336, opened 2025-02-18
Outbox: Retry failed Follower notifications

obenland:
## Proposed changes:

* Adds retry mechanism for inboxes that returned an error.
* Increases the time between retries based on the number of attempts it's on.
* Only retries inboxes that responded with error codes indicating temporary issues.

### Other information:

## Testing instructions:

* Add a snippet to your site that fakes a qualifying response:
```php
add_filter(
	'pre_http_request',
	function () {
		return new WP_Error( '503', 'Service Unavailable' );
	}
);
```
* Create an Outbox item through a qualifying event, like updating a post.
* Check the cron event queue for a `activitypub_async_batch` event with a callback of `retry_send_to_followers`: `wp cron event list --fields=hook,args`


includes/class-dispatcher.php
```diff
@@ -169,6 +160,76 @@ public static function send_to_followers( $outbox_item_id, $batch_size = 50, $of
...
+	private static function send_to_inboxes( $inboxes, $outbox_item_id ) {
+		$json    = Outbox::get_activity( $outbox_item_id )->to_json();
+		$actor   = Outbox::get_actor( \get_post( $outbox_item_id ) );
+		$retries = array();
+
+		foreach ( $inboxes as $inbox ) {
+			$result = safe_remote_post( $inbox, $json, $actor->get__id() );
+
+			if ( is_wp_error( $result ) ) {
+				$retries[] = $inbox;
```
pfefferle, line 211: should we only retry certain http error codes?

pfefferle: https://www.restapitutorial.com/advanced/responses/retries

obenland: Yes!

obenland: Fixed in [5585740](https://github.com/Automattic/wordpress-activitypub/pull/1336/commits/55857407aa55bb9b6ca3b377e3591ebabe69a221)

includes/class-dispatcher.php
```diff
@@ -169,11 +168,81 @@ public static function send_to_followers( $outbox_item_id, $batch_size = 50, $of
...
+	/**
+	 * Schedule a retry.
+	 *
+	 * @param array $retries        The inboxes to retry.
+	 * @param int   $outbox_item_id The Outbox item ID.
+	 * @param int   $attempt        Optional. The attempt number. Default 1.
+	 */
+	private static function schedule_retry( $retries, $outbox_item_id, $attempt = 1 ) {
+		\wp_schedule_single_event(
+			\time() + ( $attempt * 5 * MINUTE_IN_SECONDS ),
```
pfefferle: can we use something that increases the value exponentially? so that the last retry might be in an hour or later?

includes/class-dispatcher.php
```diff
@@ -169,11 +168,81 @@ public static function send_to_followers( $outbox_item_id, $batch_size = 50, $of
...
+	private static function send_to_inboxes( $inboxes, $outbox_item_id ) {
+		$json    = Outbox::get_activity( $outbox_item_id )->to_json();
+		$actor   = Outbox::get_actor( \get_post( $outbox_item_id ) );
+		$retries = array();
+
+		foreach ( $inboxes as $inbox ) {
+			$result = safe_remote_post( $inbox, $json, $actor->get__id() );
+
+			if ( is_wp_error( $result ) && in_array( $result->get_error_code(), self::$retry_error_codes, true ) ) {
+				$retries[] = $inbox;
```
pfefferle, line 211: What do we do with the other errors? It would be nice to somehow clean up profiles that results in errors!?!

See `Followers::add_error()` and/or https://github.com/Automattic/wordpress-activitypub/blob/ef24c8bf375dbacc19c7fe505da94239b62abc9d/includes/class-scheduler.php#L127

pfefferle: (Might not be the best cleanup method, but we can use that PR to improve it)

obenland: Yeah, not sure. `Followers::add_error()` looks promising, but I'd worry about tons of db writes when we loop through inboxes.
On the other hand, how many of these do we expect to happen at all?

includes/class-dispatcher.php
obenland: Sure. It doesn't really make a huge difference with only two retries, fwiw. But we could change the minute to hours and retry after 5 and then another 10 hours

includes/class-dispatcher.php
obenland: I'm actually not sure it's super easy to do that. Since we're only looking at inboxes, not followers, it might not be possible to track that back to an individual follower, for example if the use a shared inbox.

Maybe this is a situation where we shouldn't let perfection get in the way of progress?

includes/class-dispatcher.php
obenland: Fixed in [f73e764](https://github.com/Automattic/wordpress-activitypub/pull/1336/commits/f73e76436e83a3edfd4c8c681cb740e157bee65e)

pfefferle:
The retries seem to not work for `interactees`, but these might be the most important ones to receive the message!

pfefferle:
We should at least file an issue, that we have to work on error handling. If we have several servers that bounce our messages, it might slow down the whole precess if we re-try them three or more times. The same for the error codes that does not qualify for a re-try, what should we do with them.

mattwiebe (dismissed):
Very cool change! I like the robustness we're building towards here. 💪

This worked well and as expected in the PR description in my testing. Some notes:

1. The default retry level of 3 was respected, but since the retries began on index = 1, there were 2 subsequent retry levels. After this, no new retries were spawned.
2. Implementing exponential backoff to avoid overwhelming servers that might be having issues is A+.

obenland:
> The retries seem to not work for interactees, but these might be the most important ones to receive the message!

Fixed in [5acd235](https://github.com/Automattic/wordpress-activitypub/pull/1336/commits/5acd235a8d6f162c86b24af77f1ee53f4cfc698f)

obenland:
> The default retry level of 3 was respected, but since the retries began on index = 1, there were 2 subsequent retry levels. After this, no new retries were spawned.

The initial attempt gets counted attempt 1 but I don't have strong feelings on that. Do you have a preference?

obenland:
> Implementing exponential backoff to avoid overwhelming servers that might be having issues is A+.

Introduced in [f73e764](https://github.com/Automattic/wordpress-activitypub/pull/1336/commits/f73e76436e83a3edfd4c8c681cb740e157bee65e)

mattwiebe:
One note I didn't want in my approval was that I'd like to see a change that makes the `activitypub_async_batch` args signature a bit leaner. Eg my `args` for one local `activitypub_async_batch ` event instance with one inbox was:

```
[["Activitypub\\Dispatcher","retry_send_to_followers"],["https:\/\/mastodon.social\/inbox"],6169,2]
```

I would like to see a future enhancement that makes these args leaner somehow. A few inboxes to retry, and a moderately large string for the method we intend to call… We've seen our `cron` option blow up on dotcom before, hopefully that's a thing of the past! Perhaps saving as a `_activitypub_retry_inboxes` meta and using a keyword instead of a function name, that would reduce it to something like `['retry_followers', 6169, 2]`.

Also for the security paranoid, I don't _love_ storing the method to be called like that. A little too close to an attack injection surface.

pfefferle:
@mattwiebe do you think this should be a requirement to merge or nice for a follow up?

obenland:
Agreed, passing the list of inbox URLs is not great. For the most part I've been thinking that it's not going to be tons (famous last words), but there's no guarantee

mattwiebe:
I think it's fine to merge as is! 

And for a followup as we likely onboard more processes to the `activitypub_async_batch` infrastructure. But it can be hard to change living code! The previous signature will already still need to be supported across later enhancements while possibly-hours-in-the-future jobs wait their turn.

obenland:
@mattwiebe I updated the PR to save the inbox array in a transient, and only pass the transient key in the event args

obenland:
> Also for the security paranoid, I don't love storing the method to be called like that. A little too close to an attack injection surface.

With the event hook being available to hook into ahead of our callback I'm not sure I'd consider this a security issue. To be on the safe side though, I added an allowlist and a unit test to verify that only registered callbacks get executed.

mattwiebe:
> @mattwiebe I updated the PR to save the inbox array in a transient, and only pass the transient key in the event args

@obenland I like this! It's also a nice way to envision encapsulating arbitrary data between jobs. We could store the `retry` in there as well, other things like we see in `data` passed to jobs on wpcom. 

>  I'm not sure I'd consider this a security issue

Me either! Anybody able to hook into that level of things has already compromised your system, probably. Probably. Still the allowlist is a nice touch and maybe callbacks or keywords could also move into the transient, again more like we do things in adding jobs on wpcom.

mattwiebe (approved):
This worked as expected in my testing, and I expect it to do the same as running code.  :shipit:

PR #1338, opened 2025-02-18
Outbox: Don't federate Updates to unfederated posts.

obenland:
This will require some scrutiny. For example, this ignores changing content visibility in the Editor for old posts.
Not quite sure how to account for that in the callback, since post meta gets updated before `transition_post_status` runs and it's impossible to differentiate between a post with public visibility and one that doesn't have the post meta.

## Proposed changes:

* Bail before creating an Outbox item for update activities of unfederated posts.

### Other information:

## Testing instructions:

* Import posts into your test site (they won't get Outbox items created for them or marked as federated).
* Make an update to one of the imported posts.
* Check the Outbox and make sure no Update activity was created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * Outbox now precesses the first batch of followers right away to avoid delays in processing new Activities.
+* Updates to unfederated posts no longer trigger a new Activity.
 
 ### Fixed
 
--- a/includes/scheduler/class-post.php
+++ b/includes/scheduler/class-post.php
@@ -99,6 +99,11 @@ class Post {
 			return;
 		}
 
+		// If the post is not federated, do not send Updates.
+		if ( 'Update' === $type && 'federated' !== get_wp_object_state( $post ) ) {
+			return;
+		}
+
 		// Add the post to the outbox.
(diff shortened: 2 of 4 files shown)


pfefferle:
Should we add an "older than x month" to the check?

obenland:
What would be the benefit of that?

pfefferle:
To prevent that an up-to-date post won't be federated when the `Create` get's "lost".

obenland:
Interesting, I've not considered that. We don't currently check for an existing Create, is that even applicable?

pfefferle:
@obenland because all instances I tested so far, support `Update`s as a first Activity and handle them the same way as a `Create`, there was no need to have such a check.

pfefferle:
This is btw. nothing that can only happen to posts... we should also have a look at comments...

obenland:
Are those consuming instances or instances where there's a mix of federated and unfederated content? I'm trying to understand how that relates?

With this change we might want to introduce a way to explicitly federate a previously unfederated post. Could be an action link and/or bulk edit action in the posts list or something. I'd also be curious whether there's a way to determine whether a unfederated post had its activitypub visibility set to public, to change it that way

pfefferle:
the problem is, that we decided to have "public" the default, so we do not store anything in the metas when a post should be federate as public! we can change that though.

but the side effect of that change would be, that posts that are written using a third party app wouldn't be federated.

PR #1343, opened 2025-02-19
Migration: Re-use async batch infrastructure

obenland:
## Proposed changes:

* Replace hardcoded allowlist in Scheduler with extensible registration system
* Add `Scheduler::register_async_batch_callback()` method for dynamic callback registration  
* Remove hardcoded Migration callbacks from Scheduler class
* Update Migration class to register its callbacks through new registration system
* Add safety check to ensure callbacks are registered at proper time
* Maintain full backward compatibility - `async_batch` works exactly the same way

### Other information:

## Testing instructions:

* Run the existing test suite to verify backward compatibility:
  ```bash
  npm run env-test -- --filter Test_Scheduler
  ```
* All 12 tests should pass, confirming async batch functionality works as before
* Test that Migration callbacks are properly registered and functional:
  ```bash
  npm run env-test -- --filter test_update_comment_counts_with_existing_valid_lock
  npm run env-test -- --filter test_async_upgrade
  ```

### Usage for other plugins/classes:

Any class can now register async batch jobs without modifying core files:

```php
class My_Custom_Jobs {
        // Must be called before or during 'init' action
        Scheduler::register_async_batch_callback( 'my_custom_hook', array( self::class, 'my_batch_method' ) );
    }
    
    public static function my_batch_method( $batch_size = 50, $offset = 0 ) {
        // Your batch processing logic
        // Return array for next batch or null when done
    }
}
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add extensible async batch callback registration system to replace hardcoded allowlist

</details>


pfefferle:
@obenland what about this PR?

obenland:
Depends on #1521.

obenland:
@pfefferle This should be ready for review

obenland:
After team call: Let's make it possible to fill the Scheduler from `class-migrations.php`.

Meaning, how can we expand the batch callback allow list from the migration class.
Also move action hooks back into the migration class.
Could also be done in Migration::async_migration.

includes/class-migration.php
```diff
@@ -24,11 +24,12 @@ class Migration {
 	 * Initialize the class, registering WordPress hooks.
 	 */
 	public static function init() {
-		\add_action( 'activitypub_migrate', array( self::class, 'async_migration' ) );
-		\add_action( 'activitypub_upgrade', array( self::class, 'async_upgrade' ), 10, 99 );
-		\add_action( 'activitypub_update_comment_counts', array( self::class, 'update_comment_counts' ), 10, 2 );
-
 		self::maybe_migrate();
+
+		Scheduler::register_async_batch_callback( 'activitypub_migrate_from_0_17', array( self::class, 'migrate_from_0_17' ) );
```
pfefferle, line 29: this still feels overly complicated. you have to trigger the schedule, register the callback and implement the migration 🤔

pfefferle: why not re-use the `activitypub_migrate` logic? one hook, that is triggered with the DB version and responsible for running the right migration?

includes/class-migration.php
```diff
@@ -231,49 +232,6 @@ public static function maybe_migrate() {
 		self::unlock();
 	}
 
-	/**
-	 * Asynchronously migrates the database structure.
-	 *
-	 * @param string $version_from_db The version from which to migrate.
-	 */
-	public static function async_migration( $version_from_db ) {
```
pfefferle, line 239: why not re-use this function and add all the other migrations? this way we would simply have to extend this function the same way we do with `maybe_migrate`!?

includes/class-migration.php
obenland: It doesn't support batching, it fragments migration logic from `maybe_migrate()`, it's a single job running multiple migrations, it's ambiguous as to what will be executed when looking at the list of scheduled events, and it's not been used since 1.0.0.

With the changes proposed here, we'll re-use a common batch infrastructure, every job is clearly named with an individual action hook, there's a single place responsible for what runs when (`maybe_migrate()`), and jobs can run in separate requests.

obenland: @pfefferle Any more thoughts on this?

includes/class-migration.php
```diff
@@ -515,25 +473,11 @@ public static function migrate_to_4_7_2() {
 	 * @see Comment::pre_wp_update_comment_count_now()
 	 * @param int $batch_size Optional. Number of posts to process per batch. Default 100.
 	 * @param int $offset     Optional. Number of posts to skip. Default 0.
+	 * @return int[]|void Array with batch size and offset if there are more posts to process.
```
pfefferle, line 477:
```suggestion
	 *
	 * @return int[]|void Array with batch size and offset if there are more posts to process.
```

includes/class-migration.php
pfefferle: Let’s move forward with it and fine-tune as needed!

pfefferle (approved):

PR #1344, opened 2025-02-19
Settings: Move help tabs to settings class

obenland:
## Proposed changes:

* Consolidates help tab registration in Settings class.
* Minor improvements to make it easier for translators and RTL translations.

## Testing instructions:

* Go to Settings > Activitypub.
* Access Help in the top right corner and make sure everything looks like it did before.


pfefferle:
it is already a mix of PHP and HTML and even the help.php was no proper template, so feel free to migrate it into the class! then it is at least consistent with the rest!

pfefferle:
there seems to be a conflict with the help.php file!?!

obenland:
No more!

pfefferle (approved):

PR #1351, opened 2025-02-21
Fix Announce Activities

pfefferle:
It seems that the `Announce` Activities are sent to early or mixed in between of the original Activities, so they might not be handled properly.

This PR adds an offset of 30 sec for the `Announce`, to send them after the initial posts should be federated. Happy to postpone the `Announce`s even more.

To test it manually, I also added a CLI command to reschedule Outbox items.

Should Fix: https://wordpress.org/support/topic/no-boosts-and-likes-possible/

## Proposed changes:

* Add Cli Command to reschedule Outbox-Items
* Adds a delay to the `Announce` of the Group-Actor

### Other information:

## Testing instructions:

* Simply publish a post and check if the federated post will be Announced by the Blog-Author.

### DIFF
--- a/tests/includes/class-test-scheduler.php
+++ b/tests/includes/class-test-scheduler.php
@@ -187,7 +187,7 @@ class Test_Scheduler extends WP_UnitTestCase {
 		Scheduler::reprocess_outbox();
 
 		// Verify scheduling time.
-		$this->assertGreaterThan( 0, $scheduled_time, 'Event should be scheduled with a future timestamp' );
+		$this->assertSame( $scheduled_time, wp_next_scheduled( 'activitypub_process_outbox', array( $pending_id ) ) );
 
 		// Clean up.
 		wp_delete_post( $pending_id, true );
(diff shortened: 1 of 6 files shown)


tests/includes/class-test-scheduler.php
```diff
@@ -188,7 +188,6 @@ function ( $event ) use ( &$scheduled_time ) {
 
 		// Verify scheduling time.
 		$this->assertGreaterThan( 0, $scheduled_time, 'Event should be scheduled with a future timestamp' );
-		$this->assertGreaterThanOrEqual( time() + 10, $scheduled_time, 'Event should be scheduled at least 10 seconds in the future' );
```
obenland: These tests are both not overly useful. Maybe we should just check that the event exists.

```php
$this->assertSame( $scheduled_time, wp_next_scheduled( 'activitypub_process_outbox', array( $pending_id ) ) );
```

obenland (approved):

PR #1357, opened 2025-02-23
Improve: ActivityPub query-var check

pfefferle:
If the `is_activitypub_request` function is called before the `query_vars` are set, the query singleton stores that the request is not an AP request, even if the URL has the `?activitypub` query var.

This PR implements a query parsing as fallback.

See also: https://github.com/epiphyt/embed-privacy/issues/253

### Other information:

## Testing instructions:

* Go to '..'
*


includes/class-query.php
```diff
@@ -245,10 +269,8 @@ public function is_activitypub_request() {
 			return $this->is_activitypub_request;
 		}
 
-		global $wp_query;
-
 		// One can trigger an ActivityPub request by adding ?activitypub to the URL.
-		if ( isset( $wp_query->query_vars['activitypub'] ) ) {
+		if ( $this->has_query_var() ) {
```
obenland: Could we check for `$_GET['activitypub']` instead?

obenland:
Is it possible to unit test this change?
I understand sometimes it's not, when it comes to getting request execution timing right, I was just curious.

includes/class-query.php
pfefferle: Oh my god 🫣

obenland:
What's a scenario where `is_activitypub_request()` is called before the query_vars are set?

pfefferle:
I linked the issue in the description of the pr!

embed-privacy did exactly that.

obenland (approved):

PR #1358, opened 2025-02-24
Rewrite rules: be more specific in pattern to get author redirects

jeherve:
## Proposed changes:

Some site owners may want to use a "@author" pattern in their site permalinks, e.g. `/@%author%/%postname%/`, to get permalinks that look similar to other Fediverse tools like Mastodon.
ActivityPub's rewrite rules are currently a bit too loose to allow that, as they would force a redirect for those posts as well. I think we can affort to be a bit more strict.

### Other information:

## Testing instructions:

* Go to Settings > Permalinks
* Set your custom permalinks to `/@%author%/%postname%/`
* Save your changes.
* Load one of your posts.
    * It should load properly.
* Now load `yoursite.com/@authorslug`
    * It should redirect to the author page.


pfefferle (approved):

obenland:
Do we need to add an upgrade routine to flush permalinks for that?

pfefferle:
Oh, very good point! yes we would have to!

PR #1359, opened 2025-02-24
Improve outbox scheduling of (custom) posts

### LINKED ISSUE #1269: Race conditions e.g. on post creation: Post not "ready" when added to outbox.

pfefferle:
This PR removes the usage of `transition_post_status` hook in favour of `wp_after_insert_post`. The problem with `transition_post_status` is, that most meta informations of a post are not saved this early in the queue, when the hook is triggered.

From the docs of `wp_after_insert_post`:

> Fires once a post, its terms and meta data has been saved.

So it is literally the last `action` that is triggered in the process of CRUDing a post, either from the Classic Editor or the Block Editor. This should solve most of the race condition problems we currently have.

Fixes #1269 and #1270

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


obenland (approved):
You made it, you found a way! :)
Nice work

pfefferle:
Hopefully :)

PR #1362, opened 2025-02-24
Move: General method to move accounts

obenland:
First pass

Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-move.php
+++ b/includes/class-move.php
@@ -7,6 +7,8 @@
 
 namespace Activitypub;
 
+use Activitypub\Collection\Actors;
+
 /**
  * ActivityPub (Account) Move Class
  *
@@ -48,5 +50,41 @@ class Move {
 		}
 
 		return $actor;
+	}
+
+	/**
+	 * Move an ActivityPub account from one location to another.
+	 *
+	 * @param string $from The current account URL.
+	 * @param string $to   The new account URL.
+	 * @return int|bool The ID of the outbox item or false on failure.
+	 */
+	public static function account( $from, $to ) {
+		$user = Actors::get_by_resource( $from );
+
+		if ( ! $user ) {
+			return false;
+		}
+
+		// Update the movedTo property.
+		\update_user_option( $user->get__id(), 'activitypub_move_to', $to );
+
+		// Add the old account URL to alsoKnownAs.
+		$also_known_as   = (array) \get_user_option( 'activitypub_also_known_as', $user->get__id() );
+		$also_known_as[] = $from;
+		$also_known_as   = array_unique( $also_known_as );
+		\update_user_option( $user->get__id(), 'activitypub_also_known_as', $also_known_as );
+
+		// Create Move activity.
+		$activity = array(
+			'@context' => get_context(),
+			'type'     => 'Move',
+			'actor'    => $from,
+			'object'   => $from,
+			'target'   => $to,
+		);
+
+		// Add to outbox.
+		return add_to_outbox( $activity, 'Move', $user->get__id() );
 	}
 }
--- a/includes/class-move.php
+++ b/includes/class-move.php
@@ -62,7 +62,7 @@ class Move {
 	public static function account( $from, $to ) {
 		$user = Actors::get_by_resource( $from );
 
-		if ( ! $user ) {
+		if ( ! $user || is_wp_error( $user ) ) {
 			return false;
 		}
 
--- a/includes/class-move.php
+++ b/includes/class-move.php
@@ -57,13 +57,14 @@ class Move {
 	 *
 	 * @param string $from The current account URL.
 	 * @param string $to   The new account URL.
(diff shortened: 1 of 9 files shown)


obenland:
Is this kind of the direction we're thinking?

pfefferle:
I would use something more user editable for the also-known-as storage like URLs separated by new lines!?!

pfefferle:
And be aware that it has to also work for the blog user

includes/class-move.php
```diff
@@ -49,4 +51,40 @@ public static function extend_actor_profiles( $actor, $id, $user ) {
...
+		\update_user_option( $user->get__id(), 'activitypub_move_to', $to );
+
+		// Add the old account URL to alsoKnownAs.
+		$also_known_as   = (array) \get_user_option( 'activitypub_also_known_as', $user->get__id() );
+		$also_known_as[] = $from;
+		$also_known_as   = array_unique( $also_known_as );
+		\update_user_option( $user->get__id(), 'activitypub_also_known_as', $also_known_as );
+
+		// Create Move activity.
+		$activity = array(
```
pfefferle: we have a problem! if we use it this way, the outbox handler would set the activity as object!

This way it would look like:

```JSON
{
	"type": "Move",
	"object": {
		"type": "Move",
		"actor": "...",
		"object": "...",
		"target": "..."
	}
}
```

pfefferle:
There are two options to fix that:

1. Add the ability to store full Activities instead of "only" store Activity-Objects: https://github.com/Automattic/wordpress-activitypub/pull/1362/commits/0529a29af070fd26d7807bf37af58d970442f9f8
2. Or restructure the code to fit the current structure: https://github.com/Automattic/wordpress-activitypub/pull/1362/commits/1e6b097dcb14cc5e39b7a2e9563f7615619a7859

(I prefer the second option, because it does not require a structural change and we have to check the `also-known-as` field anyway!)

obenland:
Interesting conundrum! I'm leaning towards option 2.

Maybe we could combine that with a filter in `Outbox::get_activity()` that we can hook into to make these type-specific changes?
```php
		// Use simple Object (only ID-URI) for Like and Announce.
		if ( in_array( $type, array( 'Like', 'Delete', 'Move' ), true ) ) {
			$activity->set_object( $activity->get_object()->get_id() );
		}

		// Set origin and target for Move.
		if ( 'Move' === $type ) {
			$activity->set_origin( $activity->get_actor() );
			$activity->set_target( $activity->get_object() );
		}
```

pfefferle:
I love it! Had the exact same idea (adding a filter) today because of Announces. They seem to still not work properly!

includes/handler/class-delete.php
```diff
@@ -193,4 +179,18 @@ public static function defer_signature_verification( $defer, $request ) {
...
+
+	/**
+	 * Set the object to the object ID.
+	 *
+	 * @param \Activitypub\Activity\Activity $activity The Activity object.
+	 * @return \Activitypub\Activity\Activity The filtered Activity object.
+	 */
+	public static function outbox_activity( $activity ) {
+		if ( 'Delete' === $activity->get_type() ) {
+			$activity->set_object( $activity->get_object()->get_id() );
```
pfefferle: we should ensure that it is an object before we try to get the ID, now that it is a filter!

includes/handler/class-like.php
```diff
@@ -67,4 +63,18 @@ public static function handle_like( $like, $user_id ) {
...
+
+	/**
+	 * Set the object to the object ID.
+	 *
+	 * @param \Activitypub\Activity\Activity $activity The Activity object.
+	 * @return \Activitypub\Activity\Activity The filtered Activity object.
+	 */
+	public static function outbox_activity( $activity ) {
+		if ( 'Like' === $activity->get_type() ) {
+			$activity->set_object( $activity->get_object()->get_id() );
```
pfefferle: same as above!

includes/handler/class-move.php
```diff
@@ -101,6 +102,22 @@ public static function handle_move( $activity ) {
...
 
+	/**
+	 * Convert the object and origin to the correct format.
+	 *
+	 * @param \Activitypub\Activity\Activity $activity The Activity object.
+	 * @return \Activitypub\Activity\Activity The filtered Activity object.
+	 */
+	public static function outbox_activity( $activity ) {
+		if ( 'Move' === $activity->get_type() ) {
+			$activity->set_object( $activity->get_object()->get_id() );
```
pfefferle: same as above!

pfefferle:
Are the handlers the right place to implement this? they are currently not really involved in that process and only responsible for incoming things!

(just asking, because I am not sure about it)

obenland:
I think so! Can't think of a better place

pfefferle:
Should be a nice start to merge into the feature branch!

pfefferle (approved):

PR #1363, opened 2025-02-24
Docs: Improve type hints

obenland:
## Proposed changes:

* Update type hints for improved code traversal

Should not include functional changes.

### DIFF
--- a/includes/activity/class-activity.php
+++ b/includes/activity/class-activity.php
@@ -20,41 +20,6 @@ use function Activitypub\is_activity;
  *
  * @see https://www.w3.org/TR/activitystreams-core/#activities
  * @see https://www.w3.org/TR/activitystreams-core/#intransitiveactivities
- *
- * @method string|array|null       get_actor()          Gets one or more entities that performed or are expected to perform the activity.
- * @method string|null             get_id()             Gets the object's unique global identifier.
- * @method string                  get_type()           Gets the type of the object.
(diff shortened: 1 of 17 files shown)


includes/activity/class-activity.php
```diff
@@ -156,11 +158,11 @@ class Activity extends Base_Object {
 	 * Set the object and copy Object properties to the Activity.
 	 *
 	 * Any to, bto, cc, bcc, and audience properties specified on the object
-	 * MUST be copied over to the new Create activity by the server.
+	 * MUST be copied over to the new "Create" activity by the server.
 	 *
 	 * @see https://www.w3.org/TR/activitypub/#object-without-create
 	 *
-	 * @param array|string|Base_Object|Link|null $data Activity object.
+	 * @param array|string|Activity|Link|null $data Activity object.
```
pfefferle: It can be both... but a `Base_Object` is more common!

obenland: With `Base_Object` I get warnings for potentially polymorphic calls within the function. Maybe the better way would be to document these magic methods in `Base_Object` (as well)?

pfefferle: what IDE plugin do you use for that?

obenland: no plugin, just phpstorm itself

pfefferle (approved):

PR #1366, opened 2025-02-25
Improve: Always send `Vary` header

pfefferle:
Send Vary header also for HTML requests. This helps with content negotiation and (external) caches.

If we only send it for `application/activity+json` it might add the HTML to the wrong bucket.

Context: https://chat.indieweb.org/dev/2025-02-24#t1740415054664400

/cc @snarfed

### DIFF
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -177,11 +177,6 @@ class Activitypub {
 			if ( ! \headers_sent() ) {
 				// Send 200 status header.
 				\status_header( 200 );
-
-				if ( ACTIVITYPUB_SEND_VARY_HEADER ) {
-					// Send Vary header for Accept header.
-					\header( 'Vary: Accept' );
-				}
 			}
 
 			return $activitypub_template;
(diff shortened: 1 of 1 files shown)


pfefferle:
The question is still: Should we enable it by default? What can possibly go wrong?

obenland (approved):

mattwiebe:
> The question is still: Should we enable it by default? What can possibly go wrong?

I think we should probably enable by default, and move this into a user setting so they can disable it in a debugging process

snarfed:
This is great! Thank you @pfefferle!

As background, the original argument against turning this on by default was that it would expand caches for sites running the AP plugin. Lots of discussion in https://github.com/Automattic/wordpress-activitypub/issues/580 and https://github.com/Automattic/wordpress-activitypub/pull/683. It's a correctness vs cost tradeoff, and I generally came down on the correctness side, but I see the other side's point.

pfefferle:
thanks @mattwiebe and @snarfed ! I will work on that for the next version!

PR #1367, opened 2025-02-25
Post representation: bring back the activitypub_post_locale filter

jeherve:
## Proposed changes:

It was removed in #593, but is still neeeded to customize the language of a given post. The filter is used in the WPML integration of this plugin, for example.

### Other information:

## Testing instructions:

* In a functionality plugin, add the following snippet:

```php
/**
 * Set a post's language in its ActivityPub representation
 * based on post tags.
 * The post tags must match an array of language codes you use on your site (in my example, English, Hungarian, and French).
 * When no tags are found, the post's language uses the default (the blog's language).
 *
 * @param string  $lang    The locale of the post.
 * @param int     $post_id The post ID.
 * @param WP_Post $post    The post object.
 *
 * @return string The filtered locale of the post.
 */
function jeherve_custom_ap_language( $lang, $post_id, $post ) {
	// Get the post's hashtags.
	$post_tags = get_the_tags( $post_id );

	if ( ! empty( $post_tags ) ) {
		// Is there a "en, "hu", or "fr" tag?
		foreach ( $post_tags as $tag ) {
			if ( in_array( $tag->slug, array( 'en', 'fr', 'hu' ), true ) ) {
				$lang = $tag->slug;
				break;
			}
		}
	}

	return $lang;
}
add_filter( 'activitypub_post_locale', 'jeherve_custom_ap_language', 10, 3 );
```

* Write a new post, and add the `fr` tag to that post.
* Publish the post
* View it on the frontend, and add `/activitypub` to the post URL to view its ActivityPub representation.
    * You should see a content map for `fr`.
* Other posts should still use your site's language.


pfefferle:
We moved it to the Base class and made it a bit more general: `activitypub_locale`. Is that too generic?

https://github.com/Automattic/wordpress-activitypub/blob/d8967518876c3f346c16556721b14d1f867b82e8/includes/transformer/class-base.php#L236

jeherve:
We can certainly deprecate `activitypub_post_locale` in favor of `activitypub_locale`, but imo we would have to:

1. Update the WPML integration to use the new filter:
https://github.com/Automattic/wordpress-activitypub/blob/d8967518876c3f346c16556721b14d1f867b82e8/integration/class-wpml.php#L20

2. Keep `activitypub_post_locale` around with a deprecation warning, with `apply_filters_deprecated`:
https://developer.wordpress.org/reference/functions/apply_filters_deprecated/

Would you like me to update my PR with those changes?

pfefferle (approved):

pfefferle:
Makes sense... let's merge it and work on a deprecation then!

jeherve:
@pfefferle Here is #1370 for the deprecation alternative.

PR #1370, opened 2025-02-25
Deprecate activitypub_post_locale in favor of activitypub_locale

jeherve:
Follow-up to #1367

## Proposed changes:

As discussed in #1367, let's deprecate the old filter instead of just adding it back.

### Other information:

## Testing instructions:

```php
/**
 * Set a post's language in its ActivityPub representation
 * based on post tags.
 * The post tags must match an array of language codes you use on your site (in my example, English, Hungarian, and French).
 * When no tags are found, the post's language uses the default (the blog's language).
 *
 * @param string  $lang    The locale of the post.
 * @param int     $post_id The post ID.
 * @param WP_Post $post    The post object.
 *
 * @return string The filtered locale of the post.
 */
function jeherve_custom_ap_language( $lang, $post_id, $post ) {
	// Get the post's hashtags.
	$post_tags = get_the_tags( $post_id );

	if ( ! empty( $post_tags ) ) {
		// Is there a "en, "hu", or "fr" tag?
		foreach ( $post_tags as $tag ) {
			if ( in_array( $tag->slug, array( 'en', 'fr', 'hu' ), true ) ) {
				$lang = $tag->slug;
				break;
			}
		}
	}

	return $lang;
}
add_filter( 'activitypub_post_locale', 'jeherve_custom_ap_language', 10, 3 );
```

* Write a new post, and add the `fr` tag to that post.
* Publish the post
* View it on the frontend, and add `/activitypub` to the post URL to view its ActivityPub representation.
    * You should see a content map for `fr`.
    * In your logs, you should see a deprecation warning.
* Other posts should still use your site's language.


includes/transformer/class-base.php
```diff
@@ -236,6 +236,18 @@ public function to_activity( $type ) {
 	protected function get_locale() {
 		$lang = \strtolower( \strtok( \get_locale(), '_-' ) );
 
+		$lang = apply_filters_deprecated(
```
pfefferle: I think we would have to test if the item is a post before. I think not every item has an ID

pfefferle:
But otherwise I like it!

pfefferle (approved):

PR #1373, opened 2025-02-25
Followers: Fix backslash breaking actor json

obenland:
When a Follower has a backslash in their description, it didn't get escaped and caused an error when decoding it.

See p1740498262448489/1740495261.546209-slack-C04TJ8P900J

## Proposed changes:

* Slashes actor json before returning it in `Follower::get_post_meta_input()`.

### Other information:

## Testing instructions:

* In a Fediverse account outside of your test site, add a backslash in your description.
* Follow your test site.
* Find the actor json for your account and decode it. Make sure it doesn't cause an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Use a later hook for Posts to get published to the Outbox, to get sure all `post_meta`s and `taxonomy`s are set stored properly.
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


obenland:
@pfefferle @mattwiebe Do you anticipate any unintended consequences from this change?

pfefferle (dismissed):

pfefferle:
@obenland do we also need to add unslashes?

obenland:
I see it to be similar to #1212, where it's necessary for `json_decode` to work correctly and where it removes those slashes.

pfefferle (approved):

PR #1374, opened 2025-02-25
Add webfinger as comment author email if available

### LINKED ISSUE #1355: Likes, Reposts, and Comments via ActivityPub: "0 approved" in dashboard
### Quick summary

Can't be certain this is specific to ActivityPub, but when posts get likes, reposts, or comments from ActivityPub actors, those all show as actors with "0 approved" in the comments section of the dashboard. 

This is true even after they have been "approved" and are showing in the "approved" section of comments - nto sure what is supposed to trigger WordPress recognizing a comment author as having previously approved comments 

### Steps to reproduce

Have multiple posts someone in the fediverse has liked, commented on, or reposted. Long after (forever?) their first comment has been approved, later reactions still show as from an author with 0 approved comments. 

No real impact beyond that number - though I guess if you rely on that for approvals (approve comments automatically from authors who have previously had comments approved) this would fail 

### Site owner impact

### Severity

Minor

_No response_

_No response_

obenland:
Fixes #1355.

## Proposed changes:

* Use webfinger account as comment author email.
* Adds upgrade routine to backfill existing comments.
* Adds tests for both.

### Other information:

## Testing instructions:

* Apply this PR to your test site.
* Comment on a federated post from a federated account, like mastodon.
* Verify in comment list that the number of approved comments is accurate.

### DIFF
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -185,6 +185,9 @@ class Migration {
 		if ( \version_compare( $version_from_db, '5.3.0', '<' ) ) {
 			add_action( 'init', 'flush_rewrite_rules', 20 );
 		}
+		if ( \version_compare( $version_from_db, 'unreleased', '<' ) ) {
+			\wp_schedule_single_event( \time(), 'activitypub_upgrade', array( 'update_comment_author_emails' ) );
+		}
 
 		/*
 		 * Add new update routines above this comment. ^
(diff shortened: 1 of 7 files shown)


includes/class-migration.php
```diff
@@ -632,6 +635,57 @@ public static function create_comment_outbox_items( $batch_size = 50, $offset =
...
+			array(
+				'number'     => $batch_size,
+				'offset'     => $offset,
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+				'meta_query' => array(
+					array(
+						'key'   => 'protocol',
+						'value' => 'activitypub',
+					),
+				),
```
pfefferle, line 656: wouldn't it be easier to also check for the empty email field? this way we wouldn't have to care too much about offsets?

pfefferle:
This PR is perfectly fine, just for the sake of completeness:

* I added the spec to the PHPDOC: https://swicg.github.io/activitypub-webfinger/#reverse-discovery.
* There is a way to get the WebFinger resource through the Actor-Profile: https://codeberg.org/fediverse/fep/src/branch/main/fep/2c59/fep-2c59.md Sadly WordPress seems to be the only implementer yet, so no need to support it yet.

obenland:
> There is a way to get the WebFinger resource through the Actor-Profile: [codeberg.org/fediverse/fep/src/branch/main/fep/2c59/fep-2c59.md](https://codeberg.org/fediverse/fep/src/branch/main/fep/2c59/fep-2c59.md)

I actually looked at that during research for this PR! 🎉 
Do we have a function that looks for that information in the actor json?

pfefferle:
No, not yet.

includes/class-migration.php
obenland: That's probably good to add. We'd still have to do the meta query for protocol, yes?

obenland: Looks like [we can't search for empty comment author emails](https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/class-wp-comment-query.php#L757-L759) through `get_comments()`. We'd have to do a direct query.

pfefferle (dismissed):

pfefferle (approved):

PR #1382, opened 2025-02-26
Fix: Reply-Context link is missing

### LINKED ISSUE #138: Does this plugin support the "replies" collection?

### LINKED ISSUE #1381: Federated Reply Block: block is no longer displayed on the frontend
### Quick summary

It looks like starting in version 5.3.0, the Federated Reply link is no longer displayed on the frontend.

### Steps to reproduce

This is the view of a single post, on my site, in 5.2.0. It includes a link, with the u-in-reply-to class, to the post I'm replying to via the Federated reply block:

Image

Now if I switch to 5.3.1 and view the same post again. No more link:

Image

### Site owner impact

### Severity

Minor

No revenue impact

After running a `git bisect`, this appears to have started with #1332

pfefferle:
Fixes #1381

## Proposed changes:
With the new update of the Query, the JSON representation is called before the HTML presentation, so the `activitypub_reply_block` was removed but never re-added.
* Remove filter after JSON is rendered

### Other information:

## Testing instructions:

See #1381

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Followers with backslashes in their descriptions no longer break their actor representation.
+* Remove `activitypub_reply_block` Filter after Activity-JSON is rendered, to not affect the HTML representation.
 
 ## [5.3.1] - 2025-02-26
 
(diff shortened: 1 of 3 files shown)


jeherve (dismissed):
That fixes the problem in my tests. 🚢

jeherve (approved):

PR #1385, opened 2025-02-26
Admin: Move admin-specific callbacks into wp-admin

obenland:
## Proposed changes:

* Removes unused upgrade notice callback and left over upgrade notice.
* Moves tool_box callback to wp-admin.
* Deprecates methods in ActivityPub class.

### Other information:

## Testing instructions:

* Go to Tools section and make sure the bookmarklet is still available.
* Make sure authors don't have access.


includes/class-activitypub.php
```diff
@@ -463,25 +461,11 @@ public static function theme_compat() {
...
-				array(
-					'p'      => array(),
-					'a'      => array( 'href', 'title' ),
-					'strong' => array(),
-					'em'     => array(),
-				)
-			)
-		);
+	public static function plugin_update_message( $data ) { // phpcs:ignore
+		_deprecated_function( __METHOD__, 'unreleased' );
```
obenland: Can these methods be removed right away? Not sure if anyone would expect them to be around.

includes/wp-admin/class-admin.php
```diff
@@ -597,4 +601,11 @@ public static function add_plugin_settings_link( $actions ) {
 
 		return $actions;
 	}
+
+	/**
+	 * Adds metabox on wp-admin/tools.php.
+	 */
+	public static function tool_box() {
+		\load_template( ACTIVITYPUB_PLUGIN_DIR . 'templates/toolbox.php' );
```
obenland, line 622: Tools should only be available [to users with `edit_posts` cap](https://github.com/wordpress/wordpress-develop/blob/6.7.2/src/wp-admin/menu.php#L362) in the first place.

includes/class-activitypub.php
pfefferle: I think we can simply move them without deprecating them

obenland: Done in [1f59a28](https://github.com/Automattic/wordpress-activitypub/pull/1385/commits/1f59a2871c132205bfe4e72cef7d69fbf6a1e006)

pfefferle: Maybe I misread your question here! I thought it was about moving it and keep a deprecated version in the `class-activitypub.php`. I would vote for keeping the function in general, because it provides a nice and simple way to have upgrade notices using the readme! This was not implemented for the specific 1.0.0 only.

The tricky thing with the upgrade notice hook: it is triggered on the code of the old version, before the new version replaces it. So if we ever want to have an upgrade notice again, we would have to implement this a version before!

obenland: Core is looking for it in [the update object](https://github.com/WordPress/WordPress/blob/master/wp-admin/update-core.php#L564-L565), so it should be part of the update check response that's [passed to the hook in the second parameter](https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/update.php#L599-L616). It's kind of hard to test without an actual upgrade notice :)

I'll bring it back and update it. Let's experiment with passing an update notice in 5.4.0.

obenland: Added back in [7242941](https://github.com/Automattic/wordpress-activitypub/pull/1385/commits/7242941a3eefdc167a6970ee80937b6b5df222e8).

This is what an update notice will look like:
Screenshot 2025-02-27 at 1 35 29 PM

We're pretty limited with what we can do markup-wise. The action is run within a `<p>` tag, so we can't use block elements without breaking it.

obenland: To test:
* Download [version 1.3.0 of this plugin](https://wordpress.org/plugins/hide-and-seek-header/advanced/) and install it.
* Update the action hook `'in_plugin_update_message-' . ACTIVITYPUB_PLUGIN_BASENAME` to `'in_plugin_update_message-hide-and-seek-header/hide-and-seek-header.php'`.
* Load the plugin list.

pfefferle (approved):

PR #1389, opened 2025-02-27
Move: Add support for Blog-User

pfefferle:
Fixes #

## Proposed changes:

* Remove filter to extend profiles
* Implement getter
* Add exception for storing informations for the Blog-User

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-cli.php
+++ b/includes/class-cli.php
@@ -199,4 +199,36 @@ class Cli extends WP_CLI_Command {
 
 		WP_CLI::success( 'Rescheduled activity.' );
 	}
+
+	/**
+	 * Move the blog to a new URL.
+	 *
+	 * ## OPTIONS
+	 *
+	 * <from>
+	 *     The current URL of the blog.
+	 *
+	 * <to>
+	 *     The new URL of the blog.
+	 *
+	 * ## EXAMPLES
+	 *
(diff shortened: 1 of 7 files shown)


includes/class-cli.php
```diff
@@ -199,4 +199,36 @@ public function reschedule( $args ) {
...
+	 *
+	 * @synopsis <from> <to>
+	 *
+	 * @param array $args The arguments.
+	 */
+	public function move( $args ) {
+		$from = $args[0];
+		$to   = $args[1];
+
+		$outbox_item_id = Move::from_to( $from, $to );
```
obenland:
```suggestion
		$outbox_item_id = Move::account( $from, $to );
```
Should this be `account()`?

includes/class-move.php
```diff
@@ -69,13 +32,22 @@ public static function account( $from, $to ) {
 		}
 
 		// Update the movedTo property.
-		\update_user_option( $user->get__id(), 'activitypub_move_to', $to );
+		if ( $user->get__id() > 0 ) {
```
obenland, line 35: This is very nit-picky, so feel free to ignore: We could check explicitly for `$user->get__id() === Actors::BLOG_USER_ID` instead.

obenland (approved):

PR #1390, opened 2025-02-27
Improve `/@username` URLs

pfefferle:
With this PR, the `/@username` call will no longer be forwarded to the API and handled like for comments.

This allows us to no redirect API requests and better handle `AUTHORIZED_FETCH`es.

Fixes: If `AUTHORIZED_FETCH` is enabled, the `/@username` all fail because of missing signature, because the API does not check for `Accept` header.

Issues:

* https://wordpress.org/support/topic/link-to-more-posts-shows-as-json-error/

## Proposed changes:

* Add support for optional trailing slash in `/@username` URLs
* No longer redirect URLs to the API

### Other information:

## Testing instructions:

* There should be no change in the handling. Try to load a `/@username` URL and it should redirect to the author URL or the main page if blog user is used.


includes/rest/class-actors-controller.php
```diff
@@ -101,23 +101,15 @@ public function get_item( $request ) {
...
 		 */
 		\do_action( 'activitypub_rest_users_pre' );
 
 		$data = $user->to_array();
 
-		$response = \rest_ensure_response( $data );
+		$response    = \rest_ensure_response( $data );
+		$link_header = \sprintf( '<%1$s>; rel="alternate"; type="application/activity+json"', $user->get_id() );
 		$response->header( 'Content-Type', 'application/activity+json; charset=' . \get_option( 'blog_charset' ) );
 		$response->header( 'Link', $link_header );
```
obenland:
```suggestion
		$response = \rest_ensure_response( $data );
		$response->header( 'Content-Type', 'application/activity+json; charset=' . \get_option( 'blog_charset' ) );
		$response->header( 'Link', \sprintf( '<%1$s>; rel="alternate"; type="application/activity+json"', $user->get_id() ) );
```
Can be moved in line since it's not used anywhere else

includes/class-activitypub.php
```diff
@@ -412,12 +435,7 @@ public static function add_rewrite_rules() {
 			);
 		}
 
-		\add_rewrite_rule(
-			'^@([\w\-\.]+)$',
-			'index.php?rest_route=/' . ACTIVITYPUB_REST_NAMESPACE . '/actors/$matches[1]',
-			'top'
-		);
-
+		\add_rewrite_rule( '^@([\w\-\.]+)\/?$', 'index.php?actor=$matches[1]', 'top' );
```
obenland, line 431: Let's add an upgrade routine to flush permalinks

includes/class-migration.php
```diff
@@ -182,7 +182,7 @@ public static function maybe_migrate() {
 		if ( \version_compare( $version_from_db, '5.2.0', '<' ) ) {
 			Scheduler::register_schedules();
 		}
-		if ( \version_compare( $version_from_db, '5.3.0', '<' ) ) {
+		if ( \version_compare( $version_from_db, 'unreleased', '<' ) ) {
 			add_action( 'init', 'flush_rewrite_rules', 20 );
```
pfefferle: @obenland why can't we call the `flush_rewrite_rules` directly?

obenland: Because `Activitypub::add_rewrite_rules()` [runs after](https://github.com/Automattic/wordpress-activitypub/blob/5.3.2/includes/class-activitypub.php#L42) `maybe_migrate()`. Running it on 20 makes sure the rewrite rules have been registered.

pfefferle: why I ask: https://github.com/Automattic/wordpress-activitypub/issues/1393

includes/class-migration.php
```diff
@@ -182,7 +182,7 @@ public static function maybe_migrate() {
 		if ( \version_compare( $version_from_db, '5.2.0', '<' ) ) {
 			Scheduler::register_schedules();
 		}
-		if ( \version_compare( $version_from_db, '5.3.0', '<' ) ) {
+		if ( \version_compare( $version_from_db, 'unreleased', '<' ) ) {
```
obenland, line 185: I guess this works? Why would you not add it to the existing unreleased check?

obenland:
I get a fatal error for a missing namespace import when trying to load `@admin` on my local install. 
Maybe it would be good to add tests to make sure this does what it's supposed to do.

For non-existent users I now get a 404 instead of a json error message, I assume that's on purpose?
Are the comment-related changes separate from the @username URLs? Or is there something in conjunction that should be tested?

pfefferle:
might be a merge issue

tests/includes/class-test-query.php
```diff
@@ -265,6 +265,29 @@ public function test_comment_activitypub_object() {
 		$this->assertNull( Query::get_instance()->get_activitypub_object() );
 	}
 
+	/**
+	 * Test user at URL activity object.
+	 *
+	 * @covers ::get_activitypub_object
+	 */
+	public function test_user_at_url_activity_object() {
+		$user_id = $this->factory->user->create(
```
obenland:
```suggestion
		$user_id = self::factory()->user->create(
```
Does your IDE have access to wp unit test files to show this as [deprecated](https://github.com/WordPress/wordpress-develop/blob/6.7.2/tests/phpunit/includes/abstract-testcase.php#L29)? `self::factory()` is the intended replacement.

obenland (approved):
In a future PR, maybe we could redirect non-activitypub requests for non-existent users to their author pages (and through that a 404) as well? Low priority I think, this probably doesn't happen too often or at all, just something I noticed while testing.

PR #1395, opened 2025-02-27
Add interaction settings

### LINKED ISSUE #1140: Feature Request: Disable fediverse interactions (likes and reposts) in WP-admin
I've noticed that a recent update on ActivityPub plugin enabled interactions from fediverse to show up in WP, and those likes and reposts appears as comments. These:

Image

A few readers complained that they are getting this interactions as new comments notifications (from Subscribe to Comments Reloaded plugin). Since I don't expose fediverse interactions in my layout/template, I guess maybe adding an option to disable them could be offered?

obenland:
Fixes #1140.

## Proposed changes:

* Adds setting to disable likes and repost.
* Adds override for ACTIVITYPUB_DISABLE_INCOMING_INTERACTIONS constant.
* Runs `Migration::update_comment_counts()` when an interaction type gets disabled.
* Adds unit tests.

#### After
image

### Other information:

## Testing instructions:

* Enable/disable reaction types and attempt to send in a reaction of that type.
* Define the `ACTIVITYPUB_DISABLE_INCOMING_INTERACTIONS` constant and make sure it overrides the options.

### DIFF
--- a/includes/wp-admin/class-settings-fields.php
+++ b/includes/wp-admin/class-settings-fields.php
@@ -507,7 +507,7 @@ class Settings_Fields {
 			<p>
 				<label>
 					<input type="checkbox" name="activitypub_allow_announces" value="1" <?php checked( '1', $allow_reposts ); ?> />
-					<?php esc_html_e( 'Receive reblogs', 'activitypub' ); ?>
+					<?php esc_html_e( 'Receive reblogs (boosts)', 'activitypub' ); ?>
 				</label>
 			</p>
 			<p class="description"><?php esc_html_e( 'Types of interactions from the Fediverse your blog should accept.', 'activitypub' ); ?></p>
--- a/includes/wp-admin/class-settings-fields.php
+++ b/includes/wp-admin/class-settings-fields.php
@@ -68,6 +68,14 @@ class Settings_Fields {
 			'activitypub_activities'
 		);
 
+		add_settings_field(
+			'activitypub_allow_interactions',
+			__( 'Post interactions', 'activitypub' ),
+			array( self::class, 'render_allow_interactions_field' ),
+			'activitypub_settings',
+			'activitypub_activities'
+		);
+
 		$object_type = \get_option( 'activitypub_object_type', ACTIVITYPUB_DEFAULT_OBJECT_TYPE );
 		if ( 'note' === $object_type ) {
 			add_settings_field(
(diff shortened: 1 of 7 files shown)


obenland:
Untested at the moment, but sharing for early feedback about the general approach.

pfefferle:
@obenland I like the change!

I would like to also combine it with the `ACTIVITYPUB_DISABLE_INCOMING_INTERACTIONS` const, the same way I did it for Actor-Mode (adding the const check to `get_option`) https://github.com/Automattic/wordpress-activitypub/pull/1430

includes/wp-admin/class-settings-fields.php
```diff
@@ -75,6 +75,14 @@ public static function register_settings_fields() {
 			'activitypub_activities'
 		);
 
+		add_settings_field(
```
pfefferle: let's also check for `ACTIVITYPUB_DISABLE_INCOMING_INTERACTIONS`, to show or hide the settings.

pfefferle:
@obenland we really should get this merged! this was requested quite some time!

and for sites that have "comment subscriptions" this might really be essential!

obenland:
I can take another look at this soon

obenland:
This should be ready for an initial review!

pfefferle:
I will have a look tomorrow!

includes/wp-admin/class-settings-fields.php
```diff
@@ -478,7 +486,37 @@ public static function render_outbox_purge_days_field() {
...
+			<p>
+				<label>
+					<input type="checkbox" name="activitypub_allow_likes" value="1" <?php checked( '1', $allow_likes ); ?> />
+					<?php esc_html_e( 'Receive likes', 'activitypub' ); ?>
+				</label>
+			</p>
+			<p>
+				<label>
+					<input type="checkbox" name="activitypub_allow_announces" value="1" <?php checked( '1', $allow_reposts ); ?> />
+					<?php esc_html_e( 'Receive reblogs', 'activitypub' ); ?>
```
pfefferle:
```suggestion
					<?php esc_html_e( 'Receive reblogs (also known as boosts)', 'activitypub' ); ?>
```

I like that we use a more general term, but maybe we should mention that Mastodon calls them "boosts".

obenland:
```suggestion
					<?php esc_html_e( 'Receive reblogs (boosts)', 'activitypub' ); ?>
```
Would this provide enough context?

pfefferle (approved):

PR #1396, opened 2025-02-28
Move: Add/Remove/Modify "Account Aliases"

pfefferle:
This allows users to migrate their remote profiles to WordPress.

See: https://docs.joinmastodon.org/user/moving/#migration

## Blog-Profile

Screenshot 2025-02-28 at 11 41 18

## User-Profile

Screenshot 2025-02-28 at 11 51 48

## Proposed changes:

* Add the ability to add Aliases using the settings screens

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-move.php
+++ b/includes/class-move.php
@@ -39,18 +39,10 @@ class Move {
 		}
 
 		// Add the old account URL to alsoKnownAs.
-		$also_known_as   = (array) \get_user_option( 'activitypub_also_known_as', $user->get__id() );
-		$also_known_as[] = $from;
-		$also_known_as   = \array_unique( $also_known_as );
-
 		if ( $user->get__id() > 0 ) {
-			$also_known_as = array_map( 'trim', $also_known_as );
-			$also_known_as = array_map( 'esc_url', $also_known_as );
-			$also_known_as = array_filter( $also_known_as );
-
-			\update_user_option( $user->get__id(), 'activitypub_also_known_as', $also_known_as );
+			self::update_user_also_known_as( $user->get__id(), $from );
 		} else {
-			\update_option( 'activitypub_blog_user_also_known_as', $also_known_as );
+			self::update_blog_also_known_as( $from );
 		}
 
 		$response = Http::get_remote_object( $to );
@@ -70,5 +62,30 @@ class Move {
 
 		// Add to outbox.
 		return add_to_outbox( $actor, 'Move', $user->get__id(), ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC );
+	}
+
+	/**
+	 * Update the alsoKnownAs property of a user.
+	 *
+	 * @param int    $user_id The user ID.
+	 * @param string $from    The current account URL.
+	 */
+	private static function update_user_also_known_as( $user_id, $from ) {
+		$also_known_as   = (array) \get_user_option( 'activitypub_also_known_as', $user_id );
+		$also_known_as[] = $from;
+
+		\update_user_option( $user_id, 'activitypub_also_known_as', $also_known_as );
+	}
+
+	/**
+	 * Update the alsoKnownAs property of the blog.
+	 *
+	 * @param string $from The current account URL.
+	 */
+	private static function update_blog_also_known_as( $from ) {
+		$also_known_as   = (array) \get_option( 'activitypub_blog_user_also_known_as' );
+		$also_known_as[] = $from;
+
+		\update_option( 'activitypub_blog_user_also_known_as', $also_known_as );
 	}
 }
--- a/includes/class-move.php
+++ b/includes/class-move.php
@@ -44,6 +44,10 @@ class Move {
 		$also_known_as   = \array_unique( $also_known_as );
 
 		if ( $user->get__id() > 0 ) {
+			$also_known_as = array_map( 'trim', $also_known_as );
+			$also_known_as = array_map( 'esc_url', $also_known_as );
+			$also_known_as = array_filter( $also_known_as );
+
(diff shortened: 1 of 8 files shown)


includes/wp-admin/class-blog-settings-fields.php
```diff
@@ -224,4 +232,27 @@ public static function extra_fields_callback() {
...
 	}
+
+	/**
+	 * Also Known As field callback.
+	 */
+	public static function also_known_as_callback() {
+		$also_known_as = \get_option( 'activitypub_blog_user_also_known_as' );
+		?>
+		<p class="description">
+			<?php esc_html_e( 'If you’re moving from another account to this one, you’ll need to create an alias here first before transferring your followers. This step is safe, reversible, and doesn’t affect anything on its own. The migration itself is initiated from your old account.', 'activitypub' ); ?>
```
obenland: I ❤️ this help text!

includes/wp-admin/class-settings.php
```diff
@@ -244,6 +244,26 @@ public static function register_settings() {
 				'default'     => null,
 			)
 		);
+
+		\register_setting(
+			'activitypub_blog',
+			'activitypub_blog_user_also_known_as',
+			array(
+				'type'              => 'array',
+				'description'       => \__( 'Also Known As', 'activitypub' ),
```
obenland: Where is that description shown? I actually don't know

includes/wp-admin/class-settings.php
```diff
@@ -244,6 +244,26 @@ public static function register_settings() {
...
+				'default'           => array(),
+				'sanitize_callback' => function ( $value ) {
+					if ( ! is_array( $value ) ) {
+						$value = explode( PHP_EOL, $value );
+					}
+
+					$value = array_filter( array_map( 'trim', $value ) );
+					$value = array_filter( array_map( 'esc_url', $value ) );
```
obenland: Not yet sure whether it should be `esc_url_raw` or not, since we're about to save it (`db` context) and not display it. We could probably move the `array_filter()` to the return line and only do it once?

templates/user-settings.php
```diff
@@ -134,6 +134,26 @@ class="<?php echo \esc_attr( $classes_for_button ); ?>"
 				</p>
 			</td>
 		</tr>
+		<tr scope="row">
+			<th>
+				<label><?php \esc_html_e( 'Account Aliases', 'activitypub' ); ?></label>
```
obenland, line 139: There may be an opportunity to add this to the glossary, unless we're happy with the textarea description.

includes/wp-admin/class-admin.php
```diff
@@ -172,6 +172,17 @@ public static function save_user_settings( $user_id ) {
...
 			\delete_user_option( $user_id, 'activitypub_header_image' );
 		}
+
+		$also_known_as = ! empty( $_POST['activitypub_blog_user_also_known_as'] ) ? \sanitize_textarea_field( wp_unslash( $_POST['activitypub_blog_user_also_known_as'] ) ) : false;
+		if ( $also_known_as ) {
+			$also_known_as = explode( PHP_EOL, $also_known_as );
+			$also_known_as = array_map( 'trim', $also_known_as );
+			$also_known_as = array_filter( $also_known_as );
+			$also_known_as = array_map( 'esc_url', $also_known_as );
+			\update_user_option( $user_id, 'activitypub_also_known_as', $also_known_as );
```
obenland, line 178: If we registered this as user meta, we might be able to combine sanitization from here, class-move, and the settings update.

includes/wp-admin/class-settings.php
pfefferle: Autocomplete from AI

includes/class-move.php
```diff
@@ -71,4 +63,29 @@ public static function account( $from, $to ) {
...
+
+	/**
+	 * Update the alsoKnownAs property of a user.
+	 *
+	 * @param int    $user_id The user ID.
+	 * @param string $from    The current account URL.
+	 */
+	private static function update_user_also_known_as( $user_id, $from ) {
+		$also_known_as   = (array) \get_user_option( 'activitypub_also_known_as', $user_id );
+		$also_known_as[] = $from;
```
pfefferle, line 75: do you plan to register the meta or why have you removed the sanitisation?

obenland: Coming!

obenland: Added in [d17fed4](https://github.com/Automattic/wordpress-activitypub/pull/1396/commits/d17fed4a6ecafd27bee98febbd441a472c5bc4f2)

includes/wp-admin/class-settings.php
```diff
@@ -250,18 +251,9 @@ public static function register_settings() {
...
-					if ( ! is_array( $value ) ) {
-						$value = explode( PHP_EOL, $value );
-					}
-
-					$value = array_map( 'trim', $value );
-					$value = array_map( 'esc_url_raw', $value );
-
-					return array_filter( $value );
-				},
+				'sanitize_callback' => array( Activitypub::class, 'sanitize_textarea_urls' ),
```
pfefferle: should we have a global function for that? not sure if this fits best in the ActivityPub class?

obenland: Yeah, I've been wondering about that, too. I'm not a huge fan of stuffing more things into `functions.php`.  Maybe we could add a class for that? Maybe something that could also cover the register_post_type/meta stuff?

obenland (approved):
I think this is good to go!
We'll get unit tests for the sanitization function in #1397.

Tested adding and deleting AKA data and it works as expected

PR #1397, opened 2025-02-28
Combine sanitization functions

obenland:
Creates a central place for sanitization functions used in settings, meta, etc.

## Proposed changes:

* Adds `url_list()` sanitization to sanitize a text area list of urls.
* Moves `blog_identifier()` sanitization.
* Moves and updates `host_list()` sanitization to increase the chance of only getting host names.
* Adds unit tests for both.

### Other information:

## Testing instructions:

* Go to Activitypub settings and make sure updating the blog identifier and attribution domains still works as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Upgrade script to fix Follower json representations with unescaped backslashes.
+* Centralized place for sanitization functions.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


pfefferle:
What about https://github.com/Automattic/wordpress-activitypub/blob/b3a09d0d4caa4acca28aa9b52213a28659a10ba1/includes/functions.php#L515 ???

pfefferle:
Now that I linked it... this function looks weird and we should get rid of it, I think...

pfefferle (approved):

PR #1398, opened 2025-02-28
Profile Settings: Use Settings API

obenland:
Based on Move PR so we don't have to update the Account Aliases setting in there. This has no urgency.

## Proposed changes:

* Creates user settings fields class to register and display settings shown in user profiles.
* Updates add_profile() callback to render the new settings section.
* Removes user settings template.

### Other information:

## Testing instructions:

* Go to User > Profile and update each ActivityPub setting and make sure it still works.


pfefferle:
@obenland I fixed the `register_meta` by adding the `$wpdb->get_blog_prefix()`, because we are using `*_user_option` and not `*_user_meta`.

I also added the missing user-metas.

pfefferle (dismissed):

includes/class-activitypub.php
```diff
@@ -707,16 +707,54 @@ public static function updated_postmeta( $meta_id, $object_id, $meta_key, $meta_
...
+
+		\register_meta(
+			'user',
+			$blog_prefix . 'activitypub_description',
+			array(
+				'type'              => 'string',
+				'description'       => 'The user&#8217;s description.',
+				'single'            => true,
+				'default'           => '',
+				'sanitize_callback' => 'wp_kses_post',
```
obenland: Core is stricter for the regular user description, [limiting it to `$allowedtags`](https://github.com/WordPress/wordpress-develop/blob/6.7.2/src/wp-includes/kses.php#L898-L902). Should we do the same?

includes/class-activitypub.php
```diff
@@ -707,16 +707,54 @@ public static function updated_postmeta( $meta_id, $object_id, $meta_key, $meta_
...
 				'sanitize_callback' => array( Sanitize::class, 'url_list' ),
 			)
 		);
+
+		\register_meta(
+			'user',
+			$blog_prefix . 'activitypub_description',
+			array(
+				'type'              => 'string',
+				'description'       => 'The user&#8217;s description.',
```
obenland: This entity escaping doesn't make a ton of sense in a non-translated string that won't get displayed. How do I best allow for curly quotes in these circumstances?

pfefferle: this is UTF-8 we can use them as ASCII

includes/class-activitypub.php
```diff
@@ -707,16 +707,54 @@ public static function updated_postmeta( $meta_id, $object_id, $meta_key, $meta_
...
+
 		\register_meta(
 			'user',
-			'activitypub_also_known_as',
+			$blog_prefix . 'activitypub_also_known_as',
 			array(
 				'type'              => 'array',
 				'description'       => 'An array of URLs that the user is known by.',
-				'single'            => false,
+				'single'            => true,
```
obenland, line 736: Setting it to `false` was something Windsurf suggested and I had on my list of things to do more research on. With array types, does this make a difference?

pfefferle: it serializes the array, doesn't it? so I think it should be `single`!?

obenland:
@pfefferle I've based this PR on the Move PR so we don't have to update the Account Aliases setting in there. Should I rebase it on trunk and split it?

pfefferle:
I think this would be a bit overkill, I think the Move PR is already ready to merge!?! We should add the next features as separate PRs.

pfefferle (approved):

PR #1401, opened 2025-03-02
EMA: Remove the special handling of comments

akirk:
Accompanies https://github.com/akirk/enable-mastodon-apps/pull/225

## Proposed changes:

* There is some special handling of comments that is no longer necessary since EMA internally uses a custom post type.


integration/class-enable-mastodon-apps.php
```diff
@@ -44,6 +44,7 @@ public static function init() {
 		\add_filter( 'mastodon_api_statuses', array( self::class, 'api_statuses_external' ), 10, 2 );
 		\add_filter( 'mastodon_api_status_context', array( self::class, 'api_get_replies' ), 10, 3 );
 		\add_action( 'mastodon_api_update_credentials', array( self::class, 'api_update_credentials' ), 10, 2 );
+		\add_action( 'mastodon_api_submit_status_text', array( '\\Activitypub\\Mention', 'the_content' ), 10, 2 );
```
obenland:
```suggestion
		\add_action( 'mastodon_api_submit_status_text', array( Mention::class, 'the_content' ), 10, 2 );
```
Let's go with the constant to align with other uses in. the plugin. The Mention class might have to be imported for the suggestion to work

akirk: Thanks

obenland:
What's our back-compat policy for integrations? Is a version check needed to make sure the update is compatible with the current version of EMA? Or do we just support the most recent version of the plugin?

pfefferle:
@obenland I do not think it breaks things... in the worst case a user would not see comments in the app until he/she updates the EMA plugin.

akirk:
This already fixes the current version of EMA, I had shipped a different comment handling a while ago, and realized this code was a reason that the context for comments wasn't working properly.

pfefferle (approved):

PR #1404, opened 2025-03-03
Replace apostrophs to use ASCII instead of html-encodes

pfefferle:
only cosmetics ;)


obenland:
I'm not sure this will work with latin1 db charsets. Let's keep translatable strings that are user-facing as-is?

pfefferle:
See https://www.sttmedia.de/unicode-latin1-tabelle

obenland:
Yeah, I'm a bit out of my depth here.

Long shot: @ocean90 Based on your experience, is it safe to use curly quotes directly in translation strings, rather than their html entities?

ocean90:
> Long shot: @ocean90 Based on your experience, is it safe to use curly quotes directly in translation strings, rather than their html entities?

Core usually uses HTML entities as these are easier to read when translating. It's also noted on https://make.wordpress.org/core/handbook/best-practices/internationalization/.

pfefferle:
danke @ocean90  :)

PR #1407, opened 2025-03-04
Outbox: Log progress

obenland:
Logs progress of Outbox processing and makes it accessible through the `ap_outbox` post type API endpoint to make it easier to understand its success.

Screenshot 2025-03-03 at 7 12 43 PM

## Proposed changes:

* Adds a callback that logs Outbox processing progress.

## Testing instructions:

* Update a post.
* Make sure Outbox item gets processed.
* Query `/wp/v2/ap_outbox` and look for the `meta` key or check the Outbox UI in Debug mode.


obenland:
@mattwiebe @pfefferle Do you think this is worth adding? Helpful?

includes/class-dispatcher.php
```diff
@@ -355,6 +360,39 @@ public static function add_inboxes_of_replied_urls( $inboxes, $actor_id, $activi
...
+	 *
+	 * @param array  $inboxes        The inboxes.
+	 * @param string $json           The ActivityPub Activity JSON.
+	 * @param int    $actor_id       The actor ID.
+	 * @param int    $outbox_item_id The Outbox item ID.
+	 * @param array  $retries        The failed inboxes.
+	 * @param int    $batch_size     The batch size.
+	 * @param int    $offset         The offset.
+	 */
+	public static function log_progress( $inboxes, $json, $actor_id, $outbox_item_id, $retries, $batch_size, $offset ) {
```
pfefferle, line 374: should we move that to the debug class?

pfefferle:
I think the overall question is, how do we provide the user a way to either debug the outbox or to provide us relevant data.

Two ideas:

1. So maybe we need to think about a log-view that is visible for the admin, that show exactly that data and maybe allows to manually retry.
2. A report, that might be generated by the "Site Health", that might signal if an outbox does not properly federate any more and maybe some stats that a user can look at under: `/wp-admin/site-health.php?tab=debug`

This is not an argument against storing the data, but maybe it helps us understand what we want to store and where!?!

obenland:
Yeah, having logs in Dotcom now actually makes me feel like this change would not be super helpful. The Outbox processing looks pretty solid…

pfefferle:
@obenland should we close it then? or still a valid improvement?

obenland:
Let's close for now until we have answers to

> how do we provide the user a way to either debug the outbox or to provide us relevant data.

PR #1413, opened 2025-03-04
Move: Differentiate between `internal` and `external` `Move`

pfefferle:
To better handle different `Move` use cases, it makes sense to have a different `Move` function for internal (`Move::internally( $from, $to )`) and external (`Move::externally( $from, $to )`) moves.

~To better handle the very specific internal `Move` command, this PR also introduces the `Inherit` Activity. This `Activity` allows developers to mark an `Activity` that would normally stored as Activity-Object, to be handles as the main Activity.~

This PR depends on the changes #1474 

## Proposed changes:

* Two different functions for internal moves (user-id changes, move abandoned users) and external moves (from WordPress to Mastodon).
* ~Implement `Inherit` Activity~

### Other information:

## Testing instructions:

* Go to '..'
*


includes/class-handler.php
```diff
@@ -47,4 +49,23 @@ public static function register_handlers() {
...
 	}
+
+	/**
+	 * Filter the outbox activity.
+	 *
+	 * @param Activity $activity The activity.
+	 * @return Activity The activity.
+	 */
+	public static function outbox_activity( $activity ) {
+		if ( 'Inherit' === $activity->get_type() ) {
```
obenland: I'm not sure I understand the benefit of `Inherit` quite yet.

So far we've been able to limit ourselves to "official" types, I worry that adding a custom "fake" one will make things less clear down the line, including new contributors to this repo.

Why can't we check for `Move` instead and do all the things we do for `Inherit`?

includes/class-handler.php
```diff
@@ -47,4 +49,23 @@ public static function register_handlers() {
 		 */
 		do_action( 'activitypub_register_handlers' );
 	}
+
+	/**
+	 * Filter the outbox activity.
```
obenland: Let's be a bit more specific in docs and function name as to what the function does?

includes/class-move.php
```diff
@@ -64,6 +64,57 @@ public static function account( $from, $to ) {
...
+		$user = Actors::get_by_various( $from );
+
+		if ( \is_wp_error( $user ) ) {
+			return $user;
+		}
+
+		// Add the old account URL to alsoKnownAs.
+		if ( $user->get__id() > 0 ) {
+			self::update_user_also_known_as( $user->get__id(), $from );
+			\update_user_option( $user->get__id(), 'activitypub_move_to', $to );
```
mattwiebe: Lets be consistent with `move` vs `moved` in blog user below. Probably `moved`?

includes/class-handler.php
pfefferle: I do not want to stick with `Inherit`, but this was a nice and simple solution, for storing full Activities in the current system without completely rewriting it.

We already have the issue, that there are some special cases, where developers need to store full Activities: https://github.com/Automattic/wordpress-activitypub/issues/1290
It was possible to build a workaround for that, but we should not add too much specialities, because it will make a possible future migration more and more complex. Aside from that, storing a `Move` in a `Move` feels even worse.

From my perspective, we urgently need a way to store Activities without rewriting the whole transformer/outbox structure just yet.

mattwiebe (commented):
Does there need to be a way to save/store the JSON from the previous incarnation of the author? Does it still need to explicitly set its ID to the old domain? Like, when serving `https://olddomain.blog/?author=1`, it shouldn't have its ID set to `https://newdomain.blog/?author=1` right?

pfefferle:
@mattwiebe This has yet to be proven :)

obenland:
Pretty nice! 😊

obenland (dismissed):

obenland (approved):

PR #1414, opened 2025-03-04
Move: Always add all already alsoKnownAs fields

pfefferle:
This PR already adds all already known IDs/URLs of an Actor, like:

* `http://example.com/?author=1`
* `http://example.com/@example.com/`
* `http://example.com/author/username/`

### Other information:

## Testing instructions:

* Go to '..'
*


obenland:
Is there a way to save these defaults in the options and have them as a defacto source of truth? It would be nice to not have to keep track of multiple things

pfefferle:
sure, but then they could be deleted and maybe break things if we rely on the info

obenland:
I'd consider this true for any information we save in options or meta

pfefferle:
sure, but the rest is set by an activity the user has triggered. this was meant as a first step to unify from the different IDs we currently support, to the one we finally prefer.

includes/model/class-blog.php
```diff
@@ -559,7 +559,15 @@ public function get_attribution_domains() {
 	 * @return array The alsoKnownAs.
 	 */
 	public function get_also_known_as() {
-		return \get_option( 'activitypub_blog_user_also_known_as' );
+		$also_known_as = array(
+			\add_query_arg( 'author', $this->_id, \trailingslashit( \home_url() ) ),
```
obenland:
```suggestion
			\add_query_arg( 'author', $this->_id, \home_url( '/' ) ),
```

includes/model/class-user.php
```diff
@@ -428,7 +428,15 @@ public function get_attribution_domains() {
 	 * @return array The alsoKnownAs.
 	 */
 	public function get_also_known_as() {
-		return \get_user_option( 'activitypub_also_known_as', $this->_id );
+		$also_known_as = array(
+			\add_query_arg( 'author', $this->_id, \trailingslashit( \home_url() ) ),
```
obenland:
```suggestion
			\add_query_arg( 'author', $this->_id, \home_url( '/' ) ),
```

includes/model/class-blog.php
```diff
@@ -559,7 +559,15 @@ public function get_attribution_domains() {
...
+			$this->get_url(),
+			$this->get_alternate_url(),
+		);
+
+		$also_known_as = array_merge( $also_known_as, \get_option( 'activitypub_blog_user_also_known_as' ) ?? array() );
```
obenland:
```suggestion
		$also_known_as = array_merge( $also_known_as, \get_option( 'activitypub_blog_user_also_known_as', array() ) );
```

includes/model/class-user.php
```diff
@@ -428,7 +428,15 @@ public function get_attribution_domains() {
...
+			$this->get_url(),
+			$this->get_alternate_url(),
+		);
+
+		$also_known_as = array_merge( $also_known_as, \get_user_option( 'activitypub_also_known_as', $this->_id ) ?? array() );
```
obenland: The `??` might need to be a ternary to account for the `false` the function can return.

obenland (approved):

obenland (dismissed):

mattwiebe (dismissed):

obenland (approved):

PR #1415, opened 2025-03-04
Move: Fix `get(_user)_options` handling

pfefferle:
Fixes the handling of `get(_user)_options`.

If `\get_option( 'activitypub_blog_user_also_known_as' )` returns `false` it results in `array( false );`

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-move.php
+++ b/includes/class-move.php
@@ -71,7 +71,7 @@ class Move {
 	 * @param string $from    The current account URL.
 	 */
 	private static function update_user_also_known_as( $user_id, $from ) {
-		$also_known_as   = (array) \get_user_option( 'activitypub_also_known_as', $user_id );
+		$also_known_as   = \get_user_option( 'activitypub_also_known_as', $user_id ) ?? array();
 		$also_known_as[] = $from;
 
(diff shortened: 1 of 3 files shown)


includes/class-move.php
```diff
@@ -83,7 +83,7 @@ private static function update_user_also_known_as( $user_id, $from ) {
 	 * @param string $from The current account URL.
 	 */
 	private static function update_blog_also_known_as( $from ) {
-		$also_known_as   = (array) \get_option( 'activitypub_blog_user_also_known_as' );
+		$also_known_as   = \get_option( 'activitypub_blog_user_also_known_as' ) ?? array();
```
obenland: Would passing an empty array as the default work here?

pfefferle: the default is only used if the option was never set before. so if it is set to null or an empty string, it will not return the fallback.

pfefferle: > $default_value - Default value to return if the option does not exist.

https://developer.wordpress.org/reference/functions/get_option/#parameters

obenland (approved):
👍

PR #1416, opened 2025-03-04
Move: Add docs for Mastodon migrations

obenland:
## Proposed changes:

* Adds a help tab with migration instructions for migrating between WP and Mastodon instances. 
 
### After

Screenshot 2025-03-04 at 12 56 50 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+* Documentation for migrating from a Mastodon instance to WordPress and vice versa.
+
 ### Fixed
 
 * Updates to certain user meta fields did not trigger an Update activity.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* Documentation for migrating from a Mastodon instance to WordPress and vice versa.
+* Documentation for migrating from a Mastodon instance to WordPress.
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


obenland:
I don't believe the steps 8-11 in the WordPress-to-Mastodon instruction is correct. What, if anything, do users need to do to trigger the follower move?

mattwiebe:
> I don't believe the steps 8-11 in the WordPress-to-Mastodon instruction is correct. What, if anything, do users need to do to trigger the follower move?

It does look like `Move::account` should do the work but we don't have it hooked into anything other than the CLI command? I guess we need a UI for this part?

pfefferle:
The `Migrating from WordPress to Mastodon` describes the same process as the "Migrating from Mastodon to WordPress"!

To migrate from WordPress to a different Platform, the user has to set the remote account to `movedTo` and WordPress has to send the `Move` activities. This is currently only possible using the CLI-Command, as @mattwiebe already mentioned.

The abandoned platform always has to send the `Move` activity, because it is the only one that knows the followers at that moment.

I am not sure if we should document that process so dominant in the platform at this stage. We still have to test the `Move` CLI-Command, before we should recommend using it, because it can very easily break things if not run properly!

My 5 cent: I would remove the section from the welcome screen for now and use the help section to describe the "Migrating from Mastodon to WordPress". In this flow, the user has to only set the Alias and Mastodon has to do most of the job and they proved that it is working properly on their side!

mattwiebe:
Agreed, we can remove everything about WordPress -> Mastodon for now. 

We can possibly say something further about "this will be supported once ActivityPub for WordPress has feature parity with Mastodon" - or just a shorter note to say "unsupported at this time" ?

obenland:
Updated! Thanks for the feedback!

pfefferle (approved):

PR #1417, opened 2025-03-04
Move: Add cli command to update domains

obenland:
## Proposed changes:

* Adds CLI command to update activitypub users with a domain change.

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+* CLI command to create Move activities with a domain change.
+
 ### Fixed
 
 * Updates to certain user meta fields did not trigger an Update activity.
--- a/includes/class-cli.php
+++ b/includes/class-cli.php
@@ -7,11 +7,16 @@
 
 namespace Activitypub;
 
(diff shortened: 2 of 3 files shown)


obenland:
Pretty much untested so far.

I'm not super happy with the name of the subcommand, any ideas on how to improve that?

pfefferle:
`Move` is the technical term but I think we do not have to use it for the user facing command. What about `change_domain` instead?

Btw. Doesn't core cli use `-` as delimiter? `change-domain`?

mattwiebe:
I think in order for this to work, we'd need to be serving both domains to the current site at the time the command is run, with neither profile redirecting to their cross domain counterpart?

I also wonder if we need to add some other hooks in the case that we're serving a profile from the non-canonical domain? Are there other fields we need to keep track of and vary depending on the domain being served? 😵‍💫

pfefferle:
Aside from what @mattwiebe stated, this won't work, because the current code always fetches the Actor based on the `$from` URL, this will break, because either you would have to trigger it with the old domain still running -> this would break the lookup for the new domain, or, if you already configured the new domain for the blog, the internal Actor lookup would fail because of the domain mismatch.

Please have a look at: https://github.com/Automattic/wordpress-activitypub/pull/1413

PR #1426, opened 2025-03-05
Outbox: Don't schedule event when there are no followers

obenland:
## Proposed changes:

* Adds a check against the follower count for an outbox item.

### Other information:

## Testing instructions:

* On a test site without followers, create an Outbox item (update a post).
* Run `wp cron event list` and make sure it doesn't contain an `activitypub_process_outbox` event.
* On a site with followers, create an Outbox item.
* * Run `wp cron event list` and make sure `activitypub_process_outbox` is scheduled.

### DIFF
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -395,6 +395,13 @@ class Dispatcher {
 			in_array( $actor->get_followers(), $audience, true )
 		);
 
+		if ( $send ) {
+			$followers = Followers::get_inboxes_for_activity( $activity->to_json(), $actor->get__id() );
+
+			// Only send if there are followers to send to.
+			$send = ! is_countable( $followers ) || 0 < count( $followers );
+		}
+
 		/**
 		 * Filters whether to send an Activity to followers.
 		 *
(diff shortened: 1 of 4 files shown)


includes/class-dispatcher.php
```diff
@@ -395,6 +395,13 @@ protected static function should_send_to_followers( $activity, $actor, $outbox_i
 			in_array( $actor->get_followers(), $audience, true )
 		);
 
+		if ( $send ) {
+			$followers = Followers::get_inboxes_for_activity( $activity->to_json(), $actor->get__id() );
```
pfefferle, line 399: what about using `Followers::count_followers` instead?

obenland: That doesn't account for potentially adding blog user followers, that get_inboxes_for_activity() does.

FWIW, both follower lists should be cached on sites with object cache

pfefferle: makes sense...

pfefferle:
do you want to resolve the conflicts first? otherwise the approve will be dismissed ;)

obenland:
Done! Thank you!

pfefferle (approved):

PR #1427, opened 2025-03-05
Replies Context: Account for single user mode

obenland:
Fixes a fatal error when viewing Reply Context in single author mode.

## Proposed changes:

* Falls back to blog user when the post author is disabled.

### Other information:

### DIFF
--- a/includes/collection/class-replies.php
+++ b/includes/collection/class-replies.php
@@ -12,12 +12,14 @@ use WP_Comment;
 use WP_Error;
 
 use Activitypub\Comment;
+use Activitypub\Model\Blog;
 use Activitypub\Transformer\Post as PostTransformer;
 use Activitypub\Transformer\Comment as CommentTransformer;
 
 use function Activitypub\is_post_disabled;
 use function Activitypub\is_local_comment;
 use function Activitypub\get_rest_url_by_path;
+use function Activitypub\is_user_disabled;
 
 /**
  * Class containing code for getting replies Collections and CollectionPages of posts and comments.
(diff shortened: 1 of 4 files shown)


includes/collection/class-replies.php
```diff
@@ -171,10 +173,16 @@ public static function get_context_collection( $post_id ) {
 		$post_uri = ( new PostTransformer( $post ) )->to_id();
 		\array_unshift( $ids, $post_uri );
 
+		if ( is_user_disabled( $post->post_author ) ) {
```
mattwiebe: I think we need to also check if the Blog User is enabled here. We could be in a situation where the Blog User is disabled, and this particular author is also disabled.

I guess we probably need to bail above in that case?

includes/collection/class-replies.php
```diff
@@ -171,10 +173,16 @@ public static function get_context_collection( $post_id ) {
...
+			$author = new Blog();
+		} else {
+			$author = Actors::get_by_id( $post->post_author );
+		}
```
```suggestion
		$author = Actors::get_by_id( $post->post_author );
		if ( is_wp_error( $author ) ) {
			$author = new Blog();
		}
```
Would this be better?

includes/collection/class-replies.php
obenland: yeah, feels like we need that check in more places. As in, we probably shouldn't offer the context in the first place in that case?

obenland: ah, it only gets offered if this returns a response

obenland: Fixed in [68d01ad](https://github.com/Automattic/wordpress-activitypub/pull/1427/commits/68d01adc88155be64f3b6e6e35544dd748820e65)

includes/collection/class-replies.php
```diff
@@ -171,10 +173,19 @@ public static function get_context_collection( $post_id ) {
...
+				return false;
+			}
+
+			$author = new Blog();
+		}
+
 		return array(
 			'type'         => 'OrderedCollection',
 			'url'          => \get_permalink( $post_id ),
-			'attributedTo' => Actors::get_by_id( $post->post_author )->get_id(),
```
pfefferle, line 177: Maybe we need a function that always return the correct Actor?

mattwiebe: Yes ideally we handle those kinds of fallback inside `get_by_id` so that other areas of the code don't run into similar trouble

mattwiebe (approved):
This fixes the error in my testing. We can maybe revisit @pfefferle's comment about a refactor in `Actors::get_by_id` tomorrow :)

includes/collection/class-replies.php
pfefferle: I am not sure that changing the `get_by_id` method is a good idea, because it might have side effects.

PR #1429, opened 2025-03-06
Always use `https` URLs

mattwiebe:
The ActivityPub spec requires `https` URLs. Let's ensure that always happens.

Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1633,3 +1633,14 @@ function is_actor( $data ) {
 
 	return false;
 }
+
+/**
+ * Force the URL to use HTTPS.
+ *
+ * @param string $url The URL to force HTTPS.
+ *
+ * @return string The URL with HTTPS.
+ */
+function force_ssl( $url ) {
+	return preg_replace( '/^http:/', 'https:', $url );
+}
--- a/includes/model/class-application.php
+++ b/includes/model/class-application.php
@@ -13,6 +13,7 @@ use Activitypub\Activity\Actor;
 use Activitypub\Collection\Actors;
 
 use function Activitypub\get_rest_url_by_path;
+use function Activitypub\force_ssl;
 
 /**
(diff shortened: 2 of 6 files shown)


includes/model/class-blog.php
```diff
@@ -122,7 +123,9 @@ public function get_id() {
 			return $this->get_url();
 		}
 
-		return \add_query_arg( 'author', $this->_id, \trailingslashit( \home_url() ) );
+		$id = \add_query_arg( 'author', $this->_id, \trailingslashit( \home_url() ) );
```
obenland, line 126: Anywhere we use `\trailingslashit( \home_url() )` we could replace it with `\home_url( '/', 'https' )`

includes/model/class-application.php
```diff
@@ -87,7 +88,7 @@ public function get_id() {
 	 * @return string The User-Url.
 	 */
 	public function get_url() {
-		return $this->get_id();
+		return force_ssl( $this->get_id() );
```
obenland, line 91: For things like `get_id()`, let's go as deep as we can to add the scheme, in this case `get_rest_url_by_path()` where `\get_rest_url()` also accepts a third argument for the scheme.

includes/functions.php
```diff
@@ -1633,3 +1633,14 @@ function is_actor( $data ) {
...
 }
+
+/**
+ * Force the URL to use HTTPS.
+ *
+ * @param string $url The URL to force HTTPS.
+ *
+ * @return string The URL with HTTPS.
+ */
+function force_ssl( $url ) {
```
obenland, line 1644: How would you feel about using `set_url_scheme()` instead?

mattwiebe:
Closing this, was a bad idea :)

PR #1430, opened 2025-03-07
Lock settings based on constants

pfefferle:
This locks the Actor settings based on the `ACTIVITYPUB_SINGLE_USER_MODE`, `ACTIVITYPUB_BLOG_MODE` or `ACTIVITYPUB_ACTOR_MODE` constants.

Before:

Screenshot 2025-03-07 at 11 50 55

After:

Screenshot 2025-03-07 at 11 50 40

This is the first step to enable the settings on WordPress.com without confusing the user!

## Proposed changes:

* It locks the settings to the Actor-Mode that is defined by a constant, if it is.

### Other information:

## Testing instructions:

* Go to '..'
*


obenland:
Could we remove that settings field instead?

pfefferle:
We could :)

I like the locked screen, so that it is visible and might help with debugging!?!

obenland:
I'd love to learn more about the debugging aspect, could you expand on that?

Rather than solving it with constants, I'd suggest we register a Dotcom-specific settings field that displays these as disabled instead? I think it would be help with keeping things simple for their intended use-cases.

pfefferle:
This is not directly for dotcom! This is a general fix, that should show that this setting is already done through a constant. For dotcom I would vote for unhook the setting on top of this change!

obenland:
Ah, the issue description mentions Dotcom specifically.

In that case, I'd still come back to not showing the setting at all, rather than possibly confusing folks with something they can't edit.

pfefferle:
Ah, this was more, the part that hooks into the options! So that even the option calls care for the constants.

pfefferle:
> In that case, I'd still come back to not showing the setting at all, rather than possibly confusing folks with something they can't edit.

Funny, that is exactly the case why I kept them. I fear that it confuses users if they can't find settings and features without us explaining it. Aside from that it makes it hard for us to find out what `const` is used to overrule the user settings.

obenland:
> I fear that it confuses users if they can't find settings and features without us explaining it

Oh, you mean in the transition from one plugin version to the next. Yeah. We could show a banner with an explanatory text instead of the setting, and then remove it in the next version?

pfefferle:
No, I mean in general! A missing setting is always problematic, because people assume that it is there. The other point is, that it is harder for us to get to know the constant, because the user can't tell us if the block is hidden. This is not meant for WordPress.com users! This is if a various hoster or admin decides to configure the setting using a const.

pfefferle:
Removing the settings completely looks weird, because it is currently the only setting in `Profiles`

Screenshot 2025-03-11 at 11 30 36

obenland:
Yeah, we'd have to remove the corresponding section as well.

pfefferle:
> Yeah, we'd have to remove the corresponding section as well.

But then you might remove all other custom settings that are hooked into the Profiles section. It starts to get complicated 🫣

obenland:
Oh, right, it's not just for Dotcom. Yeah, maybe we should display an info notice instead of the field?

pfefferle:
@obenland I would recommend to keep that as is, because it already brings a lot of improvements and makes settings consistent, and work on a follow up that maybe hides the settings instead of disabling them. I can't think of a way to hide a section the correct way, without a hack or weird workaround.

So maybe we rethink that feature in the context of WordPress.com Simple Sites.

pfefferle:
@obenland so what do you think about my last comment?

obenland:
Maybe only show the notice and hide the settings themselves?

image

pfefferle:
@obenland done!

includes/wp-admin/class-settings-fields.php
```diff
@@ -223,8 +223,19 @@ public static function render_notifications_section() {
...
-		$value = get_option( 'activitypub_actor_mode', ACTIVITYPUB_ACTOR_MODE );
-		?>
+		$disabled = ( \defined( 'ACTIVITYPUB_SINGLE_USER_MODE' ) && ACTIVITYPUB_SINGLE_USER_MODE ) ||
+						( \defined( 'ACTIVITYPUB_DISABLE_USER' ) && ACTIVITYPUB_DISABLE_USER ) ||
+						( \defined( 'ACTIVITYPUB_DISABLE_BLOG_USER' ) && ACTIVITYPUB_DISABLE_BLOG_USER );
+
+		if ( $disabled ) :
+			?>
+			<p class="description">
+				<?php esc_html_e( '⚠ This setting is defined through server configuration by your blog&#8217;s administrator and cannot be changed. Please contact them if you need different settings.', 'activitypub' ); ?>
```
obenland: Optional: Limiting the message to "⚠ This setting is defined through server configuration by your blog&#8217;s administrator." could help with readers accepting the status quo and not feeling the need to change it. As in, Admin already took care of it.
Whereas communicating it cannot be changed and to get in touch with the admin to change it, might suggest something is wrong and needs changing.

includes/wp-admin/class-settings-fields.php
```diff
@@ -223,8 +223,19 @@ public static function render_notifications_section() {
...
+			</p>
+			<?php
+		else :
```
obenland: Optional: You could `return;` here as well.

obenland (dismissed):
Thanks, this make me feel much better! I think this is good to go as is, any suggestions are purely optional.

obenland (approved):

PR #1440, opened 2025-03-10
Use `add_settings_section` also for Welcome-Screen

### LINKED ISSUE #1433: Bookmarklet: better spacing
### Quick summary

Image

Should be a line break after the bookmarklet.

### Steps to reproduce

Visit ActivityPub admin. See the Bookmarklet section is ugly. Fix it.

### Site owner impact

### Severity

Minor

_No response_

_No response_

pfefferle:
Rewrite of the Welcome screen to be able to unhook or extend sections.

This PR also removes the white boxes to better match the classic WP-Admin style. 

Screenshot 2025-03-11 at 12 42 07

Fixes #1433

### Other information:

## Testing instructions:

* Check out the PR and check the "Welcome Screen"


obenland:
Rather than introducing another constant, would it be possible to make the Troubleshooting block a settings section that could be unset?

pfefferle:
It is no setting and I am not sure if it is a good idea to bend the setting api only to be able to hide it!?! We can use do_actions/add_actions instead, but from my experience constants are the easiest way to enable/disable features, because you can do it in the `wp-config.php`.

pfefferle:
@obenland I rewrote the complete "Welcome Screen" to use `add_settings_section` and tried to remove the white boxes.

obenland:
We _should_ be able to keep the boxes by passing this as the fifth argument to register_settings_section:
```
			array(
				'before_section' => '<div class="box">',
				'after_section'  => '</div>',
			)
```

In my testing though, for some reason `after_section` doesn't get rendered. Not sure why

pfefferle:
No need to keep the boxes! I think your point is valid to build back to the classic design to prepare for the admin rebranding some day in the future!

includes/wp-admin/class-welcome-fields.php
```diff
@@ -0,0 +1,274 @@
...
+	public static function init() {
+		add_action( 'load-settings_page_activitypub', array( self::class, 'register_welcome_fields' ) );
+	}
+
+	/**
+	 * Register welcome fields.
+	 */
+	public static function register_welcome_fields() {
+		// Add settings sections.
+		add_settings_section(
```
obenland: We're a bit inconsistent with backslashing global functions in this file. Not sure if it's worth fixing?

includes/wp-admin/class-welcome-fields.php
```diff
@@ -0,0 +1,274 @@
...
+		);
+
+		add_settings_section(
+			'activitypub_bookmarklet',
+			\__( 'Bookmarklet', 'activitypub' ),
+			array( self::class, 'render_bookmarklet_section' ),
+			'activitypub_welcome'
+		);
+
+		if ( ! is_user_disabled( Actors::BLOG_USER_ID ) ) :
```
obenland: No need to change it, it's fine as it is, this is just an FYI: The alternative syntax (if () : endif;) is typically used in template files where PHP and HTML are mixed as it's considered more readable.

includes/wp-admin/class-welcome-fields.php
```diff
@@ -0,0 +1,274 @@
...
+			/* translators: %s is the domain of this site */
+			$reply_from_template = __( 'Reply from %s', 'activitypub' );
+			$button              = sprintf(
+				'<a href="%s" class="button">%s</a>',
+				esc_attr( $bookmarklet_js ), // Need to escape quotes for the bookmarklet.
+				sprintf( $reply_from_template, \wp_parse_url( \home_url(), PHP_URL_HOST ) )
+			);
+			/* translators: %s is where the button HTML will be rendered. */
+			$button_and_explanation_template = \__(
+				'%s Save this bookmarklet to reply to posts on other sites from your own blog! When visiting a post on another site, click the bookmarklet to start a reply.',
```
obenland: This could be an opportunity to fix #1433 and display the button above this string? There it wouldn't need to be accounted for for RTL languages either.

includes/wp-admin/class-welcome-fields.php
```diff
@@ -0,0 +1,274 @@
...
+					<th scope="row"><?php \esc_html_e( 'Profile URL', 'activitypub' ); ?></th>
+					<td>
+						<input type="text" class="large-text code" id="activitypub-blog-url" value="<?php echo \esc_attr( $blog_user->get_url() ); ?>" readonly />
+					</td>
+				</tr>
+			</tbody>
+		</table>
+		<p>
+			<?php \esc_html_e( 'This blog profile will federate all posts written on your blog, regardless of the author who posted it.', 'activitypub' ); ?>
+			<a href="<?php echo \esc_url_raw( \admin_url( '/options-general.php?page=activitypub&tab=blog-profile' ) ); ?>">
```
```suggestion
			<a href="<?php echo \esc_url( \admin_url( '/options-general.php?page=activitypub&tab=blog-profile' ) ); ?>">
```
I think it should be `esc_url()` since we're in a display context?

includes/wp-admin/class-welcome-fields.php
```diff
@@ -0,0 +1,274 @@
...
+			</a>
+		</p>
+		<?php
+	}
+
+	/**
+	 * Render author profile section.
+	 */
+	public static function render_author_profile_section() {
+		$user = Actors::get_by_id( wp_get_current_user()->ID );
```
```suggestion
		$user = Actors::get_by_id( wp_get_current_user_id() );
```
Not sure if it makes a big difference

includes/wp-admin/class-welcome-fields.php
```diff
@@ -0,0 +1,274 @@
...
+					<th scope="row"><?php \esc_html_e( 'Profile URL', 'activitypub' ); ?></th>
+					<td>
+						<input type="text" class="large-text code" id="activitypub-user-url" value="<?php echo \esc_attr( $user->get_url() ); ?>" readonly />
+					</td>
+				</tr>
+			</tbody>
+		</table>
+		<p>
+			<?php \esc_html_e( 'Authors who can not access this settings page will find their username on the "Edit Profile" page.', 'activitypub' ); ?>
+			<a href="<?php echo \esc_url_raw( \admin_url( '/profile.php#activitypub' ) ); ?>">
```
```suggestion
			<a href="<?php echo \esc_url( \admin_url( '/profile.php#activitypub' ) ); ?>">
```
Same as above

templates/welcome.php
```diff
@@ -8,181 +8,6 @@
...
-			<p><?php \esc_html_e( 'NodeInfo is an effort to create a standardized way of exposing metadata about a server running one of the distributed social networks. The two key goals are being able to get better insights into the user base of distributed social networking and the ability to build tools that allow users to choose the best fitting software and server for their needs.', 'activitypub' ); ?></p>
-			<p><?php \esc_html_e( 'The ActivityPub plugin comes with a simple NodeInfo endpoint. If you need more configuration options and compatibility with other Fediverse plugins, please install the NodeInfo plugin.', 'activitypub' ); ?></p>
-			<p><a href="<?php echo \esc_url( \admin_url( 'plugin-install.php?tab=plugin-information&plugin=nodeinfo&TB_iframe=true' ) ); ?>" class="thickbox open-plugin-details-modal button install-now" target="_blank"><?php \esc_html_e( 'Install the NodeInfo Plugin', 'activitypub' ); ?></a></p>
-		</div>
-		<?php endif; ?>
-	</div>
-	<?php endif; ?>
+	<?php settings_fields( 'activitypub_welcome' ); ?>
```
obenland: This might not be needed if there's no form that we're submitting and need a nonce for.

obenland (approved):
Thank you!

PR #1443, opened 2025-03-10
Transformer: Make sure field exists before using it

obenland:
I've not been able to reproduce this myself, unfortunately.

See p1741641194159339-slack-C04TJ8P900J (PHP Warning on Dotcom)

## Proposed changes:

* Adds a check for `id` before using it.


mattwiebe (approved):
Looks good!

PR #1444, opened 2025-03-11
Rest API: Use Collection and CollectionPage more accurately

obenland:
What started as a first pass at standardizing collection links across API endpoints ended up as a second pass on how we structure `Collection`s and `CollectionPage`s. In this form it's much closer to [how the spec describes them](https://www.w3.org/TR/activitystreams-core/#collection).

## Proposed changes:

* Adds a `Collection` trait for reusability across API endpoints.
* Updates `page` parameters to no longer have a default, so we can differentiate between `Collection` and `CollectionPage` requests.
* `Collection` responses include items when the amount of total items is below the `per_page` parameter (if it doesn't need to be paged).
* `Collection` responses don't include items when there are so many items that they need to be paged, only links to `CollectionPage`s.
* `Collection` responses don't include `partOf`, `next`, and `prev` properties anymore, as they are exclusive to `CollectionPage` objects.

### Other information:

### Before
* No differentiation between `OrderedCollection` and `OrderedCollectionPage`.
* `page` parameter was ambiguous for first page.
* `id` didn't account for `page` parameter.
* `id` and `partOf` were always the same and inaccurate.

```json
{
	"@context": [
		"https://www.w3.org/ns/activitystreams"
	],
	"id": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox",
	"generator": "https://wordpress.org/?v=6.7.2",
	"actor": "http://localhost:8888/?author=0",
	"type": "OrderedCollectionPage",
	"partOf": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox",
	"totalItems": 37,
	"orderedItems": [],
	"first": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=1",
	"last": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=4",
	"next": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=2"
}
```
#### Outbox `Collection`
```json
{
	"@context": [
		"https://www.w3.org/ns/activitystreams"
	],
	"id": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox",
	"generator": "https://wordpress.org/?v=6.7.2",
	"actor": "http://localhost:8888/?author=0",
	"type": "OrderedCollection",
	"totalItems": 37,
	"first": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=1",
	"last": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=2"
}
```

#### Outbox `CollectionPage`

```json
{
	"@context": [
		"https://www.w3.org/ns/activitystreams"
	],
	"id": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=1",
	"generator": "https://wordpress.org/?v=6.7.2",
	"actor": "http://localhost:8888/?author=0",
	"type": "OrderedCollectionPage",
	"totalItems": 37,
	"orderedItems": [ ... ],
	"first": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=1",
	"last": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=2",
	"partOf": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox",
	"next": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/outbox?page=2"
}
```

#### Followers `Collection` with fewer items than `per_page`

```json
{
	"@context": [
		"https://www.w3.org/ns/activitystreams"
	],
	"id": "http://localhost:8888/wp-json/activitypub/1.0/actors/0/followers",
	"generator": "https://wordpress.org/?v=6.7",
	"actor": "http://localhost:8888/?author=0",
	"type": "OrderedCollection",
	"totalItems": 1,
	"orderedItems": [
		"https://mastodon.social/users/obenland"
	]
}
```

## Testing instructions:

* Apply the patch
* Go through API responses and make sure they return a format [as described in the spec](https://www.w3.org/TR/activitystreams-core/#collection).

### DIFF
--- a/includes/rest/class-actors-inbox-controller.php
+++ b/includes/rest/class-actors-inbox-controller.php
@@ -22,6 +22,8 @@ use function Activitypub\get_masked_wp_version;
  * @see https://www.w3.org/TR/activitypub/#inbox
  */
 class Actors_Inbox_Controller extends Actors_Controller {
+	use Collection_Links;
+
 	/**
 	 * Register routes.
 	 */
@@ -121,9 +123,6 @@ class Actors_Inbox_Controller extends Actors_Controller {
 		 */
 		\do_action( 'activitypub_rest_inbox_pre' );
 
-		$page     = $request->get_param( 'page' );
-		$per_page = $request->get_param( 'per_page' );
-
 		$response = array(
 			'@context'     => get_context(),
 			'id'           => get_rest_url_by_path( \sprintf( 'actors/%d/inbox', $user->get__id() ) ),
@@ -132,7 +131,6 @@ class Actors_Inbox_Controller extends Actors_Controller {
 			'partOf'       => get_rest_url_by_path( \sprintf( 'actors/%d/inbox', $user->get__id() ) ),
 			'totalItems'   => 0,
 			'orderedItems' => array(),
-			'first'        => get_rest_url_by_path( \sprintf( 'actors/%d/inbox', $user->get__id() ) ),
 		);
 
 		/**
@@ -142,24 +140,9 @@ class Actors_Inbox_Controller extends Actors_Controller {
 		 */
 		$response = \apply_filters( 'activitypub_rest_inbox_array', $response );
 
-		$max_pages = \ceil( $response['totalItems'] / $per_page );
-
-		if ( $page > $max_pages ) {
-			return new \WP_Error(
-				'rest_post_invalid_page_number',
-				'The page number requested is larger than the number of pages available.',
-				array( 'status' => 400 )
-			);
-		}
-
-		$response['last'] = \add_query_arg( 'page', \max( $max_pages, 1 ), $response['partOf'] );
-
-		if ( $max_pages > $page ) {
-			$response['next'] = \add_query_arg( 'page', $page + 1, $response['partOf'] );
-		}
-
-		if ( $page > 1 ) {
-			$response['prev'] = \add_query_arg( 'page', $page - 1, $response['partOf'] );
+		$response = $this->add_collection_links( $response, $request );
+		if ( is_wp_error( $response ) ) {
+			return $response;
 		}
 
 		/**
--- a/includes/rest/class-actors-inbox-controller.php
+++ b/includes/rest/class-actors-inbox-controller.php
@@ -48,6 +48,7 @@ class Actors_Inbox_Controller extends Actors_Controller {
 						'page'     => array(
 							'description' => 'Current page of the collection.',
 							'type'        => 'integer',
+							'minimum'     => 1,
 							// No default so we differentiate between Collection and CollectionPage requests.
 						),
 						'per_page' => array(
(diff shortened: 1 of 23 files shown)


pfefferle:
`Collection Links` is very generic, what about `Collection Pager`?

obenland:
I made it more generic :)

But! It also opens the door for having more methods in that trait in the future.

obenland:
@mattwiebe @pfefferle This feels like quite a fundamental change? Should we find ways to test that more heavily and closer to production before release?

obenland:
@pfefferle @mattwiebe This should be ready for a final review, now that 5.5.0 is out

pfefferle:
The APIs currently throw no error when the pager is out of the page range? Is that an expected case?

Updated: It seems like it behaves the exact same way as with Mastdodon... so probably we should keep it that way (for now).

pfefferle:
But it seems not implemented consistently, because `/wp-json/activitypub/1.0/actors/1/followers` is throwing an error if outside of the page range!?

obenland:
`/wp-json/activitypub/1.0/actors/1/followers` throws the error on an empty list with page 2 and above. Page 1 returns the Collection, not the collection page. I guess that is unexpected?

obenland:
I think I'll have to take another look at this, thanks for the test

obenland:
@pfefferle Can you try again?

pfefferle:
`/wp-json/activitypub/1.0/actors/1/followers` shows followers even on the OrderedCollection (not yet) paged endpoint.

obenland:
Yes, that would be expected. When there's no need to page, this PR returns it as part of the Collection, as in [Example 18](https://www.w3.org/TR/activitystreams-core/#collection).

pfefferle:
A, I see! It's a bug, not a feature ;)

obenland:
`/posts/28/replies`, `/posts/28/likes`, and `/posts/28/shares` are a bit special. Likes and Shares shouldn't even accept a page parameter as we don't list out the individual entities, just the count. Replies are unique as well, with the collections being put together in the `Replies` class, split up in Collection and CollectionPage methods.

I think I'd want to do a follow-up PR breaking out Likes and Shares into separate endpoints and bringing the way we assemble the Replies Collection closer to the rest of the endpoints (short of moving that into the endpoint).

obenland:
@pfefferle Final-final review?

includes/rest/class-actors-inbox-controller.php
```diff
@@ -142,24 +139,9 @@ public function get_items( $request ) {
...
-		$response['last'] = \add_query_arg( 'page', \max( $max_pages, 1 ), $response['partOf'] );
-
-		if ( $max_pages > $page ) {
-			$response['next'] = \add_query_arg( 'page', $page + 1, $response['partOf'] );
-		}
-
-		if ( $page > 1 ) {
-			$response['prev'] = \add_query_arg( 'page', $page - 1, $response['partOf'] );
+		$response = $this->prepare_collection_response( $response, $request );
+		if ( is_wp_error( $response ) ) {
```
pfefferle:
```suggestion
		if ( \is_wp_error( $response ) ) {
```

pfefferle (approved):
really nice improvement! I love it ❤️

PR #1446, opened 2025-03-11
Separate init hook for WP-Admin

pfefferle:
This PR introduces a separate hook for all admin hooks (`admin_menu` and `admin_init`) to allow unhooking the complete ActivityPub WP-Admin area.

props @obenland 

## Proposed changes:

* encapsulate the `admin_init` and `admin_menu` in a `plugin_admin_init` hook

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -102,7 +102,7 @@ function plugin_admin_init() {
 	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Blog_Settings_Fields', 'init' ) );
 	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\User_Settings_Fields', 'init' ) );
 }
-\add_action( 'plugins_loaded', __NAMESPACE__ . '\plugin_admin_init', 11 );
+\add_action( 'plugins_loaded', __NAMESPACE__ . '\plugin_admin_init' );
 
 \register_activation_hook(
 	__FILE__,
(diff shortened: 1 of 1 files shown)


activitypub.php
```diff
@@ -98,6 +89,21 @@ function plugin_init() {
...
+	// Menus are registered before `admin_init`, because of course they are.
+	\add_action( 'admin_menu', array( __NAMESPACE__ . '\WP_Admin\Menu', 'admin_menu' ) );
+	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Admin', 'init' ) );
+	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Health_Check', 'init' ) );
+	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Settings', 'init' ) );
+	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Settings_Fields', 'init' ) );
+	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Blog_Settings_Fields', 'init' ) );
+	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\User_Settings_Fields', 'init' ) );
+}
+\add_action( 'plugins_loaded', __NAMESPACE__ . '\plugin_admin_init', 11 );
```
obenland: Any reason they can't run on 10?

pfefferle: to get sure to run after the main init, to avoid possible race conditions!?!

obenland: I think if there's no explicit race condition that we're aware of it's fine to keep it at the default priority. Hooks on the same priority get executed in the order they're added.

Setting a non-default priority frequently leads to uncertainty about underlying reasons for it when changing code at a later date. Like [we had with the rewrite rules recently](https://github.com/Automattic/wordpress-activitypub/pull/1390#discussion_r1974081124), where I should have documented that reason.

obenland (approved):

PR #1448, opened 2025-03-11
Version: Use masked version in more places

obenland:
## Proposed changes:

* Uses masked version in Outbox and NodeInfo endpoints.


obenland:
@pfefferle Is there anything that'll break when we use the masked version for NodeInfo?

pfefferle:
Oh, good catch! No, it's meant directly for that!

pfefferle (approved):

PR #1449, opened 2025-03-12
Advanced Settings

pfefferle:
Not all settings are meant for every user, because they might change the behaviour of a site drastically! So in a discussion with @jeherve we came up with the idea of advanced settings. These settings are not public by default (for now it requires to know the query param, but we can also add a setting to the default settings: "Enable Advanced Settings", to always show the tab).

For now I only migrated the settings, I think might sense to not be permanently user facing and I would love to add more in the future.

Other "Advanced Settings":

* Enable/Disable `Vary` header
* Enable/Disable `Content Negotiation` for pretty permalinks

Screenshot 2025-03-12 at 10 47 43

props @obenland for the nice restructuring of the settings, that makes allows adding new tabs so easily!

## Proposed changes:

* Add a new (hidden) tab for "Advanced Settings".
* Migration to always show the advanced settings, if they were already changed.

### Other information:

## Testing instructions:

* Go to `/wp-admin/options-general.php?page=activitypub&tab=advanced` and check the settings


includes/class-migration.php
```diff
@@ -897,4 +901,14 @@ public static function migrate_actor_mode() {
...
 		}
 	}
+
+	/**
+	 * Migrate the advanced settings.
+	 */
+	public static function migrate_advanced_settings() {
+		// Check if one of the advanced settings is set and show the tab.
+		if ( '1' === \get_option( 'activitypub_authorized_fetch' ) || '180' !== \get_option( 'activitypub_outbox_purge_days' ) ) {
+			\update_option( 'activitypub_show_advanced_tab', '1' );
```
obenland: Could we make this a user meta, too, like for the welcome screen? Otherwise this will be enabled/disabled for every user on the site.

includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -0,0 +1,115 @@
...
+	 */
+	public static function init() {
+		add_action( 'load-settings_page_activitypub', array( self::class, 'register_advanced_fields' ) );
+	}
+
+	/**
+	 * Register settings.
+	 */
+	public static function register_advanced_fields() {
+		\add_settings_section(
```
obenland, line 26: Not sure if we've discussed this yet, but I think it would be preferable to have this as a section in the regular settings tab. We have limited horizontal space in the navigation, but unlimited vertical space.

includes/wp-admin/class-settings-fields.php
```diff
@@ -528,4 +440,24 @@ class="large-text"
...
+	 */
+	public static function render_blocklist_field() {
+		?>
+		<p>
+			<?php
+			echo wp_kses(
+				sprintf(
+					// translators: %s is a URL.
+					__( 'To block servers, add the host of the server to the "<a href="%s">Disallowed Comment Keys</a>" list.', 'activitypub' ),
+					esc_url( admin_url( 'options-discussion.php#disallowed_keys' ) )
```
obenland: There are some missing backslashes here and in other files

includes/wp-admin/class-settings.php
```diff
@@ -491,6 +518,11 @@ public static function add_screen_option( $screen_settings, $screen ) {
 				<input name="activitypub_show_welcome_tab" type="checkbox" id="activitypub_show_welcome_tab" value="1" ' . \checked( 1, \get_user_meta( \get_current_user_id(), 'activitypub_show_welcome_tab', true ), false ) . ' />
 				' . \__( 'Welcome Page', 'activitypub' ) . '
 			</label>
+			<label for="activitypub_show_advanced_tab">
+				<input name="activitypub_show_advanced_tab" type="hidden" value="0" />
+				<input name="activitypub_show_advanced_tab" type="checkbox" id="activitypub_show_advanced_tab" value="1" ' . \checked( 1, \get_option( 'activitypub_show_advanced_tab' ), false ) . ' />
+				' . \__( 'Advanced Settings', 'activitypub' ) . '
```
obenland:
```suggestion
				' . \esc_html__( 'Advanced Settings', 'activitypub' ) . '
```
This should also be added in line 513 and 519.

templates/admin-header.php
```diff
@@ -16,6 +16,10 @@
 	<nav class="activitypub-settings-tabs-wrapper" aria-label="<?php \esc_attr_e( 'Secondary menu', 'activitypub' ); ?>">
 		<?php
 		foreach ( $args['tabs'] as $slug => $label ) :
+			if ( isset( $args[ $slug ]['visible'] ) && ! $args[ $slug ]['visible'] ) {
```
obenland: I couldn't find any other reference to `visible`, is this outdated?

obenland (changes requested):

includes/class-migration.php
pfefferle: It is limited to the admins, but yes, I can change it.

obenland:
Not sure if I'm blocking progress for this PR?

pfefferle:
@obenland I am not sure if "blocking" is the right term here. I think we have different views on this and need a way to find a solution on how to finally build it!

For now I simply try to keep the PR up to date, so that the merge conflicts are not getting worse.

obenland:
I can disagree with and commit to the separate tab. Let's move from the `activitypub_show_advanced_tab` option to user meta for parity with how we show the welcome screen, though

includes/class-migration.php
pfefferle: @obenland done!

includes/class-activitypub.php
```diff
@@ -498,6 +499,25 @@ public static function theme_compat() {
...
+	public static function pre_option_activitypub_authorized_fetch( $pre ) {
+		if ( ! \defined( 'ACTIVITYPUB_AUTHORIZED_FETCH' ) ) {
+			return $pre;
+		}
+
+		if ( ACTIVITYPUB_AUTHORIZED_FETCH ) {
+			return '1';
+		}
+
+		return '0';
```
```suggestion
		return (bool) ACTIVITYPUB_AUTHORIZED_FETCH;
```
Given the type cast in `use_authorized_fetch` we're only interested in true or false here?

includes/class-activitypub.php
```diff
@@ -498,6 +499,25 @@ public static function theme_compat() {
...
 	}
 
+	/**
+	 * Pre-get option filter for the Authorized Fetch.
+	 *
+	 * @param string $pre The pre-get option value.
+	 *
+	 * @return string If the constant is defined, return the value, otherwise return the pre-get option value.
+	 */
+	public static function pre_option_activitypub_authorized_fetch( $pre ) {
```
obenland: Maybe this could be moved underneath `pre_option_activitypub_actor_mode` to group them and keep with the order the callbacks are registered?

includes/functions.php
```diff
@@ -1414,14 +1414,7 @@ function get_upload_baseurl() {
...
 function use_authorized_fetch() {
-	$use = false;
-
-	// Prefer the constant over the option.
-	if ( \defined( 'ACTIVITYPUB_AUTHORIZED_FETCH' ) ) {
-		$use = ACTIVITYPUB_AUTHORIZED_FETCH;
-	} else {
-		$use = (bool) \get_option( 'activitypub_authorized_fetch', '0' );
-	}
+	$use = (bool) \get_option( 'activitypub_authorized_fetch', '0' );
```
```suggestion
	$use = (bool) \get_option( 'activitypub_authorized_fetch' );
```
Since false is fine here the default is fine.

includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Advanced Settings Fields class.
+ */
+class Advanced_Settings_Fields {
+
+	/**
+	 * Initialize.
+	 */
+	public static function init() {
+		add_action( 'load-settings_page_activitypub', array( self::class, 'register_advanced_fields' ) );
```
obenland: There's an opportunity to prefix function calls with backslashes here and in later function calls.

includes/wp-admin/class-settings-fields.php
```diff
@@ -445,124 +448,53 @@ class="large-text"
...
 			id="activitypub_relays"
 			name="activitypub_relays"
 			class="large-text"
 			cols="50"
 			rows="5"
-		><?php echo esc_textarea( implode( PHP_EOL, $value ) ); ?></textarea>
+		><?php echo \esc_textarea( implode( PHP_EOL, $value ) ); ?></textarea>
 		<p class="description">
-			<?php echo wp_kses( __( 'A <strong>Fediverse-Relay</strong> distributes content across instances, expanding reach, engagement, and discoverability, especially for smaller instances.', 'activitypub' ), 'default' ); ?>
+			<?php echo \wp_kses( __( 'A <strong>Fediverse-Relay</strong> distributes content across instances, expanding reach, engagement, and discoverability, especially for smaller instances.', 'activitypub' ), 'default' ); ?>
```
obenland:
```suggestion
			<?php echo \wp_kses( \__( 'A <strong>Fediverse-Relay</strong> distributes content across instances, expanding reach, engagement, and discoverability, especially for smaller instances.', 'activitypub' ), 'default' ); ?>
```

includes/wp-admin/class-settings.php
```diff
@@ -282,6 +273,14 @@ public static function settings_page() {
 			'template' => ACTIVITYPUB_PLUGIN_DIR . 'templates/settings.php',
 		);
 
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( ( isset( $_GET['tab'] ) && 'advanced' === $_GET['tab'] ) || $show_advanced_tab ) {
+			$settings_tabs['advanced'] = array(
+				'label'    => __( 'Advanced', 'activitypub' ),
```
obenland:
```suggestion
				'label'    => \__( 'Advanced', 'activitypub' ),
```

includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -0,0 +1,115 @@
...
+		) . '</p>';
+	}
+
+	/**
+	 * Render use Authorized Fetch field.
+	 */
+	public static function render_authorized_fetch_field() {
+		$disabled = '';
+		if ( defined( 'ACTIVITYPUB_AUTHORIZED_FETCH' ) ) {
+			$disabled = 'disabled';
```
obenland: Instead of disabling it, should we replace it with the same message as we do for actor mode and the interactions setting?

includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -0,0 +1,115 @@
...
+		?>
+		<p>
+			<?php
+			$allowed_html = array(
+				'a' => array(
+					'href'   => true,
+					'target' => true,
+				),
+			);
+			echo wp_kses( __( 'Advanced settings allow deep customization but can affect your site&rsquo;s functionality, security, or performance if misconfigured. Only proceed if you fully understand the changes, and always back up your site beforehand. If unsure, consult <a href="https://github.com/Automattic/wordpress-activitypub/tree/trunk/docs" target="_blank">documentation</a> or seek <a href="https://wordpress.org/support/plugin/activitypub/" target="_blank">expert advice</a>.', 'activitypub' ), $allowed_html );
```
obenland:
```suggestion
			echo wp_kses( __( 'Advanced settings allow deep customization but can affect your site&#8217;s functionality, security, or performance if misconfigured. Only proceed if you fully understand the changes, and always back up your site beforehand. If unsure, consult <a href="https://github.com/Automattic/wordpress-activitypub/tree/trunk/docs" target="_blank">documentation</a> or seek <a href="https://wordpress.org/support/plugin/activitypub/" target="_blank">expert advice</a>.', 'activitypub' ), $allowed_html );
```
Just to double-check, we're fine with making the URLs translatable?
I don't know if it's a convention or something others have done for security purposes, but I've mostly seen URLs formatted in to avoid translators accidentally breaking them, unless there was a localized version, like for `https://wordpress.org/support/plugin/activitypub/` for example.
Not sure what's the most right thing here, just want to make sure we make that decision consciously.

includes/class-activitypub.php
obenland: If we go ahead with that change, function docs would need to be updated, too

pfefferle: I would keep it as is, to be compatible with how `get_option` and `get_user_meta` works...

includes/class-activitypub.php
obenland: You want to keep it as is?

obenland (dismissed):
I'd still move `pre_option_activitypub_authorized_fetch` but is probably not a blocker

pfefferle:
What do you mean by move?

pfefferle:
I think I've done that 😳

obenland:
I meant to move `pre_option_activitypub_authorized_fetch()` below `pre_option_activitypub_actor_mode()` to group them and keep with the order the callbacks are registered

pfefferle:
I moved it, but to the wrong place 😂

obenland (approved):

PR #1450, opened 2025-03-12
Recommend two new plugins

pfefferle:
Add "Enable Mastodon Apps" by @akirk and "Event Bridge for ActivityPub" by @Menrath to the recommended plugins list.

Screenshot 2025-03-12 at 11 37 36

### DIFF
--- a/includes/wp-admin/class-welcome-fields.php
+++ b/includes/wp-admin/class-welcome-fields.php
@@ -232,6 +232,30 @@ class Welcome_Fields {
 				<p><a href="<?php echo \esc_url( \admin_url( 'plugin-install.php?tab=plugin-information&plugin=friends&TB_iframe=true' ) ); ?>" class="thickbox open-plugin-details-modal button install-now" target="_blank"><?php \esc_html_e( 'Install the Friends Plugin', 'activitypub' ); ?></a></p>
 			</div>
 			<?php endif; ?>
+			<?php if ( ! \defined( 'EVENT_BRIDGE_FOR_ACTIVITYPUB_PLUGIN_VERSION' ) ) : ?>
+			<h4 class="activitypub-settings-accordion-heading">
+				<button aria-expanded="false" class="activitypub-settings-accordion-trigger" aria-controls="activitypub-settings-accordion-block-event-bridge-for-activitypub-plugin" type="button">
+					<span class="title"><?php \esc_html_e( 'Federate Events', 'activitypub' ); ?></span>
+					<span class="icon"></span>
+				</button>
+			</h4>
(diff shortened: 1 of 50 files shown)


includes/wp-admin/class-welcome-fields.php
```diff
@@ -232,6 +232,30 @@ public static function render_recommended_plugins_section() {
 				<p><a href="<?php echo \esc_url( \admin_url( 'plugin-install.php?tab=plugin-information&plugin=friends&TB_iframe=true' ) ); ?>" class="thickbox open-plugin-details-modal button install-now" target="_blank"><?php \esc_html_e( 'Install the Friends Plugin', 'activitypub' ); ?></a></p>
 			</div>
 			<?php endif; ?>
+			<?php if ( ! \defined( 'EVENT_BRIDGE_FOR_ACTIVITYPUB_PLUGIN_VERSION' ) ) : ?>
```
Menrath, line 235: I could provide an array of constants to evaluate whether a supported event plugin is installed.

obenland (approved):

includes/wp-admin/class-welcome-fields.php
```diff
@@ -232,6 +232,30 @@ public static function render_recommended_plugins_section() {
...
+			<?php endif; ?>
+			<?php if ( ! \defined( 'ENABLE_MASTODON_APPS_VERSION' ) ) : ?>
+			<h4 class="activitypub-settings-accordion-heading">
+				<button aria-expanded="false" class="activitypub-settings-accordion-trigger" aria-controls="activitypub-settings-accordion-block-enable-mastodon-apps-plugin" type="button">
+					<span class="title"><?php \esc_html_e( 'Use Mastodon Apps', 'activitypub' ); ?></span>
+					<span class="icon"></span>
+				</button>
+			</h4>
+			<div id="activitypub-settings-accordion-block-enable-mastodon-apps-plugin" class="activitypub-settings-accordion-panel plugin-card-block-enable-mastodon-apps" hidden="hidden">
+				<p><?php \esc_html_e( 'Use Mastodon Apps to connect your WordPress site to Mastodon. This plugin allows you to use Mastodon apps to authenticate users on your WordPress site.', 'activitypub' ); ?></p>
```
akirk, line 255: I am unsure about the text here. I wouldn't say it connects WordPress to Mastodon. Also, it is more than just authentication. What about
>Interact with your WordPress site using Mastodon apps. This gives you a wide choice of apps that you can use to read or write posts that will be federated using the ActivityPub plugin.

pfefferle: @akirk would you mind filing a PR for that?

PR #1452, opened 2025-03-12
Release process: add Changelogger

jeherve:
## Proposed changes:

> [!NOTE]
> This is a follow-up to #988 and #1159.

In the past few months we've made some changes to improve the development and release process. It is now recommended for everyone to include a changelog entry with each one of their PRs, so we have a running changelog that can be compiled on release day.

The current process has a few shortcomings that we could aim to improve:
- We currently have to update both `CHANGELOG.md` and `readme.txt` when adding a changelog entry.
- There is no set up to ensure a changelog entry is present in both files.
- There is no standardized way to create a changelog entry.
- On release day, the release lead must decide what version they want to tag, based on what they know will be in the release.

I'd like to suggest the following changes:

1. Start using Jetpack Changelogger in the repo. The package includes tools to create standardized changelog entries, and compile a standardized `CHANGELOG.md` file.
2. Update the release script so one does not have to update the changelog in 2 places.
3. Update the release script so the version is automatically defined based on the different changelog entries that were created during a release cycle.

> [!NOTE]
> In practice, that means that from now on, you'll need to run `composer changelog:add` before you open a PR. That will give you a prompt you can follow to create a standardized changelog entry. That is all you will have to do.
> The release process will still happen via `npm run release`, but you will not have to decide the version number you want to release as part of that process. That number will be automatically defined.

This PR brings in the following commits:

**Changelog generation**

- `composer.json`: we bring in the new dependency, configure it, and add new scripts that we will be able to run locally to create a changelog entry. One of those scripts will also be used by the release script.
- Create a new `.github/changelog` directory where our individual changelog entries will live. That directory will always be there, and must not be deleted. It includes a `.gitkeep` file to ensure the directory is never removed, even when it is empty.
- Update the existing GitHub action workflow that looks for a changelog entry and allows you to set a "Skip Changelog" label.
- Bring in all existing changelog entries for the unreleased version of the plugin into the new system.
- Remove those unreleased entries from the changelogs.
- Fix the existing `CHANGELOG.md`: some of the release dates were missing, some of the headings had the wrong level, and it used `*` instead of `-` for lists ; the changelog format we follow uses `-`.

**Release management**

- Update the release script to use the new composer command (and thus Jetpack changelogger) to generate the updated changelog.
- When running the release script, copy `CHANGELOG.md` into `readme.txt` after a bit of data massaging so the changelog entry format matches the expected format in `readme.txt`.

### Notes

This PR does not:

- Include a way to generate a changelog entry from the PR template you fill in when you open a PR. That would be an alternative to running `composer changelog:add` locally. If you are interested in such a feature, I would be happy to implement it.
- Include tests for the updates in the release script. I think it may be interesting to add tests for the whole process, but that may be something for a follow-up PR. :)

### Other information:

## Testing instructions:

- I created this PR to test out the process: https://github.com/Automattic/wordpress-activitypub/pull/1451
- On top of that, I would recommend pasting the generated `readme.txt` file into https://wpreadme.com/ to confirm that the new readme file will be parsed properly on WordPress.org.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,4 +1,4 @@
-name: "Changelog Updated"
+name: "Changelog entry"
 on:
   pull_request:
     # The specific activity types are listed here to include "labeled" and "unlabeled"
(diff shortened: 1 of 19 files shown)


.github/workflows/changelog.yml
jeherve, line 1: You can see the output of that updated workflow here:
https://github.com/Automattic/wordpress-activitypub/actions/runs/13816945939/job/38652511035?pr=1452

.github/changelog/.gitkeep
jeherve, line 1: This empty file is necessary because on each release day, the release script will delete all the single changelog entries from that directory, to place them into the generated changelogs. We want to keep the directory around though, since it will be where future changelog entries will go. This empty file ensures the empty directory remains.

CHANGELOG.md
jeherve, line 1: This is a big diff  because I've reformatted the whole file. Some of the release dates were missing, some of the headings had the wrong level, and most importantly the file used `*` instead of `-` for lists ; the changelog format we follow (Keep a Changelog) uses `-`.

obenland:
I think this looks good! Thanks so much for your work on this. 

> The release process will still happen via npm run release, but you will not have to decide the version number you want to release as part of that process. That number will be automatically defined.

Can you tell me more about how that works, what that means? Does it suggest the next major, minor, patch version numbers or is it more prescriptive than that?

> Include a way to generate a changelog entry from the PR template you fill in when you open a PR.

I think that would be ideal. How much work would it be to add that?
As I read through the PR description I noted the “additional” CLI step for myself, as that breaks out of any GitHub Desktop-driven workflow. Making it part of the PR template would bring it back into an existing step.

jeherve:
> Does it suggest the next major, minor, patch version numbers or is it more prescriptive than that?

When you create a changelog entry, you're asked to define your change's significance:

https://github.com/user-attachments/assets/aacdbcc5-c556-48f8-b065-0ad79f031770

Your answer to that question is reflected in the saved changelog entry file.

- In this example, I said the change was "minor":
https://github.com/Automattic/wordpress-activitypub/pull/1452/files#diff-9a1168fca396a732cfdc290c200095e2d8b5488b6078bae2de25693631653c56R1

- In this example, I said it was a "patch" level change:

https://github.com/Automattic/wordpress-activitypub/pull/1452/files#diff-10f4a0951940c58b5ac816602f5c04795e62cf8a36c7432d4e66d5a881866796R1

When the release script compiles the changelog, it looks at:

- The current stable version.
- All the provided changelog entries, and the significance of each.

Let's consider a few examples and the outcome, if the current stable version is `5.4.1`:

- If all the provided changelog entries are defined as "patch" levels, it's going to create version `5.4.2`.
- If among all changelog entries, there were many "patch" levels and at least one "minor" level, it's going to create version `5.5.0`.
- If among all the changelog entries there is at least one "major" level entry, then the significance of the other entries won't matter: it's going to create version `6.0.0`.

> I think that would be ideal. How much work would it be to add that?
> As I read through the PR description I noted the “additional” CLI step for myself, as that breaks out of any GitHub Desktop-driven workflow. Making it part of the PR template would bring it back into an existing step.

I'd be happy to add it. It would involve changing the existing PR template to add new fields, and then create a new GitHub action that would pull the changelog data from those fields and push a new commit on the PR with the generated changelog entry.

I'm happy to work on that, but I would prefer to do it in a follow-up PR to avoid having this PR sitting open for too long ; it involves rebases every time someone merges something that changes the changelog. :)

obenland (approved):
Yes, absolutely. Let’s get this in and do follow-up PRs

pfefferle:
That was a fast merge! 😳

@jeherve what would happen if we need to release a patch with some cherry picks?

@obenland what about the release script? What about the Version replacements in the code ("unreleased").

How to release new versions now?

Does this block the next release for now?

obenland:
We use the release script as before, see #1451. Nothing on the release side changed other than how the changelog gets populated.

pfefferle:
Was this done with the script?

jeherve:
> Was this done with the script?

Yes, that PR was generated with the script.

jeherve:
I opened a PR to update the docs and mention the changelog tool here:
https://github.com/Automattic/wordpress-activitypub/pull/1454

jeherve:
> Include a way to generate a changelog entry from the PR template you fill in when you open a PR.

I opened a PR to do that -> #1456

PR #1456, opened 2025-03-13
Changelogger: automatically create a changelog entry from info extracted from PR description

jeherve:
Follow-up to #1452

## Proposed changes:

Now that we have a CLI tool that allows us to create changelog entries on the fly, let's introduce a way to create those entries directly from the PR description. It should help folks who may not use the CLI in their daily development flow.

This PR makes the following changes:

1. It updates the existing PR template to include new fields the PR author (or a reviewer) can fill in.
2. It adds a new step to the existing changelog workflow. When no changelog is detected, and when the "Skip Changelog" label is not found on the PR, look for changelog info in the PR body.
3. It adds another step that will run when info is found in the PR body ; it will create a changelog file with the extracted information, and commit it to that branch for you.

### Other information:

## Testing instructions:

This is better tested directly in a PR: https://github.com/Automattic/wordpress-activitypub/pull/1457

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,3 +20,41 @@ Fixes #
 * Go to '..'
 *
 
+### Changelog entry
+
+<!-- You can optionally choose to enter a changelog entry by checking the box below and supplying data. -->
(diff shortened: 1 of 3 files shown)


obenland (approved):
Thank you, this is great!

PR #1466, opened 2025-03-16
Add PHPstan to CI.

Menrath:
Greetings from Cloudfest Hackathon :)

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -34,6 +34,7 @@ npm-debug.log
 package-lock.json
 package.json
 phpcs.xml
+phpstan.neon.dist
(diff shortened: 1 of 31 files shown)


obenland:
Thanks so much for putting this together @Menrath! It was a valuable experience for me better understand what it offers and how it could benefit the plugin.

At this point, I'm not convinced it's the best use of time to go through all existing errors and keep up with adding exceptions in future PRs. I'm happy for us to come back to it at some point, but for now I think the sweet spot of effort and value would be to cherry-pick the code changes and punt the GitHub action.

PR #1468, opened 2025-03-17
HTTP: Remove abstractions in favor of direct access

obenland:
## Proposed changes:

* Uses `Http::post` directly, instead of wrapper `safe_remote_post()`.
* Deprecated unused `safe_remote_get()` and `safe_remote_post()` functions.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Deprecated `safe_remote_get()` and `safe_remote_post()` in favor of using `Http` methods directly.
</details>


obenland:
Do we need to deprecate these functions? Do we think they're used anywhere else?
A GitHub search didn't really show any instances besides old versions in ActivityPub forks.

obenland:
@pfefferle This should be ready for review.

pfefferle:
I am (still) not sure if it is a good idea to deprecate them... I would bet, that we would have to change the HTTP Class in the next 3 month and these functions allow us to simply do that without forcing third party implementers to update their code!

obenland:
Sounds good, let's wait

PR #1471, opened 2025-03-18
Update the text for Enable Mastodon Apps

akirk:
Follow-up to #1450, as discussed in https://github.com/Automattic/wordpress-activitypub/pull/1450/files/9ec32dfaa0a0dbe4575be8e8529f26149c9b379d#r1998156842

### DIFF
--- a/includes/wp-admin/class-welcome-fields.php
+++ b/includes/wp-admin/class-welcome-fields.php
@@ -252,7 +252,22 @@ class Welcome_Fields {
 				</button>
 			</h4>
 			<div id="activitypub-settings-accordion-block-enable-mastodon-apps-plugin" class="activitypub-settings-accordion-panel plugin-card-block-enable-mastodon-apps" hidden="hidden">
-				<p><?php \esc_html_e( 'Use Mastodon Apps to connect your WordPress site to Mastodon. This plugin allows you to use Mastodon apps to authenticate users on your WordPress site.', 'activitypub' ); ?></p>
+				<p>
+					<?php
+					echo \wp_kses(
+						\sprintf(
+							// translators: %s is a URL.
+							\__( 'Interact with your WordPress site using <a href="%s" target="_blank">Mastodon apps</a>. This gives you a wide choice of apps that you can use to read or write posts that will be federated using the ActivityPub plugin.', 'activitypub' ),
+							'https://joinmastodon.org/apps'
+						),
+						array(
+							'a' => array(
+								'href' => true,
+							),
+						)
+					);
+					?>
+				</p>
 				<p><a href="<?php echo \esc_url( \admin_url( 'plugin-install.php?tab=plugin-information&plugin=enable-mastodon-apps&TB_iframe=true' ) ); ?>" class="thickbox open-plugin-details-modal button install-now" target="_blank"><?php \esc_html_e( 'Install the Enable Mastodon Apps Plugin', 'activitypub' ); ?></a></p>
 			</div>
 			<?php endif; ?>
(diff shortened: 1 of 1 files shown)


includes/wp-admin/class-welcome-fields.php
```diff
@@ -252,7 +252,22 @@ public static function render_recommended_plugins_section() {
 				</button>
 			</h4>
 			<div id="activitypub-settings-accordion-block-enable-mastodon-apps-plugin" class="activitypub-settings-accordion-panel plugin-card-block-enable-mastodon-apps" hidden="hidden">
-				<p><?php \esc_html_e( 'Use Mastodon Apps to connect your WordPress site to Mastodon. This plugin allows you to use Mastodon apps to authenticate users on your WordPress site.', 'activitypub' ); ?></p>
+				<p>
+					<?php
+					echo \wp_kses(
+						\sprintf(
+							// translators: %s is a URL.
+							\__( 'Interact with your WordPress site using <a href="%s" target="_blank">Mastodon apps</a>. This gives you a wide choice of apps that you can use to read or write posts that will be federated using the ActivityPub plugin.', 'activitypub' ),
```
obenland: > Use <a href="%s" target="_blank">Mastodon apps</a> to read fediverse posts and publish to your WordPress blog, with federation handled by the ActivityPub plugin.

How do you feel about that? I was trying to think of a way to make it super obvious that it still publishes to the blog

includes/wp-admin/class-welcome-fields.php
```diff
@@ -252,7 +252,22 @@ public static function render_recommended_plugins_section() {
...
+							'https://joinmastodon.org/apps'
+						),
+						array(
+							'a' => array(
+								'href' => true,
```
```suggestion
								'href'   => true,
								'target' => true,
```

includes/wp-admin/class-welcome-fields.php
akirk: Well the "read fediverse posts" is only true when you have Friends installed. EMA+ActivityPub will show your own posts. What about:

> Interact with your WordPress site using [Mastodon apps](https://joinmastodon.org/apps). This gives you a wide choice of apps that you can use to read or write WordPress posts that will be federated using the ActivityPub plugin.

And maybe add an addition sentence?

> You can also add the Friends plugin to read posts from the Fediverse.

akirk: Or
> Interact with your WordPress site using [Mastodon apps](https://joinmastodon.org/apps). This gives you a wide choice of apps that you can use to read your posts and publish to your WordPress blog, with federation handled by the ActivityPub plugin. When you also add the Friends plugin, you'll be able to read posts from the Fediverse in Mastodon apps as well.

obenland: > Engage with your WordPress site through a variety of [Mastodon apps](https://joinmastodon.org/apps). This plugin connects you to a broad selection of apps to write posts, which will then be seamlessly federated via the ActivityPub plugin.

I've been struggling to get AI to come up with something less clunky—this was my favorite. What do you think?

akirk: "Connects you to" doesn't feel right to me and the double mention apps makes it clunky. What about

> Enable the use of a wide variety of [Mastodon apps](https://joinmastodon.org/apps) to interact with your WordPress site, for example write posts that can then be federated via the ActivityPub plugin.

obenland: Sounds good!

akirk: Done!

obenland (approved):

PR #1474, opened 2025-03-18
Outbox: Add support for full activities

obenland:
Moves the Outbox from saving activity objects to full Activities with back compat for 3rd-party plugins.

## Proposed changes:

* `Outbox::add()` only accepts full Activities now.
* `add_to_outbox()` now transforms data to full objects, based on what shape of data it receives.
* Updated unit tests to account for the new data shape.

### Other information:

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Outbox items now contain the full activity, not just activity objects.

</details>

### DIFF
--- a/includes/collection/class-outbox.php
+++ b/includes/collection/class-outbox.php
@@ -35,8 +35,8 @@ class Outbox {
 		$object_id  = false;
 		$title      = '';
 		if ( $activity->get_object() ) {
-			$object_id  = is_string( $activity->get_object() ) ? $activity->get_object() : $activity->get_object()->get_id();
-			$title      = self::recursively_get_title( $activity->get_object() );
+			$object_id = is_string( $activity->get_object() ) ? $activity->get_object() : $activity->get_object()->get_id();
+			$title     = self::recursively_get_title( $activity->get_object() );
 		}
 
 		$outbox_item = array(
--- a/includes/collection/class-outbox.php
+++ b/includes/collection/class-outbox.php
@@ -34,18 +34,8 @@ class Outbox {
 	 * @return false|int|\WP_Error The added item or an error.
 	 */
 	public static function add( $activity_object, $activity_type, $user_id, $content_visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
-		$actor_type            = Actors::get_type_by_id( $user_id );
-		$title                 = $activity_object->get_name() ?? $activity_object->get_content();
-		$activitypub_object_id = $activity_object->get_id();
-
-		if ( ! $title && is_activity( $activity_object ) && $activity_object->get_object() instanceof \Activitypub\Activity\Base_Object ) {
-			$title                 = $activity_object->get_object()->get_name() ?? $activity_object->get_object()->get_content();
-			$activitypub_object_id = $activity_object->get_object()->get_id();
-		}
-
-    if ( null === $activity_type ) {
-			$activity_type = $activity_object->get_type();
-		}
+		$actor_type                        = Actors::get_type_by_id( $user_id );
+		[ $title, $activitypub_object_id ] = self::recursively_get_activity_meta( $activity_object );
 
 		$outbox_item = array(
 			'post_type'    => self::POST_TYPE,
@@ -288,5 +278,22 @@ class Outbox {
 		}
 
 		return self::get_activity( $outbox_item );
+	}
+
+	/**
+	 * Get the metadata of an activity recursively.
+	 *
+	 * @param \Activitypub\Activity\Base_Object $activity The activity object.
+	 * @return array The meta data.
+	 */
+	private static function recursively_get_activity_meta( $activity ) {
+		$title                 = $activity->get_name() ?? $activity->get_content();
+		$activitypub_object_id = $activity->get_id();
+
+		if ( ! $title && is_activity( $activity ) && $activity->get_object() instanceof \Activitypub\Activity\Base_Object ) {
+			return self::recursively_get_activity_meta( $activity->get_object() );
+		}
+
+		return array( $title, $activitypub_object_id );
 	}
 }
--- a/includes/collection/class-outbox.php
+++ b/includes/collection/class-outbox.php
@@ -66,11 +66,11 @@ class Outbox {
 		// Update the activity ID if the post was inserted successfully.
 		if ( $id && ! \is_wp_error( $id ) ) {
 			$activity->set_id( \get_the_guid( $id ) );
-			$content = \wp_slash( $activity->to_json() );
+
 			\wp_update_post(
 				array(
 					'ID'           => $id,
-					'post_content' => $content,
+					'post_content' => \wp_slash( $activity->to_json() ),
 				)
 			);
 		}
--- a/includes/collection/class-outbox.php
+++ b/includes/collection/class-outbox.php
@@ -11,7 +11,6 @@ use Activitypub\Dispatcher;
 use Activitypub\Scheduler;
 use Activitypub\Activity\Activity;
 
-use function Activitypub\is_activity;
 use function Activitypub\add_to_outbox;
 
 /**
@@ -32,8 +31,9 @@ class Outbox {
 	 * @return false|int|\WP_Error The added item or an error.
 	 */
 	public static function add( $activity, $user_id, $visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
-		$actor_type                        = Actors::get_type_by_id( $user_id );
-		[ $title, $activitypub_object_id ] = self::recursively_get_activity_meta( $activity );
+		$actor_type = Actors::get_type_by_id( $user_id );
+		$object_id  = is_string( $activity->get_object() ) ? $activity->get_object() : $activity->get_object()->get_id();
+		$title      = self::recursively_get_title( $activity->get_object() );
 
 		$outbox_item = array(
 			'post_type'    => self::POST_TYPE,
@@ -48,7 +48,7 @@ class Outbox {
 			'post_author'  => \max( $user_id, 0 ),
 			'post_status'  => 'pending',
 			'meta_input'   => array(
-				'_activitypub_object_id'         => $activitypub_object_id,
+				'_activitypub_object_id'         => $object_id,
 				'_activitypub_activity_type'     => $activity->get_type(),
 				'_activitypub_activity_actor'    => $actor_type,
 				'activitypub_content_visibility' => $visibility,
@@ -87,7 +87,7 @@ class Outbox {
 			return false;
 		}
 
-		self::invalidate_existing_items( $activitypub_object_id, $activity->get_type(), $id );
+		self::invalidate_existing_items( $object_id, $activity->get_type(), $id );
 
 		return $id;
 	}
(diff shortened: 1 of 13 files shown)


includes/collection/class-outbox.php
```diff
@@ -214,13 +219,18 @@ public static function get_activity( $outbox_item ) {
...
+
+		if ( in_array( $type, Activitypub::$object_types, true ) ) {
+			$activity = new Activity();
+			$activity->set_type( $type );
+			$activity->set_id( $outbox_item->guid );
+			$activity->set_actor( $actor->get_id() );
+			// Pre-fill the Activity with data (for example cc and to).
+			$activity->set_object( \json_decode( $outbox_item->post_content, true ) );
+		} else {
+			$activity = Activity::init_from_array( \json_decode( $outbox_item->post_content, true ) );
```
obenland: This is what the result looks like:
```
[18-Mar-2025 10:52:28 UTC] Activitypub\Activity\Activity Object
(
    [id:protected] => https://localhost:8888/author/admin#activity-move-1742295027
    [type:protected] => Move
    [attachment:protected] => 
    [attributed_to:protected] => 
    [audience:protected] => 
    [content:protected] => 
    [context:protected] => 
    [content_map:protected] => 
    [name:protected] => 
    [name_map:protected] => 
    [end_time:protected] => 
    [generator:protected] => 
    [icon:protected] => 
    [image:protected] => 
    [in_reply_to:protected] => 
    [location:protected] => 
    [preview:protected] => 
    [published:protected] => 
    [start_time:protected] => 
    [summary:protected] => 
    [summary_map:protected] => 
    [tag:protected] => 
    [updated:protected] => 
    [url:protected] => 
    [to:protected] => 
    [bto:protected] => 
    [cc:protected] => Array
        (
        )

    [bcc:protected] => 
    [media_type:protected] => text/html
    [duration:protected] => 
    [source:protected] => 
    [replies:protected] => 
    [likes:protected] => 
    [shares:protected] => 
    [sensitive:protected] => 
    [object:protected] => https://localhost:8888/author/admin
    [actor:protected] => http://localhost:8888/@admin
    [target:protected] => https://localhost:8888/author/admin
    [result:protected] => 
    [origin:protected] => http://localhost:8888/@admin
    [instrument:protected] => 
)
```

obenland:
@pfefferle When Undo-ing an activity, it saves a full activity as the activity object, making it hard to differentiate self-contained activities from activity objects. We'd have to check whether the activity type is an Undo type?

pfefferle:
An undo is different from a create, because you undo an activity and not an object. So for that we have to store the undo as activity with the create activity as object.

includes/collection/class-outbox.php
```diff
@@ -26,21 +26,15 @@ class Outbox {
...
 	 *
 	 * @param \Activitypub\Activity\Base_Object $activity_object    The object of the activity that will be added to the outbox.
-	 * @param string                            $activity_type      The activity type.
+	 * @param null|string                       $activity_type      The activity type. null if the full activity is provided in $activity_object.
 	 * @param int                               $user_id            The real or imaginary user ID of the actor that published the activity that will be added to the outbox.
 	 * @param string                            $content_visibility Optional. The visibility of the content. Default: `ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC`. See `constants.php` for possible values: `ACTIVITYPUB_CONTENT_VISIBILITY_*`.
 	 *
 	 * @return false|int|\WP_Error The added item or an error.
 	 */
 	public static function add( $activity_object, $activity_type, $user_id, $content_visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
```
pfefferle:
```suggestion
	 * @param int                               $user_id            The real or imaginary user ID of the actor that published the activity that will be added to the outbox.
	 * @param string                            $content_visibility Optional. The visibility of the content. Default: `ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC`. See `constants.php` for possible values: `ACTIVITYPUB_CONTENT_VISIBILITY_*`.
	 *
	 * @return false|int|\WP_Error The added item or an error.
	 */
	public static function add( $activity_object, $user_id, $content_visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
```

I made a GitHub search and no other plugin seem to use the function directly. Should we remove the `$activity_type` completely then and only have it in the `add_to_outbox` function?

includes/collection/class-outbox.php
```diff
@@ -25,41 +25,33 @@ class Outbox {
...
-	public static function add( $activity_object, $activity_type, $user_id, $content_visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
-		$actor_type            = Actors::get_type_by_id( $user_id );
-		$title                 = $activity_object->get_name() ?? $activity_object->get_content();
-		$activitypub_object_id = $activity_object->get_id();
-
-		if ( ! $title && is_activity( $activity_object ) && $activity_object->get_object() instanceof \Activitypub\Activity\Base_Object ) {
-			$title                 = $activity_object->get_object()->get_name() ?? $activity_object->get_object()->get_content();
-			$activitypub_object_id = $activity_object->get_object()->get_id();
-		}
+	public static function add( $activity, $user_id, $visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
```
pfefferle:
```suggestion
	public static function add( Activity $activity, $user_id, $visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
```

Should we add a type hint for `$activity`?

pfefferle: We require an `Activity` now, so we could skip verification in the function?

obenland: We took away an argument already, so we might as well

tests/includes/rest/class-test-outbox-controller.php
```diff
@@ -698,7 +698,7 @@ public function test_get_items_meta_query_for_non_privileged_users() {
 		$this->assertEquals( 200, $response->get_status() );
 		$this->assertSame( 1, (int) $data['totalItems'] );
 		$this->assertCount( 1, $data['orderedItems'] );
-		$this->assertSame( 'https://example.org/activity/1', $data['orderedItems'][0]['object']['id'] );
+		$this->assertSame( 'https://example.org/note/1', $data['orderedItems'][0]['object']['id'] );
```
obenland, line 701: @pfefferle This is still changed…

pfefferle: I try to find out if it is a bug now or was before... 🤔

pfefferle: @obenland this is good to go! the test stores the full `Activity`, so the old code took that and surrounded it with another `Activity`, that is why it returned the `Activity` ID. Now that we do no longer surround the stored `Objects` (`_activitypub_activity_type` is the same as the `Activity` Type), it returns the `Object` ID.

(I hope it was the AI that messed that up in the first place 🫣)

pfefferle:
* [X] Outbox API Collection

pfefferle (dismissed):
this turned out to be a pretty nice improvement and I am glad we did that!

but because I also worked on that PR, maybe it is good that @mattwiebe has a second look at the changes!

includes/collection/class-outbox.php
```diff
@@ -25,41 +24,34 @@ class Outbox {
...
-		$title                 = $activity_object->get_name() ?? $activity_object->get_content();
-		$activitypub_object_id = $activity_object->get_id();
-
-		if ( ! $title && is_activity( $activity_object ) && $activity_object->get_object() instanceof \Activitypub\Activity\Base_Object ) {
-			$title                 = $activity_object->get_object()->get_name() ?? $activity_object->get_object()->get_content();
-			$activitypub_object_id = $activity_object->get_object()->get_id();
-		}
+	public static function add( Activity $activity, $user_id, $visibility = ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC ) {
+		$actor_type = Actors::get_type_by_id( $user_id );
+		$object_id  = is_string( $activity->get_object() ) ? $activity->get_object() : $activity->get_object()->get_id();
```
pfefferle:
```suggestion
		$object_id = false;
		if ( $activity->get_object() ) {
			$object_id  = is_string( $activity->get_object() ) ? $activity->get_object() : $activity->get_object()->get_id();
		}
```

An `object` is not required!

obenland: Any repercussions from not having an object id saved?

includes/collection/class-outbox.php
```diff
@@ -296,19 +289,44 @@ public static function maybe_get_activity( $outbox_item ) {
...
 	 */
-	private static function recursively_get_title( $activity_object ) {
+	private static function get_object_title( $activity_object ) {
 		if ( is_string( $activity_object ) ) {
 			$post_id = url_to_postid( $activity_object );
 
 			return $post_id ? get_the_title( $post_id ) : '';
 		}
 
+		if ( null === $activity_object ) {
```
pfefferle: I would make this check more general and use it to fail early:

```PHP
if ( ! $activity_object ) { 
	return '';
}
```

obenland: Fixed in [794e231](https://github.com/Automattic/wordpress-activitypub/pull/1474/commits/794e2316b52a17896a09102dcfe3e013702869e2)

includes/collection/class-outbox.php
```diff
@@ -278,4 +288,51 @@ public static function maybe_get_activity( $outbox_item ) {
...
+			return '';
+		}
+
+		if ( is_string( $activity_object ) ) {
+			$post_id = url_to_postid( $activity_object );
+
+			return $post_id ? get_the_title( $post_id ) : '';
+		}
+
+		$title = $activity_object->get_name() ?? $activity_object->get_content();
```
pfefferle, line 330: maybe also have a check for `is_object` there might be the rare case that this might be an `array`!?!

obenland: If there's a case where that's an array we should fix that, I think. We know that `$activity_object` is the result of `Generic_Object::init_from_array()` if the data is an array, so if that ends up being something else we'd want to know?

includes/collection/class-outbox.php
```diff
@@ -278,4 +288,51 @@ public static function maybe_get_activity( $outbox_item ) {
...
+			return $activity->get_object()->get_id();
+		}
+
+		// Rare.
+		if ( is_string( $activity->get_object() ) ) {
+			return $activity->get_object();
+		}
+
+		// Exceptional.
+		return $activity->get_actor() ?? $activity->get_id();
```
pfefferle, line 310: wouldn't this return the generated ID with the hash when the actor is empty?

obenland: Yes, if it ever got to that. I'm not sure it will?

pfefferle: this is a waste of data and I would even recommend to remove this generation of a default ID from the code!

obenland: Follow-up PR?

pfefferle: the default ID was only because we had no persistent outbox activities and it was required to have a unique id by remote services.

pfefferle: but then, this snippet would return `null` in that case and store an empty id and we would again have to deal with it! I think it is important to have this discussion before merging this!

obenland: Sure! What would you suggest as a more appropriate fallback to `actor` in Activities without `object`?

includes/collection/class-outbox.php
pfefferle: but this seems to not be done by `Outbox::get_activity`!?!

includes/collection/class-outbox.php
pfefferle: not setting it!

obenland: I'm not sure we even need a fallback for it? [Reviewing the spec](https://www.w3.org/TR/activitystreams-vocabulary/#activity-types), it looks like only `Question` doesn't have either an `object` or `actor` field.

obenland: We could use `name` in that case?

pfefferle: It is not about if an `Activity` supports an `object`, it is about if it provides one. And because it is not required it might fail or return something unexpected.

The `name` is no valid ID and might cause issues if we might use the meta as ID in the future.

tests/includes/collection/class-test-outbox.php
```diff
@@ -120,7 +127,7 @@ public function activity_object_provider() {
...
 			),
```
pfefferle, line 166: I think we should also test adding full activities, like:

```PHP
array(
	array(
		'@context' => 'https://www.w3.org/ns/activitystreams',
		'type'     => 'Create',
		'object'   => array(
			'id'       => 'https://example.com/4',
			'type'     => 'Note',
			'content'  => '<p>Test content</p>',
		),
	),
	null,
	1,
	'{...}',
)
```

But this needs refactoring of the function or we need a new to test especially full activities.

pfefferle: I added a simple test... should be enough to improve over time!

pfefferle (approved):

PR #1477, opened 2025-03-19
Fix spacing issue

edent:
Current text looks like this:

image

## Proposed changes:

* Add a space for readability

-   [x] Automatically create a changelog entry from the details below.

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Added - for new features
-   [ ] Changed - for changes in existing functionality
-   [ ] Deprecated - for soon-to-be removed features
-   [ ] Removed - for now removed features
-   [x] Fixed - for any bug fixes
-   [ ] Security - in case of vulnerabilities

#### Message

Improved readability in Mastodon Apps plugin string.

</details>

PR #1479, opened 2025-03-20
Changelog: allow generating a changelog entry from the PR description in a fork, too.

jeherve:
## Proposed changes:

See https://github.com/Automattic/wordpress-activitypub/pull/1477#issuecomment-2739498385

External contributors should be able to generate a changelog entry in their fork using the workflow. Let's allow that by switching to triggering the workflow on `pull_request_target`, where the event has the necessary access to be able to run the workflow, but doesn't have the possibility of running the workflow from code they may have modified in their fork.

### Other information:

## Testing instructions:

This cannot really be tested until it gets merged, I'm afraid, since we are changing the workflow itself in this PR.

### DIFF
--- a/.github/changelog/fix-changelog-workflow-forks
+++ b/.github/changelog/fix-changelog-workflow-forks
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Changelog entries: allow automating changelog entry generation from forks as well.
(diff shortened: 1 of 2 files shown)


obenland:
Oh neat, not a big change 👍

obenland (approved):

PR #1482, opened 2025-03-20
Changelog: invite external contributors to add changelog file

jeherve:
Follow-up to #1479.

## Proposed changes:

I was wrong in #1479. We cannot add a file to their branch automatically, unfortunately. We'll consequently take a different approach: if the PR is from a fork and the contributor provided changelog information in the PR description, we'll leave a comment on the PR inviting them to add the changelog file manually.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

This can only be tested in a fork until this PR is merged. Here is how the comment would look like:

image

PR #1480, opened 2025-03-20
A quick change to test changelog fixes

jeherve:
> [!WARNING]
> **Do not merge**

## Proposed changes:

This small PR is only meant to test the changes made in #1479.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

We'll test by trying to provide changelog info below.

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Added - for new features
-   [ ] Changed - for changes in existing functionality
-   [ ] Deprecated - for soon-to-be removed features
-   [ ] Removed - for now removed features
-   [x] Fixed - for any bug fixes
-   [ ] Security - in case of vulnerabilities

#### Message

Changelog: testing changes to the changelog generation flow.

</details>

PR #1483, opened 2025-03-20
Changelog: don't post duplicate comments for external contributors

jeherve:
Follow-up to #1479 and #1482

## Proposed changes:

Hopefully that's the last one in this series! Sorry for the extra review work!

See https://github.com/Automattic/wordpress-activitypub/pull/1480#issuecomment-2739831347

When posting comments to invite contributors to add a changelog entry, we now first check if such a comment was already posted, to avoid duplication.

### Other information:

## Testing instructions:

This can only be tested in a fork.


obenland (approved):
The devil is always in the details! :) Thanks for sticking with it

PR #1484, opened 2025-03-20
Fix PHPStan errors

obenland:
Cherry-picked from 1466.

## Proposed changes:

* Comments and type fixes across the codebase

This is probably about as much as we can get out of there at this point without compromising other things.

### DIFF
--- a/includes/rest/class-nodeinfo-controller.php
+++ b/includes/rest/class-nodeinfo-controller.php
@@ -156,8 +156,8 @@ class Nodeinfo_Controller extends \WP_REST_Controller {
 			'usage'             => array(
 				'users'         => array(
 					'total'          => get_total_users(),
-					'activeHalfyear' => get_active_users( '6 month ago' ),
-					'activeMonth'    => get_active_users( '1 month ago' ),
+					'activeHalfyear' => get_active_users( 6 ),
+					'activeMonth'    => get_active_users(),
 				),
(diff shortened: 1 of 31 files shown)


includes/rest/class-nodeinfo-controller.php
```diff
@@ -156,8 +156,8 @@ public function get_version_2_0() {
 			'usage'             => array(
 				'users'         => array(
 					'total'          => get_total_users(),
-					'activeHalfyear' => get_active_users( '6 month ago' ),
-					'activeMonth'    => get_active_users( '1 month ago' ),
+					'activeHalfyear' => get_active_users( 6 ),
```
pfefferle, line 159: Have you changed the function lately, without changing the calls? This looks so new to me!

pfefferle (approved):

includes/rest/class-nodeinfo-controller.php
obenland: Introduced [two years ago](https://github.com/Automattic/wordpress-activitypub/pull/530/files#diff-1c08b27bcf43583c75cbc442bc401638be3427586fe544f3389e22c9979fa1afR484-R486), and [not changed since](https://github.com/Automattic/wordpress-activitypub/blame/53626f62b8be4b42836aa91c0ab59bc6f2733f26/includes/functions.php#L586) :)

PR #1485, opened 2025-03-20
Interaction: Fix deleting comments

obenland:
While reading through code for the PHPStan changed I came across `get_interactions_by_actor()` and noticed the use of `WP_Comment_Query` in a way that I've not considered it before. Decided to write a test to make sure it works and found an unrelated bug :D

## Proposed changes:

* Updates comment query arg to catch comment author url.
* Adds test for `get_interactions_by_actor()`

### Other information:

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Comments from Fediverse actors will now be purged as expected.

</details>

### DIFF
--- a/.github/changelog/1485-from-description
+++ b/.github/changelog/1485-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Comments from Fediverse actors will now be purged as expected.
--- a/includes/collection/class-interactions.php
+++ b/includes/collection/class-interactions.php
@@ -179,20 +179,19 @@ class Interactions {
 			$actor = object_to_uri( $meta['url'] );
 		}
 
-		$args          = array(
-			'nopaging'   => true,
-			'author_url' => $actor,
+		$args = array(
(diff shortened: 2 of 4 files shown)


pfefferle:
So haven't this worked before?

obenland:
I don't think so? Not sure.

pfefferle (dismissed):

obenland:
I did some more testing. Turns out I had the unit test set up wrong. Setting `user_id` when creating the comment led to false results. Not sure why, but it probably doesn't matter either, as fediverse followers won't have a user_id either way.

So in the end, this is a bit of a nothingburger. The code got changed and we have a unit test for it, but it's not an improvement or fix. We can probably just close this.

pfefferle (approved):
I am happy to also merge small improvements!

(and I am glad that the code was not completely wrong ☺️)

(and I like the word "nothingburger" a lot 😍)

PR #1486, opened 2025-03-20
Use real Attchment types instead of `Document`

pfefferle:
Use `Audio` and `Video` instead of `Document`.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Use `Audio` and `Video` type for Attachments, instead of the very generic `Document` type.

</details>


obenland:
Would a unit test help to make sure we get the expected output? Probably more cosmetics, as it's really more about third-party consumers understanding it correctly?

pfefferle:
We had some isses with attachments not showing up properly on Mastodon and I thought it might be because of them changing the requirments... but it seems to be not the solution to that because Mastodon is still using `Document` 🤔

mattwiebe:
I like the idea of sending proper Attachment types. But maybe not if it doesn't fix anything heh

includes/transformer/class-post.php
```diff
@@ -1018,7 +1018,7 @@ public function wp_attachment_to_activity_attachment( $media ) {
 			case 'audio':
 			case 'video':
 				$attachment = array(
-					'type'      => 'Document',
+					'type'      => \ucfirst( $mime_type_parts[0] ),
```
obenland: Should we use `strtok()` to get the media type directly?

obenland (approved):

PR #1493, opened 2025-03-21
Do a little advertising for WordPress

### LINKED ISSUE #916: Feature Request: Add a "Powered By WordPress" link in the profile
### What

Mockup
Screenshot 2024-10-01 at 10 37 47

### Why

To make more people aware that you can use WordPress+ActivityPub to appear in Mastodon.

### How

Add this new link by default but the user can edit or delete it.

pfefferle:
Add

| Powered by | WordPress |
| -------- | ------- |

As an extra-field.

This is only for new installations/new users and can be removed.

Fixes #916

## Proposed changes:

* Add a migration that adds it as a default

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

A default Extra-Filed to do a little advertising for WordPress.

</details>

### DIFF
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -815,19 +815,6 @@ class Migration {
 
 		// Add a default extra field for each user.
 		foreach ( $users as $user ) {
-			// check if the extra field already exists.
-			$extra_field = get_posts(
-				array(
-					'post_type'      => Extra_Fields::USER_POST_TYPE,
-					'author'         => $user->ID,
-					'posts_per_page' => 1,
-				)
-			);
-
-			if ( $extra_field ) {
-				continue;
-			}
-
 			wp_insert_post(
 				array(
 					'post_type'    => Extra_Fields::USER_POST_TYPE,
@@ -837,19 +824,6 @@ class Migration {
 					'post_content' => $content,
 				)
 			);
-		}
-
-		// Add a default extra field for the blog user.
-		$extra_field = get_posts(
-			array(
-				'post_type'      => Extra_Fields::BLOG_POST_TYPE,
-				'author'         => 0,
-				'posts_per_page' => 1,
-			)
-		);
-
-		if ( $extra_field ) {
-			return;
 		}
 
 		wp_insert_post(
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -811,7 +811,7 @@ class Migration {
 		);
 
 		$title   = __( 'Powered by', 'activitypub' );
-		$content = __( 'WordPress ❤️', 'activitypub' );
+		$content = __( 'WordPress', 'activitypub' );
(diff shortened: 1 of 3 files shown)


includes/class-migration.php
```diff
@@ -798,6 +800,69 @@ private static function add_notification_defaults() {
...
+	 */
+	private static function add_default_extra_field() {
+		$users = \get_users(
+			array(
+				'capability__in' => array( 'activitypub' ),
+			)
+		);
+
+		$title   = __( 'Powered by', 'activitypub' );
+		$content = __( 'WordPress ❤️', 'activitypub' );
```
obenland: I'd probably leave out the emoji, as quirky as it may feel :)

includes/class-migration.php
```diff
@@ -798,6 +800,69 @@ private static function add_notification_defaults() {
...
+					'author'         => $user->ID,
+					'posts_per_page' => 1,
+				)
+			);
+
+			if ( $extra_field ) {
+				continue;
+			}
+
+			wp_insert_post(
```
obenland: Do you anticipate this running any other time, aside from first activation? If not, I wouldn't bother looking up existing posts and instead just add them.

tests/includes/class-test-migration.php
```diff
@@ -733,4 +734,165 @@ public function mock_webfinger() {
...
+		// Check the extra field for the user.
+		$user_fields = get_posts(
+			array(
+				'post_type'      => Extra_Fields::USER_POST_TYPE,
+				'author'         => $user_id,
+				'posts_per_page' => -1,
+			)
+		);
+
+		$this->assertCount( 1, $user_fields, 'Es sollte ein Extra-Feld für den Benutzer erstellt worden sein' );
```
obenland: Those messages are in German :)

includes/class-migration.php
pfefferle: I want to get sure to only add the default if no other extra-fields are added already. This also helps with tests 🫣

includes/class-migration.php
pfefferle: This is the fediverse!!! YOU HAVE TO USE EMOJIS!!! 😉

includes/class-migration.php
obenland: You could run `_delete_all_posts()` in `tear_down()` to account for that.
What are scenarios where you'd see that happening?

obenland (dismissed):
Thanks! I think this is good to go as is. Optionally you could remove the translation function around `WordPress` and backslash the functions that don’t have it yet. No blockers.

obenland (approved):
✨

PR #1494, opened 2025-03-21
Add: `#FediBlog` Hashtag support

pfefferle:
I really like the idea of adding a default hashtag to all posts, so that we can make blog-posts more visible and improve discoverability.

https://mastodon.social/@tchambers@indieweb.social/114197659082334414

Screenshot 2025-03-21 at 15 17 10

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

The option to automatically add the `#FediBlog` to every federated Blog-Post (only Articles, not Notes), to improve the visibility of long form content. 

</details>


pfefferle:
@obenland @mattwiebe what do you think?

jeherve:
In that Fediverse discussion I asked:

> should we add the tag to long-form content only, or to any posts published via Ghost or WordPress (like this one)?

Adding it to everything may end up being a bit noisy, and highlight some irrelevant content (like replies to unrelated conversations, for example).

I would personally keep it a conscious choice on a per-post basis for site owners.

pfefferle:
Fair enough! Maybe I try to make a sidebar item to add it more easily!?!

jeherve:
More discussion about this happening here:
https://mastodon.social/@pfefferle/114205164158375166

includes/transformer/class-post.php
jeherve, line 1: I'm curious why you went with that dummy URL? Wouldn't it be useful to point to the right tag archive page, especially on sites that will have a mix of short content and longform articles?

pfefferle: It depends! the current implementation does not really add the tag to the post, it simply adds it to the JSON output. This way it is findable on mastodon, but will not be added locally.

I am not sure that this is a good or bad idea, but I had some discussions, that blog owners does not want to have "default" tags in their blog system.

pfefferle: @jeherve but I see that this might be inconsistent and maybe not be a good idea. Having tags that are only visible on the fediverse might be the wrong signal, because it might be misused by spammers. 🤔

pfefferle:
@obenland @jeherve should this be part of the core plugin or is this something that could live in a custom extension?

jeherve:
> @obenland @jeherve should this be part of the core plugin or is this something that could live in a custom extension?

Could we have it in the core plugin, but as an option one can toggle from the Advanced settings panel for now?

pfefferle:
🤔

I am not sure that this is something that should live in the Advanced Settings. I would keep the Advanced Settings for critical options that might cause problems when used wrong.

obenland:
Let's close it

PR #1500, opened 2025-03-24
Fix: Announces

pfefferle:
With the re-write of how we store Activities in the Outbox, the Announces do no longer work. This PR fixes that.

## Proposed changes:

* Handle `Activity` instead of `Activity`>`Object`

### Other information:

## Testing instructions:

* Publish a new Post
* Check Outbox for `Create` and `Announce` Activity

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

`Scheduler::schedule_announce_activity` to handle Activities instead of Activity-Objects.

</details>

### DIFF
--- a/includes/class-scheduler.php
+++ b/includes/class-scheduler.php
@@ -7,6 +7,7 @@
 
 namespace Activitypub;
 
+use Activitypub\Activity\Activity;
 use Activitypub\Scheduler\Post;
 use Activitypub\Scheduler\Actor;
 use Activitypub\Scheduler\Comment;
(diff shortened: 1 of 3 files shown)


obenland:
Oh, so this is a bug in trunk only, not production, yes?

includes/class-scheduler.php
```diff
@@ -411,28 +412,26 @@ public static function schedule_announce_activity( $outbox_activity_id, $activit
...
 
-		// Check if the object is an article, image, audio, video, event or document and ignore profile updates and other activities.
-		if ( ! in_array( $activity_object->get_type(), array( 'Note', 'Article', 'Image', 'Audio', 'Video', 'Event', 'Document' ), true ) ) {
+		if ( ! is_object( $activity->get_object() ) ) {
 			return;
 		}
 
-		$transformer = Factory::get_transformer( $activity_object );
-		if ( ! $transformer || \is_wp_error( $transformer ) ) {
+		// Check if the object is an article, image, audio, video, event or document and ignore profile updates and other activities.
```
obenland:
```suggestion
		// Check if the object is an article, image, audio, video, event, or document and ignore profile updates and other activities.
```
Oxford comma FTW!

obenland:
How did you find out they were broken? We should add a test for that.

pfefferle:
Oh, we have not released that yet! 😮‍💨

I am glad I found it, before we released it!

obenland:
Me too! 

I'm curious, should it not have been backwards compatible and still work with the previous function signature of `add_to_outbox()`? What's missing that made them not work?

pfefferle:
The Action now always passes an Activity and no longer an Activity-Object.

mattwiebe:
Happy to approve it if it fixes a bug but unclear when this was broken?

pfefferle:
This is still WIP! It fixes a bug in trunk but I will clean up and add tests tomorrow! I will ping you when I think it is ready to review.

Thanks for already having a look @mattwiebe and @obenland

obenland (approved):

PR #1501, opened 2025-03-24
Fix: (Pixelfed) Follow

pfefferle:
`add_to_outbox` only supports generic visibilities like for example `ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE` or 'ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC'. For custom audiences, we would have to add the `to`, `cc`, ... fields of the Activity. With the former implementation, we were only able to set the audience for the `Activity`>`Object` and these were up-merged to the Activity afterwards. In the case of an `Accept` of a `Follow` we were forced to update the `Follow`-`Activity`. But because the `Follow` was sent by a remote platform, adding a new field (`to` for example) would change the scope and could cause problems for remote verification.

With the rewrite to store the full `Activity` instead of the `Object`, we are now able to add the audience to this `Activity` and keep the `Object` untouched. 

## Proposed changes:

* Do not rely on the Transformer
* Create the full Activity
* Add audience to Activity

### Other information:

## Testing instructions:

* Send Follow `Activity` to inbox
* Check Outbox for `Accept`-`Activity`

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Properly set `to` audience of `Activity` instead of changing the `Follow` Object.

</details>

### DIFF
--- a/tests/includes/activity/class-test-activity.php
+++ b/tests/includes/activity/class-test-activity.php
@@ -159,12 +159,12 @@ class Test_Activity extends \WP_UnitTestCase {
 		$activity->set_actor( Actors::get_by_id( $user_id )->get_id() );
 		$activity->set_object( $activity_object );
 
-		$this->assertEquals( 'https://example.com/author/123', $activity->get_to()[0] );
+		$this->assertContains( 'https://example.com/author/123', $activity->get_to() );
 
 		$activity->set_to( array( 'https://example.com/author/456' ) );
 		$activity->set_object( $activity_object );
 
-		$this->assertEquals( 'https://example.com/author/456', $activity->get_to()[0] );
+		$this->assertContains( 'https://example.com/author/456', $activity->get_to() );
 
 		// Delete user.
 		\wp_delete_user( $user_id );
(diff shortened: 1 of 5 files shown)


tests/includes/activity/class-test-activity.php
```diff
@@ -132,4 +133,40 @@ public function test_activity_object_id() {
...
+			'object' => 'https://example.com/post/123',
+			'to'     => array( 'https://example.com/author/123' ),
+		);
+
+		$activity = new Activity();
+		$activity->set_type( 'Accept' );
+		$activity->set_actor( Actors::get_by_id( $user_id )->get_id() );
+		$activity->set_object( $activity_object );
+
+		$this->assertEquals( 'https://example.com/author/123', $activity->get_to()[0] );
```
obenland: I think this is fine, no need to change. For future tests, where you might not have control over the array that you want to test, `assertContains` might be helpful.

obenland (dismissed):

obenland (approved):

PR #1502, opened 2025-03-24
Move: Add a Mastodon importer

obenland:
## Proposed changes:

* Adds an importer for Mastodon archives.

### Other information:

## Testing instructions:

* Go to Mastodon > Preferences > Import and Export and request your archive.
* Activate this PR.
* Go to Tools > Import and select the Mastodon importer.
* Upload the archive zip file and click Import.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Added a Mastodon importer to move your Mastodon posts to your WordPress site.

</details>

@krafit


includes/wp-admin/import/class-mastodon.php
```diff
@@ -0,0 +1,242 @@
...
+			}
+
+			if ( ! $user_id ) {
+				$users = get_users(
+					array(
+						'fields'     => 'ID',
+						// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+						'meta_query' => array(
+							array(
+								'key'     => $GLOBALS['wpdb']->get_blog_prefix() . 'activitypub_also_known_as',
```
mattwiebe: So this will require this value to have been set first? 

And what about sites that only have the Blog User active?

obenland: > So this will require this [also known as] value to have been set first?

Yes! We should probably add an author selection step, like the WP importer, to overwrite that, with the user we can find through this query as the default.

> And what about sites that only have the Blog User active?

Yeah, great question! Maybe another reason to add the author selection step?

obenland:
@mattwiebe @pfefferle My Mastodon archive is pretty basic, no attachments, polls, etc. Not sure if you have a test user with a richer archive that I could test with?

pfefferle:
Happy to test this tomorrow morning with a backup of my account!

obenland:
Heads up that it won't import attachments as of yet, since I didn't have an example to code against.

pfefferle:
~I can't really test it. It fails with:~

~Sorry, there has been an error.~
~Missing author.~

## Update

It was the missing `alsoKnownAs` filed. We should explain that requirement in the description and also what "version" of the URL the tool needs and/or the script should be able to support the URL `https://mastodon.social/@<username>` and the ID `https://mastodon.social/users/<username>`.

When running the importer, I see a ton of `Warnings` like this:

```
Warning: Undefined property: stdClass::$replies in /var/www/html/wp-content/plugins/activitypub/includes/wp-admin/import/class-mastodon.php on line 205 Warning: Attempt to read property "first" on null in /var/www/html/wp-content/plugins/activitypub/includes/wp-admin/import/class-mastodon.php on line 205 Warning: Attempt to read property "items" on null in /var/www/html/wp-content/plugins/activitypub/includes/wp-admin/import/class-mastodon.php on line 205 Warning: Undefined property: stdClass::$replies in /var/www/html/wp-content/plugins/activitypub/includes/wp-admin/import/class-mastodon.php on line 205
```

Except from the missing attachments, all notes seem to show up nicely!

## Problem

The script does not care about private or hidden posts and imports all of them as public!

I think for a first beta it might be enough to check for `"to":["https://www.w3.org/ns/activitystreams#Public"]` and ignore all others.

## Feature Request

If you have a "normal" blog with a focus on `Articles` this can mess up your post table, so we should at least provide an option to select a post-format (to be able to set them as `Status` or `Aside` posts) or a custom-post-type to store the import. 

## Kudos

Very cool feature!

...and the code checks if a post was already imported, nice! ❤️

includes/wp-admin/import/class-mastodon.php
```diff
@@ -0,0 +1,242 @@
...
+							return ltrim( $tag->name, '#' );
+						}
+
+						return '';
+					},
+					$post->object->tag
+				),
+			);
+
+			$post_exists = post_exists( '', $post_data['post_content'], $post_data['post_date'], $post_data['post_type'] );
```
pfefferle: what about using the original ID URL as GUID? might be faster?

includes/wp-admin/import/class-mastodon.php
```diff
@@ -0,0 +1,242 @@
...
+	public static function import_posts() {
+		$user_id = 0;
+
+		// phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+		foreach ( self::$outbox->orderedItems as $post ) {
+			// Skip boosts.
+			if ( 'Announce' === $post->type ) {
+				continue;
+			}
+
```
pfefferle, line 285: Check if post is public!

pfefferle (changes requested):

includes/wp-admin/import/class-mastodon.php
obenland: What would be the best way to do that? See whether `https://www.w3.org/ns/activitystreams#Public` is in either `to` or `cc`?
Are there any non-public posts in the outbox?

obenland:
> If you have a "normal" blog with a focus on Articles this can mess up your post table, so we should at least provide an option to select a post-format (to be able to set them as Status or Aside posts) or a custom-post-type to store the import.

It's currently set to [save all of them as Status](https://github.com/Automattic/wordpress-activitypub/pull/1502/files#diff-34b205a466245fe788fad447d191ed46b6c18a6d57b02224cc1ca72936da0e63R203), is that what you'd expect?

includes/wp-admin/import/class-mastodon.php
obenland: Ah, just saw you recommended just that!

pfefferle: In the `outbox.json` export file are non public direct messages!

Does that help? https://github.com/Automattic/wordpress-activitypub/blob/93630017325512bacfad3d6b79128d6c3aca4fd1/includes/functions.php#L562

includes/wp-admin/import/class-mastodon.php
mattwiebe: Author selection step sounds like a very good idea 👍🏻

obenland: It exists now, if you want to give it another spin!

includes/wp-admin/import/class-mastodon.php
```diff
@@ -0,0 +1,411 @@
+<?php
+/**
+ * Mastodon importer file.
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\WP_Admin\Import;
+
+use function Activitypub\site_supports_blocks;
```
pfefferle, line 11:
```suggestion
use function Activitypub\is_activity_public;
use function Activitypub\site_supports_blocks;
```

includes/wp-admin/import/class-mastodon.php
```diff
@@ -0,0 +1,411 @@
...
+	 */
+	public static function import_posts() {
+		// phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+		foreach ( self::$outbox->orderedItems as $post ) {
+			// Skip boosts.
+			if ( 'Announce' === $post->type ) {
+				continue;
+			}
+
+			if ( ! \Activitypub\is_activity_public( \get_object_vars( $post ) ) ) {
```
pfefferle:
```suggestion
			if ( ! is_activity_public( \get_object_vars( $post ) ) ) {
```

includes/wp-admin/import/class-mastodon.php
```diff
@@ -0,0 +1,412 @@
...
+					function ( $tag ) {
+						if ( 'Hashtag' === $tag->type ) {
+							return \ltrim( $tag->name, '#' );
+						}
+
+						return '';
+					},
+					$post->object->tag
+				),
+			);
```
pfefferle, line 310: Maybe we can add a filter, so that implementers can change params (for example changing the `post_type` or add `meta_input`!?!

```suggestion
			);
			
			$post_data = \apply_filters( 'activitypub_import_post_data', $post_data );
```

includes/wp-admin/import/class-mastodon.php
```diff
@@ -0,0 +1,412 @@
...
+				'tags_input'   => \array_map(
+					function ( $tag ) {
+						if ( 'Hashtag' === $tag->type ) {
+							return \ltrim( $tag->name, '#' );
+						}
+
+						return '';
+					},
+					$post->object->tag
+				),
```
pfefferle, line 309: What about adding an indicator that it is/was an import, like we have with comments?

```PHP
'meta_input' => array(
	'protocol' => 'activitypub',
	'source_id' => 'https://mastodon.social/...'
)
```

So that we are able to query these posts and to `(C)RUD` them later on?

includes/wp-admin/import/class-mastodon.php
obenland: Added in [9f41674](https://github.com/Automattic/wordpress-activitypub/pull/1502/commits/9f41674246105690ab68fd2a532838feeaf623e8)

includes/wp-admin/import/class-mastodon.php
obenland: I've added `_source_id` d5e633d45524e503e0131d85659a22348f1674e6 for now. Not sure that's really needed?

We'd want any meta key to be "private" (prefixed with an underscore) and ideally they should also be registered I guess. I've skipped the registration for now, since we don't expect it to be used/updated/managed.

pfefferle:
I get an error:

```PHP
Warning: file_get_contents(/var/www/html/wp-content/import/outbox.json_/outbox.json): Failed to open stream: No such file or directory in /var/www/html/wp-content/plugins/activitypub/includes/wp-admin/import/class-mastodon.php on line 198 Warning: Attempt to read property "orderedItems" on null in /var/www/html/wp-content/plugins/activitypub/includes/wp-admin/import/class-mastodon.php on line 241 Warning: foreach() argument must be of type array|object, null given in /var/www/html/wp-content/plugins/activitypub/includes/wp-admin/import/class-mastodon.php on line 241 
```

obenland:
Interesting, looks like it didn't create a temp folder for you… I'll take a look

pfefferle:
Might be an issue with my local env then?

obenland:
Possibly. What's the return value of `\unzip_file()` in line 189? Does it return true?

Do you upload the export zip file, or just the outbox.json?

pfefferle:
Oh! Maybe I should re-phrase my issue: We should check for a ZIP file after the upload!

😂

pfefferle:
Sadly we can't limit the upload to `zip`, there seems to be no filter to use `<input type="file" accept=".zip" />`

Maybe we should change the output a bit? It looks quite overwhelming and at first it looked like errors/warnings ;)

Screenshot 2025-03-31 at 15 14 41

Maybe we should limit the output to something like: `343 posts imported`!?!

obenland:
@pfefferle Could you try again and see if the line break helps with how the output looks?

pfefferle:
Now with the line breaks I can see the errors/warnings better :)

`Warning: Trying to access array offset on value of type bool in /var/www/html/wp-content/plugins/activitypub/includes/wp-admin/import/class-mastodon.php on line 400 Deprecated: ltrim(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/wp-includes/formatting.php on line 4467 Imported post https://mastodon.social/users/pfefferle/statuses/110859046889772039.`

Still not sure if it helps to see (in my case) 3271 lines of imported URLs, but should not be a blocker.

obenland:
We could display `.` for each post, like SVN does for each file when committing. I was just looking for a low-key progress indicator.

pfefferle (approved):
@obenland awesome! ❤️

@krafit thanks for inspiring @obenland 😉

PR #1504, opened 2025-03-25
Onboarding: Allow skipping the welcome page

pfefferle:
The "Welcome" page should be the page, the admin ends up, after the onboarding process.

This PR allows to skip this page, to directly call the `settings` page instead.

Screenshot 2025-03-26 at 09 13 24

Fixes #

## Proposed changes:

* Re-use the same mechanism that Core uses to dismiss the welcome banner on the dashboard.
* When hidden, the page can be re-activated through the screen-options.
* The tab-menu only shows up if there is more than one tab.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

The option to show/hide the "Welcome Page".

</details>

### DIFF
--- a/assets/css/activitypub-admin.css
+++ b/assets/css/activitypub-admin.css
@@ -227,8 +227,7 @@ input.blog-user-identifier {
 	top: 0px;
 	right: 0px;
 	font-size: 13px;
-	padding: 2px 15px 10px 20px;
-	line-height: 1.23076923;
+	padding: 0 5px 0 20px;
 	text-decoration: none;
 	z-index: 1;
 }
--- a/assets/css/activitypub-admin.css
+++ b/assets/css/activitypub-admin.css
@@ -1,6 +1,7 @@
 .activitypub-settings {
 	max-width: 800px;
 	margin: 0 auto;
+	position: relative;
 }
 
 .settings_page_activitypub .notice {
@@ -219,4 +220,25 @@ input.blog-user-identifier {
 .repost .dashboard-comment-wrap .comment-author,
 .like .dashboard-comment-wrap .comment-author {
 	margin-block: 0;
+}
+
+.activitypub-settings .welcome-tab-close {
+	position: absolute;
+	top: 0px;
+	right: 0px;
+	font-size: 13px;
+	padding: 2px 15px 10px 20px;
+	line-height: 1.23076923;
+	text-decoration: none;
+	z-index: 1;
+}
+
+.activitypub-settings .welcome-tab-close::before {
+	position: absolute;
+	top: 0px;
+	left: 0;
+	transition: all .1s ease-in-out;
+	font: normal 16px/20px dashicons;
+	content: '\f335';
+	font-size: 20px;
 }
--- a/includes/wp-admin/class-settings.php
+++ b/includes/wp-admin/class-settings.php
@@ -482,7 +482,7 @@ class Settings {
 		}
 
 		$screen_settings = '<fieldset>
-		<legend class="screen-layout">' . __( 'Settings Pages', 'activitypub' ) . '</legend>
+		<legend class="screen-layout">' . \esc_html__( 'Settings Pages', 'activitypub' ) . '</legend>
 		<p>
 			' . __( 'Some settings pages can be shown or hidden by using the checkboxes.', 'activitypub' ) . '
 		</p>
--- a/includes/wp-admin/class-settings.php
+++ b/includes/wp-admin/class-settings.php
@@ -23,8 +23,6 @@ class Settings {
 		\add_action( 'admin_init', array( self::class, 'register_settings' ), 11 );
 		\add_action( 'admin_menu', array( self::class, 'add_settings_page' ) );
 
-		\add_filter( 'screen_settings', array( self::class, 'add_screen_option' ), 10, 2 );
-
 		\add_filter(
 			'screen_options_show_submit',
 			function ( $show_submit, $screen ) {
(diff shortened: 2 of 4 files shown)


obenland:
Interesting! When I first used the plugin I remember thinking that this would be nice option to have, to just hide the welcome.

Some initial thoughts:
* Maybe we could re-use the same mechanism that Core uses to dismiss notices? It feels to me like a user-level setting.
* Instead of the paragraph and primary button, how about a secondary button or link-button floated to the top right that just says "Hide"?
* When hidden, should we add that information to the hep tab?
* When Welcome is hidden and actor mode doesn't include the blog user, we're left with Settings being the only tab—making tabs obsolete. In that case we shouldn't show them.

pfefferle:
thanks for the detailed and early feedback! I agree with all 4 points!

pfefferle:
@obenland I added all the recommended changes.

* [X] Maybe we could re-use the same mechanism that Core uses to dismiss notices? It feels to me like a user-level setting.
* [X] Instead of the paragraph and primary button, how about a secondary button or link-button floated to the top right that just says "Hide"?

I used the same mechanism as used for the Welcome banner on the dashboard.

Screenshot 2025-03-26 at 09 03 58

* [X] When hidden, should we add that information to the hep tab?

I enabled the "Screen Options" tab and added a checkbox to show/hide the "Welcome Page". I would love to also use that to show/hide the "Advanced Settings": #1449 

Screenshot 2025-03-26 at 09 04 53

* [X] When Welcome is hidden and actor mode doesn't include the blog user, we're left with Settings being the only tab—making tabs obsolete. In that case we shouldn't show them.

Screenshot 2025-03-26 at 09 02 09

obenland:
SCREEN OPTIONS!! 😍

includes/wp-admin/class-settings.php
```diff
@@ -226,25 +241,44 @@ public static function register_settings() {
...
 		);
+
+		\register_setting(
+			'activitypub_welcome',
+			'activitypub_welcome_skip',
+			array(
+				'type'    => 'boolean',
+				'default' => false,
+			)
+		);
```
obenland:
```suggestion
```
Looks like this is unused now?

includes/wp-admin/class-settings.php
```diff
@@ -22,6 +22,21 @@ class Settings {
...
+
+		\add_filter(
+			'screen_options_show_submit',
+			function ( $show_submit, $screen ) {
+				if ( 'settings_page_activitypub' !== $screen->id ) {
+					return $show_submit;
+				}
+
+				return true;
+			},
```
obenland: If it's the same to you, let's make that a class method for consistency and so it could be unhooked if need be?

includes/wp-admin/class-settings.php
```diff
@@ -226,25 +241,44 @@ public static function register_settings() {
...
+				'default' => false,
+			)
+		);
 	}
 
 	/**
 	 * Load settings page.
 	 */
 	public static function settings_page() {
+		$default_tab = \get_user_meta( \get_current_user_id(), 'activitypub_show_welcome_tab', true ) ? 'welcome' : 'settings';
```
obenland: There are three calls to get `activitypub_show_welcome_tab` in this method, let's save that in a variable?

includes/wp-admin/class-settings.php
```diff
@@ -226,25 +241,44 @@ public static function register_settings() {
...
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		$tab = isset( $_GET['tab'] ) ? sanitize_key( $_GET['tab'] ) : 'welcome';
+		$tab = isset( $_GET['tab'] ) ? sanitize_key( $_GET['tab'] ) : $default_tab;
 
-		$settings_tabs = array(
-			'welcome'  => array(
+		// Redirect welcome tab to settings if skipped.
+		if ( 'welcome' === $tab && ! \get_user_meta( \get_current_user_id(), 'activitypub_show_welcome_tab', true ) ) {
```
obenland: When would this condition be true? When someone manually added `tab=welcome` to URL?

includes/wp-admin/class-settings.php
```diff
@@ -417,4 +461,49 @@ public static function add_settings_help_tab() {
...
+			$welcome = sanitize_text_field( wp_unslash( $_POST['activitypub_show_welcome_tab'] ) );
+			// Verify screen options nonce.
+			if ( \wp_verify_nonce( $nonce, 'screen-options-nonce' ) ) {
+				$welcome_checked = empty( $welcome ) ? 0 : 1;
+				\update_user_meta( \get_current_user_id(), 'activitypub_show_welcome_tab', $welcome_checked );
+			}
+		}
+
+		$screen_settings = '<fieldset>
+		<legend class="screen-layout">' . __( 'Settings Pages', 'activitypub' ) . '</legend>
```
obenland:
```suggestion
		<legend class="screen-layout">' . \esc_html__( 'Settings Pages', 'activitypub' ) . '</legend>
```
This and the following translation functions should be escaped

assets/css/activitypub-admin.css
```diff
@@ -220,3 +221,24 @@ input.blog-user-identifier {
...
 	margin-block: 0;
 }
+
+.activitypub-settings .welcome-tab-close {
+	position: absolute;
+	top: 0px;
+	right: 0px;
+	font-size: 13px;
+	padding: 2px 15px 10px 20px;
+	line-height: 1.23076923;
```
obenland:
```suggestion
	padding: 0 5px 0 20px;
```
How would you feel about this change? It would help with active and focus styles:

Before | After
--- | ---
Screenshot 2025-03-26 at 8 37 38 AM | Screenshot 2025-03-26 at 8 37 19 AM

includes/wp-admin/class-settings.php
```diff
@@ -226,25 +241,44 @@ public static function register_settings() {
...
-				'label'    => __( 'Settings', 'activitypub' ),
-				'template' => ACTIVITYPUB_PLUGIN_DIR . 'templates/settings.php',
-			),
+			);
+		}
+
+		$settings_tabs['settings'] = array(
+			'label'    => __( 'Settings', 'activitypub' ),
+			'template' => ACTIVITYPUB_PLUGIN_DIR . 'templates/settings.php',
 		);
```
obenland, line 260: Nitpick, feel free to ignore: we could leave that in the initial `$settings_tabs` declaration

includes/wp-admin/class-welcome-fields.php
```diff
@@ -83,6 +83,7 @@ public static function register_welcome_fields() {
 	 */
 	public static function render_welcome_intro_section() {
 		?>
+		<a class="welcome-tab-close" href="<?php echo \esc_url( \admin_url( 'options-general.php?page=activitypub&welcome=0' ) ); ?>" aria-label="Dismiss the welcome page">Dismiss Welcome Page</a>
```
```suggestion
		<a class="welcome-tab-close" href="<?php echo \esc_url( \admin_url( 'options-general.php?page=activitypub&welcome=0' ) ); ?>" aria-label="<?php \esc_attr_e( 'Dismiss the welcome page' ); ?>"><?php \esc_html_e( 'Dismiss Welcome Page' ); ?></a>
```

obenland (changes requested):

obenland:
Oh, there are some inconsistencies with backslashing global functions, I noticed

includes/wp-admin/class-settings.php
pfefferle: but then the welcome page would be second or we would have to order it afterwards

includes/wp-admin/class-settings.php
pfefferle: yes

obenland (approved):
This turned out pretty nice, kudos!

PR #1506, opened 2025-03-25
Return error when Outbox item is invalid

obenland:
`Warning: Attempt to read property "ID" on null`
`Warning: Attempt to read property "post_author" on null`

We use `Outbox::get_activity()` in quite a few places without checking for WP_Errors being returned. So far it looks like it works just fine—just something I noticed while working on this.

## Proposed changes:

* Returns WP_Error when passing a an invalid post to `get_actor()`.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

PHP Warnings when attempting to get an author rom an invalid outbox item.

</details>

### DIFF
--- a/.github/changelog/1506-from-description
+++ b/.github/changelog/1506-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+PHP Warnings when attempting to get an author rom an invalid outbox item.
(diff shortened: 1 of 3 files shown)


pfefferle:
I think it is fine to fight the symptoms, but I am curious why/how this even happens!?!

pfefferle (dismissed):

obenland:
> I am curious why/how this even happens!?!

I am unsure. Logs show it happening on one Outbox API request, seemingly returning `null` for each post, instead of `WP_Post`. The site behaves fine now, so I wonder if there was a db connection issue for that one request?

Happy to punt this for now and see if it's something that happens again. We can always reopen if it does.

pfefferle:
I am super happy to merge this, to at least prevent against errors... I am just curious, because it is a really weird one 😳

PR #1508, opened 2025-03-25
Interactions: Fallback when actor is not defined

obenland:
`Warning: Undefined array key "actor"`

## Proposed changes:

* Adds fallback when `actor` is not part of the activity.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

PHP Warning when there's no actor information in comment activities.

</details>

### DIFF
--- a/includes/collection/class-interactions.php
+++ b/includes/collection/class-interactions.php
@@ -229,7 +229,7 @@ class Interactions {
 	 */
 	public static function activity_to_comment( $activity ) {
 		$comment_content = null;
-		$actor           = object_to_uri( $activity['actor'] );
+		$actor           = object_to_uri( $activity['actor'] ?? null );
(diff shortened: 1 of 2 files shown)


includes/collection/class-interactions.php
```diff
@@ -229,7 +229,7 @@ public static function allowed_comment_html( $allowed_tags, $context = '' ) {
 	 */
 	public static function activity_to_comment( $activity ) {
 		$comment_content = null;
-		$actor           = object_to_uri( $activity['actor'] );
```
pfefferle, line 232: I think it should fail if there is no actor!

obenland: It fails ~a few lines~ one line later, in `get_remote_metadata_by_actor()`

pfefferle (approved):

PR #1510, opened 2025-03-25
Mentions: Standardize around only displaying username

### LINKED ISSUE #1499: Mentions: inconsistent ActivityPub and site representation
### Quick summary

In some of the ActivityPub representations, we display mentions as `<a rel="mention" class="u-url mention" href="https://instance.tld/@example">@<span>example</span></a>`. In others, we display the full account name: `<a rel="mention" class="u-url mention" href="https://instance.tld/@example">@example@instance.tld</a>`. It would be nice if we were consistent in all our implementations.

Similarly, in some situations we include the mention in the ActivityPub representation, but not on the site. 

**References**

https://github.com/Automattic/wordpress-activitypub/blob/00a57c39a4e844949bc4f64a45714d1f21d9645d/includes/class-mention.php#L85
https://github.com/Automattic/wordpress-activitypub/blob/a920b1dc17e8e7baf56871ca68fff13be149344d/includes/transformer/class-comment.php#L124-L126

### Steps to reproduce

1. Write a new post, where you'll mention someone: `@pfefferle@mastodon.social hello there`. 
    * On the site, the mention will be displayed as `<a rel="mention" class="u-url mention" href="https://mastodon.social/@pfefferle">@<span>pfefferle</span></a>`
2. Have that person reply to you ; their reply will appear as a comment.
3. From your site's frontend, reply to them ; do not include any mention in the body of your comment.
    * You will not see any mention in the comment displayed on your site
    * The ActivityPub representation, however, will display `<a rel="mention" class="u-url mention" href="https://mastodon.social/@pfefferle">@pfefferle@mastodon.social</a>`

### Site owner impact

### Severity

Minor

No revenue impact

It would be nice if:

- When a mention is included in the ActivityPub representation, it should also be visible on the site.
- All mentions follow the same standard display. imo it should be `<a rel="mention" class="u-url mention" href="https://mastodon.social/@pfefferle">@<span>pfefferle</span></a>`, where the instance name is **not** displayed. This is the standard in use on Mastodon for example.

obenland:
Fixes #1499.

## Proposed changes:

* Removes host from webfinger to create a username mention in comments.

### Other information:

## Testing instructions:

* On a post with comment from the Fediverse: Reply to that comment and mention the comment author.
* Make sure the mention is username-only.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Standardized mentions to use usernames only.

</details>

### DIFF
--- a/includes/class-mention.php
+++ b/includes/class-mention.php
@@ -82,7 +82,7 @@ class Mention {
 
 			$url = isset( $metadata['url'] ) ? object_to_uri( $metadata['url'] ) : object_to_uri( $metadata['id'] );
 
-			return \sprintf( '<a rel="mention" class="u-url mention" href="%s">@<span>%s</span></a>', esc_url( $url ), esc_html( $username ) );
+			return \sprintf( '<a rel="mention" class="u-url mention" href="%1$s">@%2$s</a>', esc_url( $url ), esc_html( $username ) );
 		}
 
 		return $result[0];
--- a/includes/class-mention.php
+++ b/includes/class-mention.php
@@ -72,15 +72,15 @@ class Mention {
 			! is_wp_error( $metadata ) &&
 			( ! empty( $metadata['id'] ) || ! empty( $metadata['url'] ) )
 		) {
-			$url = isset( $metadata['url'] ) ? object_to_uri( $metadata['url'] ) : object_to_uri( $metadata['id'] );
-
 			$username = ltrim( $result[0], '@' );
 			if ( ! empty( $metadata['name'] ) ) {
 				$username = $metadata['name'];
 			}
 			if ( ! empty( $metadata['preferredUsername'] ) ) {
-				$username = sprintf( '%1$s@%2$s', $metadata['preferredUsername'], wp_parse_url( $url, PHP_URL_HOST ) );
+				$username = $metadata['preferredUsername'];
 			}
+
+			$url = isset( $metadata['url'] ) ? object_to_uri( $metadata['url'] ) : object_to_uri( $metadata['id'] );
 
 			return \sprintf( '<a rel="mention" class="u-url mention" href="%s">@<span>%s</span></a>', esc_url( $url ), esc_html( $username ) );
 		}
--- a/includes/class-mention.php
+++ b/includes/class-mention.php
@@ -72,15 +72,15 @@ class Mention {
 			! is_wp_error( $metadata ) &&
 			( ! empty( $metadata['id'] ) || ! empty( $metadata['url'] ) )
 		) {
+			$url = isset( $metadata['url'] ) ? object_to_uri( $metadata['url'] ) : object_to_uri( $metadata['id'] );
+
 			$username = ltrim( $result[0], '@' );
 			if ( ! empty( $metadata['name'] ) ) {
 				$username = $metadata['name'];
 			}
 			if ( ! empty( $metadata['preferredUsername'] ) ) {
-				$username = $metadata['preferredUsername'];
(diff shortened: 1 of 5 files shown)


includes/class-mention.php
```diff
@@ -72,16 +72,16 @@ public static function replace_with_links( $result ) {
...
 				$username = $metadata['name'];
 			}
 			if ( ! empty( $metadata['preferredUsername'] ) ) {
-				$username = $metadata['preferredUsername'];
+				$username = sprintf( '%1$s@%2$s', $metadata['preferredUsername'], wp_parse_url( $url, PHP_URL_HOST ) );
 			}
 
-			$url = isset( $metadata['url'] ) ? object_to_uri( $metadata['url'] ) : object_to_uri( $metadata['id'] );
-
 			return \sprintf( '<a rel="mention" class="u-url mention" href="%s">@<span>%s</span></a>', esc_url( $url ), esc_html( $username ) );
```
pfefferle:
```suggestion
			return \sprintf( '<a rel="mention" class="u-url mention" href="%s">@%s</a>', esc_url( $url ), esc_html( $username ) );
```

There is still a difference in HTML. One is adding the `<span>` and the transformer seems to not add it.

jeherve:
>  genuinely don't know whether this is the best path forward, looking for feedback on the approach.

Looking at how Mastodon does things, wouldn't it be best to use the username only, instead of the full account name?

e.g. your mention [here](https://mastodon.social/@obenland/114210414908667355) is displayed as `@evan`, not `@evan@cosocial.ca`:

image

obenland:
Ha, sure! I had a 50:50 shot 😄

obenland:
The only downside I can think of is when you're responding to multiple users with the same username, like `@index@ghost.org` and `@index@activitypub.ghost.org`.

jeherve:
> The only downside I can think of is when you're responding to multiple users with the same username, like `@index@ghost.org` and `@index@activitypub.ghost.org`.

That's a good point. Do you know how Mastodon handles this? Do they detect that use-case?

pfefferle:
Mastodon has Gravatar like pop-overs, that show details about the person.

Screenshot 2025-03-26 at 16 29 59

jeherve:
> Mastodon has Gravatar like pop-overs, that show details about the person.

Good point. And I see my desktop app (Semaphore) does have a `title` attribute on the link that shows the full handle. Elk, however, has nothing.

pfefferle:
I think having a title attribute would bridge the gap here!?!

pfefferle:
I think this is fine but still: https://github.com/Automattic/wordpress-activitypub/pull/1510#pullrequestreview-2716285440

obenland:
Is the comment you linked to not solved?

pfefferle:
It looked like, but it seems that it was somehow reverted... maybe because of a merge conflict... or the GitHub UI does not show it properly... I can see only changes in three files and the Mention Class is not one of them 🤔

obenland:
Right, that's the change from full account names to only usernames, as requested in that comment. There are no changes necessary in the Mentions class.

Let me update the PR description to reflect the change of direction.

pfefferle:
But that is what I meant, both implementations produce different HTML and I would love to have it normalized! See the `<span />`.

obenland:
> But that is what I meant, both implementations produce different HTML and I would love to have it normalized! See the <span />.

Fixed in [43636dd](https://github.com/Automattic/wordpress-activitypub/pull/1510/commits/43636dd1ae2f62d6d839cd22300e40858d9a5f13).

mattwiebe (approved):
Tests well, nice change 👍🏻

PR #1514, opened 2025-03-26
Dispatcher: Adhere to passed batch size.

obenland:
## Proposed changes:

* Adds test to make sure passed batch size gets honored.
* Updates send_to_followers to compare to the correct batch size.

### Other information:

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Outbox batch processing adheres to passed batch size.

</details>

### DIFF
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -125,7 +125,7 @@ class Dispatcher {
 	 *
 	 * @return array|void The next batch of followers to process, or void if done.
 	 */
-	public static function send_to_followers( $outbox_item_id, $batch_size = 50, $offset = 0 ) {
+	public static function send_to_followers( $outbox_item_id, $batch_size = ACTIVITYPUB_OUTBOX_PROCESSING_BATCH_SIZE, $offset = 0 ) {
 		$json    = Outbox::get_activity( $outbox_item_id )->to_json();
 		$actor   = Outbox::get_actor( \get_post( $outbox_item_id ) );
 		$inboxes = Followers::get_inboxes_for_activity( $json, $actor->get__id(), $batch_size, $offset );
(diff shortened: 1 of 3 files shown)


obenland:
I noticed this while working on hook unit tests on wpcom, that passing a batch size didn't really result in subsequent batches be returned as expected because it gets compared to the default of 50.

includes/class-dispatcher.php
```diff
@@ -125,7 +125,7 @@ public static function process_outbox( $id ) {
 	 *
 	 * @return array|void The next batch of followers to process, or void if done.
 	 */
-	public static function send_to_followers( $outbox_item_id, $batch_size = 50, $offset = 0 ) {
+	public static function send_to_followers( $outbox_item_id, $batch_size = ACTIVITYPUB_OUTBOX_PROCESSING_BATCH_SIZE, $offset = 0 ) {
```
pfefferle, line 128:
```suggestion
	public static function send_to_followers( $outbox_item_id, $batch_size = self::$batch_size, $offset = 0 ) {
```

Shouldn't it be `self::$batch_size` to be more consistent with the rest of the code?

pfefferle: not possible :(

pfefferle, line 128: You also have to change the phpdoc to `Default ACTIVITYPUB_OUTBOX_PROCESSING_BATCH_SIZE.`.

obenland: Fixed in [2e16efd](https://github.com/Automattic/wordpress-activitypub/pull/1514/commits/2e16efdf34a80ffdff76e27796ebd27f404ae5e9)

pfefferle (approved):

PR #1515, opened 2025-03-27
Comment: Send announce for all new comments

### LINKED ISSUE #1001: Comments left by mastodon do not get pushed to lemmy
### Quick summary

When leaving a comment to a WP post via mastodon, if that post has been federated to lemmy already, the WP comment won't be sent to the lemmy "community".

### Steps to reproduce

I opened this blogpost: https://dbzer0.com/blog/this-blog-is-now-federated-natively-to-lemmy/

It got federated to lemmy here: https://lemmy.dbzer0.com/post/31734668

I left one comment from lemmy, which federated back to WP

I then replied to that comment via mastodon: https://hachyderm.io/@db0/113503918837384265

The comment appears in wordpress, correctly threaded bellow the lemmy comment. However it does not appear in the lemmy community.

Lemmy communities normally send through any comments received to all their subscribing instances. In mastodon it thus appears as if they're "boosting" all their comment. Something similar should happen in this plugin as well.

### What actually happened

The comment from mastodon does not sync to lemmy.

### Impact

All

### Available workarounds?

### Logs or notes

_No response_

obenland:
Fixes #1001.

## Proposed changes:

* Creates an Announce outbox item for every incoming comment.

## Testing instructions:

* Apply this change to a WP test site that's accessible online.
* Comment on a post from a fediverse instance of your choice.
* Check the Outbox and make sure it created an Announce activity.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Incoming comments create an Announce activity so other instances get notified about it.

</details>

### DIFF
--- a/includes/handler/class-create.php
+++ b/includes/handler/class-create.php
@@ -9,6 +9,7 @@ namespace Activitypub\Handler;
 
 use Activitypub\Collection\Interactions;
 
+use function Activitypub\add_to_outbox;
 use function Activitypub\is_self_ping;
 use function Activitypub\is_activity_reply;
 use function Activitypub\is_activity_public;
@@ -78,6 +79,8 @@ class Create {
 		if ( $state && ! \is_wp_error( $state ) ) {
 			$reaction = \get_comment( $state );
 		}
+
+		add_to_outbox( $activity, 'Announce', $user_id, ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC );
 
 		/**
 		 * Fires after a Create activity has been handled.
--- a/includes/handler/class-create.php
+++ b/includes/handler/class-create.php
@@ -80,6 +80,7 @@ class Create {
 			$reaction = \get_comment( $state );
 		}
 
+		/* @ticket https://github.com/Automattic/wordpress-activitypub/issues/1001 */
 		add_to_outbox( $activity, 'Announce', $user_id, ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC );
 
 		/**
--- a/.github/changelog/1515-from-description
+++ b/.github/changelog/1515-from-description
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Incoming comments create an Announce activity so other instances get notified about it.


obenland:
I guess this could create quite a bit of load on sites with lots of followers that comment a lot. Is that something we need to be concerned about?

db0:
If @obenland comment is a concern, maybe a switch to turn it off.

includes/handler/class-create.php
```diff
@@ -79,6 +80,9 @@ public static function handle_create( $activity, $user_id, $activity_object = nu
 			$reaction = \get_comment( $state );
 		}
 
+		/* @ticket https://github.com/Automattic/wordpress-activitypub/issues/1001 */
+		add_to_outbox( $activity, 'Announce', $user_id, ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC );
+
```
pfefferle, line 85: This makes only sense if you use the `ACTIVITYPUB_ACTOR_AND_BLOG_MODE` and the `Announce` should always done by the Blog-User.

```suggestion
		/* @ticket https://github.com/Automattic/wordpress-activitypub/issues/1001 */
		if ( ACTIVITYPUB_ACTOR_AND_BLOG_MODE === \get_option( 'activitypub_actor_mode' ) ) {
			add_to_outbox( $activity, 'Announce', Actors::BLOG_USER_ID, ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC );
		}

```

I would also recommend to only boost the Object-ID URL instead of the complete Activity. Not sure though if that is enough information.

includes/handler/class-create.php
```diff
@@ -9,6 +9,7 @@
 
 use Activitypub\Collection\Interactions;
 
+use function Activitypub\add_to_outbox;
```
pfefferle, line 12:
```suggestion
use Activitypub\Collection\Actors;

use function Activitypub\add_to_outbox;
```

includes/handler/class-create.php
```diff
@@ -9,6 +9,7 @@
 
 use Activitypub\Collection\Interactions;
```
pfefferle, line 10:
```suggestion
use Activitypub\Collection\Actors;
use Activitypub\Collection\Interactions;
```

obenland:
Closing for now until we have a better understanding what's needed here. See https://github.com/Automattic/wordpress-activitypub/issues/1001#issuecomment-2761637216

pfefferle:
Let's keep it open and set to WIP instead!

pfefferle:
@obenland have you tested it with the changes I suggested?

pfefferle:
We would also need to respect the moderation queue and only announce reactions that are published on the blog.

pfefferle:
closed in favour of #1562

Issue #1513, opened 2025-03-26
Move: support domain changes


mattwiebe:
When the `home` domain is changed, we should detect it and take action to call `Move::internally`, which will need enhancing to support this case (or possibly another specific function like `Move::domain` for this case.)

I worked with Windsurf to spec out what would be needed.

---

1. **Domain Change Detection System**

   1.1. **Create a Domain Change Monitor**
   - Create a new class `Domain_Change_Monitor` to track domain changes  
   - Store the current domain in the WordPress options table  
   - Add a hook to `update_option_home` to detect when the site URL changes  
   - Compare the old and new domains to determine if a domain change occurred  

   1.2. **Domain Change Event Handling**
   - When a domain change is detected, trigger a domain migration process  
   - Log the change with timestamp and details for future reference  
   - Store both old and new domains in a persistent option for reference  

2. **Extend the `internally` Function**

   2.1. **Add Domain Change Detection Parameter**
   - Modify the `internally` function to accept an optional `$domain_change` parameter  
   - Update function signature: `public static function internally( $from, $to, $domain_change = false )`  
   - Add documentation explaining the new parameter's purpose  

   2.2. **Domain-Specific Logic**
   - Add conditional logic to handle domain change scenarios differently  
   - For domain changes, generate URLs with both old and new domains  
   - Update the `Activity` object to include domain change information  

3. **Actor Profile Management**

   3.1. **Update Actor Identifiers**
   - Modify the Blog profile to handle both old and new domains  
   - Update user profiles to maintain both `@user@olddomain.com` and `@user@newdomain.com` identifiers  
   - Store the mapping between old and new identifiers  

   3.2. **Extend `alsoKnownAs` Property**
   - Ensure all old domain URLs are added to the `alsoKnownAs` property  
   - Create a helper method to generate domain-specific actor URLs  
   - Update both user and blog `alsoKnownAs` properties  

4. **Request Routing and Profile Serving**

   4.1. **Modify `get_by_username` Method**
   - Extend to check for usernames with both old and new domains  
   - Add domain-aware username parsing to extract the base username  
   - Implement domain detection to determine which profile version to serve  

   4.2. **Update `get_by_various` Method**
   - Enhance to handle resources from both old and new domains  
   - Add domain detection logic to map between domain variants  
   - Maintain backward compatibility with existing implementations  

   4.3. **Enhance `template_redirect` Hook**
   - Modify to detect requests coming to the old domain  
[...]

PR #1516, opened 2025-03-27
Move: use `$from` for object in `Move::internally`

mattwiebe:
See #1513 especially [this comment](https://github.com/Automattic/wordpress-activitypub/issues/1513#issuecomment-2759529509)

According to the [Draft Data Portability spec, section 4.1](https://swicg.github.io/activitypub-data-portability/#process-0):

> The user sends a Move activity from the old actor to all followers.
```
{
  "type": "Move",
  "actor": "https://old.example/users/username",
  "object": "https://old.example/users/username",
  "target": "https://new.example/users/username"
}
```

## Proposed changes:

* Set `object` to the `$from` param rather than the `$to` param in `Move::internally`

### Other information:

## Testing instructions:

The testing is performed in the tests. You can manually call `Move::internally` yourself to verify if you wish.

### Changelog entry

<details>

<summary>Use the `$from` account for the object in Move activity for internal Moves</summary>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Use the `$from` account for the object in Move activity for internal Moves

</details>

### DIFF
--- a/.github/changelog/1516-from-description
+++ b/.github/changelog/1516-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Use the `$from` account for the object in Move activity for internal Moves
--- a/includes/class-move.php
+++ b/includes/class-move.php
@@ -130,7 +130,7 @@ class Move {
 		$activity->set_type( 'Move' );
 		$activity->set_actor( $actor );
 		$activity->set_origin( $actor );
-		$activity->set_object( $to );
+		$activity->set_object( $actor );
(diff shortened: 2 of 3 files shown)


mattwiebe:
I guess we should probably change this in `Move::externally` as well? Has it been working properly there?

pfefferle (approved):
Just re-checked the mastodon docs and you are right, `object` references the source id, not the target! Good find!

pfefferle:
And we should definitely also change it for `Move::externally`!

PR #1517, opened 2025-03-28
Fix: Do not send self-replies

pfefferle:
While debugging the Outbox-Scheduler I noticed, that the blogs own outbox was listed in the `send_to_additional_inboxes` call, after writing a local comment. This is because the post is listed in the `in_reply_to` of the comment and the code check this field to ping all recipients.

With this PR, self domain replies will be ignored for the `inbox` retrieval process.

## Proposed changes:

* Ignore replies to a local URL

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Prevent self-replies on local comments.

</details>

### DIFF
--- a/tests/includes/class-test-dispatcher.php
+++ b/tests/includes/class-test-dispatcher.php
@@ -382,49 +382,38 @@ class Test_Dispatcher extends ActivityPub_Outbox_TestCase {
 		// Create a test actor.
 		$actor_id = self::$user_id;
 
-		// Mock the HTTP response for the remote object.
-		add_filter(
-			'pre_http_request',
-			function ( $pre, $parsed_args, $url ) {
-				if ( 'https://mastodon.social/@user/123456789' === $url ) {
-					return array(
-						'response' => array( 'code' => 200 ),
-						'body'     => \wp_json_encode(
-							array(
-								'type'         => 'Note',
-								'id'           => 'https://mastodon.social/@user/123456789',
-								'attributedTo' => 'https://mastodon.social/@user',
-							)
-						),
-					);
-				}
-				return $pre;
-			},
-			10,
-			3
-		);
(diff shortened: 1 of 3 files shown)


tests/includes/class-test-dispatcher.php
```diff
@@ -344,4 +344,95 @@ public function add_follower( $pre, $actor ) {
...
+			},
+			10,
+			3
+		);
+
+		// Mock the HTTP response for the actor.
+		add_filter(
+			'pre_http_request',
+			function ( $pre, $parsed_args, $url ) {
+				if ( 'https://mastodon.social/@user' === $url ) {
```
obenland: If you want to, this could probably be moved in with the first callback, alongside the `'https://mastodon.social/@user/123456789'` check.

tests/includes/class-test-dispatcher.php
```diff
@@ -344,4 +344,95 @@ public function add_follower( $pre, $actor ) {
...
+		);
+
+		// Get inboxes for the activity.
+		$inboxes = Dispatcher::add_inboxes_of_replied_urls( array(), $actor_id, $activity );
+
+		// Verify that the inbox was added.
+		$this->assertContains( 'https://mastodon.social/inbox', $inboxes, 'Inbox should be added for different domain in_reply_to URLs' );
+
+		// Clean up.
+		remove_all_filters( 'pre_http_request' );
```
obenland: I worry this would remove the `http_disable_request` callback in `bootstrap.php`. 
Alternatively the callbacks in this function could be saved to a variable and then removed individually.

obenland (approved):

PR #1518, opened 2025-03-28
Fix: Properly traverse the object hierarchy to get the ID

pfefferle:
While debugging the Outbox-Scheduler, I realized that the `Announce` Outbox Items had the wrong `_activitypub_object_id`. It was the `Activity` ID instead of the original `Object` ID of the post/comment/...

This PR checks recursively for the deepest nested ID possible. This way we always have the `Object` ID which is needed to invalidate Outbox-Activities.

## Proposed changes:

* Properly search for the ID in nested Activities/Objects

### Other information:

## Testing instructions:

* Write a new Post
* Check the Outbox-Items

Before the PR, the `Announce` Activity would use the ID of the `Create` Activity instead of the `Object` ID of the Post.

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Support multiple layers of nested Outbox activities when searching for the Object ID.

</details>

### DIFF
--- a/includes/activity/class-activity.php
+++ b/includes/activity/class-activity.php
@@ -9,7 +9,8 @@
 
 namespace Activitypub\Activity;
 
-use Activitypub\Link;
+use Activitypub\Activity\Extended_Object\Event;
+use Activitypub\Activity\Extended_Object\Place;
 
 /**
  * \Activitypub\Activity\Activity implements the common
@@ -21,6 +22,43 @@ use Activitypub\Link;
 class Activity extends Base_Object {
 	const JSON_LD_CONTEXT = array(
 		'https://www.w3.org/ns/activitystreams',
+	);
+
+	/**
+	 * The default types for Activities.
+	 *
+	 * @see https://www.w3.org/TR/activitystreams-vocabulary/#activity-types
+	 *
+	 * @var array
+	 */
+	const DEFAULT_TYPES = array(
+		'Accept',
+		'Add',
+		'Announce',
+		'Arrive',
+		'Block',
+		'Create',
+		'Delete',
+		'Dislike',
+		'Follow',
+		'Flag',
+		'Ignore',
+		'Invite',
+		'Join',
+		'Leave',
+		'Like',
+		'Listen',
+		'Move',
+		'Offer',
+		'Read',
+		'Reject',
+		'Remove',
+		'TentativeAccept',
+		'TentativeReject',
+		'Travel',
+		'Undo',
+		'Update',
+		'View',
 	);
 
 	/**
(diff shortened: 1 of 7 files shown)


includes/collection/class-outbox.php
```diff
@@ -296,28 +296,35 @@ public static function maybe_get_activity( $outbox_item ) {
 	/**
 	 * Get the object ID of an activity.
 	 *
-	 * @param Activity $activity The activity object.
+	 * @param Generic_Object $data The activity object.
```
obenland: Should this be `Base_Object`? My understanding was that `Generic_Object` was just for cases where we needed dynamic properties.

includes/collection/class-outbox.php
```diff
@@ -296,28 +296,35 @@ public static function maybe_get_activity( $outbox_item ) {
...
+	 *
 	 * @return string The object ID.
 	 */
-	private static function get_object_id( $activity ) {
+	private static function get_object_id( $data ) {
+		// If the object is an array, convert it to a Generic_Object.
+		if ( is_array( $data->get_object() ) ) {
```
obenland: Are there instance outside of the unit tests, where we'd expect the object to be an array?

pfefferle: What happens if you use an array as param for set_object? Will it be transformed?

obenland: Sounds like something we'd want to handle in set_object

includes/collection/class-outbox.php
pfefferle: It can be an Activity or Object, so Base_Object would drop params. Aside from that, a base Object does not have an object attribute as far as I know!?!

includes/collection/class-outbox.php
pfefferle: Nevertheless there could be arrays already in the outbox!?! So we might want to have both for some time?

pfefferle: It will be done, if we properly cast to the regarding object type: https://github.com/Automattic/wordpress-activitypub/blob/7e1c9e14729ac4e578deb88de0a2c1ccede81fea/includes/activity/class-activity.php#L129

obenland: What do you think of [this exploration](https://github.com/Automattic/wordpress-activitypub/compare/fix/outbox-object-id...suggest/object_id)?

It uses Activities and nested Activities instead of generic objects, in the hopes that that is a bit closer to what we'll see in real life?

pfefferle: I will check it out on Monday, thanks for the input!

...and you challenged me once again and I had an idea how to properly cast to the correct object in the init_from_array function while having a bike-ride ☺️

obenland: Not sure that's necessary if we call it on the right class

pfefferle: Let's take the case where you want to fill the whole traverse that means the surrounding object with an activity as object with an object as object! We would have to build something that can handle that "filling" process recursively.

Hard to explain... but I also had no detailed look at your proposal...

obenland: I think I know what you mean. In the proposal it creates an Activity and then passes it as an object to another Activity.

So far, I think, we always had an activity that we want to wrap in another one, never a multi-level activity that needed transforming?

pfefferle:
@obenland I merged your PR (thanks for the input) and improved the `set_object` function of the `Activity`.

obenland:
ade7adcf545b76a2f38ec9917e4508eadceaec1e and ccd8c2a60125d5750470becb9bc2fc5bd3deba13 feel unrelated to the original premise of this PR, unless I'm missing something?

I'm also not sure it's needed, necessarily? Is it used anywhere yet?
If we want class-specific initialization, why can't we call the the specific class directly instead of relying on Activity->set_object to handle that? Feels like that would provide more clarity.

pfefferle:
@obenland this is to ensure that an object is always an object and not accidentally an array. It is used by every `init_from_array` that means by a lot of transformers and by the dispatcher, to migrate the db data to a proper object tree. So if we do not fill the objects properly (Generic_Object does not transform anything on `set_object`) we would need the fallback I had before.

pfefferle:
There are still places that needs to be optimized, but I think this provides a solid foundation to ensure objects.

obenland:
> this is to ensure that an object is always an object and not accidentally an array

Does [this part](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/activity/class-activity.php#L130-L133) not do that?

> So if we do not fill the objects properly 

Are there any instances where we don't do that? If there are, we should fix them specifically. Creating a big fallback for an eventuality, I think, would do us a disservice as it removes any confidence in data integrity in the future.

Objects are not spec'd to contain other objects, only activities can be nested (according to my current understanding of things, please correct me if I'm wrong). Let's make make sure we deal with Activities as appropriate so this is not needed.

pfefferle:
This is true if you see only one level of inheritance... We currently have the tree of

`Activity (Announce) > Activity (Create) > Object (Note)`

The `Generic_Object` has no special handling for `set_object`, so the object might end up as an array. Things like this are very risky to produce artefacts like that: https://github.com/Automattic/wordpress-activitypub/pull/1515/files#diff-2b79b0caa05a137854d850f948bf52e99bf50ad9636c834bc300894dd0b49a5eR84

pfefferle:
> Are there any instances where we don't do that? If there are, we should fix them specifically. Creating a big fallback for an eventuality, I think, would do us a disservice as it removes any confidence in data integrity in the future.

This is not eventually. Every Activity that comes through the inbox is an array and we need something to properly transform these, if we do not want to have array handlers as fallbacks everywhere.

pfefferle:
This code example:

```PHP
<?php

$test = array(
	'type' => 'Announce',
	'id' => 'https://example.com/test',
	'object' => array(
		'id' => 'https://example.com/test2',
		'type' => 'Create',
		'object' => array(
			'id' => 'https://example.com/test3',
			'type' => 'Note',
			'content' => 'Hello, world!',
		),
	)
);

$object = \Activitypub\Activity\Activity::init_from_array( $test );

echo '<pre>';
var_dump( $object );
echo '</pre>';
```

currently returns:

```PHP
object(Activitypub\Activity\Activity)#1327 (41) {
  ["id":protected]=>
  string(24) "https://example.com/test"
  ["type":protected]=>
  string(8) "Announce"
  ["object":protected]=>
  object(Activitypub\Activity\Generic_Object)#1353 (3) {
    ["id":protected]=>
    string(25) "https://example.com/test2"
    ["type"]=>
    string(8) "Activity"
    ["object"]=>
    array(3) {
      ["id"]=>
      string(25) "https://example.com/test3"
      ["type"]=>
      string(4) "Note"
      ["content"]=>
      string(13) "Hello, world!"
    }
  }
}
```

So the nested object is still an array!

With the change the result is:

```PHP
object(Activitypub\Activity\Activity)#1332 (41) {
  ["id":protected]=>
  string(24) "https://example.com/test"
  ["type":protected]=>
  string(8) "Announce"
  ["object":protected]=>
  object(Activitypub\Activity\Activity)#1358 (41) {
    ["id":protected]=>
    string(25) "https://example.com/test2"
    ["type":protected]=>
    string(6) "Create"
    ["object":protected]=>
    object(Activitypub\Activity\Base_Object)#1352 (35) {
      ["id":protected]=>
      string(25) "https://example.com/test3"
      ["type":protected]=>
      string(4) "Note"
      ["attachment":protected]=>
      NULL
      ["attributed_to":protected]=>
      NULL
      ["audience":protected]=>
      NULL
      ["content":protected]=>
      string(13) "Hello, world!"
    }
  }
}
```

mattwiebe (commented):
This is a lot of complexity! Could we not just run a migration on our `Announce`s instead to fix the underlying issue?

pfefferle:
@mattwiebe this is always an issue! this IS the migration we always have to do, because we get data through the inbox and have to transform it. This is not a backwards compatibility thing! This is to handle Activity-JSONs properly!

pfefferle:
We use this mechanism quite often and if the result is not properly transformed, before we store it in the DB or if we do not process it properly by retrieving it from the DB, we will get issue when we strictly rely on objects:

https://github.com/search?q=repo%3AAutomattic%2Fwordpress-activitypub+%3A%3Ainit_from_array&type=code

pfefferle:
Even `Outbox::get_activity` has this issue and does not properly nest the data from the db for the `Announces` of the Blog-User, and we are responsible for the whole path there!

includes/activity/class-activity.php
```diff
@@ -124,58 +162,91 @@ class Activity extends Base_Object {
...
+						break;
+				}
+			} else {
+				$object = Generic_Object::init_from_array( $data );
+			}
+		}
+
+		$this->set( 'object', $object );
+
+		if ( $pre_fill ) {
```
obenland: What are circumstances where we would not want the object to be prefilled?

includes/activity/class-activity.php
```diff
@@ -23,6 +24,43 @@ class Activity extends Base_Object {
...
 	);
 
+	/**
+	 * The default types for Activities.
+	 *
+	 * @see https://www.w3.org/TR/activitystreams-vocabulary/#activity-types
+	 *
+	 * @var array
+	 */
+	const DEFAULT_TYPES = array(
```
obenland: Could we call these just `TYPES` (similar in `Actor` and `Base_Object`)? I don't anticipate we'll have non-default types?

This is a personal preference, but consider keeping these in one line? It's a standard set that we don't anticipate to change (or even read) much, you have a link to further docs above, and I think it helps with readability of the rest of the file.

includes/functions.php
```diff
@@ -1537,35 +1538,7 @@ function is_activity( $data ) {
...
-			'Reject',
-			'Remove',
-			'TentativeAccept',
-			'TentativeReject',
-			'Travel',
-			'Undo',
-			'Update',
-			'View',
-		)
+		Activity::DEFAULT_TYPES
```
obenland: With this replacement we should be able to bring the whole `apply_filters()` call into one line.

includes/functions.php
```diff
@@ -1600,13 +1573,7 @@ function is_actor( $data ) {
...
 	$types = apply_filters(
 		'activitypub_actor_types',
-		array(
-			'Application',
-			'Group',
-			'Organization',
-			'Person',
-			'Service',
-		)
+		Actor::DEFAULT_TYPES
```
obenland: Same as above ^.

obenland (commented):
Thanks for the examples, that was helpful

includes/activity/class-activity.php
pfefferle: can remove that

obenland (approved):

PR #1521, opened 2025-03-28
Scheduler: Process batches for individual hooks

obenland:
## Proposed changes:

* Adds specific event hooks for sending outbox items to followers and retries.
* Adds back-compat to existing events that still pass an event callback.
* Makes checks for existing jobs post-specific, since that's now possible.
* Removes `next_scheduled_hook()` method, as it is no longer used.
* Updates unit test.

### Other information:

## Testing instructions:

* Update a post, create a comment, or update your user profile.
* Trigger registered events: `npm run env -- run cli wp cron event run --all`
* Make sure there are no errors or warnings and Outbox items still get processed as expected.
* Switch to trunk and generate events with the old signature (I defined `ACTIVITYPUB_OUTBOX_PROCESSING_BATCH_SIZE = 1` to trigger batching)
* With events with the old signature scheduled, switch to this branch and run the events.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Batch processing jobs can now be scheduled with individual hooks.

</details>


obenland:
First pass. It'll need a bit more testing.

includes/class-scheduler.php
```diff
@@ -47,6 +46,8 @@ public static function init() {
 
 		// Event callbacks.
 		\add_action( 'activitypub_async_batch', array( self::class, 'async_batch' ), 10, 99 );
+		\add_action( 'activitypub_send_followers', array( self::class, 'async_batch' ), 10, 3 );
+		\add_action( 'activitypub_retry_followers', array( self::class, 'async_batch' ), 10, 3 );
```
pfefferle: this sounds a lot like these were actions on the followers, not on Activites that will be sent to the followers 🤔

what about 

```suggestion
		\add_action( 'activitypub_send_activities', array( self::class, 'async_batch' ), 10, 3 );
		\add_action( 'activitypub_retry_activities', array( self::class, 'async_batch' ), 10, 3 );
```

obenland: Oh, I like it, why did I not think of that. Let's go with the singular!

pfefferle:
I will run some tests on my live system tomorrow morning.

pfefferle:
I have the same issue when "flooding" the db with comments. It seems that only 2 of 5 to 7 comments will be queued properly. If I trigger the `activitypub_reprocess_outbox` process afterwards, all comments will be queued properly and also sent to all followers.

Might there be a flood control mechanism that still interferes somehow?

obenland:
Yeah, I wouldn't expect this PR to help with that directly. But it should help with more fine-grain control over identifying specific jobs

pfefferle:
Except from the flooding issue, everything seems to work as before... Have to check the retries though... I have no broken followers on my second blog 🫣

pfefferle:
I have some really weird experiences on my live system. Have to observe that a bit.

pfefferle:
Removed my changes... will think about logging in general...

obenland:
@pfefferle I gave this another round of testing and it doesn't seem to show its age :)

pfefferle (approved):

PR #1525, opened 2025-03-31
Importer: Format import count number.

obenland:
Before | After
--- | ---
Screenshot 2025-03-31 at 16 41 29 | Screenshot 2025-03-31 at 10 22 56 AM

### DIFF
--- a/includes/wp-admin/import/class-mastodon.php
+++ b/includes/wp-admin/import/class-mastodon.php
@@ -437,7 +437,7 @@ class Mastodon {
 		}
 
 		/* translators: %d: Number of posts */
-		echo '<p>' . \esc_html( \sprintf( \_n( 'Imported %d post.', 'Imported %d posts.', $imported, 'activitypub' ), $imported ) ) . '</p>';
+		echo '<p>' . \esc_html( \sprintf( \_n( 'Imported %s post.', 'Imported %s posts.', $imported, 'activitypub' ), number_format_i18n( $imported ) ) ) . '</p>';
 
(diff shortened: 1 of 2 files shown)


includes/wp-admin/import/class-mastodon.php
```diff
@@ -437,7 +437,7 @@ function ( $tag ) {
 		}
 
 		/* translators: %d: Number of posts */
-		echo '<p>' . \esc_html( \sprintf( \_n( 'Imported %d post.', 'Imported %d posts.', $imported, 'activitypub' ), $imported ) ) . '</p>';
+		echo '<p>' . \esc_html( \sprintf( \_n( 'Imported %s post.', 'Imported %s posts.', $imported, 'activitypub' ), number_format_i18n( $imported ) ) ) . '</p>';
```
pfefferle:
```suggestion
		echo '<p>' . \esc_html( \sprintf( \_n( 'Imported %s post.', 'Imported %s posts.', $imported, 'activitypub' ), \number_format_i18n( $imported ) ) ) . '</p>';
```

pfefferle (dismissed):

pfefferle (approved):

obenland:
@jeherve Should the gardening action have picked up the label change and added the `[Feature] Import` label? Or does that not work in the same PR?

pfefferle:
I think it uses the actions on/from the trunk.

pfefferle:
Might be a security thing, so that you can't do something harmful with PRs.

PR #1528, opened 2025-03-31
Scheduler: Don't federate Deletes of unfederated posts.

obenland:
Also see #1338.

While testing imports and deleting imported posts between test runs, I noticed that we create a Delete activity for every deleted post, even when it wasn't federated. These posts were not federated because they were created as part of an import. 

In actor & blog mode, it even added another Announcement activity for it.

## Proposed changes:

* Checks federation status for a post before setting `$type` to `Delete`.
* Adds unit test.
* Update unit tests for a typo.

### Other information:

## Testing instructions:

* Import a few posts.
* Delete one of these posts.
* Make sure there's no Outbox item created for that Delete.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

No longer federates `Delete` activities for posts that were not federated.

</details>

### DIFF
--- a/includes/scheduler/class-post.php
+++ b/includes/scheduler/class-post.php
@@ -8,6 +8,7 @@
 namespace Activitypub\Scheduler;
 
 use function Activitypub\add_to_outbox;
+use function Activitypub\get_wp_object_state;
 use function Activitypub\is_post_disabled;
 
 /**
@@ -62,7 +63,7 @@ class Post {
 				break;
 
 			case 'trash':
-				$type = 'Delete';
+				$type = 'federated' === get_wp_object_state( $post ) ? 'Delete' : false;
 				break;
 
 			default:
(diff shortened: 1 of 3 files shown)


includes/scheduler/class-post.php
```diff
@@ -62,7 +63,7 @@ public static function schedule_post_activity( $post_id, $post, $update, $post_b
 				break;
 
 			case 'trash':
-				$type = 'Delete';
+				$type = 'federated' === get_wp_object_state( $post ) ? 'Delete' : false;
```
pfefferle, line 66: `get_wp_object_state` is a function I kept for backwards compatibility and it feels no longer valid. we shouldn't rely to heavily on it, but think about a proper alternative, that fits better with the new outbox flow.

pfefferle (approved):

PR #1529, opened 2025-03-31
Introduce user_can_activitypub

obenland:
Replaces `is_user_disabled()` to avoid double meaning.

## Proposed changes:

* Introduces `user_can_activitypub()` to determine whether the passed user_id is enabled.
* Deprecates `is_user_disabled()`.
* Replaces all uses of `is_user_disabled()` with `user_can_activitypub()`.

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -125,13 +125,7 @@ class Comment {
 			$current_user = Actors::BLOG_USER_ID;
 		}
 
-		$is_user_disabled = is_user_disabled( $current_user );
-
-		if ( $is_user_disabled ) {
-			return false;
-		}
-
-		return true;
+		return user_can_activitypub( $current_user );
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


obenland:
Is `user_can_activitypub()` too easy to confuse with `user_can( $user_id, 'activitypub' )`?

mattwiebe:
I fully support deprecating `is_user_disabled`! 

> Is `user_can_activitypub()` too easy to confuse with `user_can( $user_id, 'activitypub' )`?

Yeah it doesn't roll off the tongue. Maybe make it action-oriented, like `user_can_federate()` ?

pfefferle:
> Is user_can_activitypub() too easy to confuse with user_can( $user_id, 'activitypub' )?

But isn't that exactly what the function is checking (with some extras because of the virtual users)?

obenland:
Either is fine with me! If we're not going to be confused with the cap check function, I'm happy to leave it as is. I also like `user_can_federate()` 🤷

pfefferle:
I would avoid `fediverse` here. This is about the protocol and not user facing, so I would use the technical term.

What about `actor_can_activitypub`? Or even simpler `is_actor` or `is_user_actor` or `can_user_actor`?

obenland:
Let's stick with `user_can_activitypub` then @mattwiebe?

pfefferle:
can you maybe explain the purpose of the renaming a bit more? is it because the initial one could be a bit misleading?

obenland:
The old function can be read like we're asking whether a user is handicapped

pfefferle (approved):

PR #1530, opened 2025-04-01
Move: support same-server domain migrations

### LINKED ISSUE #1513: Move: support domain changes

mattwiebe:
Very much a WIP, utterly untested, but the general shape of things is there. There's no listener on the `home` option yet, just working to have a `Move::change_domain` method that would ultimately connect to it, and the beginnings of serving the `$from` Actor distinctly from the `$to` Actor.

Fixes #1513 

## Proposed changes:

*

### Other information:

## Testing instructions:

* Apply this PR to a WoA test blog.
* Go to https://wordpress.com/domains/manage/SITE and update the primary domain.
* Dump the saved old actor json and make sure the domains are correct (the old primary domain)
```
var_dump( get_option( 'activitypub_old_domain' ), json_decode( \get_option( 'activitypub_blog_user_old_domain_data' ) ), json_decode( \get_user_option( 'activitypub_old_domain_data', 12082294 ) ) );exit;
```
* Make sure there's a Move activity in the outbox with the correct `target` property (new domain).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [x] Major

#### Type

#### Message

Support same-server domain migrations

</details>


includes/class-move.php
```diff
@@ -161,4 +161,109 @@ private static function update_blog_also_known_as( $from ) {
...
+	 * It also caches the JSON representation of the old Actor for future lookups.
+	 *
+	 * @param string $from The old domain.
+	 * @param string $to   The new domain.
+	 *
+	 * @return array Array of results from Move::internally calls.
+	 */
+	public static function change_domain( $from, $to ) {
+		// Store the old domain for future reference.
+		\update_option( 'activitypub_old_domain', $from );
```
obenland: I genuinely don't know whether we need to/should support more than one old domain, just noting here in case we do/should.

obenland: Oh, let's also add this option to the list of options in `Activitypub::deactivate()`

includes/class-move.php
```diff
@@ -161,4 +161,109 @@ private static function update_blog_also_known_as( $from ) {
...
+		$actor_data = self::replace_urls_in_array( $actor_data, $new_domain, $old_domain );
+
+		// Don't show movedTo if it matches the current actor ID.
+		if ( isset( $actor_data['movedTo'] ) && $actor_data['movedTo'] === $actor_id ) {
+			unset( $actor_data['movedTo'] );
+		}
+
+		// Cache the actor data.
+		$option_key = 'activitypub_moved_' . md5( $actor_id );
+		\update_option( $option_key, wp_json_encode( $actor_data ), false );
```
obenland: How would you feel about saving that in (a registered) user meta/option instead?

includes/class-move.php
```diff
@@ -161,4 +161,109 @@ private static function update_blog_also_known_as( $from ) {
...
+	 */
+	private static function cache_old_actor_data( $actor, $old_domain, $new_domain ) {
+		// Get the actor data as an array.
+		$actor_data = $actor->to_array();
+
+		// Get the actor ID.
+		$actor_id = $actor->get_id();
+
+		// Replace the new domain with the old domain in all URLs.
+		$actor_data = self::replace_urls_in_array( $actor_data, $new_domain, $old_domain );
```
obenland: Why is that needed? Shouldn't `$actor_data` still be the old data at that point?

includes/model/class-cached-actor.php
```diff
@@ -0,0 +1,99 @@
+<?php
```
pfefferle: why having another actor? can't we use the already existing models?

includes/class-move.php
mattwiebe: > I genuinely don't know whether we need to/should support more than one old domain, just noting here in case we do/should.

I don't think we'll ever need more than one at a time. I have thought about it given that we'll see some folks on dotcom possibly go through a *.wp.com -> custom domain -> *.wp.com flow over time. 

> Oh, let's also add this option to the list of options in Activitypub::deactivate()

Good call 👍

includes/class-move.php
mattwiebe: I'm not sure. The place I anticipated that failing is when migrating the Blog User account from the old default domain-based ID, like `https://wp.test/@wp.test`. Once I've moved to, say `https://newsite.test/?author=0`, how do I know how to resolve the old Actor? Although now as I'm typing this, I think we'd have enough to guess that based on the value of `activitypub_old_domain`. Hmm.

But then also user meta is impossible for the Blog User since it's not even a WP User. So again I was hesitant, and this all assumes that we can resolve the User first, and then discover if there's an old domain Actor to serve instead. I was trying to bypass that and be able to look that old Actor up directly. 

I will think on this more. I do like relating the fallback to the User better than this. I just had my doubts we could resolve it that way. I'm going to explore this a bit more

includes/class-move.php
mattwiebe: I'm not sure. Might not be needed. Claude got a little over-eager there 😅

includes/model/class-cached-actor.php
mattwiebe: My assumption was that those models have way too much to override, given the way that they dynamically generate properties and instead I'm trying to rely on static cached Actor data. But maybe the Model could be responsible for its own fallback? Could we actually do `Actor::int_from_json( $cached_data )` and then bypass the `get_*` dynamic methods once those properties have been set?

includes/class-move.php
obenland: For `moved_to` and `also_known_as` we have [a switch to account for blog user vs wp user](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/class-move.php#L55-L60). It might make it harder to resolve in `get_by_username` though, we'd have to try that out

obenland: Hm, I guess we could just add it to the meta query later-on in `get_by_username`

includes/class-move.php
```diff
@@ -161,4 +161,109 @@ private static function update_blog_also_known_as( $from ) {
...
+
+		// Don't show movedTo if it matches the current actor ID.
+		if ( isset( $actor_data['movedTo'] ) && $actor_data['movedTo'] === $actor_id ) {
+			unset( $actor_data['movedTo'] );
+		}
```
obenland, line 268: Should we keep `movedTo` in the cached data, so requests to it point to the new profile?

mattwiebe: Yes. I got rid of that.

mattwiebe:
Sorry I overwrote your changes @obenland  - but I changed the approach radically enough that that function doesn't exist anymore :)

includes/class-move.php
```diff
@@ -161,4 +161,74 @@ private static function update_blog_also_known_as( $from ) {
...
+
+		return $results;
+	}
+
+	/**
+	 * Cache the old actor data for future lookups.
+	 *
+	 * @param object $actor The actor object.
+	 */
+	private static function cache_old_actor_data( $actor ) {
```
obenland: Should we call it `save_old_actor_data()`?

includes/collection/class-actors.php
```diff
@@ -76,6 +76,20 @@ public static function get_by_id( $user_id ) {
 	 * @return User|Blog|Application|WP_Error The Actor or WP_Error if user not found.
 	 */
 	public static function get_by_username( $username ) {
+		// Check if we're dealing with an old domain request.
+		$old_domain      = \get_option( 'activitypub_old_domain', '' );
+		$request_domain  = isset( $_SERVER['HTTP_HOST'] ) ? \sanitize_text_field( \wp_unslash( $_SERVER['HTTP_HOST'] ) ) : '';
+		$old_domain_host = ! empty( $old_domain ) ? \wp_parse_url( $old_domain, PHP_URL_HOST ) : '';
```
obenland: Let's save it as just the host? Or do we need the full domain somewhere else?

includes/model/class-user.php
```diff
@@ -30,6 +30,34 @@ class User extends Actor {
...
+				if ( $request_domain === $old_domain_host ) {
+					$cached_data = \get_user_option( 'activitypub_old_domain_data', $user_id );
+
+					if ( ! empty( $cached_data ) ) {
+						$this->from_json( $cached_data );
+					}
+				}
+			}
+		}
+	}
```
obenland: Let's move this method to below the class properties

includes/class-move.php
mattwiebe: Sounds good to me

includes/collection/class-actors.php
mattwiebe: Yeah we should just need the host, will save some busywork

mattwiebe:
Some remaining tasks:

- [x] Handle the lifecycle of receiving follows on the new ID and unfollows on the old ID
- [x] Ensure that the old domain ID is used when POSTing  the `Move` Activity to Follower inboxes
- [x] Add a `home` option listener, gated behind a constant for further testing

obenland:
> Ensure that the old domain ID is used when POSTing to Follower inboxes

Could you share more on that? I would have expected no changes since followers will send new Follow requests to the new account.

mattwiebe:
> > Ensure that the old domain ID is used when POSTing to Follower inboxes
> 
> Could you share more on that? I would have expected no changes since followers will send new Follow requests to the new account.

When we send the `Move` activity from the outbox to our Followers, we need to make sure we're using that we're [using the correct Actor ID here](https://github.com/Automattic/wordpress-activitypub/blob/776d67d4db731432f37aeced7025e9e05736acd3/includes/class-dispatcher.php#L213). The "main"/default Actor at this point will be the new domain, but these requests need to be sent/signed by the old domain.

obenland:
Ah, in the context of Move activities, only. I thought in general. Thank you!

mattwiebe:
Also note that some concerns we're raising here probably also apply to `Move::internally`. 

I think we need to serve `$from` and `$to` differently in that case as well. So I think we ultimately need to follow a very similar caching behaviour there as well, but it's going to be harder to detect say `https://wp.test/?author=1` vs `https://wp.test/@username` in that case.

includes/class-dispatcher.php
```diff
@@ -126,9 +126,18 @@ public static function process_outbox( $id ) {
...
+		$actor    = Outbox::get_actor( \get_post( $outbox_item_id ) );
+		$is_move  = $activity->get_type() === 'Move';
+		$batch    = $is_move ? -1 : $batch_size; // -1 for Move activities to get all inboxes at once before deleting below.
+		$inboxes  = Followers::get_inboxes_for_activity( $json, $actor->get__id(), $batch, $offset );
+
+		if ( $is_move && ! empty( $inboxes ) ) {
+			// We are sending a `Move` which means that we need to remove all Followers from the old Actor.
+			// The new Actor will receive the `Follow` activity from any Followers that choose to Follow it.
+			Followers::remove_all_followers( $actor->get__id() );
+		}
```
obenland: Hm, I'm not sure this feels like the right place for that. I guess there's a race condition we're trying to avoid with this, where waiting until the batches are complete would be too late?

obenland: What if we added a post meta for Followers that keeps track of the actor id and we differentiate `Undo` requests based on that? Add an upgrade routine that sets it for all followers?

pfefferle: I think it is fine to keep them! They will re-follow and this might update some, but if the user does not change, I would not delete anything!

obenland: My understanding is that follower instance will start sending Undo Follow requests to the old ID and Follow requests to the new ID (which it probably expects to be accepted).

The expectation that the follower list for the old ID is empty [is part of the draft spec](https://swicg.github.io/activitypub-data-portability/#results-0).

pfefferle: But this is done by the process (sending the unfollow and refollow)! We should not delete something "by hand". If we mess up something with the process, the followers are gone!

includes/collection/class-actors.php
```diff
@@ -76,6 +76,16 @@ public static function get_by_id( $user_id ) {
 	 * @return User|Blog|Application|WP_Error The Actor or WP_Error if user not found.
 	 */
 	public static function get_by_username( $username ) {
+		// Check if we're dealing with an old domain request.
+		$old_domain     = \get_option( 'activitypub_old_domain', '' );
+		$request_domain = isset( $_SERVER['HTTP_HOST'] ) ? \sanitize_text_field( \wp_unslash( $_SERVER['HTTP_HOST'] ) ) : '';
```
pfefferle: What about if we encapsulate this in a function and test also for the referrer? this way, we could also handle domains that point to the same server?

obenland: Fixed in [698e2ac](https://github.com/Automattic/wordpress-activitypub/pull/1530/commits/698e2ac7e247d26842f83c3a644243fed0e469bb)

includes/class-dispatcher.php
obenland: Fixed in [a733aca](https://github.com/Automattic/wordpress-activitypub/pull/1530/commits/a733aca3e0231f08e2bebbb2188f39c632b6335b)

includes/collection/class-actors.php
```diff
@@ -76,6 +77,14 @@ public static function get_by_id( $user_id ) {
 	 * @return User|Blog|Application|WP_Error The Actor or WP_Error if user not found.
 	 */
 	public static function get_by_username( $username ) {
+		$old_domain = \get_option( 'activitypub_old_domain' );
+
+		// Special case for Blog Actor on old domain.
+		if ( $old_domain && $old_domain === $username && Http::is_domain_match( $old_domain ) ) {
+			// Return a new Blog instance which will load the cached data in its constructor.
+			return new Blog();
+		}
```
pfefferle: I would prefer to use filters for that! we should not (hard)code the functionality in too many places!

obenland: Why not? Once the feature is enabled, the filter would run every time, too.

Maybe it shouldn't be the first thing we do here? Maybe it could be the last thing?

pfefferle: it makes things way more complex, because the integration seems to be everywhere. through a filter it is easier to have all integrations in one class and it is unobtrusive. you can easily unhook the feature without doing too much changes in every possible class.

pfefferle: I think you need this code maybe once in the livetime of your website, so it would be nice to even load the whole migration integration only if `$domain_moves_enabled` is enabled or `'activitypub_old_domain'` is set or based on some other indicator.

obenland:
EOD update:
* I moved custom username handling and actor model instantiation into hook callbacks.
* The hooks only get registered if the `activitypub_old_domain` option has a value.
* All of that only gets executed when the "feature flag" filter is enabled.
* The naming of hooks and functions are up for debate, I think there's an opportunity to have them follow more of a pattern.
* The blog and user actor instantiation now has tests, so feel free to mess with them—the tests will let you know when you broke something.

Two things I learned from writing the unit tests:
* `( new Blog() )->to_array()` [returns an old user as expected](https://github.com/Automattic/wordpress-activitypub/pull/1530/files#diff-76854826e5270175ed145d030a6140113b2aacf33323d7e6d1947d743cb19692R62-R63), `( new Blog() )->get_id()` does not, as it calls `get_home_url()`, which returns the new domain.
* We can't hook `Move::change_domain()` into `update_option_home` because home_url is already updated at that point and we can't create "old" actors to save. That's why I moved it to [pre_update_option_home](https://github.com/Automattic/wordpress-activitypub/pull/1530/files#diff-9b77fd1b96427950be8e815af422057b6227ae193f0821da6970207433fad760R36).

obenland:
Tested this on WoA a bit, with little success. [Primary domain changes happen there through direct SQL updates](https://github.com/Automattic/jetpack/blob/87610763aa51359f7b8aca1fa65264897d1c87a7/projects/plugins/wpcomsh/class-wpcomsh-cli-commands.php#L380-L397), bypassing the `pre_update_option_*` hook we rely on. 

Maybe we should keep track of the current domain ourselves and use that to populate our `$from` values. It would also allow us to switch back to the `update_option_home` action, that fires after it's actually been updated.

obenland:
I think I made a fair bit of progress today:

* Found a way to make `change_domain()` work on the `update_option_home` hook. It's not very elegant, it just replaces all instances of the new domain in the actor json with the old domain, before saving it. Maybe that's too naïve? Do we need. toworry about the signature we're storing?
* Made [a few updates to domain checks](https://github.com/Automattic/wordpress-activitypub/pull/1530/commits/765073012bb2981811328adbedf635adabf7ebb0) in order to allow actor identification with the old domain.
* Updated the test to reflect the changes in hooks used.

@pfefferle I'd love for you to test this on a WoA sites or something more real world than localhost and see what you find!

includes/class-move.php
```diff
@@ -10,13 +10,42 @@
...
+	/**
+	 * Initialize the Move class.
+	 */
+	public static function init() {
+		/**
+		 * Filter to enable automatically moving Fediverse accounts when the domain changes.
+		 *
+		 * @param bool $domain_moves_enabled Whether domain moves are enabled.
+		 */
+		$domain_moves_enabled = apply_filters( 'activitypub_enable_primary_domain_moves', false );
```
pfefferle, line 32: @obenland I know that you are not a fan of constants, but this would be the third possibility to enable a feature, aside from advanced settings and constants.

I think we should have a certain level of consistency. I am not arguing for a constant her, maybe we should think about a `Developer Preview` or `Beta Features` section in the Advanced settings to enable beta features!?!

pfefferle:
I think we should merge this PR! It is behind a feature flag, so it is not breaking something and with the functionality in the core plugin, it would allow us to more easily test it on WordPress.com or self hosted installations.

It is a solid state we can improve on.

includes/class-move.php
obenland: I'd be happy to deprecate feature constants in favor of filters or advanced settings

obenland (approved):

PR #1539, opened 2025-04-02
Fix: Default value for empty `movedTo` field

pfefferle:
`movedTo` returns `false` if it is not set, but the spec defines it as `@id` so it should be either empty or not existent instead: https://www.w3.org/TR/json-ld/#node-identifiers

## Proposed changes:

* Do not show `movedTo` if not set.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

So not show `movedTo` attribute instead of setting it to `false` if empty.

</details>

### DIFF
--- a/.github/changelog/1539-from-description
+++ b/.github/changelog/1539-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+So not show `movedTo` attribute instead of setting it to `false` if empty.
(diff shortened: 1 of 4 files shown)


obenland:
Let's add a unit test for it, so it can't fail again with any future changes.

pfefferle:
@obenland too late :)

obenland (approved):

PR #1543, opened 2025-04-02
Tests: Make sure test data gets deleted properly

obenland:
We missed calling `parent:: tear_down_after_class()` in a few instances, which caused data not being deleted properly.

Turns out it's safer for us to not even attempt to delete our own data. 😄

`tear_down_after_class` [calls](https://github.com/WordPress/wordpress-develop/blob/7474af7ae7826ff7f9da044d61cf9e92d259b50a/tests/phpunit/includes/abstract-testcase.php#L91) `_delete_all_data`, which then [does just that](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/includes/functions.php#L110-L137).

## Proposed changes:

* Removes all calls to `tear_down_after_class()`

### Other information:

### DIFF
--- a/.github/changelog/add-advanced-settings
+++ b/.github/changelog/add-advanced-settings
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Advanced Settings tab, with special settings for advanced users.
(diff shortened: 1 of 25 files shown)


obenland:
yeah, confirmed, it was missing `parent::tear_down_after_class();` calls.

pfefferle (approved):

PR #1546, opened 2025-04-03
Onboarding: Improve troubleshooting UI

### LINKED ISSUE #1467: Feature Request: Make sure users know that ActivityPub won't work due to failing preconditions
### What

Right now the Activity-Pub plugin can be activated and configured without realizing that preconditions are failing.

The plugin checks preconditions via the WP-healthcheck but the results are not shown prominently and cause users to not check the healthcheck in the first place.

This will cause users to expect things to work even though they can't and will fail with weird feedback (or none at all)

In a specific case the Activity-Pub settings do not show any issues
Image

Even though the health-check shows a clear problem:
Image

### Why

To make the onboarding process for new users less painful and help users to fix errors themselves without having to raise issues here or (probably more often) just leave frustrated

### How

Show a prominent message whether health-checks failed or not directly on top of the settings page.

Add to failing messages that this means that the plugin will not work before they are fixed.

pfefferle:
Check for ActivityPub Site-Health issues and show them more dominant on the Welcome Page. This should help users to better understand and fix issues with their installations.

Fixes #1467 and CMF-77

Screenshot 2025-04-03 at 10 38 27

## Proposed changes:

* Add function count tests by result
* Show troubleshooting banner only if critical errors exist

### Other information:

## Testing instructions:

* The Site-Health should already throw errors on your local test env, so simply go to the Welcome Page and check the banner.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Improve the troubleshooting UI and show Site-Health stats in ActivityPub settings.

</details>


obenland:
How would you feel about shortening it to an admin notice that's dismissable?

pfefferle:
I am not sure if it feels like "if it is skippable it might not be important" then?

...and the welcome screen is already dismissible!?!

includes/wp-admin/class-welcome-fields.php
```diff
@@ -194,22 +196,51 @@ public static function render_author_profile_section() {
...
-						'activitypub'
+		<div class="notice notice-warning inline">
+			<p>
+				<span class="dashicons dashicons-warning"></span>
+				<?php
+				echo wp_kses(
+					\sprintf(
+						/* translators: the placeholders are the number of critical and recommended issues on the site. */
+						\__(
+							'<strong>Important:</strong> You have <span class="count">%1$d</span> critical and <span class="count">%2$d</span> recommended issues on your site. These issues may affect your site\'s compatibility with the fediverse.',
```
```suggestion
							'<strong>Important:</strong> There are <span class="count">%1$d</span> critical and <span class="count">%2$d</span> recommended issues affecting your site&#8217;s compatibility with the fediverse.',
```
Let's make it less personal

obenland (approved):

obenland:
```
There was 1 warning:

1) Warning
No tests found in class "Activitypub\Tests\WP_Admin\Test_Admin".
```

It looks like this PR removed all tests from that file. We should either delete the file or add another test in there to avoid the warning.

PR #1553, opened 2025-04-07
Add: Shared Inbox setting

pfefferle:
Allow users to easily enable/disable the shared inbox feature without having to add constants.

## Proposed changes:

* Add option to enable/disable the shared inbox feature through the "Advanced Settings".

### Other information:

## Testing instructions:

* Go to "Advanced Settings"
* Enable shared inbox
* Check Actor JSON

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Add option to enable/disable the "shared inbox" to the "Advanced Settings".

</details>


includes/class-activitypub.php
```diff
@@ -51,6 +51,7 @@ public static function init() {
 		\add_action( 'added_post_meta', array( self::class, 'updated_postmeta' ), 10, 4 );
 		\add_filter( 'pre_option_activitypub_actor_mode', array( self::class, 'pre_option_activitypub_actor_mode' ) );
 		\add_filter( 'pre_option_activitypub_authorized_fetch', array( self::class, 'pre_option_activitypub_authorized_fetch' ) );
+		\add_filter( 'pre_option_activitypub_shared_inbox', array( self::class, 'pre_option_activitypub_shared_inbox' ) );
```
obenland, line 54: The "fourth" callback! :) 
Should we create a separate class for these?

includes/class-activitypub.php
```diff
@@ -100,6 +101,7 @@ public static function uninstall() {
 		delete_option( 'activitypub_allow_replies' );
 		delete_option( 'activitypub_attribution_domains' );
 		delete_option( 'activitypub_authorized_fetch' );
+		delete_option( 'activitypub_shared_inbox' );
```
obenland: Let's move it to line 126 so the list stays alphabetized.

includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -49,6 +49,17 @@ public static function register_advanced_fields() {
 				array( 'label_for' => 'activitypub_authorized_fetch' )
 			);
 		}
+
+		if ( ! defined( 'ACTIVITYPUB_SHARED_INBOX_FEATURE' ) ) {
+			\add_settings_field(
+				'activitypub_shared_inbox',
+				\__( 'Shared Inbox', 'activitypub' ) . ' ' . \__( '(beta)', 'activitypub' ),
```
obenland: This won't work for RTL languages. Let's just keep it in the same translation string.

includes/wp-admin/class-settings.php
```diff
@@ -187,6 +187,16 @@ public static function register_settings() {
 			)
 		);
 
+		\register_setting(
+			'activitypub_advanced',
+			'activitypub_shared_inbox',
+			array(
+				'type'        => 'boolean',
+				'description' => \__( 'Enable the shared inbox.', 'activitypub' ),
```
obenland, line 195: Do these need to be translated? Where are these stings visible?

includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -110,4 +121,22 @@ public static function render_authorized_fetch_field() {
...
+	/**
+	 * Render shared inbox field.
+	 */
+	public static function render_shared_inbox_field() {
+		$value = \get_option( 'activitypub_shared_inbox', '0' );
+		?>
+		<p>
+			<label>
+				<input type="checkbox" id="activitypub_shared_inbox" name="activitypub_shared_inbox" value="1" <?php checked( '1', $value ); ?> />
+				<?php \esc_html_e( 'Enable the shared inbox.', 'activitypub' ); ?>
```
obenland: "Use a shared inbox for incoming messages"

includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -110,4 +121,22 @@ public static function render_authorized_fetch_field() {
...
+			</label>
+		</p>
+		<p class="description">
+			<?php \esc_html_e( 'This enables the use of the shared inbox for receiving ActivityPub messages, which is especially useful for sites with a large user base and may be required by certain platforms. This feature is still in beta, so please be aware that you might encounter issues.', 'activitypub' ); ?>
```
obenland: I asked AI for help to make the strings more end-user friendly. This is not a huge jump I think, but might make it a bit more readable? 

"Allows your site to handle incoming ActivityPub messages more efficiently, especially helpful for busy or multi-user sites. This feature is still in beta and may encounter issues."

includes/class-activitypub.php
pfefferle: ...if I make it by alphabet, you want it by order of the hooks, if I make it by the order of the hooks, ... 😘

obenland: Gotta keep you on your toes! :)

includes/wp-admin/class-settings.php
pfefferle: all settings are translated, that's why AI wanted to keep consistency. I am fine with both, but we should change the rest too then.

obenland (approved):

PR #1554, opened 2025-04-07
Actors: Don't convert non-numeric strings to Blog user id

obenland:
Fixes a bug in our Enable Mastodon Apps integration that can create hundreds of Outbox items.

The `mastodon_api_account` filter [passes a user_id or author_url as the second parameter](https://github.com/akirk/enable-mastodon-apps/blob/36707ccb7ce16b28d37c294f835df787f62ffed0/includes/class-comment-cpt.php#L245-L249), which [we currently don't account for in `Enable_Mastodon_Apps::api_account_internal()`](https://github.com/Automattic/wordpress-activitypub/blob/trunk/integration/class-enable-mastodon-apps.php#L274-L279).

Requests like the following end up creating two extra fields post for every comment from Activitypub commenters, which in turn create an Outbox item each:
```
Debugger [2025-04-06 14:31:22.067] 200 GET [/api/v1/timelines/public?only_media=false&since_id=8452](https://obietester.blog/api/v1/timelines/public?only_media=false&since_id=8452&_wpnonce=NONCE&_pretty=1)
params array (
  'limit' => 20,
  'only_media' => 'false',
  'since_id' => '8452',
)
current_user 0
user_agent 'Apache-HttpClient/4.5.13  Java/11.0.20 '
app \Enable_Mastodon_Apps\Mastodon_App::__set_state(array(
   'term' => 
  \WP_Term::__set_state(array(
     'term_id' => 1370,
     'name' => 'enable-mastodon-apps',
     'slug' => 'enable-mastodon-apps',
     'term_group' => 0,
     'term_taxonomy_id' => 16,
     'taxonomy' => 'mastodon-app',
     'description' => '',
     'parent' => 0,
     'count' => 0,
     'filter' => 'raw',
  )),
))
```

Stack trace:
```
require('wp-blog-header.php'), 
wp, 
WP->main, 
WP->parse_request, 
do_action_ref_array('parse_request'), 
WP_Hook->do_action, 
WP_Hook->apply_filters, 
rest_api_loaded, 
WP_REST_Server->serve_request, 
WP_REST_Server->dispatch, 
WP_REST_Server->respond_to_request, 
Enable_Mastodon_Apps\Mastodon_API->api_public_timeline, 
apply_filters('mastodon_api_public_timeline'), 
WP_Hook->apply_filters, 
Enable_Mastodon_Apps\Handler\Timeline->api_public_timeline, 
Enable_Mastodon_Apps\Handler\Handler->get_posts, 
apply_filters('mastodon_api_status'), 
WP_Hook->apply_filters, 
Enable_Mastodon_Apps\Comment_CPT->api_status, 
apply_filters('mastodon_api_account'), WP_Hook->apply_filters, Activitypub\Integration\Enable_Mastodon_Apps::api_account_internal, Activitypub\Integration\Enable_Mastodon_Apps::get_extra_fields, 
Activitypub\Collection\Extra_Fields::get_actor_fields, 
apply_filters('activitypub_get_actor_extra_fields'), 
WP_Hook->apply_filters, 
Activitypub\Collection\Extra_Fields::default_actor_extra_fields, 
wp_insert_post, 
wp_transition_post_status, 
do_action('transition_post_status'), 
WP_Hook->do_action, 
WP_Hook->apply_filters, 
Activitypub\Scheduler\Actor::schedule_post_activity, 
Activitypub\Scheduler\Actor::schedule_profile_update
```

The `author_url` we receive in `Enable_Mastodon_Apps::api_account_internal()` gets passed to `Actors::get_by_id()`, where [this check converts it to `0`](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/collection/class-actors.php#L49-L51), clearing the `user_can_activitypub()` check on blogs with the blog user enabled.

Further down in `Enable_Mastodon_Apps::api_account_internal()` [it gets the extra fields for the user](https://github.com/Automattic/wordpress-activitypub/blob/trunk/integration/class-enable-mastodon-apps.php#L315), which in this case is still the author_url and not an internal user_id. This non-ID id gets passed down all the way to `Extra_Fields::default_actor_extra_fields()`, where [it can't find cached extra fields for it](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/collection/class-extra-fields.php#L206-L212) and [fails to save the default extra fields it just created](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/collection/class-extra-fields.php#L260-L262), since it doesn't have a valid user_id.

Every time an Extra Field post gets created, [it creates an Update activity for the blog user](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/scheduler/class-actor.php#L117-L123).

## Proposed changes:

* Removes the `is_string()` check that on its own was enough to continue with the process.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Sites with comments from the Fediverse no longer create uncached extra fields posts that flood the Outbox.

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -329,10 +329,14 @@ function is_post_disabled( $post ) {
 /**
  * This function checks if a user is enabled for ActivityPub.
  *
- * @param int $user_id The user ID.
+ * @param int|string $user_id The user ID.
  * @return boolean True if the user is enabled, false otherwise.
  */
 function user_can_activitypub( $user_id ) {
+	if ( ! is_numeric( $user_id ) ) {
+		return false;
+	}
+
 	switch ( $user_id ) {
 		case Actors::APPLICATION_USER_ID:
 			$enabled = true; // Application user is always enabled.
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -333,27 +333,16 @@ function is_post_disabled( $post ) {
  * @return boolean True if the user is enabled, false otherwise.
  */
 function user_can_activitypub( $user_id ) {
-	switch ( $user_id ) {
-		case Actors::APPLICATION_USER_ID:
-			$enabled = true; // Application user is always enabled.
-			break;
-
-		case Actors::BLOG_USER_ID:
-			$enabled = ! is_user_type_disabled( 'blog' );
-			break;
-
-		default:
-			if ( ! \get_user_by( 'id', $user_id ) ) {
-				$enabled = false;
-				break;
-			}
-
-			if ( is_user_type_disabled( 'user' ) ) {
-				$enabled = false;
-				break;
-			}
-
-			$enabled = \user_can( $user_id, 'activitypub' );
+	if ( Actors::APPLICATION_USER_ID === $user_id ) {
+		$enabled = true; // Application user is always enabled.
+	} elseif ( Actors::BLOG_USER_ID === $user_id ) {
+		$enabled = ! is_user_type_disabled( 'blog' );
+	} elseif ( ! \get_user_by( 'id', $user_id ) ) {
+		$enabled = false;
+	} elseif ( is_user_type_disabled( 'user' ) ) {
+		$enabled = false;
+	} else {
+		$enabled = \user_can( $user_id, 'activitypub' );
 	}
 
 	/**
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -333,16 +333,27 @@ function is_post_disabled( $post ) {
  * @return boolean True if the user is enabled, false otherwise.
  */
 function user_can_activitypub( $user_id ) {
-	if ( Actors::APPLICATION_USER_ID === $user_id ) {
-		$enabled = true; // Application user is always enabled.
-	} elseif ( Actors::BLOG_USER_ID === $user_id ) {
-		$enabled = ! is_user_type_disabled( 'blog' );
-	} elseif ( ! \get_user_by( 'id', $user_id ) ) {
-		$enabled = false;
-	} elseif ( is_user_type_disabled( 'user' ) ) {
-		$enabled = false;
-	} else {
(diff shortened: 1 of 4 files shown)


includes/functions.php
```diff
@@ -333,27 +333,16 @@ function is_post_disabled( $post ) {
...
-				$enabled = false;
-				break;
-			}
-
-			if ( is_user_type_disabled( 'user' ) ) {
-				$enabled = false;
-				break;
-			}
-
-			$enabled = \user_can( $user_id, 'activitypub' );
```
obenland: It turns out that in PHP versions below 8.0, PHP runs an `(int) $user_id` call internally, when comparing a string in the switch condition with an int in the case statement, leading to the same problem we're trying to fix in `Actors:: get_by_id()`.

Making all of these checks strict will change the functions behavior when `$user_id` is a numeric string, so I'm not sure this is the way to go. 

We should probably bail early if `$user_id` is not numeric—but should that still be filterable?

pfefferle (approved):
Nice find!

akirk:
What do you suggest shall we do about those erroneously created entries? Can we identify the duplicate entries in a migration step and delete them?

akirk:
Oh, and thanks so much for identifying and fixing this! It was baffling to me what was causing this.

obenland:
I wouldn't do anything about them. They'll be deleted automatically with the Outbox purge job.

akirk:
These yes, but as a side effect I've see lots of duplicate "default fields" being generated, specifically custom post types `ap_extrafield` with `post_author = 0`.

obenland:
Oh, I didn't see that on my test site. Yes, those we should take care of. I can take a stab at that

PR #1558, opened 2025-04-09
Actors: Extend collection method

obenland:
Follow-up to #1530, proposing an extended `Actors::get_collection()` method instead of a separate one.

## Proposed changes:

* Extends get_collection to optionally include the Blog user
* Removes `Actors:all()`
* Updates function call in `Move::change_domain`
* Makes extending User class compatible
* Adds unit test

### Other information:

### DIFF
--- a/includes/class-move.php
+++ b/includes/class-move.php
@@ -205,7 +205,7 @@ class Move {
 	 */
 	public static function change_domain( $from, $to ) {
 		// Get all actors that need to be migrated.
-		$actors = Actors::get_all();
+		$actors = Actors::get_collection( 'all' );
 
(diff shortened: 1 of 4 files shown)


pfefferle (commented):
can we use a different naming for the function then?

obenland:
I didn't realize "collection" only included users. Maybe it makes sense then to keep it as-is and have `get_all()` re-use it?

pfefferle:
I named it collection, because it returns exactly what I needed in the collections API endpoints. In the followers collection we use `get_followers` and `get_follower` but the `follower` and `followers` are redundant because the context is already given by the class.

PR #1560, opened 2025-04-09
Reply block: Improve fallback embed

obenland:
## Proposed changes:

* Removes top and bottom margins when displaying in the Editor so it doesn't overflow the iframe.
* Only shows repost and like count if they're passed in the ActivityPub object.
* Accounts for `Image` attachments for compatibility with Pixelfed.

#### Before
Screenshot 2025-04-09 at 3 06 12 PM

#### After
Screenshot 2025-04-09 at 3 05 53 PM

## Testing instructions:

* Go to the Editor and insert a Federated Reply block.
* Paste a URL for a Pixelfed post and make sure it displays correctly.


pfefferle (approved):
Tested with misskey and pixelfed, works like a charm!

PR #1562, opened 2025-04-10
Comment: Send announce for all new comments

### LINKED ISSUE #1001: Comments left by mastodon do not get pushed to lemmy

pfefferle:
Supersedes #1515 and fixes #1001

## Proposed changes:

* Stores the Activity in the Comment-Meta, because we have to get sure, that the comment will be approved or not before announcing it
* Hook into `approve` process, to late federate the `Announce`

### Other information:

## Testing instructions:

* Apply this change to a WP test site that's accessible online.
* Comment on a post from a fediverse instance of your choice.
* Check the Outbox and make sure it created an Announce activity.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Incoming interactions create an Announce activity so other instances get notified about it.

</details>


includes/class-comment.php
```diff
@@ -33,6 +33,8 @@ public static function init() {
 		\add_action( 'update_option_activitypub_allow_likes', array( self::class, 'maybe_update_comment_counts' ), 10, 2 );
 		\add_action( 'update_option_activitypub_allow_reposts', array( self::class, 'maybe_update_comment_counts' ), 10, 2 );
 		\add_filter( 'pre_wp_update_comment_count_now', array( static::class, 'pre_wp_update_comment_count_now' ), 10, 3 );
+
+		\add_action( 'transition_comment_status', array( self::class, 'maybe_announce_interaction' ), 20, 3 );
```
obenland: Could this be moved to `/scheduler/class-comment.php`? We already schedule Outbox items for comments there, this feels like a good fit.

pfefferle: I HAVE NO IDEA! I had a look at it and I don't know why I decided that this wouldn't be the right place... 🫣

(no irony!!!)

includes/class-dispatcher.php
```diff
@@ -33,7 +33,7 @@ class Dispatcher {
 	 * @see https://github.com/tfredrich/RestApiTutorial.com/blob/fd08b0f67f07450521d143b123cd6e1846cb2e3b/content/advanced/responses/retries.md
 	 * @var int[]
 	 */
-	public static $retry_error_codes = array( 408, 429, 500, 502, 503, 504 );
+	public static $retry_error_codes = array( 400, 408, 429, 500, 502, 503, 504 );
```
obenland: I have similar concerns as CoPilot. What makes adding 400 necessary?

includes/scheduler/class-comment.php
```diff
@@ -81,6 +86,54 @@ public static function schedule_comment_activity( $new_status, $old_status, $com
...
+		if ( 'approved' !== $new_status || 'approved' === $old_status ) {
+			return;
+		}
+
+		if ( ! Comment_Utils::was_received( $comment ) ) {
+			return;
+		}
+
+		// Get activity from comment meta.
+		$activity = \get_comment_meta( $comment->comment_ID, '_activitypub_activity', true );
```
obenland, line 127: Could we get this from the Inbox, instead of saving it in comment meta?
Alternatively, could we use the `source_id` meta as the object instead of offering the full object?

includes/class-dispatcher.php
pfefferle: that is what lemmy is returning 😒

includes/scheduler/class-comment.php
pfefferle: We can postpone this a bit until we have the inbox activated by default. PR incoming.

includes/class-dispatcher.php
obenland: I assume "that is what lemmy is returning" is referring to [these](https://github.com/Automattic/wordpress-activitypub/issues/1001#issuecomment-2830253853) [examples](https://github.com/Automattic/wordpress-activitypub/issues/1001#issuecomment-2854513673)? If so, I'd expect sending the correct Announce would not trigger a 400.
Are there other circumstances where we'd expect a 400 that reflects something temporary?

Jiwoon-Kim:
Some thoughts:

1. As far as I understand, comment synchronization seems to rely on the **group actor’s `Announce`** activity.
   What happens if the **group actor does not exist** in a given implementation?

2. What about comments whose **visibility is not public** — for example, those visible **only to followers or mentioned users**?
   Should such comments still be **federated to unrelated servers**, or should delivery be restricted only to the **intended audience’s inboxes**?

3. In Misskey, all comments can be browsed regardless of context — isn’t this simply a **design limitation of Lemmy**?
   As far as I know, in the Fediverse, it’s actually **normal behavior** that **not all comments are federated**, depending on visibility or thread scope.

4. For **WordPress-to-WordPress comment synchronization**,
   which model would be more appropriate — a **push** model, a **pull** model, or something similar to **RSS polling**?

5. Personally, I find the **current group actor model** somewhat **spammy**.
   It doesn’t really improve feed readability, and often makes threads harder to follow.

---

As you know, I don’t think a **blog profile** is suitable as a **group actor**, since it isn’t a forum-type entity.
Wouldn’t it make more sense to discuss federation with Lemmy **after** a proper **forum profile** is implemented in WordPress?

After all, Lemmy’s `post` type is based on **Page**, and since `Page` inherits the properties of a **Document**, it could potentially **conflict with WordPress’s media library** or related document structures.

Jiwoon-Kim:
By the way, has Lemmy’s **WebFinger ambiguity issue** been resolved?
[https://socialhub.activitypub.rocks/t/are-actors-and-webfingers-1-to-1/4539](https://socialhub.activitypub.rocks/t/are-actors-and-webfingers-1-to-1/4539)

> Lemmy is a good example here, as users and communities may in fact use the same name, and the WebFinger is therefore ambiguous as to whether you meant the user or the community.

Wouldn’t a conflict occur in WordPress if the blog profile handle and the user profile handle are identical?

Jiwoon-Kim:
I have a simple proposal to improve the **Group Actor model**:

1. If a Group Actor is treated entirely as a forum, then even when an object's visibility is `public`, make it so that a **Person Actor sends a `Create` activity targeting only that specific Group Actor**.
   The Group Actor can then announce the post via an `Announce` activity.
   This ensures that if other users do not follow the forum, the object will generally not appear in their feeds.

2. **Prevent Group Actors from announcing comments.**
   This clutters the feed unnecessarily.
   Federated comments should be synchronized differently — for example, through the `context` of the top-level post.

mediaformat:
A few trust and safety questions: 

- If a benign comment is approved, and later edited will it return to moderation?
- Is there a notification for status updates?
- Is there a mechanism to undo an Announce?

If not, this could become a spam or misinfo vector.

I understand the desire to bootstrap WP federation, but I think Announces might best be left to author/editor control.

PR #1563, opened 2025-04-10
Implement FEP-c180: Problem Details for ActivityPub

pfefferle:
This PR provides the foundation for a unified error format, based on the FEP-c180. It implements it for a set of known errors and [uses the `about:blank` as fallback](https://codeberg.org/fediverse/fep/src/branch/main/fep/c180/fep-c180.md#specification).

The code is already fully functional and can be merged. Improvements for specific formats or new error `types` that might be defined in the future, can be done as follow up PRs.

See the proposal here: https://codeberg.org/fediverse/fep/src/branch/main/fep/c180/fep-c180.md

The main user stories from the spec:

* As an ActivityPub API client developer, I want a machine-readable description of an error from the server, so that I can provide actionable and localised error messages to the user.
* As an ActivityPub API client developer, I want to know if I can recover from an erroneous request, so that I can provide a more robust experience.
* As a server developer, I want to know if I can recover from an erroneous request, so that I can provide more reliable communications.
* As a server developer, I want to know if a remote server would ever accept the activity I am sending, so that I can save resources by not sending activities that will be rejected.

## Proposed changes:

* Change error output to the schema defined in FEP-c180

### Other information:

## Testing instructions:

* Call a various API endpoint and try to make an error nous call

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Use FEP-c180 schema for error responses.

</details>


includes/rest/class-server.php
```diff
@@ -170,4 +172,64 @@ public static function request_parameter_order( $order, $request ) {
...
+	public static function filter_output( $response, $server, $request ) {
+		$route = $request->get_route();
+
+		// Check if it is an activitypub request and exclude webfinger and nodeinfo endpoints.
+		if ( ! \str_starts_with( $route, '/' . ACTIVITYPUB_REST_NAMESPACE ) ) {
+			return $response;
+		}
+
+		// Only alter responses that return an error status code.
+		if ( $response->get_status() < 300 ) {
```
obenland: Could this be replaced with an `is_wp_error()` check?

includes/rest/class-server.php
```diff
@@ -170,4 +172,64 @@ public static function request_parameter_order( $order, $request ) {
...
+			case 'activitypub_signature_verification':
+				$error['title'] = 'Invalid Signature';
+				break;
+			case 'activitypub_comment_not_found':
+			case 'activitypub_local_only_comment':
+			case 'activitypub_post_not_found':
+			case 'activitypub_user_not_found':
+				$error['type']   = 'https://w3id.org/fep/c180#object-does-not-exist';
+				$error['status'] = 400;
+				$error['title']  = 'Object does not exist';
```
obenland: If I understand the spec correctly, this is about activities referencing a non-existing object, which is/might be different from what these `*_not_found` error codes refer to?

> The activity refers to an object in one of its properties, such as object, target, or an addressing property, but the object does not exist.

includes/rest/class-server.php
pfefferle: the problem is, that this is already a `WP_REST_Response`. I searched for an earlier hook that would allow me to customize the "from WP_Error to WP_REST_Response" process, but sadly there is none :(

includes/rest/class-server.php
pfefferle: > Note that this type is distinct from an endpoint returning a 404 Not Found status code for a GET request for an object that does not exist, or for posting to an endpoint that does not exist.

🤔

This makes it even more complicated :(

pfefferle:
@obenland I simplified it to simply reformat the WP_Error response to map the key/value pairs of FEP-c180

includes/rest/class-server.php
```diff
@@ -170,4 +172,48 @@ public static function request_parameter_order( $order, $request ) {
...
+	public static function filter_output( $response, $server, $request ) {
+		$route = $request->get_route();
+
+		// Check if it is an activitypub request and exclude webfinger and nodeinfo endpoints.
+		if ( ! \str_starts_with( $route, '/' . ACTIVITYPUB_REST_NAMESPACE ) ) {
+			return $response;
+		}
+
+		// Only alter responses that return an error status code.
+		if ( $response->get_status() < 300 ) {
```
obenland:
```suggestion
		if ( $response->get_status() < 400 ) {
```
I think changing that would be good, before merging

obenland (approved):

PR #1564, opened 2025-04-10
Fix: Missing Actor in Outbox-Activities

pfefferle:
In some cases the Actor of an Activity will not properly set and we wrongly set the `inReplyTo` for the Activity.

These fixes are cherry picked from #1562 

## Proposed changes:

* Add Actor before storing the Activity in the Outbox
* Add Actor before federating the Activity (for backwards compatibility)
* Remove reply-to from Activity context

### Other information:

## Testing instructions:

* Publish a post
* Check outbox item if Activity has an Actor

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Ensure that an `Activity` is properly built before adding it to the Outbox.

</details>


includes/activity/class-activity.php
```diff
@@ -233,10 +233,6 @@ public function pre_fill_activity_from_object() {
 			$this->set( 'actor', $object->get_attributed_to() );
 		}
 
-		if ( $object->get_in_reply_to() && ! $this->get_in_reply_to() ) {
-			$this->set( 'in_reply_to', $object->get_in_reply_to() );
-		}
```
obenland, line 238: How does this relate to the missing actor field?

obenland (commented):
> In some cases the Actor of an Activity will not properly set.

Do we know which cases those are?

pfefferle:
If you use the `add_to_outbox()` and let the function build the Activity, like with `add_to_outbox( $activity, 'Announce', $id )`.

includes/activity/class-activity.php
pfefferle: Updated readme and title

obenland (approved):

PR #1566, opened 2025-04-10
Upgrades: Delete orphaned extra fields

obenland:
Follow-up to #1554.

## Proposed changes:

* Adds an upgrade routine that deletes any user extra fields where the post author is the blog user.

## Testing instructions:

* Apply this PR on a test site with user extra fields that were generated through the bug in #1554.
* Manually adjust the version constant to a higher version (5.6.1.1, for example) and use the same version in the version check.
* Load wp-admin.
* Check in phpmyadmin that those posts no longer exist.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Upgrade routine that removes any erroneously created extra field entries.
</details>

### DIFF
--- a/.github/changelog/1566-from-description
+++ b/.github/changelog/1566-from-description
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Upgrade routine that removes any erroneously created extra field entries.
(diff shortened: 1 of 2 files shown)


akirk (approved):
Tested directly calling `Activitypub\Migration::delete_mastodon_api_orphaned_extra_fields();` from CLI and it worked! Thanks!

pfefferle (approved):

PR #1567, opened 2025-04-10
Transformers: Allow setting properties to false.

### LINKED ISSUE #1565: Bug: Transformer does not allow setting attribute/property to `false`
It should be possible to set an attribute to `false`. Within the `Base::transform_object_properties()`  function this does not work anymore:

https://github.com/Automattic/wordpress-activitypub/blob/1dbcc8ec37e5a0ccc5313fca46389ba207c140a1/includes/transformer/class-base.php#L93

Introduced in https://github.com/Automattic/wordpress-activitypub/pull/1301

> Updates Base::transform_object_properties() and Base_Object::to_array() to not only discard null values but also empty values unless they're false.

Seems like `empty( false )` evaluates to true, which I also did not know.

obenland:
Fixes #1565.

## Proposed changes:

* Adds an explicit check for false values when evaluating getter results.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Transformers allow settings values to false again, a regression from 5.5.0.

</details>

@Menrath


pfefferle:
I think we should unify all of the transformation methods!

See https://github.com/Automattic/wordpress-activitypub/pull/1568

pfefferle:
I think it is fine to use `isset( $value )` for everything. We should skip `null`, but I think `false`, `''` and `array()` are fine.

pfefferle:
and we should also check the rest of the code. there is at least the from_array, that needs to be also fixed.

includes/transformer/class-base.php
```diff
@@ -90,7 +90,7 @@ protected function transform_object_properties( $activity_object ) {
 			if ( \method_exists( $this, $getter ) ) {
 				$value = \call_user_func( array( $this, $getter ) );
 
-				if ( ! empty( $value ) ) {
+				if ( ! empty( $value ) || false === $value ) {
```
pfefferle:
```suggestion
				if ( isset( $value ) ) {
```

obenland: I went with an explicit `null` check, if that's okay

pfefferle: why not isset? the result is the same and we are sure that the var exists (not always guaranteed when iterating through the array!

obenland: Just to be explicit. I wouldn't expect there to be a situation where `$value` is truly not set after calling `call_user_func()`.

pfefferle: but then we would at least have exactly the same behaviour everywhere (to_array is using isset)... but I do not want to be pedantic...

pfefferle (approved):

Menrath:
> I think it is fine to use `isset( $value )` for everything. We should skip `null`, but I think `false`, `''` and `array()` are fine.

Note that we in the events world have a case where a decidated null might have information, but as long as one can still can the setter manually this is totally fine.

PR #1586, opened 2025-04-16
Make notification settings user-specific

### LINKED ISSUE #1435: Notifications: better alignment
### What

Image

This looks not great.

The checkboxes should be moved left and aligned.

### Why

Inconsistent look and feel

### How

Image

Should look more like the checkboxes above.

obenland:
The updated emails link to notification settings to make it easier to update them when receiving these notifications.
This PR makes them user-specific, with the default being the current global settings.
The global settings become the settings for the blog user.

## Proposed changes:

* Moves global notification settings to blog settings.
* Adds user profile settings fields.
* Adds user options to manage notification options.
* Update Mailer to check for user and blog options.
* Updates existing unit tests and adds new ones.
* Deprecates Notification class.

### Other information:

## Testing instructions:

* While still in trunk, set a few notification settings.
* Switch to this branch and make sure the setting values for the blog and your user are the same.
* Send a DM, a new follow, and/or a mention from a remote test user.
* Make sure notification settings are being respected.
* Update notification settings and keep testing them.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Updated notification settings to be user-specific for more personalization.

</details>

Fixes #1435.

### DIFF
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -942,14 +942,9 @@ class Migration {
 
 		// Add the actor notification options.
 		foreach ( Actors::get_collection() as $actor ) {
-			\update_user_option( $actor->get__id(), 'activitypub_mailer_new_mention', 1 );
-
-			if ( '1' === $new_dm ) {
-				\update_user_option( $actor->get__id(), 'activitypub_mailer_new_dm', 1 );
-			}
-			if ( '1' === $new_follower ) {
-				\update_user_option( $actor->get__id(), 'activitypub_mailer_new_dm', 1 );
-			}
+			\update_user_option( $actor->get__id(), 'activitypub_mailer_new_dm', $new_dm );
+			\update_user_option( $actor->get__id(), 'activitypub_mailer_new_follower', $new_follower );
+			\update_user_option( $actor->get__id(), 'activitypub_mailer_new_mention', '1' );
 		}
 
 		// Delete the old notification options.
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -743,7 +743,6 @@ class Migration {
 	 */
 	public static function add_default_settings() {
 		self::add_activitypub_capability();
-		self::add_notification_defaults();
 		self::add_default_extra_field();
 	}
 
@@ -796,15 +795,6 @@ class Migration {
 		foreach ( $users as $user ) {
 			$user->add_cap( 'activitypub' );
 		}
-	}
-
-	/**
-	 * Add default notification settings.
-	 */
-	private static function add_notification_defaults() {
-		\add_option( 'activitypub_blog_user_mailer_new_follower', '1' );
-		\add_option( 'activitypub_blog_user_mailer_new_dm', '1' );
-		\add_option( 'activitypub_blog_user_mailer_new_mention', '1' );
 	}
 
 	/**
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -929,7 +929,7 @@ class Migration {
 	}
 
 	/**
-	 * Add the new mention notification option.
+	 * Update notification options.
 	 */
 	public static function update_notification_options() {
 		$new_dm       = \get_option( 'activitypub_mailer_new_dm', '1' );
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -192,7 +192,7 @@ class Migration {
 			self::delete_mastodon_api_orphaned_extra_fields();
 		}
 		if ( \version_compare( $version_from_db, 'unreleased', '<' ) ) {
-			self::add_mention_notification_option();
+			self::update_notification_options();
 		}
 
 		/*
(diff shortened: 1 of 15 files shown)


pfefferle:
@obenland the notification class is not necessarily used for mailings, it was initially used for jetpack notifications (on dotcom), so we have to get sure that these will still work or add an alternative hook system!

I still think a notification object is a nice idea, because their provide a nice internal API and different messaging systems (including the mailer and jetpack notifications) can easily subscribe and send their pings.

includes/class-migration.php
```diff
@@ -802,9 +802,9 @@ private static function add_activitypub_capability() {
 	 * Add default notification settings.
 	 */
 	private static function add_notification_defaults() {
-		\add_option( 'activitypub_mailer_new_follower', '1' );
-		\add_option( 'activitypub_mailer_new_dm', '1' );
-		\add_option( 'activitypub_mailer_new_mention', '1' );
+		\add_option( 'activitypub_blog_user_mailer_new_follower', '1' );
```
pfefferle: isn't it better to have a proper default instead of writing so many default entries?

includes/wp-admin/class-admin.php
```diff
@@ -482,6 +491,10 @@ public static function handle_bulk_request( $sendback, $action, $users ) {
 			$user = new \WP_User( $user_id );
 			if ( 'add_activitypub_cap' === $action ) {
 				$user->add_cap( 'activitypub' );
+
+				\update_user_option( $user_id, 'activitypub_mailer_new_dm', 1 );
```
pfefferle: same as above: I would prefer to have a proper default instead of adding the default to the db.

includes/wp-admin/class-blog-settings-fields.php
```diff
@@ -187,6 +195,34 @@ class="blog-user-description large-text"
...
+			</p>
+			<p>
+				<label>
+					<input type="checkbox" name="activitypub_blog_user_mailer_new_dm" id="activitypub_blog_user_mailer_new_dm" value="1" <?php \checked( '1', \get_option( 'activitypub_blog_user_mailer_new_dm', '0' ) ); ?> />
+					<?php \esc_html_e( 'Direct Messages', 'activitypub' ); ?>
+				</label>
+			</p>
+			<p>
+				<label>
+					<input type="checkbox" name="activitypub_blog_user_mailer_new_mention" id="activitypub_blog_user_mailer_new_mention" value="1" <?php \checked( '1', \get_option( 'activitypub_blog_user_mailer_new_mention', '1' ) ); ?> />
```
pfefferle, line 218: why have we decided to have dms and followers as default off?

maybe we should re-review our default settings!?!

includes/wp-admin/class-settings.php
```diff
@@ -272,6 +242,36 @@ public static function register_settings() {
 			)
 		);
 
+		\register_setting(
+			'activitypub_blog',
+			'activitypub_mailer_new_dm',
+			array(
+				'type'        => 'boolean',
+				'description' => 'Send a notification when someone sends a user of the blog a direct message.',
+				'default'     => false,
```
pfefferle: this is really weird... we set the default to `false` and add an (user_)option instead. Let's fix that as part of this PR. we should not rely on that after the re-write.

includes/class-migration.php
obenland: For options, sure! I updated these to stick with the precedence.

includes/wp-admin/class-admin.php
obenland: Unfortunately, user options don't have the same defaults mechanism that options have, but I could probably build something similar with the filter in get_user_option(). Will give that a try.

includes/wp-admin/class-blog-settings-fields.php
obenland: Yes, it looks like there were different defaults introduced in #1109. Happy to change them all to `1`.

includes/wp-admin/class-settings.php
obenland: Updated in [1b93797](https://github.com/Automattic/wordpress-activitypub/pull/1586/commits/1b937971ab5c7fbc3d193b3db474ba53144e1736)

includes/wp-admin/class-blog-settings-fields.php
obenland: Updated in [1b93797](https://github.com/Automattic/wordpress-activitypub/pull/1586/commits/1b937971ab5c7fbc3d193b3db474ba53144e1736)

includes/class-migration.php
obenland: Updated in [1b93797](https://github.com/Automattic/wordpress-activitypub/pull/1586/commits/1b937971ab5c7fbc3d193b3db474ba53144e1736)

includes/wp-admin/class-admin.php
obenland: Updated in [e208e0d](https://github.com/Automattic/wordpress-activitypub/pull/1586/commits/e208e0dcdbedcf9bdd29c89615e975f550352ca9)

pfefferle:
I do not receive any direct-messages or @-mentions for the user or the blog user, regardless of the setting.

pfefferle (approved):
All mails seem to still work... have not tested the migration...

PR #1590, opened 2025-04-17
Inbox: Apply disallowed list before processing

### LINKED ISSUE #1589: Feature Request: Apply disallowed keys in Inbox
### What

Do the thing WordPress Core does, to send comments/pings to trash that have a hit from disallowed keys, already in the Inbox, so these requests don't even get processed.

### Why

Disallowed keys prevent comments, likes, and reposts from becoming approved, but they don't prevent notifications from being sent for DMs, mentions, and new followers. They also don't prevent new followers, I believe?

### How

_No response_

obenland:
Fixes #1589.

## Proposed changes:

* Adds `wp_check_comment_disallowed_list()` check in Inbox endpoint.
* Adds unit tests.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Added WordPress disallowed list filtering to block unwanted ActivityPub interactions.
</details>

### DIFF
--- a/includes/rest/class-inbox-controller.php
+++ b/includes/rest/class-inbox-controller.php
@@ -8,6 +8,7 @@
 namespace Activitypub\Rest;
 
 use Activitypub\Activity\Activity;
+use Activitypub\Debug;
 
 /**
  * Inbox_Controller class.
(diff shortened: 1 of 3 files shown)


includes/rest/class-inbox-controller.php
```diff
@@ -127,27 +128,31 @@ public function register_routes() {
...
-		 * @param array              $data     The data array.
-		 * @param int|null           $user_id  The user ID.
-		 * @param Activity|\WP_Error $activity The Activity object.
-		 */
-		\do_action( 'activitypub_inbox_' . $type, $data, null, $activity );
+		$type     = \strtolower( $request->get_param( 'type' ) );
+
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput
+		if ( wp_check_comment_disallowed_list( $activity->to_json( false ), '', '', '', $_SERVER['REMOTE_ADDR'], $_SERVER['HTTP_USER_AGENT'] ?? '' ) ) {
+			Debug::write_log( 'Blocked activity from: ' . $activity->get_actor() );
```
pfefferle, line 135: Let's prepare the responses in the if and else and prepare a `forbidden` error in this case (currently we always return a `202`).

Maybe also: https://codeberg.org/fediverse/fep/src/branch/main/fep/c180/fep-c180.md#actor-not-authorized (but this can be done as part of #1563 )

pfefferle:
So simple and smart! I love it!

includes/rest/class-inbox-controller.php
obenland: I tried to find out whether there were expectations around response codes etc when activities from a blocked user make it to an inbox but couldn't really find much.

My understanding of the desired behavior is that any blocked user shouldn't be notified/see that they're blocked, so returning the general success message while we deal with the content internally seemed prudent to me.

The request itself was successful after all, we just decide to not do anything with the contents.

pfefferle: An API error is not user-facing and might help the API to limit requests, but I am fine with the PR as a first step.

pfefferle (approved):

Issue #1295, opened 2025-02-08
Reaction Block language.


nickbohle:
### Quick summary

Language. My backend is English, my Frontend is German. The Frontend reaction block is English (and I am quite sure it is already translated)

### Steps to reproduce

German WordPress (frontend) with English Backend.

Reactions block is English and it should be German.

### Site owner impact

Fewer than 20% of the total website/platform users

### Severity

Moderate

mattwiebe:
Interesting bug! Maybe the `@wordpress/i18n` package doesn't distinguish between admin and front end like it should, assuming everywhere is admin.

In the meantime, the title can be directly edited inside the Site Editor, in the Posts template (where it can also be toggled off entirely if desired).

pfefferle:
@mattwiebe I am not sure if it is an issue with wp-admin. I have both sides set to German and my frontend-blocks also only show up in english.

obenland:
I'm not sure this is a bug with translation loading, as much as it may be an unexpected translation?

The strings `%d likes` and `%d reposts` are translated with [`%s Likes`](https://translate.wordpress.org/projects/wp-plugins/activitypub/dev/de/default/?filters%5Bterm%5D=%25d+likes) and [`%s Reposts`](https://translate.wordpress.org/projects/wp-plugins/activitypub/dev/de/default/?filters%5Bterm%5D=%25d+reposts).

So if you see a capitalized version of those words, they are the translation. The title of the block can be changed in the post/site editor.

pfefferle:
But it is not only that, the complete overlays are also not translated!

obenland:
Do you mean the Follow Me overlay? Or another? Are language packs up-to-date?

<img width="975" alt="Image" src="https://github.com/user-attachments/assets/0a69549c-2565-46f7-bb49-9d932127e95e" />

pfefferle:
🤔

PR #1592, opened 2025-04-17
Block i18n improvements

obenland:
See #1295.

## Proposed changes:

* Adds missing translators comments.
* Adds missing translations for reactions placeholders.
* Adds missing dev dependencies.
* Removes unused imports.


obenland:
I'm hoping this helps with pulling in more strings to be translated. The German language pack for the Reaction block only contains translations for the title of the block, nothing else.

Running `wp i18n make-pot . languages/activitypub.pot --include=src` locally includes all other strings now, so I hope that'll be the case in GlotPress, too.

/cc @jeherve in case you have an idea what else might be missing to make block translations work properly.

src/follow-me/follow-me.js
```diff
@@ -124,6 +124,7 @@ function Follow( {
 	buttonSize,
 } ) {
 	const [ isOpen, setIsOpen ] = useState( false );
+	/* translators: %s: profile name */
 	const title = sprintf( __( 'Follow %s', 'activitypub' ), profile?.name );
```
jeherve: For the translator comment to be properly picked up by GlotPress, it would need to be right before the translatable string:

```suggestion
	const title = sprintf(
		/* translators: %s: profile name */
		__( 'Follow %s', 'activitypub' ),
		profile?.name
	);
```

src/reactions/edit.js
```diff
@@ -82,11 +82,13 @@ const generateDummyReaction = ( index ) => {
  */
 const generateDummyReactions = () => ( {
 	likes: {
-		label: '9 likes',
+		/* translators: %d: Number of likes */
+		label: sprintf( _x( '%d likes', 'number of likes', 'activitypub' ), 9 ),
```
jeherve: Same like above.

src/reactions/edit.js
```diff
@@ -82,11 +82,13 @@ const generateDummyReaction = ( index ) => {
...
 		items: Array.from( { length: 9 }, ( _, i ) => generateDummyReaction( i ) ),
 	},
 	reposts: {
-		label: '6 reposts',
+		/* translators: %d: Number of reposts */
+		label:  sprintf( _x( '%d reposts', 'number of reposts', 'activitypub' ), 6 ),
```
jeherve: Same like above.

jeherve (commented):
> The German language pack for the Reaction block only contains translations for the title of the block, nothing else.

This is odd. I see the translations in the language pack I downloaded from translate.wordpress.org:

image

I also do not experience any issues with the translations on my own site (set to French).

> Running `wp i18n make-pot . languages/activitypub.pot --include=src` locally includes all other strings now, so I hope that'll be the case in GlotPress, too.

A better test may be to run this against the `build` directory, since the `src` directory does not ship to SVN. GlotPress will be looking at the build directory for those strings.

I should also note that the translations for the number of likes and reposts displayed on the frontend come from different strings than the ones you edited here, in the `src` directory. On the frontend, those numbers come from an API call, and thus the translations are set in PHP:
https://github.com/Automattic/wordpress-activitypub/blob/2f347b19827210fe55c2dc8cb82eda4dba389bc2/src/reactions/reactions.js#L317-L321
https://github.com/Automattic/wordpress-activitypub/blob/007877cf32f2d85ff70a449b4c97f072fc82aa63/includes/rest/class-post.php#L101-L111
https://github.com/Automattic/wordpress-activitypub/blob/3166870da5be7e2ca1128e7348d460d50ccd6217/includes/class-comment.php#L672

obenland:
Running `wp i18n make-pot . languages/activitypub.pot` includes this:
```
#. translators: %d: Number of reposts
#: includes/class-comment.php:655
#: build/reactions/index.js:3
#: src/reactions/edit.js:95
msgctxt "number of reposts"
msgid "%d reposts"
msgstr ""
```
Which looks right.

pfefferle (approved):

PR #1598, opened 2025-04-23
Directly link to a help tab.

pfefferle:
This PR allows you to directly link to help tabs. The idea is to use it for the Onboarding as part of a launchpad.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Open the ActivityPub admin page with the hash of the help tab you want to open. The link for the "Account Migration" help tab is `/wp-admin/options-general.php?page=activitypub#tab-link-account-migration` for example.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Enable direct linking to Help Tabs.

</details>

### DIFF
--- a/includes/wp-admin/class-admin.php
+++ b/includes/wp-admin/class-admin.php
@@ -612,21 +612,27 @@ class Admin {
 		\load_template( ACTIVITYPUB_PLUGIN_DIR . 'templates/toolbox.php' );
 	}
 
+	/**
+	 * Open the help tab.
+	 *
+	 * This function is used to open the help tab,
+	 * it is triggered by the hash in the URL.
+	 */
 	public static function open_help_tab() {
-	?>
-	<script type="text/javascript">
-	jQuery(document).ready(function($) {
-		var hash = window.location.hash;
-		if (hash) {
-			// Small delay to ensure the help tab is loaded
-			setTimeout(function() {
-				$('#contextual-help-link').click();
-				console.log(hash + ' a');
-				$(hash + ' a').click();
-			}, 500);
-		}
-	});
-	</script>
-	<?php
+		?>
+		<script type="text/javascript">
+		jQuery(document).ready(function($) {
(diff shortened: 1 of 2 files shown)


includes/wp-admin/class-admin.php
```diff
@@ -60,6 +60,8 @@ public static function init() {
 		if ( site_supports_blocks() ) {
 			\add_action( 'tool_box', array( self::class, 'tool_box' ) );
 		}
+
+		\add_action( 'admin_print_footer_scripts', array( self::class, 'open_help_tab' ) );
```
obenland: Would it still work if we used `"admin_print_footer_scripts-{$hook_suffix}"` and limited it to our settings page?
If we can limit it to the settings page, maybe the callback should live in class-settings?

includes/wp-admin/class-admin.php
```diff
@@ -609,4 +611,28 @@ public static function plugin_update_message( $data, $update ) {
...
+			var hash = window.location.hash;
+			if (hash) {
+				// Small delay to ensure the help tab is loaded
+				setTimeout(function() {
+					$('#contextual-help-link').click();
+					console.log(hash + ' a');
+					$(hash + ' a').click();
+				}, 500);
+			}
+		});
```
obenland, line 644: I got a jQery migrate warning in my testing, `jQuery.fn.click() event shorthand is deprecated`.
Could we just use vanilla JS for this snippet?

obenland (commented):
Nice! Small, yet powerful.

I tested it with the path you suggested in the issue description but it didn't work immediately, I had to change the hash to `#tab-link-account-migration`. 

That made me think that we might want to add an allow list of links that can automatically be clicked, so users can't be tricked into clicking unintended links by getting passed malicious hashes?

obenland (approved):

PR #1601, opened 2025-04-24
First draft of a launchpad

pfefferle:
This is a first draft of the launchpad. Because there are no reusable methods, I think we should start with a simple PHP/HTML/CSS version and improve on that.

Screenshot 2025-04-24 at 13 26 37

## Todo

* [ ] Launchpad should load under the welcome tab and not introduce a separate one
* [ ] Maybe we should load the launchpad above the welcome settings when visiting the site the second time with still open items missing.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/activitypub.php
+++ b/activitypub.php
@@ -101,6 +101,7 @@ function plugin_admin_init() {
 	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Admin', 'init' ) );
 	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Health_Check', 'init' ) );
 	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Settings', 'init' ) );
+	\add_action( 'admin_init', array( __NAMESPACE__ . '\WP_Admin\Launchpad_Fields', 'init' ) );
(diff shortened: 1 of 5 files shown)


obenland:
Maybe it makes sense to support/re-use the same data structure as the Dotcom Launchpad? I assume we'd want a Calypso-based Launchpad for Simple sites?

pfefferle:
I am Not Site about that. It isnt really much more than links and a Status, why using a complex react implementation?

obenland:
Is that a pun? 😄

With data structure I mean using the same option names and data types, save progress in the same way, etc. Just so in case the Launchpad gets extended to wp-admin in the future, it'll be compatible. We still don't know what the Launchpad for Simple sites will look like, or if we want to do something for WoA sites, etc.

Happy to build it with PHP for the time being.

pfefferle:
oh, sorry... totally! let's reuse as much as possible from the options/settings that is used by other launchpads... and sorry for the cryptic comment... I had to buy a new handy yesterday and it seems that I forgot to add an en keyboard and the german auto-correct went a bit nuts 🫣

PR #1609, opened 2025-04-24
Release: Add prompt for upgrade notice

obenland:
I keep wanting to develop the habit of adding upgrade notices to our version updates, but every time we do one, I forget. It happened again today, when we released 5.8.0.

This change asks for an upgrade notice and adds it to the readme. If we want to skip it, we can.

This is from [my test run just now](https://github.com/Automattic/wordpress-activitypub/pull/1608/files#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6L370-L379):

```
$ npm run release

> release
> node bin/release.js

Switched to a new branch 'release/5.8.1'
Updated readme.txt with changelog entries for version 5.8.1 and other entries from major version 5

Would you like to add an upgrade notice for this version? (y/n): y
Enter the upgrade notice (leave empty to skip): Intoduces new email templates and notification settings. Update now and win!
Added upgrade notice for version 5.8.1 to readme.txt
[…]
```

## Proposed changes:

* Adds optional prompt to add an upgrade notice.

### DIFF
--- a/bin/release.js
+++ b/bin/release.js
@@ -150,6 +150,48 @@ const updateReadmeWithChangelog = (version) => {
 	console.log(`Updated readme.txt with changelog entries for version ${version} and other entries from major version ${majorVersion}`);
 };
 
+const updateReadmeWithUpgradeNotice = (version) => {
+	return new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


pfefferle (approved):
nice one!

PR #1619, opened 2025-04-28
Editor: Add control for max image attachments

### LINKED ISSUE #1354: Feature Request: Option to set the number of media attachments and template per post
### What

Rather than having a single global set of settings for the post template and number of media attachments to attach to a Fediverse post from this plugin, it would be nice to be able to set these per post when you're publishing a Wordpress blog post that will federate.

### Why

The type of blog posts I write varies: some are just text and can federate in their entirety, but others are longer and may have images interspersed throughout, but these end up looking pretty terrible when federated because the images aren't inline as they are in the blog post and there are only a limited number of them that will be attached. In the latter case it would be really nice to be able to change the ActivityPub template just for that specific post, to show just the excerpt with a link to the full post and not attach any images at all.

### How

The current Settings > Activities > "Post content" and "Media attachments"  sections could be included in the "Fediverse" panel when writing a blog post and would default to what's set in the plugin settings but would be able to be set per post.

obenland:
Fixes #1354.

Adds a per-post control over how many images to share to the Fediverse.

## Proposed changes:

* Adds Range control in ActivityPub post options.
* Adds new post meta to hold the max number of attachments to share.
* Updates Post transformer to look for the post meta information and fall back to the global setting.

### Other information:

## Testing instructions:

* Create a new post in the Edior.
* Insert a few images.
* Use the new selector to decide how many images to share with the Fediverse.
* Go to the Fediverse preview and make sure that number gets respected.
* Change the max number of images to be shared in Settings > Activitypub.
* Create a new post in the Editor and make sure the default number for the control reflects that settings.

#### After

image

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

The number of images shared to the Fediverse can now be chosen on a per-post basis.

</details>


src/editor-plugin/plugin.js
```diff
@@ -65,6 +65,17 @@ const EditorPlugin = () => {
...
 			/>
 
+			<RangeControl
+				label={ __( 'Maximum Image Attachments', 'activitypub' ) }
+				value={ meta?.activitypub_max_image_attachments || window._activityPubOptions?.maxImageAttachments || 4 }
+				onChange={ ( value ) => {
+					setMeta( { ...meta, activitypub_max_image_attachments: value } );
+				} }
+				min={ 0 }
+				max={ 10 }
```
obenland, line 75: Not sure if we ever thought about a max number of images/attachments that should be allowed? Should this be even limited?

includes/transformer/class-post.php
```diff
@@ -280,6 +280,9 @@ protected function get_attachment() {
 			return array();
 		}
 
+		// phpcs:ignore Universal.Operators.DisallowShortTernary
+		$max_media = \get_post_meta( $this->item->ID, 'activitypub_max_image_attachments', true ) ?: \get_option( 'activitypub_max_image_attachments', ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS );
```
pfefferle, line 284: this looks a bit ugly... can we maybe add a function for that?

pfefferle:
Is this really a common usecase? I know there was an issue, but is this really a problem that needs to be solved?

If we go this path, I would recommend to remove the global setting then in favour of fixed number (maybe as a const ☺️) to simplify things.

Or maybe an either/or?

* If you use the block editor -> show the block setting
* If not -> use the global

src/editor-plugin/plugin.js
pfefferle: ...if a user wants to mess up their posts...

includes/transformer/class-post.php
obenland:
```suggestion
		$max_media = \get_post_meta( $this->item->ID, 'activitypub_max_image_attachments', true );
		if ( ! is_numeric( $max_media ) ) {
			$max_media = \get_option( 'activitypub_max_image_attachments', ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS );
		}
```
Would this be better?

obenland:
> Is this really a common usecase? I know there was an issue, but is this really a problem that needs to be solved?

It adds more customizability. Maybe you have two galleries and don't what to have images from both show up in federated posts, etc.

> If we go this path, I would recommend to remove the global setting then in favour of fixed number (maybe as a const ☺️) to simplify things.

Sure, I'd be fine with that. I had left it for now as I looked at it as setting the default for that across posts, instead of having to adjust it for every post. But we already have a constant that could be used for that.

> Or maybe an either/or?
>
> If you use the block editor -> show the block setting
> If not -> use the global

I'd expect that to already work that way. The block editor setting would not show up in the classic editor.

obenland:
@pfefferle This should be ready for another look

includes/wp-admin/class-settings-fields.php
```diff
@@ -80,14 +80,16 @@ public static function register_settings_fields() {
...
 
-		add_settings_field(
-			'activitypub_max_image_attachments',
-			__( 'Media attachments', 'activitypub' ),
-			array( self::class, 'render_max_image_attachments_field' ),
-			'activitypub_settings',
-			'activitypub_activities',
-			array( 'label_for' => 'activitypub_max_image_attachments' )
-		);
+		if ( \is_plugin_active( 'classic-editor/classic-editor.php' ) ) {
```
pfefferle: can we reuse the `site_supports_blocks` function and add the plugin check there?

pfefferle (approved):

PR #1620, opened 2025-04-28
Inbox: Replace debug call with action

### LINKED ISSUE #1618: Inbox: Replace debugs with actions
Let's avoid calling Debug directly, and offer a general action instead.

obenland:
Fixes #1618, CMF-103.

No funcional changes.

## Proposed changes:

* Replaces Debug call in Inbox endpoints with action.
* Adds debug call to new action.


includes/rest/class-actors-inbox-controller.php
```diff
@@ -178,7 +178,15 @@ public function create_item( $request ) {
 
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput
 		if ( \wp_check_comment_disallowed_list( $activity->to_json( false ), '', '', '', $_SERVER['REMOTE_ADDR'], $_SERVER['HTTP_USER_AGENT'] ?? '' ) ) {
-			Debug::write_log( 'Blocked activity from: ' . $activity->get_actor() );
+			/**
```
pfefferle, line 181: What about simply changing the `$type` and removing the `else`? This way we could use the actions in both cases?

pfefferle: What do you think about using `block` as `$type`?

This is a valid ActivityType, so we would be somehow consistent with the spec and it is yet only used in C2S, that means that it should only be sent to the outbox and we would not receive a `Block` through the inbox: https://www.w3.org/TR/activitypub/#block-activity-outbox

pfefferle: But either way, I think we can at least remove the `_activity` at the end.

obenland: I tried to clearly differentiate this action from the wildcard activity type action. With using `block` I worry that it would be confusing with actual incoming Block activities that should be processed. My preference would be for it to be something completely different, so activity processing can be done reliably.

What if we broke with the `_inbox` namespace and called it `activitypub_disallowed_activity`?

pfefferle: This is a bit generic... We could use that if we provide a `$context` to give a hint where it happened?

obenland: In the Outbox endpoint we have a `acrivitypub_rest_outbox_*` namespece, what about `activitypub_rest_inbox_disallowed`?

obenland: Fixed in [339ca2f](https://github.com/Automattic/wordpress-activitypub/pull/1620/commits/339ca2f04c34f0ecd41d3fbe31ca13a82d782dfe).

pfefferle (approved):

PR #1625, opened 2025-04-29
Improve Welcome page with the focus on better onboarding.

pfefferle:
- [x] Remove hash after opening tab.
- [x] Set #id in profiles to link to.
- [x] Plugin cards in help tab should be same height 
- [x] Keep site health always, mark it as checked when there are no critical.
- [x] Make sure step count updates with option step updates.

Screenshot 2025-04-30 at 12 31 17

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

A guided onboarding flow after plugin activation to help users make key setup decisions and understand Fediverse concepts.

</details>


pfefferle:
@obenland can you have a look if that is the right direction?

includes/wp-admin/class-welcome-fields.php
```diff
@@ -117,35 +100,68 @@ public static function render_welcome_intro_section() {
...
-				sprintf( esc_html( $reply_from_template ), esc_html( \wp_parse_url( \home_url(), PHP_URL_HOST ) ) )
-			);
-			?>
-		</p>
-		<p>
-			<?php
-			/* translators: %s is where the button HTML will be rendered. */
 			\esc_html_e(
-				'Save this bookmarklet to reply to posts on other sites from your own blog! When visiting a post on another site, click the bookmarklet to start a reply.',
+				'Every beginning is a little tough—but you’re not alone. Start by following the checklist below. Explore the documentation, fine-tune your profile settings, and visit the help section for tips on connecting your site to the fediverse. For the best experience, make sure your site is healthy and your profile info is up to date.',
```
```suggestion
				'New beginnings can feel daunting—but you’re not alone. Start by following the checklist below. Explore the documentation, fine-tune your profile settings, and visit the help section for tips on connecting your site to the fediverse. For the best experience, make sure your site is healthy and your profile info is up to date.',
```

obenland:
I think it is! 

A few early notes that came to mind for me:
* I wouldn't call it Launchpad, even under the hood—it's not the same and it would probably be good to not create that expectation.
* The sections are being displayed incredibly densely, maybe because we don't have actual settings in them that gives it more whitespace? Probably not an issue with future visual iterations.
* I wonder if we could make the profile information part of a checklist item, too, to only be left with the list?
* I'm sure you have more visual polish planned for the list. If not, let's plan on it.

pfefferle:
> I'm sure you have more visual polish planned for the list. If not, let's plan on it.

Yes :)

Just wanted to have some early feedback.

pfefferle:
> I wonder if we could make the profile information part of a checklist item, too, to only be left with the list?

This is the second item!?! Or do you mean inline?

obenland:
> This is the second item!?! Or do you mean inline?

I mean everything below the site health section, starting with Blog Profile.
Maybe Site Health could be the first section in Settings?

includes/wp-admin/class-settings.php
```diff
@@ -350,6 +350,9 @@ public static function settings_page() {
 				\add_thickbox();
 				\wp_enqueue_script( 'updates' );
 				break;
+			case 'settings':
+				\update_option( 'activitypub_launchpad_settings_visited', true );
```
obenland: I wonder if this option is even needed. Could we use the `activitypub_actor_mode` option instead and see if it has a value? The task would be completed once they selected an actor mode and saved it.

pfefferle: what if the user does not change it, because they are fine with the default?

obenland: We could consider removing the default in the settings callback so the radio buttons are all empty. I guess it depends on whether we want them to understand it and make a conscious choice, or just consider it.

pfefferle: But this would force the user to go through every step, no "work out of the box", no easy skip.

obenland: Yeah, agreed.

obenland:
What do you have in mind in terms of visual improvements to how these steps are displayed? Closer to what Dave shared on P2? I'd be happy to help with making progress on this

pfefferle:
I have no idea tbh. Feel free to take over!

pfefferle:
Some context: The problem with the UI drafts so far, is that they were made with a launchpad in mind and we have decided to not build one, but a checklist instead. that it is why I added the checkboxes in the draft.

pfefferle:
and I would not move the site health check. the settings should only be for settings and the welcome screen should exactly show these informations in my opinion. kind of like a cheat sheet.

obenland:
How about making the Health Check stuff part of the list? Add a step to fix them if there are issues?

If all checkboxes are checked off and there are still health check issues, we could display them separately, like when they break after they were okay previously.

pfefferle:
I think this is a good idea. Showing the health check section only if the Checklist is done and gone and the site has still open issues.

obenland (approved):
Let's get it in and iterate as needed. Huzzah!

PR #1626, opened 2025-04-29
Fix: Set the correct Activity Actor for Outbox Items

pfefferle:
See changes

## Proposed changes:

* Fixed duplicate Announce check
* Fixed Handling of nested Activities
* Properly set Activity-Actor in `add_to_outbox` function

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Improved the handling of Shares and Boosts.

</details>

### DIFF
--- a/.github/changelog/1626-from-description
+++ b/.github/changelog/1626-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Improved the handling of Shares and Boosts.
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1477,6 +1477,15 @@ function is_self_ping( $id ) {
  * @return boolean|int The ID of the outbox item or false on failure.
  */
 function add_to_outbox( $data, $activity_type = null, $user_id = 0, $content_visibility = null ) {
+	// If the user is disabled, fall back to the blog user when available.
+	if ( ! user_can_activitypub( $user_id ) ) {
+		if ( user_can_activitypub( Actors::BLOG_USER_ID ) ) {
+			$user_id = Actors::BLOG_USER_ID;
+		} else {
+			return false;
+		}
+	}
+
 	$transformer = Transformer_Factory::get_transformer( $data );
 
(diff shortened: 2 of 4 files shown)


obenland:
I'm afraid it's not really obvious to me what this fixes.

> Fixed duplicate Announce check

What were the circumstances where we'd create duplicate announces?

> Fixed Handling of nested Activities

What used to go wrong?

Could we add unit tests that fail on trunk but pass with these changes to make that clearer?

pfefferle:
> Fixed duplicate Announce check

We currently use the object ID (which is the ID of the post) to check if the Announce is already received and stored, but we store the Activity-ID. So this check never really worked.

> > Fixed Handling of nested Activities

> What used to go wrong?

I have no idea, because it might be that it never really worked. this was something I found while fixing the main issue: The wrong actor when letting the `add_to_outbox` build an Activity.

> Could we add unit tests that fail on trunk but pass with these changes to make that clearer?

I will try!

pfefferle:
@obenland I added tests!

pfefferle:
@obenland Can you have another look?

obenland (approved):
Apologies, I did yesterday and then got sidetracked.

Thank you for adding the tests! I'm afraid I'm still not entirely clear on what's being fixed in Outbox, but I'm not sure I need to be. Existing tests not breaking is probably good enough on that. Don't know if it matters, but running the new tests in trunk has only `test_maybe_save_announce` failing, all `test_handle_announces` tests are passing.

In future PRs, let's try adding more context to the PR description, including testing steps, if possible. It'll also help our future selves retrace why changes were made in the past. For our new quest to make changelog entries more informative, I wonder if it would help to run them through a couple of rounds of ["So what?" tests](https://butlerscicomm.com/so-what-a-simple-test-for-concise-interesting-writing/)?

PR #1630, opened 2025-04-30
Support for WP REST Cache

obenland:
See https://wordpress.org/support/topic/caching-activity-pub-endpoints/
See https://epiph.yt/en/blog/2025/accidental-ddos-through-activitypub-plugin/

## Proposed changes:

* Adds integration with Rest Cache plugin.
* Adds Activitypub endpoints to allowlist and manages cache invalidation.

### Other information:

## Testing instructions:

* Install the WP Rest Cache plugin and set it up.
* Send requests to various Activitypub REST API endpoints and make sure they get cached.
* Create a user update, something that doesn't create a new post, and make sure it flushes the cache.
* Send requests to various Activitypub REST API endpoints and make sure they get cached.
* Comment on a post and make sure it flushes the cache.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Added support for the WP Rest Cache plugin to help with caching REST API responses.

</details>

/cc @MatzeKitt


obenland:
Proper cache invalidation will be the biggest challenge. New posts, new comments, user meta updates, blog user option updates, etc.

It would also be nice if we could avoid blanket-invalidating all cached ActivityPub endpoints for all of these, and be more targeted.

MatzeKitt:
So we need to keep track of any change regarding any of the endpoints? Also, do we need to check whether the change does actually change the API endpoint response or should we always invalidate the cache when it _might_ change the response?

I guess we currently have no decorated list of changes we need to look for?

obenland:
> Also, do we need to check whether the change does actually change the API endpoint response or should we always invalidate the cache when it might change the response?

Looking a bit more at the plugin, I think we'll have to be pretty broad about invalidating caches. Most of the endpoints are user/actor-based, and WP REST Cache doesn't have a concept for that. It's post type and post_id, and that seems to be pretty much it.

I think your approach of defining a custom type and invalidating all associated endpoints is probably the safest.

obenland:
Need to do a bit more testing before this is ready for review.

pfefferle:
What about user updates? Shouldn't we have hooks to invalideate the User endpoints on changes? Example: https://notiz.blog/wp-api/activitypub/1.0/actors/1

integration/load.php
```diff
@@ -123,6 +123,10 @@ function ( $transformer, $data, $object_class ) {
 		WPML::init();
 	}
 
+	if ( class_exists( 'WP_Rest_Cache_Plugin\Includes\Plugin' ) ) {
```
pfefferle:
```suggestion
	if ( \class_exists( 'WP_Rest_Cache_Plugin\Includes\Plugin' ) ) {
```

integration/class-wp-rest-cache.php
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Add ActivityPub endpoints to the list of allowed endpoints.
+	 *
+	 * @param array $endpoints List of allowed endpoints.
+	 *
+	 * @return array
+	 */
+	public static function add_activitypub_endpoints( $endpoints ) {
+		$endpoints[ ACTIVITYPUB_REST_NAMESPACE ] = array( 'actors', 'collections', 'comments', 'interactions', 'nodeinfo', 'posts', 'users' );
```
pfefferle, line 50: isn't the namespace simply `activitypub` or `activitypub/1.0`?

integration/class-wp-rest-cache.php
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	/**
+	 * Set object type for ActivityPub.
+	 *
+	 * @param string $object_type Object type.
+	 * @param string $cache_key   Object key.
+	 * @param mixed  $data        Data to cache.
+	 * @param string $uri         Request URI.
+	 * @return string Updated object type
```
pfefferle:
```suggestion
	 *
	 * @return string Updated object type.
```

integration/class-wp-rest-cache.php
```diff
@@ -0,0 +1,145 @@
...
+		}
+
+		Caching::get_instance()->delete_object_type_caches( 'ActivityPub' );
+	}
+
+	/**
+	 * Test, whether the current endpoint is an ActivityPub endpoint.
+	 *
+	 * @param string $uri URI to test.
+	 * @return bool Whether the current endpoint is an ActivityPub endpoint
```
pfefferle:
```suggestion
	 *
	 * @return bool Whether the current endpoint is an ActivityPub endpoint.
```

integration/class-wp-rest-cache.php
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Set whether the cache represents a single item.
+	 *
+	 * Always return false for ActivityPub endpoints, since cache entries cannot be flushed otherwise.
+	 *
+	 * @param bool   $is_single Whether the current cache represents a single item.
+	 * @param mixed  $data      Data to cache.
+	 * @param string $uri       Request URI.
+	 * @return bool Whether the cache represents a single item
```
pfefferle:
```suggestion
	 *
	 * @return bool Whether the cache represents a single item.
```

integration/class-wp-rest-cache.php
```diff
@@ -0,0 +1,145 @@
...
+		\add_action( 'transition_post_status', array( self::class, 'transition_post_status' ), 10, 3 );
+		\add_action( 'transition_comment_status', array( self::class, 'transition_comment_status' ), 10, 3 );
+	}
+
+	/**
+	 * Add ActivityPub endpoints to the list of allowed endpoints.
+	 *
+	 * @param array $endpoints List of allowed endpoints.
+	 *
+	 * @return array
```
pfefferle:
```suggestion
	 * @return array Filtered list of allowed endpoints.
```

obenland:
> What about user updates? Shouldn't we have hooks to invalideate the User endpoints on changes?

They should work, it's a step in the testing instructions "Create a user update, something that doesn't create a new post, and make sure it flushes the cache."

User updates should trigger a flush through the post status change in the Outbox post they produce.

integration/class-wp-rest-cache.php
obenland: Yes, `activitypub/1.0`

pfefferle: I meant, can we wildcard all the subpathes?

pfefferle: with this specific list, we would have to take care that every new endpoint will also be added here... that is a bit fragile!?!

obenland: Ah! Yeah, unfortunately the plugin is pretty literal in its allowlist processing:

https://github.com/RedCastor/wp-rest-cache/blob/c6f94c7b61f30770424fa5dd5f4ccd4e0d7f52b9/includes/api/class-endpoint-api.php#L241-L251

We _could_ grab all of our registered routes and format them to fit this list, but it would be like using a sledgehammer to crack a nut. I'm not too concerned, our endpoint routes have been pretty stable and chances are that any new endpoints are likely to be sub-paths of `actors`, `users`, etc

pfefferle (approved):

PR #1631, opened 2025-04-30
Max Image Attachments: Default to integer

obenland:
It's possible for the `register_post_meta` call to trigger a `_doing_it_wrong()` notice, when the default value is not of type integer.

## Proposed changes:

* Adds unit test.
* Adds option callback to make sure it always returns an integer.
* Tightens sanitization to make sure it never saves an empty string.

### Other information:

## Testing instructions:

* While on trunk, activate Classic Editor plugin to make the setting show up.
* Delete the number in the max image attachments setting and save.
* Make sure the input field is empty and a `_doing_it_wrong()` notice appears (DEBUG enabled).
* Switch to this branch. The notice should no longer appear.

### DIFF
--- a/includes/class-sanitize.php
+++ b/includes/class-sanitize.php
@@ -119,15 +119,4 @@ class Sanitize {
 
 		return $value;
 	}
-
-	/**
-	 * Sanitize the max image attachments.
-	 *
-	 * @param string $value The value to sanitize.
-	 *
-	 * @return int The sanitized value.
-	 */
-	public static function max_image_attachments( $value ) {
-		return \is_numeric( $value ) ? \absint( $value ) : ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS;
-	}
 }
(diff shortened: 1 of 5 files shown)


includes/class-sanitize.php
```diff
@@ -119,4 +119,15 @@ public static function constant_value( $value ) {
...
 	}
+
+	/**
+	 * Sanitize the max image attachments.
+	 *
+	 * @param string $value The value to sanitize.
+	 *
+	 * @return int The sanitized value.
+	 */
+	public static function max_image_attachments( $value ) {
```
pfefferle: this is very specific to the `max_image_attachment`, do we really want to have that in the generic Sanitize class?

obenland: I was on the fence about it. Wasn't sure we wanted to capture all sanitization callbacks here. I can change it

obenland: Fixed in [967ba2f](https://github.com/Automattic/wordpress-activitypub/pull/1631/commits/967ba2f3dff7fe9cb477c28cd76ded397e4915eb).

pfefferle (approved):

PR #1632, opened 2025-04-30
Onboarding: Isolate profile settings during onboarding

obenland:
Depends on #1625.

## Proposed changes:

* Creates separate registration callback for profile section and setting.
* Hooks settings sections conditionally on onboarding task having been completed.

## Testing instructions:

* Go to Settings > Activitypub.
* Without #1625 the onboarding option should not be set and it should only display the Profile setting.
* After testing #1625, the option should be set and the full settings screen should show.

#### After
Screenshot 2025-04-30 at 2 50 50 PM

### DIFF
--- a/includes/class-sanitize.php
+++ b/includes/class-sanitize.php
@@ -16,12 +16,12 @@ class Sanitize {
 	/**
 	 * Sanitize a list of URLs.
 	 *
-	 * @param string|array $value The value to sanitize.
+	 * @param string|array|null $value The value to sanitize.
 	 * @return array The sanitized list of URLs.
 	 */
 	public static function url_list( $value ) {
 		if ( ! \is_array( $value ) ) {
-			$value = \explode( PHP_EOL, $value );
+			$value = \explode( PHP_EOL, (string) $value );
 		}
 
 		$value = \array_filter( $value );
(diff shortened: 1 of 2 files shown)


pfefferle:
I am not sure that this is a good idea... it might confuse more that it helps... and what is the problem if the user also checks the other entries?

obenland:
It's about focus and removing distractions. If we want new users to understand actor mode, let's just give them that and not confuse them with all the other things we have there.

pfefferle:
but you will see different versions of the settings site based on a state... and that is also bad ux, because how does the user know that we have more settings?

pfefferle:
Sorry for updateing the branch, but it showed all changes from the docs PR and it was hard to see the real changes 🫣

pfefferle:
If we want to do that, I would prefer highlighting the profile settings instead of removing the rest. It should not be impossible to also tweak the rest of the settings, but we want to get sure, that the user at least changes the profile settings.

obenland:
How strongly do you feel about that? I genuinely believe it'll make for a better user experience if we show the profile settings on their own.

We could decouple it from the actor mode setting, and only show it that way when someone gets there through clicking the onboarding link?

pfefferle:
strongly, because this feels very inconsistent and confusing!

pfefferle:
And the current link even opens the help page with the description of the Actor-Settings, so it should be self explaining like it is.

If people are confused, by the current settings, we still can improve.

PR #1635, opened 2025-05-01
Fix minor code changes

obenland:
## Proposed changes:

* Documents and adds missing return values.
* Removes unnecessary namespace imports.
* Replaces some full namespace references with imports.
* Adds missing `alt` attributes.
* Updates variables to use snake_case.
* Removes function call parameters that are defaults.
* Removes unnecessary namespace qualifiers.
* Fixes some more comment typos.

### DIFF
--- a/includes/constants.php
+++ b/includes/constants.php
@@ -7,19 +7,19 @@
 
 // The following constants can be defined in your wp-config.php file to override the default values.
 
-\defined( 'ACTIVITYPUB_REST_NAMESPACE' ) || \define( 'ACTIVITYPUB_REST_NAMESPACE', 'activitypub/1.0' );
-\defined( 'ACTIVITYPUB_EXCERPT_LENGTH' ) || \define( 'ACTIVITYPUB_EXCERPT_LENGTH', 400 );
-\defined( 'ACTIVITYPUB_NOTE_LENGTH' ) || \define( 'ACTIVITYPUB_NOTE_LENGTH', 400 );
-\defined( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS' ) || \define( 'ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS', 4 );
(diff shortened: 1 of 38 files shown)


includes/constants.php
```diff
@@ -61,15 +61,15 @@
...
-\define( 'ACTIVITYPUB_ACTOR_AND_BLOG_MODE', 'actor_blog' );
+define( 'ACTIVITYPUB_ACTOR_MODE', 'actor' );
+define( 'ACTIVITYPUB_BLOG_MODE', 'blog' );
+define( 'ACTIVITYPUB_ACTOR_AND_BLOG_MODE', 'actor_blog' );
 
 // Post visibility constants.
-\define( 'ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC', '' );
-\define( 'ACTIVITYPUB_CONTENT_VISIBILITY_QUIET_PUBLIC', 'quiet_public' );
-\define( 'ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE', 'private' );
-\define( 'ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL', 'local' );
```
pfefferle, line 75: it does not hurt to have it also in non-namespaced files ;)

includes/handler/class-announce.php
```diff
@@ -85,7 +85,7 @@ public static function handle_announce( $announcement, $user_id, $activity = nul
 		 * @param int                                 $user_id  The id of the local blog-user.
 		 * @param \Activitypub\Activity\Activity|null $activity The activity object.
 		 */
-		\do_action( "activitypub_inbox_{$type}", $object, $user_id, $activity );
```
pfefferle: isn't that how WordPress is doing it?

pfefferle (dismissed):

includes/handler/class-announce.php
obenland: Sorry, I missed this one in reverting the curly brace changes. Fixed in [8741903](https://github.com/Automattic/wordpress-activitypub/pull/1635/commits/874190358356a5c8aed718325023673605f5eb7d).

I don't know if WP uses curly braces in cases where they're not needed, I'd have to check.

pfefferle (approved):

PR #1636, opened 2025-05-01
Editor: Show media in Preview for drafts

obenland:
While posts are in draft, any media included in the post will not show up in the Fediverse Preview. This PR changes that, so the post can be previewed with media, before publishing it.

Introduced in #780.

## Proposed changes:

* Removes post status check for Drafts when generating attachements list.
* Removes function exists check.

## Testing instructions:

* Add a media item of your choice.
* Select Fediverse Preview and make sure it's being displayed there.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Fediverse Previews in the Block Editor now show media items, even if the post has not been published yet.
</details>

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -297,7 +297,7 @@ class Post extends Base {
 		$id    = $this->item->ID;
 
 		// List post thumbnail first if this post has one.
-		if ( \function_exists( 'has_post_thumbnail' ) && \has_post_thumbnail( $id ) ) {
+		if ( \has_post_thumbnail( $id ) ) {
 			$media['image'][] = array( 'id' => \get_post_thumbnail_id( $id ) );
 		}
 
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -275,11 +275,6 @@ class Post extends Base {
 	 * @return array The Attachments.
 	 */
 	protected function get_attachment() {
-		// Remove attachments from drafts.
-		if ( 'draft' === \get_post_status( $this->item ) ) {
-			return array();
-		}
-
 		// phpcs:ignore Universal.Operators.DisallowShortTernary
 		$max_media = \get_post_meta( $this->item->ID, 'activitypub_max_image_attachments', true ) ?: \get_option( 'activitypub_max_image_attachments', ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS );
 
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -275,6 +275,14 @@ class Post extends Base {
 	 * @return array The Attachments.
 	 */
 	protected function get_attachment() {
+		/*
+		 * Remove attachments from the Fediverse if a post was federated and then set back to draft.
+		 * Except in preview mode, where we want to show attachments.
+		 */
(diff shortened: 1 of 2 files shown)


includes/transformer/class-post.php
```diff
@@ -302,7 +297,7 @@ protected function get_attachment() {
 		$id    = $this->item->ID;
 
 		// List post thumbnail first if this post has one.
-		if ( \function_exists( 'has_post_thumbnail' ) && \has_post_thumbnail( $id ) ) {
```
obenland, line 305: @pfefferle Do you remember why this check was necessary? Does this code run in an environment where that function might not exist?

pfefferle:
This `draft` check is to remove images from the fediverse if a post was published and will be set back to draft afterwards. We can not remove the check like this!

includes/transformer/class-post.php
pfefferle: I think we can remove this check

includes/transformer/class-post.php
```diff
@@ -275,11 +275,6 @@ protected function get_icon() {
 	 * @return array The Attachments.
 	 */
 	protected function get_attachment() {
-		// Remove attachments from drafts.
-		if ( 'draft' === \get_post_status( $this->item ) ) {
-			return array();
-		}
```
pfefferle, line 281: This `draft` check is to remove images from the fediverse if a post was first published, federated and then set back to draft mode. We can not remove the check like this, because we have to get sure to remove every media item from the now draft post!

pfefferle (changes requested):

includes/transformer/class-post.php
pfefferle: We could re-write the code to `Undo` the publish instead of sending an `Update` with a placeholder though.

pfefferle, line 281: This `draft` check is to remove images from the fediverse if a post was first published, federated and then set back to draft mode. We can not remove the check like this, because we have to get sure to remove every media item from the now draft post!

We could re-write the code to `Undo` the publish instead of sending an `Update` with a placeholder though.

Or this should also do the trick:

```PHP
if ( ! $this->is_preview() && 'draft' === \get_post_status( $this->item ) ) {
	return array();
}
```

includes/transformer/class-post.php
```diff
@@ -275,11 +275,6 @@ protected function get_icon() {
...
-
```
pfefferle, line 282: This `draft` check is to remove images from the fediverse if a post was first published, federated and then set back to draft mode. We can not remove the check like this, because we have to get sure to remove every media item from the now draft post!

We could re-write the code to `Undo` the publish instead of sending an `Update` with a placeholder though.

Or this should also do the trick:

```PHP
if ( ! $this->is_preview() && 'draft' === \get_post_status( $this->item ) ) {
	return array();
}
```

pfefferle (changes requested):

includes/transformer/class-post.php
obenland: That makes sense! Thank you!
Fixed in [71d45ee](https://github.com/Automattic/wordpress-activitypub/pull/1636/commits/71d45ee866e3a4c76ca33eaaf97fcec5034dc79c)

pfefferle (approved):

PR #1642, opened 2025-05-02
Embed: Check it's an activity object before using it

### LINKED ISSUE #1641: Undefined array key in class-embed.php
### Quick summary

```
Warning at shkspr.mobi/blog/wp-admin/admin-ajax.php?action=oembed-cache&post=32092:
[2] Undefined array key "id" in /wp-content/plugins/activitypub/includes/class-embed.php on line 108
```

https://github.com/Automattic/wordpress-activitypub/blob/7837721dc479d9a6d73e4f14b80ff16c30fd2134/includes/class-embed.php#L108

### Steps to reproduce

No idea. Just started seeing it in my logs.

### Site owner impact

### Severity

Minor

_No response_

I'm not sure why this function is being called. I have a normal oEmbed representation of my posts.

The post itself is fairly old, but was edited recently - `https://shkspr.mobi/blog/?p=32092`

obenland:
Fixes #1641.

I have not been able to reproduce the notice described in #1641. Adding a check to make sure we're dealing with an activity object before using it, should account for that scenario, though.

## Proposed changes:

* Checks whether the object returned is an activity object.
* Introduces `is_activity_object()`.
* Adds unit tests.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Added a check to make sure we only attempt to embed activity objects, when processing fallback embeds.

</details>

### DIFF
--- a/includes/class-embed.php
+++ b/includes/class-embed.php
@@ -32,7 +32,8 @@ class Embed {
 	public static function get_html( $url, $inline_css = true ) {
 		// Try to get ActivityPub representation.
 		$object = Http::get_remote_object( $url );
-		if ( is_wp_error( $object ) ) {
+
+		if ( is_wp_error( $object ) || ! is_activity_object( $object ) ) {
 			return false;
 		}
 
--- a/includes/class-embed.php
+++ b/includes/class-embed.php
@@ -33,7 +33,7 @@ class Embed {
 		// Try to get ActivityPub representation.
 		$object = Http::get_remote_object( $url );
 
-		if ( is_wp_error( $object ) || ! is_activity_object( $object ) ) {
+		if ( \is_wp_error( $object ) || ! is_activity_object( $object ) ) {
 			return false;
 		}
 
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1547,19 +1547,27 @@ function is_activity( $data ) {
 	 */
 	$types = apply_filters( 'activitypub_activity_types', Activity::TYPES );
 
-	if ( is_string( $data ) ) {
-		return in_array( $data, $types, true );
-	}
+	return _is_type_of( $data, $types );
+}
 
-	if ( is_array( $data ) && isset( $data['type'] ) ) {
-		return in_array( $data['type'], $types, true );
-	}
+/**
+ * Check if an `$data` is an Activity Object.
+ *
+ * @see https://www.w3.org/TR/activitystreams-vocabulary/#object-types
+ *
+ * @param array|object|string $data The data to check.
+ *
+ * @return boolean True if the `$data` is an Activity Object, false otherwise.
+ */
+function is_activity_object( $data ) {
+	/**
+	 * Filters the activity object types.
+	 *
+	 * @param array $types The activity object types.
+	 */
+	$types = apply_filters( 'activitypub_activity_object_types', Base_Object::TYPES );
 
-	if ( is_object( $data ) && $data instanceof Base_Object ) {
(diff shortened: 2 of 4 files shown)


includes/class-embed.php
```diff
@@ -32,7 +32,8 @@ public static function init() {
 	public static function get_html( $url, $inline_css = true ) {
 		// Try to get ActivityPub representation.
 		$object = Http::get_remote_object( $url );
-		if ( is_wp_error( $object ) ) {
+
+		if ( is_wp_error( $object ) || ! is_activity_object( $object ) ) {
```
pfefferle:
```suggestion
		if ( \is_wp_error( $object ) || ! is_activity_object( $object ) ) {
```

includes/functions.php
```diff
@@ -1579,6 +1587,17 @@ function is_actor( $data ) {
...
 
+	return _is_type_of( $data, $types );
+}
+
+/**
+ * Private helper to check if $data is of a given type set.
+ *
+ * @param array|object|string $data  The data to check.
+ * @param array               $types The types to check against.
+ * @return boolean True if $data is of one of the types, false otherwise.
```
pfefferle:
```suggestion
 * @param array               $types The types to check against.
 *
 * @return boolean True if $data is of one of the types, false otherwise.
```

includes/functions.php
```diff
@@ -1579,6 +1587,17 @@ function is_actor( $data ) {
...
+ */
+function _is_type_of( $data, $types ) {
```
pfefferle, line 1601: why prefixing?

includes/functions.php
```diff
@@ -1547,19 +1547,27 @@ function is_activity( $data ) {
...
+ *
+ * @return boolean True if the `$data` is an Activity Object, false otherwise.
+ */
+function is_activity_object( $data ) {
+	/**
+	 * Filters the activity object types.
+	 *
+	 * @param array $types The activity object types.
+	 */
+	$types = apply_filters( 'activitypub_activity_object_types', Base_Object::TYPES );
```
pfefferle:
```suggestion
	$types = \apply_filters( 'activitypub_activity_object_types', Base_Object::TYPES );
```

includes/functions.php
obenland: It's WordPress' (imperfect) way to denote a private function that's not part of an API and shouldn't be used outside of Core ([example](https://github.com/WordPress/wordpress-develop/blob/6.8.1/src/wp-includes/cron.php#L1010-L1015)).

pfefferle: sure but why does it have to be private? we even use private WordPress functions ;)

I would not go that path.

obenland: It's just a helper function. If this was part of a class I'd make it a private method. In the end nothing changes, prefix or not, people can still use it. But at least they know they shouldn't.

pfefferle: but why shouldn't they?

pfefferle (approved):

includes/functions.php
obenland: Because it's a helper function and not part of an API. I think there's a general expectations that functions in this file are fine to be used in third-party plugins. `_is_type_of()` is not part of an API, it's just a helper for the three API functions that use them.

obenland: Ha! I said the same thing twice 😄

PR #1643, opened 2025-05-02
Only show "Reply to the post" when using block-editor

pfefferle:
When user is not using the block editor the link would not work.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Hide interaction buttons in emails when the Classic Editor is used.

</details>

### DIFF
--- a/templates/emails/new-mention.php
+++ b/templates/emails/new-mention.php
@@ -62,7 +62,7 @@ require __DIR__ . '/parts/header.php';
  *
  * @param array $args The template arguments.
  */
-\do_action( 'activitypub_new_mention_email', $args );
+do_action( 'activitypub_new_mention_email', $args );
 
 // Load footer.
 require __DIR__ . '/parts/footer.php';
--- a/templates/emails/new-mention.php
+++ b/templates/emails/new-mention.php
@@ -11,7 +11,7 @@ use Activitypub\Embed;
 use function Activitypub\site_supports_blocks;
 
 /* @var array $args Template arguments. */
-$args = \wp_parse_args( $args ?? array() );
+$args = wp_parse_args( $args ?? array() );
 
 // Load header.
 require __DIR__ . '/parts/header.php';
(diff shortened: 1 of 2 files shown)


templates/emails/new-mention.php
```diff
@@ -8,8 +8,10 @@
 use Activitypub\Collection\Actors;
 use Activitypub\Embed;
 
+use function Activitypub\site_supports_blocks;
+
 /* @var array $args Template arguments. */
-$args = wp_parse_args( $args ?? array() );
+$args = \wp_parse_args( $args ?? array() );
```
obenland: Templates don't use namespaces so these shouldn't be necessary.

templates/emails/new-mention.php
```diff
@@ -46,19 +48,21 @@
...
 
+<?php if ( site_supports_blocks() && ! \is_plugin_active( 'classic-editor/classic-editor.php' ) ) : ?>
 <p>
-	<a class="button" href="<?php echo esc_url( admin_url( 'post-new.php?in_reply_to=' . $args['activity']['object']['id'] ) ); ?>">
-		<?php esc_html_e( 'Reply to the post', 'activitypub' ); ?>
+	<a class="button" href="<?php echo \esc_url( \admin_url( 'post-new.php?in_reply_to=' . $args['activity']['object']['id'] ) ); ?>">
+		<?php \esc_html_e( 'Reply to the post', 'activitypub' ); ?>
 	</a>
 </p>
+<?php endif; ?>
```
obenland, line 57: What's an alternative action that we could suggest here? View the post?

pfefferle: I think simply removing is fine... The classic editor is the exception, so we should simply care about not producing any errors.

obenland (approved):

PR #1646, opened 2025-05-02
Welcome: Try new UI

obenland:
## Proposed changes:
localhost_8888_wp-admin_options-general php_page=activitypub (4)

### DIFF
--- a/assets/css/activitypub-admin.css
+++ b/assets/css/activitypub-admin.css
@@ -222,47 +222,6 @@ input.blog-user-identifier {
 	margin-block: 0;
 }
 
-.activitypub-settings .welcome-tab-close {
-	position: absolute;
-	top: 0px;
-	right: 0px;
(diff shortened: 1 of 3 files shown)


pfefferle:
Isn't that exactly what I built and closed?

obenland:
Oh yeah, it's pretty similar! I guess that's just where AI takes us when trying to combine Dave's wireframe and what we already have. I'm pretty happy with this one

pfefferle:
Early Feedback:

* I think we can remove the enable/disable welcome page checkbox in "Screen Options". There is no need to see this page again after going through the process.
* What I liked about the current welcome page is, that it provided a place to show important informations and that it could be removed. This way it was possible to keep the settings free from (maybe) distracting informations. The welcome page also provides the option to re-enable it for every new plugin version we release, to highlight new features (like with the release pages in WordPress core).

PR #1648, opened 2025-05-05
Add Cache Config to enable ActivityPub support for Surge

pfefferle:
This adds the needed config file, so that Surge will work nicely with ActivityPub. Thanks to @kovshenin and @ocean90 for the fix:

https://dominikschilling.de/notes/http-accept-header-wordpress-cache-activitypub/

## Proposed changes:

* Add guidance to the Site-Health check, how to properly configure surge, to play nicely with ActivityPub.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Guidance for configuring Surge to support ActivityPub caching.

</details>


includes/wp-admin/class-health-check.php
```diff
@@ -448,4 +455,46 @@ public static function test_pretty_permalinks() {
...
+			$result['badge']['color'] = 'red';
+			$result['description']    = \sprintf(
+				'<p>%s</p>',
+				\__( 'Surge isn’t currently set up to work with ActivityPub. While this isn’t a major problem, it’s a good idea to enable support. Without it, some technical files (like JSON) might accidentally show up in your website’s cache and be visible to visitors.', 'activitypub' )
+			);
+			$result['actions']        = \sprintf(
+				'<p>%s</p>',
+				sprintf(
+					// translators: %s: Plugin directory path.
+					\__( 'To enable the ActivityPub integration with Surge, add the following line to your <code>wp-config.php</code> <file></file>: <br /><code>define( \'WP_CACHE_CONFIG\', %sintegration/surge.php );</code>', 'activitypub' ),
```
obenland:
```suggestion
					\__( 'To enable the ActivityPub integration with Surge, add the following line to your <code>wp-config.php</code> file: <br /><code>define( \'WP_CACHE_CONFIG\', %sintegration/surge.php );</code>', 'activitypub' ),
```

Let's [attempt to do it for them](https://github.com/kovshenin/surge/blob/b9bbb79421a49761cf42f6dcb81cd1b5fad3c8db/include/install.php#L36-L63) before giving them a task?

pfefferle: I tried. This is needed so early in the process, that it hast to be added to the `wp-config.php`. Every place in the ActivityPub plugin was too late.

obenland: It doesn't have to be there on the first page load. We could attempt to write it to the file and keep track of the attempt with an option. If the option is set and the constant doesn't exist, we'd then show the check?

pfefferle: Is this edge case worth the work?

obenland: Which edge case?

I'm proposing to attempt to do write that constant definition to wp-config, so users don't have to do it manually. The Surge plugin does the same with the WP_CACHE constant, so I assume the chances are high we could do the same with WP_CACHE_CONFIG.

pfefferle: Edge Case: Not a lot people might use both plugins

pfefferle: But maybe a nice POC that could be reused for other integrations... I will give it a try...

integration/class-surge.php
```diff
@@ -0,0 +1,108 @@
...
+ * @see https://wordpress.org/plugins/surge/
+ */
+class Surge {
+	/**
+	 * The file to add the Surge cache config to.
+	 *
+	 * @var string
+	 */
+	private static $config_file = ABSPATH . 'wp-config.php';
+	/**
```
obenland:
```suggestion

	/**
```

integration/class-surge.php
```diff
@@ -0,0 +1,108 @@
...
+	 *
+	 * @var string
+	 */
+	private static $cache_config_pattern = '/define\s*\(\s*[\'"](WP_CACHE_CONFIG)[\'"]\s*,\s*[\'"](.*?)[\'"]\s*\)\s*;/i';
+
+	/**
+	 * Initialize the Surge integration.
+	 */
+	public static function init() {
+		if ( \is_plugin_active( 'surge/surge.php' ) && ! \defined( 'WP_CACHE_CONFIG' ) ) {
```
obenland: This would run on every request to wp-admin, even if we found the file not writeable. Surge itself uses an option to keep track of their attempt to update wp-config—it would probably be good to do something similar.

integration/class-surge.php
```diff
@@ -0,0 +1,108 @@
...
+		if ( ! \is_plugin_active( 'surge/surge.php' ) && \defined( 'WP_CACHE_CONFIG' ) ) {
+			self::remove_cache_config();
+		}
+	}
+
+	/**
+	 * Add the Surge cache config.
+	 */
+	public static function add_cache_config() {
+		$file = self::$config_file;
```
obenland: wp-config can be placed outside of the WP installation folder, too. [Core has a check for it](https://github.com/WordPress/wordpress-develop/blob/6.8.1/src/wp-load.php#L47-L57), maybe we can adopt that here, too.

obenland (commented):
Nice, that looks great!

I think it's fine to keep the Site Health stuff for cases where we were unable to set the constant ourselves. Just thought we should try ourselves first.

integration/class-surge.php
pfefferle: do you mean `update_option( 'surge_installed', $bytes ? 1 : 2 );`?

obenland: Exactly. Not sure we need to replicate `$bytes ? 1 : 2`, but just having an indicator that this ran.

pfefferle: but this checks a bit more then only the config? is it really more effective to check an option instead of if the plugin is installed and the const is set?

I am currently working on re-activating the site-health integration if the config file is not writable.

pfefferle:
Thanks for the feedback @obenland... this version is so much cleaner now!

props!!

integration/load.php
```diff
@@ -162,3 +162,12 @@ function ( $post_types ) {
...
 add_action( 'bp_include', array( __NAMESPACE__ . '\Buddypress', 'init' ), 0 );
+
+/**
+ * Load the Surge integration.
+ *
+ * Only load code that needs Surge to run once Surge is loaded and initialized.
+ *
+ * @see https://wordpress.org/plugins/surge/
+ */
+add_action( 'init', array( __NAMESPACE__ . '\Surge', 'init' ) );
```
obenland: Adding this to `init` runs it after the "regular" integration inits, which run on `plugins_loaded`. I think it's fine to move this into `plugin_init()` and just call it directly `Surge::init()`.

That said, to account for situations where Surge is already active, I think we should add activation/deactivation callbacks for activitypub here:
```
\add_action( 'activate_activitypub/activitypub.php', array( __NAMESPACE__ . '\Surge', 'add_cache_config' ) );
\add_action( 'deactivate_activitypub/activitypub.php', array( __NAMESPACE__ . '\Surge', 'remove_cache_config' ) );
```

tests/integration/class-test-surge.php
```diff
@@ -0,0 +1,219 @@
...
+ * Test Surge integration.
+ */
+class Test_Surge extends \WP_UnitTestCase {
+	/**
+	 * Test file path.
+	 *
+	 * @var string
+	 */
+	private $test_file;
+	/**
```
obenland, line 23:
```suggestion

	/**
```

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+			return;
+		}
+
+		global $wp_filesystem;
+		\WP_Filesystem();
+
+		$config = $wp_filesystem->get_contents( $file );
+
+		// Remove the define line.
+		$config = preg_replace( self::$cache_config_pattern, '', $config );
```
obenland: Nitpick: We're adding new lines in `add_cache_config()` that we don't seem to be removing here.

obenland (commented):
This is turning out really well. Nice work @pfefferle!

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+			'description' => \sprintf(
+				'<p>%s</p>',
+				\__( 'Surge is well configured to work with ActivityPub.', 'activitypub' )
+			),
+			'actions'     => '',
+			'test'        => 'test_surge_integration',
+		);
+
+		if ( ! \defined( 'WP_CACHE_CONFIG' ) ) {
+			$result['status']         = 'critical';
```
obenland, line 168: With the checklist coming together now, should this be an issue that blocks completion of the Site Health step?

integration/load.php
pfefferle: can a plugin listen to it's own activation/deactivation? I think we would have to use the `register_activation_hook` instead?

this might get complicated and I think we should deal with that in a follow up!?!

it still will be caught with the site-health check.

integration/class-surge.php
pfefferle: It can mess up a lot, but I can downgrade it to a recommendation.

integration/class-surge.php
pfefferle: I feared that if someone changes the config by hand afterwards and the whitespaces/breaks are no longer the same, that it will no longer be properly removed!?

integration/load.php
obenland: This is all `register_activation_hook` does, adding that action. It just makes it easier by not requiring plugin authors to know their plugin's hook. But it's a good point, we could probably just use that here:

```
\register_activation_hook( ACTIVITYPUB_PLUGIN_FILE, array( __NAMESPACE__ . '\Surge', 'add_cache_config' ) );
\register_deactivation_hook( ACTIVITYPUB_PLUGIN_FILE, array( __NAMESPACE__ . '\Surge', 'remove_cache_config' ) );
```

> this might get complicated and I think we should deal with that in a follow up!?!

I tested it (with the previous hooks), it should be as easy as adding those two lines.

integration/class-surge.php
obenland: That's a good point. Let's add _one_ line break, so the line this is on gets removed?

integration/load.php
pfefferle: where have you added these lines? in `load.php`?

obenland: Yeah, right below this init callback.

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
+<?php
+/**
+ * Surge integration file.
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\Integration;
+
+use function Activitypub\user_can_activitypub;
```
obenland:
```suggestion
```
This looks to be unused now

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * @see https://wordpress.org/plugins/surge/
+ */
+class Surge {
+	/**
+	 * The define for the Surge cache config.
+	 *
+	 * @var string
+	 */
+	public static $cache_config = 'define( \'WP_CACHE_CONFIG\', \'' . ACTIVITYPUB_PLUGIN_DIR . 'integration/surge-cache-config.php\' );';
```
obenland:
```suggestion
	public static $cache_config = "define( 'WP_CACHE_CONFIG', '' . ACTIVITYPUB_PLUGIN_DIR . 'integration/surge-cache-config.php' );";
```
Let's use double-quotes to make this easier to read. phpcs is fine with double quotes if they're used to contain single quotes.

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+		if ( ! \function_exists( 'WP_Filesystem' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/file.php';
+		}
+
+		global $wp_filesystem;
+		\WP_Filesystem();
+
+		$config = $wp_filesystem->get_contents( $file );
+
+		// Check if the define already exists.
```
obenland:
```suggestion
		// Check if the constant already exists.
```
My IDE complained about grammar on this one? "constant" fixes it.

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+
+	/**
+	 * Get the config file.
+	 *
+	 * @return string|false The config file or false.
+	 */
+	public static function get_config_file_path() {
+		$config_file = false;
+
+		// phpcs:ignore
```
obenland:
```suggestion
		// phpcs:ignore WordPress.PHP.NoSilencedErrors
```

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+		if ( @file_exists( ABSPATH . 'wp-config.php' ) ) {
+
+			/** The config file resides in ABSPATH */
+			$config_file = ABSPATH . 'wp-config.php';
+		// phpcs:ignore
```
obenland:
```suggestion
		// phpcs:ignore WordPress.PHP.NoSilencedErrors
```

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+		} elseif ( @file_exists( dirname( ABSPATH ) . '/wp-config.php' ) && ! @file_exists( dirname( ABSPATH ) . '/wp-settings.php' ) ) {
+
+			/** The config file resides one level above ABSPATH but is not part of another installation */
+			$config_file = dirname( ABSPATH ) . '/wp-config.php';
+		}
+
+		return \apply_filters( 'activitypub_surge_cache_config_file', $config_file );
```
obenland, line 123:
```suggestion
		/**
		 * Filter the config file path.
		 *
		 * @param string|false $config_file The config file path.
		 */
		return \apply_filters( 'activitypub_surge_cache_config_file', $config_file );
```

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+			$result['status']         = 'critical';
+			$result['label']          = \__( 'Surge might not be properly configured.', 'activitypub' );
+			$result['badge']['color'] = 'red';
+			$result['description']    = \sprintf(
+				'<p>%s</p>',
+				\__( 'Surge isn&#8217;t currently set up to work with ActivityPub. While this isn&#8217;t a major problem, it&#8217;s a good idea to enable support. Without it, some technical files (like JSON) might accidentally show up in your website&#8217;s cache and be visible to visitors.', 'activitypub' )
+			);
+			$result['actions']        = \sprintf(
+				'<p>%s</p>',
+				sprintf(
```
obenland:
```suggestion
				\sprintf(
```

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+			return;
+		}
+
+		global $wp_filesystem;
+		\WP_Filesystem();
+
+		$config = $wp_filesystem->get_contents( $file );
+
+		// Remove the define line.
+		$config = preg_replace( PHP_EOL . self::$cache_config_pattern . PHP_EOL, '', $config );
```
obenland:
```suggestion
		$config = \preg_replace( PHP_EOL . self::$cache_config_pattern . PHP_EOL, '', $config );
```

integration/class-surge.php
```diff
@@ -0,0 +1,184 @@
...
+			require_once ABSPATH . 'wp-admin/includes/file.php';
+		}
+
+		global $wp_filesystem;
+		\WP_Filesystem();
+
+		$config = $wp_filesystem->get_contents( $file );
+
+		// Check if the define already exists.
+		if ( preg_match( self::$cache_config_pattern, $config ) ) {
```
obenland:
```suggestion
		if ( \preg_match( self::$cache_config_pattern, $config ) ) {
```

integration/surge-cache-config.php
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @see https://dominikschilling.de/notes/http-accept-header-wordpress-cache-activitypub/
+ *
+ * @package Activitypub
+ */
+
+$representation = 'html';
+
+if ( isset( $_SERVER['HTTP_ACCEPT'] ) ) {
+	// phpcs:ignore
```
obenland:
```suggestion
	// phpcs:ignore WordPress.Security.ValidatedSanitizedInput
```

integration/surge-cache-config.php
```diff
@@ -0,0 +1,32 @@
...
+	} elseif (
+		str_contains( $accept, 'application/json' ) ||
+		str_contains( $accept, 'application/activity+json' ) ||
+		str_contains( $accept, 'application/ld+json' )
+	) {
+		$representation = 'json';
+	}
+}
+
+// phpcs:ignore
```
obenland:
```suggestion
// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
```

integration/surge-cache-config.php
```diff
@@ -0,0 +1,32 @@
...
+$config['variants']['representation'] = $representation;
+unset( $accept, $representation );
+
+// phpcs:ignore
```
obenland:
```suggestion
// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
```

integration/load.php
```diff
@@ -162,3 +162,15 @@ function ( $post_types ) {
...
 add_action( 'bp_include', array( __NAMESPACE__ . '\Buddypress', 'init' ), 0 );
+
+/**
+ * Load the Surge integration.
+ *
+ * Only load code that needs Surge to run once Surge is loaded and initialized.
+ *
+ * @see https://wordpress.org/plugins/surge/
+ */
+Surge::init();
```
obenland: Sorry, I meant this could be called with the other integrations in `plugin_init()`.

integration/class-surge.php
pfefferle: that was phpcbf I think or phpcs! I would love to!

integration/load.php
obenland: @pfefferle Only the lines marked in this ^ comment. The plugin activation calls need to stay outside of `plugin_init()` so they get caught when ActivityPub first gets activated.

obenland (approved):
Nice work!

PR #1651, opened 2025-05-05
Checklist: Keep better track of step counts

obenland:
## Proposed changes:

* Adds profile link check option.
* Updates value checks to what these options will actually return.


pfefferle:
Is this more effective than a DB select?

obenland:
I think so. They should be autoloaded and cached.

pfefferle (approved):

PR #1655, opened 2025-05-05
Checklist: Mark help tab visits as done

obenland:
This is not quite ideal. On clicking the button the page reloads and it flashes the updated checklist before the help tab gets toggled.

I think I'd like to try the ajax route before committing to this.


obenland:
Updated to use ajax for these steps. I think it'll need a bit more iteration, but it's getting closer.

pfefferle (approved):
nice :)

PR #1657, opened 2025-05-06
 Implements Heading Customization for the Reactions Block

### LINKED ISSUE #1096: Reactions Block: add option to customize and / or remove the heading
### What

It would be nice if we could customize and / or remove the heading of the Reactions block.

On my site, I would like to be able to:

1. Use the heading level I want
2. Change the text
3. Customize the display of the heading (top border, style, ...)
4. Remove it entirely if necessary

Image

### Why

The more customizable the block is, the easier it will be for folks to integrate it on their sites.

### How

Would it be possible to auto-insert a core group block instead, and that group block would have a heading block and a Reactions block? This way folks would then be able to customize it to their liking.

vk17-starlord:
This PR introduces enhancements to the Reactions block, enabling users to customize the heading using Gutenberg's InnerBlocks. The updates provide flexibility and control over the heading display.

#### Key Features:
- **Editable Heading Text:** Users can modify the text of the heading.
- **Changeable Heading Levels:** Users can select heading levels (H1-H6) to suit their requirements.
- **Removable Heading:** Provides the option to remove the heading entirely if desired.
- **Standard Gutenberg Styles:** Integrates with standard Gutenberg heading styles for consistent design.

This implementation is inspired by Jetpack's Related Posts block and aims to provide a similar level of customization.

#### Fixes:
- Resolves [#1096](https://github.com/Automattic/wordpress-activitypub/issues/1096).

### Proposed Changes:
- Adds support for heading customization in the Reactions block.
- Uses Gutenberg's InnerBlocks for flexibility and ease of customization.
- Ensures compatibility with existing Gutenberg styling standards.

#### Significance:
- Minor

#### Type:
- Added: New customization options for the Reactions block heading.


src/reactions/reactions.js
```diff
@@ -284,72 +284,66 @@ const ReactionGroup = ( { items, label } ) => {
...
-						key={ key }
-						items={ group.items }
-						label={ group.label }
-					/>
-				);
-			} ) }
-		</div>
-	);
+export function Reactions({
+    postId = null,
```
pfefferle: please use tabs instead of spaces.

pfefferle:
Thanks for your contribution @vk17-starlord 

Can you please use `npm run build` to run the build? It seems that your build script changed a lot of the untouched files, simply because of a different configuration.

vk17-starlord:
ok @pfefferle will implement those changes

vk17-starlord:
@pfefferle - when i run 'NPM RUN BUILD' it still generated build files for all blocks and other unnecessary files which I haven't touched

vk17-starlord:
@pfefferle - Have converted spaces to tabs using vscode shortcut please check formatting and logic once

pfefferle:
@vk17-starlord I just ran the npm script on your branch and it removed all the map files and minified the JS/CSS as it should. Have you stopped the `npm run dev` task before?

vk17-starlord:
> @vk17-starlord I just ran the npm script on your branch and it removed all the map files and minified the JS/CSS as it should. Have you stopped the `npm run dev` task before?

Yes I had stopped NPM RUN DEV before running NPM RUN BUILD . The NPM RUN BUILD script is building all blocks which haven't been changed

vk17-starlord:
I created a new branch and copied the reaction block code. Initially, VSCode showed only 3 file changes in the top-left source control panel:

image

However, after running `npm run build`, the number of changed files increased to 52:

image

obenland:
>  The NPM RUN BUILD script is building all blocks which haven't been changed

That should be fine, it should just update the hashes. Please go ahead and just run `npm run build` and push the results. Thanks!

vk17-starlord:
@obenland - have already pushed the code you can check it out and review

src/reactions/reactions.js
```diff
@@ -284,72 +284,66 @@ const ReactionGroup = ( { items, label } ) => {
...
+		setLoading(true);
+		apiFetch({
+			path: `/${namespace}/posts/${postId}/reactions`,
+		})
+			.then((response) => {
+				setReactions(response);
+				setLoading(false);
+			})
+			.catch(() => setLoading(false));
+	}, [postId, providedReactions, namespace]);
```
obenland: I'm not sure we need to add `namespace` here? It should be constant.

obenland (commented):
This is pretty close. There are quite a few whitespace changes currently, would you mind reverting these to stay closer to [WP coding standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#spacing)?

src/reactions/block.json
```diff
@@ -18,12 +18,7 @@
 			}
 		}
 	},
-	"attributes": {
```
jeherve, line 21: This seems problematic as it will create validation errors for existing users of the block, whenever they try to visualize the block in the site editor.
I would recommend implementing [deprecation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/) to avoid that.

src/reactions/reactions.js
```diff
@@ -284,72 +284,66 @@ const ReactionGroup = ( { items, label } ) => {
...
 				}
 
 				return (
 					<ReactionGroup
-						key={ key }
-						items={ group.items }
-						label={ group.label }
+						key={key}
+						items={group.items}
+						label={group.label}
```
jeherve: Your editor doesn't seem to pick up the configuration from this project's `.editorconfig` file. Could you keep the spaces inside brackets and parenthesis in the file?

jeherve (commented):
Thanks for your contribution! I left some comments below.

vk17-starlord:
@pfefferle @obenland – I've added deprecation handling for the Reaction block and resolved the related linting issues. When you have a moment, I'd appreciate your review and any feedback on the deprecation implementation.

Apologies for any formatting or minor issues you may notice — I'm still getting up to speed with contributing in the WordPress ecosystem, but I'm excited to learn and continue contributing!

obenland:
@jeherve When you get a chance, could you give this one a spin? Based on your comment I assume you have a few existing Reactions blocks you don't want to break?

jeherve (commented):
Thanks for adding the deprecation handling. Unfortunately the migration doesn't seem to happen for me. If I set a custom title, save my changes then switch to your branch and refresh the site editor, that custom title is not set as the new heading contents in the block ; I only see the default title instead.

obenland:
Hm, that's annoying. It works locally for me! Could you share the block markup so I could try to reproduce?
What about in the frontend, for Reactions blocks that have not been transformed yet? Do they still display properly?

https://github.com/user-attachments/assets/67bd4e37-fe44-4a9f-b6e6-69aec5a35615

jeherve (approved):
Hm, it must have been a caching issue. It works well now! This is looking good, should be good to merge I think.

obenland:
Nice work @vk17-starlord! Thank you for helping with this improvement!

PR #1660, opened 2025-05-06
Extra Fields: Don't send more than 20

obenland:
Limits the amount of extra fields that get returned to 100, when outside of wp-admin to prevent response size errors in clients.

See https://mastodon.social/@manualdousuario/114461466878024369

## Proposed changes:

* Limits the amount of extra fields that get returned to 100, when outside of wp-admin.

## Testing instructions:

* Make sure you have more than one Extra Field in `/wp-admin/edit.php?post_type=ap_extrafield`.
* Manually change the limit of extra fields in this PR to 1.
* Call your author profile link and make sure it only shows one extra field, equalling two attachments.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Author profiles will cap the amount of extra fields they return to 100, to avoid response size errors in clients.

</details>

### DIFF
--- a/includes/collection/class-extra-fields.php
+++ b/includes/collection/class-extra-fields.php
@@ -39,7 +39,7 @@ class Extra_Fields {
 			$args['author'] = $user_id;
 		}
 
-		// Limit to 100 fields to prevent response size issues.
+		// Limit to 20 fields to prevent response size issues.
 		if ( ! is_admin() ) {
 			/**
 			 * Filters the number of extra fields to retrieve for an ActivityPub actor.
--- a/includes/collection/class-extra-fields.php
+++ b/includes/collection/class-extra-fields.php
@@ -39,6 +39,12 @@ class Extra_Fields {
 			$args['author'] = $user_id;
 		}
 
+		// Limit to 100 fields to prevent response size issues.
+		if ( ! is_admin() ) {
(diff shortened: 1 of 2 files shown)


includes/collection/class-extra-fields.php
```diff
@@ -39,6 +39,12 @@ public static function get_actor_fields( $user_id ) {
 			$args['author'] = $user_id;
 		}
 
+		// Limit to 100 fields to prevent response size issues.
+		if ( ! is_admin() ) {
+			$args['nopaging']       = false;
+			$args['posts_per_page'] = 100;
```
pfefferle:
```suggestion
			$args['posts_per_page'] = 20;
```

I would go even lower!

obenland: Fixed in [8724146](https://github.com/Automattic/wordpress-activitypub/pull/1660/commits/8724146796ce01630453f72987bd8d81dc3de6e6).

pfefferle (approved):

obenland:
@pfefferle Should we revert this? [It didn't end up being an issue](https://mastodon.social/@obenland/114461944860384252).

pfefferle:
I would keep it. It is definitely an improvement.

PR #1670, opened 2025-05-07
Checklist: Fix Help-Tabs

pfefferle:
This is a follow up of #1666

## Proposed changes:

* Help tab query string for "Getting Started"
* Load the update_options call, to save the new state, on every ActivityPub WP-Admin page, because "Getting Started" is triggered on the welcome page

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


obenland (approved):

includes/wp-admin/class-settings.php
```diff
@@ -358,7 +358,7 @@ public static function settings_page() {
 				\wp_enqueue_media();
 				\wp_enqueue_script( 'activitypub-header-image' );
 				break;
-			case 'settings':
+			default:
```
obenland, line 364: If we run it on default, we should probably run it much earlier. Can you help me understand why you don't want to scope it to the settings tab?

obenland (changes requested):

includes/wp-admin/class-settings.php
pfefferle: because we call "Getting Started" on the welcome page and if the options will only be set on settings, "Getting Started" will never get marked as resolved.

obenland: Ah! Part of #1666 was the proposal to have help links link to the settings tab, resulting in a new pageload and avoiding the flash of an updated checklist. So they should always end up here

pfefferle: but it does not work like that

obenland: It looks like [the change that added the tab parameter to the getting-started link](https://github.com/Automattic/wordpress-activitypub/pull/1666/commits/f134a9fd3a6a7df9dab1b6013105a47311aed6ab) was [reverted somehow in a merge commit](https://github.com/Automattic/wordpress-activitypub/pull/1666/commits/c85fb7d5205af67f193aaf3be33df9ca08ab4308) :/

includes/wp-admin/class-welcome-fields.php
```diff
@@ -262,7 +262,7 @@ public static function render_step_fediverse_intro() {
 					<p><?php \esc_html_e( 'Learn what the Fediverse is and why it matters.', 'activitypub' ); ?></p>
 				</div>
 				<div class="step-action">
-					<a href="<?php echo \esc_url( \admin_url( 'options-general.php?page=activitypub#tab-link-getting-started' ) ); ?>" class="button <?php echo \esc_attr( $button_class ); ?>">
+					<a href="<?php echo \esc_url( \admin_url( 'options-general.php?page=activitypub&help-tab=getting-started#tab-link-getting-started' ) ); ?>" class="button <?php echo \esc_attr( $button_class ); ?>">
```
obenland, line 275:
```suggestion
					<a href="<?php echo \esc_url( \admin_url( 'options-general.php?page=activitypub&tab=settings&help-tab=getting-started#tab-link-getting-started' ) ); ?>" class="button <?php echo \esc_attr( $button_class ); ?>">
```

includes/wp-admin/class-settings.php
pfefferle: a this was because of a conflict! then feel free to open another one and close this one.

that said, why do we want to redirect the users to the settings only to view the getting started? isn't it a smart thing to keep them on the welcome page to easier go to the next step?

includes/wp-admin/class-welcome-fields.php
pfefferle: as said in the comment above, I think we should do the change I propose. Why should we want to move the user away from the welcome page only to show the Getting Started video?

includes/wp-admin/class-settings.php
obenland: If we keep them on the Welcome page we need some sort of JS to update the option and make all the UI changes that come with it, otherwise it will load the updated page, with the step marked as complete, before the help tab slides down.
By linking to the settings tab, we avoid the flash of updated checklist, but still get a page refresh

pfefferle: with my change not... you simply redirect to the welcome page with the help-tab param... so technically it is a redirect, but to the same page..

obenland: Interesting. So it saves the option after rendering the checklist?
It's still a bit funky, with it being not checked off after opening the help tab.

https://github.com/user-attachments/assets/051f185d-c2c3-43c3-913d-06146bfcb417

pfefferle: I think it is fine to redirect the others to the settings pages, because these are about settings... and we might still need to improve things... but what turned out to be a messed up merge, seems like a feature to me ;)

obenland: The checklist item not updating after clicking the button feels pretty unpolished, no?

pfefferle: true... but can we make it happen? 👀

obenland (approved):

PR #1673, opened 2025-05-07
JS: Update format to follow WP standards

obenland:
Taking inspiration from #1657, let's update our js an json files to correspond to wp coding standards.

## Proposed changes:

* Adds npm commands to format and lint js/css.
* Format updates in JS, json, yml, md files.
* Adds formatting step to build script.


tests/fixtures/example-com-well-known-webfinger.json
```diff
@@ -1 +1,18 @@
...rset=utf-8\" \/>\n    <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\" \/>\n    <style type=\"text\/css\">\n    body {\n        background-color: #f0f0f2;\n        margin: 0;\n        padding: 0;\n        font-family: -apple-system, system-ui, BlinkMacSystemFont, \"Segoe UI\", \"Open Sans\", \"Helvetica Neue\", Helvetica, Arial, sans-serif;\n        \n    }\n    div {\n        width: 600px;\n        margin: 5em auto;\n        padding: 2em;\n        background-color: #fdfdff;\n        border-radius: 0.5em;\n        box-shadow: 2px 3px 7px 2px rgba(0,0,0,0.02);\n    }\n    a:link, a:visited {\n        color: #38488f;\n        text-decoration: none;\n    }\n    @media (max-width: 700px) {\n        div {\n            margin: 0 auto;\n            width: auto;\n        }\n    }\n    <\/style>    \n<\/head>\n\n<body>\n<div>\n    <h1>Example Domain<\/h1>\n    <p>This domain is for use in illustrative examples in documents. You may use this\n    domain in literature without prior coordination or asking for permission.<\/p>\n    <p><a href=\"https:\/\/www.iana.org\/domains\/example\">More information...<\/a><\/p>\n<\/div>\n<\/body>\n<\/html>\n","response":{"code":404}}
```
pfefferle: Can we exclude the test folder? there is no need to have the fixtures formatted properly. This only takes more space and we should keep the original file from the endpoints we test against to test as near as possible to the live environment.

assets/js/activitypub-header-image.js
```diff
@@ -23,34 +23,36 @@
...
 				ratio = cropDetails.width / cropDetails.height;
 
 			// Use crop measurements when flexible in both directions.
 			if ( control.params.flex_width && control.params.flex_height ) {
-				cropDetails.dst_width  = cropDetails.width;
+				cropDetails.dst_width = cropDetails.width;
 				cropDetails.dst_height = cropDetails.height;
 
-			// Constrain flexible side based on image ratio and size of the fixed side.
+				// Constrain flexible side based on image ratio and size of the fixed side.
```
pfefferle: this makes no sense

.wordpress-org/blueprints/blueprint.json
```diff
@@ -1,43 +1,43 @@
...
-    {
-      "step": "setSiteOptions",
-      "options": {
-        "blogname": "ActivityPub Demo",
-        "blogdescription": "Democratize Publishing in the Fediverse",
-        "activitypub_enable_blog_user": "1"
-      }
-    }
-  ]
+    "landingPage": "/wp-admin/options-general.php?page=activitypub",
```
pfefferle, line 2: spaces in JSON and tabs in JS? 😳😱

pfefferle:
Can we make this as a step by step thing?

Risks that I see with such a big re-formatting PR:

* We would have to test every JS widget properly even if it is "only" re-formatting
* It breaks the history for all JS files and makes it hard to track changes before the reformatting
* Some of the auto formatting does not make sense, so we should also update for example inline comments (see my comments) to match the new style
* If we get issues/bugs because of this task, it is hard to track them down, because of the big change (this might be a more general risk of big PRs)

Can we do this incremental? Every time we have to change something in a file, we also re-format it before?

assets/js/activitypub-header-image.js
jeherve: I think the tool doesn't expect a comment right before a closing brace in general.

pfefferle: ...but this means we should go through the code and tweak these things, I think!?!

.wordpress-org/blueprints/blueprint.json
obenland: I'm worried that tabs in json might be unexpected by third-party tools/parsers. Spaces seemed "safer" to me. Would you prefer tabs? I'm not feeling strongly. about it.

obenland:
> Can we do this incremental? Every time we have to change something in a file, we also re-format it before?

Sure! I'd love for us to run `wp-scripts format` as part of `npm run build`, so we never have to worry about formatting again. 

To make that work, I'd propose doing it similar to how we updated phpcs stuff: Starting with a long `.prettierignore` file and filing separate PRs for each line. Would that work for you?

.wordpress-org/blueprints/blueprint.json
pfefferle: It's mostly because the monk in me cried: "consistency"... if this is the coding standard so I have to take it... but I couldn't do it without comment 😂

pfefferle:
I think I can live with both (in this specific case)... merging that as is (for now) or having separate PRs (maybe not for each line, but for each frontent project/area)...

What's important to me is, that we have a strategy to work with things like that in general... I hope this will not be the last restyling and it is important to do it, but step by step...

pfefferle:
But we would have to go through the code and check all changes.

.wordpress-org/blueprints/blueprint.json
obenland: I understand 🙂 Maybe consistency by filetype? YAML files would break with tabs

obenland:
Happy to break it up to make it easier to review, for sure

tests/fixtures/example-com-well-known-webfinger.json
obenland: Fixed in [60f540a](https://github.com/Automattic/wordpress-activitypub/pull/1673/commits/60f540a57de17ec73e2e1b589bc25d843944961f).

obenland:
This PR now only covers changes to internal files and the new prettier configs.

All subsequent PRs should be as easy as
1. Remove a line from `.prettierignore`.
2. Run `npm run build`.

pfefferle (approved):

pfefferle:
One step closer to a perfectly formatted repo ☺️

PR #1675, opened 2025-05-07
Checklist: Make health check number an option

obenland:
With health check issues being tracked in an option, we can add a callback to all of these options and determine when the last of them was completed. At that point we can consider the checklist complete and update user meta to no longer show the welcome tab.

## Proposed changes:

* Adds an option to keep track of Site Health Check issues.
* Adds calls to delete all new options when plugin gets deleted.
* Adds "profile setup" to the list of next steps.
* Accounts for button class in profile setup step, based on whether it's considered the next step.

### DIFF
--- a/includes/class-activitypub.php
+++ b/includes/class-activitypub.php
@@ -106,6 +106,11 @@ class Activitypub {
 		\delete_option( 'activitypub_blog_user_public_key' );
 		\delete_option( 'activitypub_blog_description' );
 		\delete_option( 'activitypub_blog_identifier' );
+		\delete_option( 'activitypub_checklist_blocks_visited' );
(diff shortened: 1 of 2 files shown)


pfefferle:
I am not sure what the improvement here is? That we limit the function calls that runs the checks?

obenland:
It gives us a place where we can determine the moment that all site health issues were fixed and that step can be marked as done. Adding it to the list of other options, we then know when all steps are done and the checklist can be hidden.

pfefferle:
But this can also be done with the version before or not?

obenland:
I don't know how. What would you suggest?

pfefferle (approved):

PR #1679, opened 2025-05-08
Checklist: Resolve once all steps are complete

obenland:
## Proposed changes:

* Adds callbacks to checklist option updates to determine whether the checklist is complete.
* Moves `activitypub_checklist_health_check_issues` update to below the step registration so we know how many steps there are to compare against.
* `activitypub_checklist_health_check_issues` is completed when it's at `0`, not something other than `0`.

### DIFF
--- a/includes/wp-admin/class-welcome-fields.php
+++ b/includes/wp-admin/class-welcome-fields.php
@@ -20,16 +20,8 @@ class Welcome_Fields {
 		\add_action( 'load-settings_page_activitypub', array( self::class, 'register_welcome_fields' ) );
 		\add_action( 'admin_print_styles-settings_page_activitypub', array( self::class, 'enqueue_styles' ) );
 
-		\add_action( 'add_option_activitypub_checklist_health_check_issues', array( self::class, 'resolve_checklist' ) );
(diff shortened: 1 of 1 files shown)


includes/wp-admin/class-welcome-fields.php
```diff
@@ -19,6 +19,9 @@ class Welcome_Fields {
 	public static function init() {
 		\add_action( 'load-settings_page_activitypub', array( self::class, 'register_welcome_fields' ) );
 		\add_action( 'admin_print_styles-settings_page_activitypub', array( self::class, 'enqueue_styles' ) );
+
+		\add_action( 'added_option', array( self::class, 'resolve_checklist' ) );
+		\add_action( 'updated_option', array( self::class, 'resolve_checklist' ) );
```
obenland: Hm, these would run for _every_ option addition or upgrade.

What did you not like about the option-specific versions?

pfefferle: it's quite noise and there is a high change that you miss one or have a typo.

> Hm, these would run for every option addition or upgrade.

but that is a valid point.

obenland: Let's revert and merge? should be ready for final testing then

pfefferle (approved):

PR #1683, opened 2025-05-09
Integration: Add support for the LightSpeed Cache plugin

pfefferle:
This PR adds integration support for the LiteSpeed Cache plugin to prevent ActivityPub JSON responses from being cached incorrectly. The integration automatically adds Apache rewrite rules to the `.htaccess` file and includes a Site Health check to verify proper configuration.

## Proposed changes:

- Adds automatic `.htaccess` rule management for LiteSpeed Cache
- Implements Site Health integration to monitor configuration status
- Includes comprehensive test coverage for all functionality

### Other information:

## Testing instructions:

1. Install LiteSpeed Cache plugin
2. Verify htaccess rules were added to the TOP of the file
3. Should see the `# BEGIN ActivityPub LiteSpeed Cache` section at the top, before WordPress rules
4. Check Site Health shows green status
5. Cleanup

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Add LiteSpeed Cache integration to prevent ActivityPub JSON responses from being cached incorrectly. Includes automatic .htaccess rules and Site Health check to ensure proper configuration.

</details>


integration/class-litespeed-cache.php
```diff
@@ -0,0 +1,143 @@
...
+RewriteCond %{HTTP:Accept} application
+RewriteRule ^ - [E=Cache-Control:vary=%{ENV:LSCACHE_VARY_VALUE}+isjson]
+</IfModule>';
+
+	/**
+	 * The option name to store the htaccess rules.
+	 *
+	 * @var string
+	 */
+	private static $option_name = 'activitypub_integration_litespeed_cache_htaccess_rules';
```
obenland: It might be personal preference, but I always feel like abstracting option names makes things harder to understand, not easier. I don't think we do it anywhere else?

If you feel like the key is too long, maybe we can find a shorter one?

integration/class-litespeed-cache.php
```diff
@@ -0,0 +1,143 @@
...
+			),
+			'description' => \sprintf(
+				'<p>%s</p>',
+				\__( 'Litespeed Cache is well configured to work with ActivityPub.', 'activitypub' )
+			),
+			'actions'     => '',
+			'test'        => 'test_litespeed_cache_integration',
+		);
+
+		if ( 0 === \get_option( self::$option_name, '0' ) ) {
```
obenland:
```suggestion
		if ( '0' === \get_option( self::$option_name, '0' ) ) {
```

integration/class-litespeed-cache.php
```diff
@@ -0,0 +1,143 @@
...
+	 * @param array $tests The site health tests.
+	 *
+	 * @return array The site health tests with the Litespeed Cache config test.
+	 */
+	public static function maybe_add_site_health( $tests ) {
+		if ( ! \is_plugin_active( 'litespeed-cache/litespeed-cache.php' ) ) {
+			return $tests;
+		}
+
+		$tests['direct']['activitypub_test_litespeed_cache_integration'] = array(
```
obenland, line 113: It would be good to test whether we're on Apache before adding the test.

obenland (commented):
I'm sensing a theme 🙂

Is there anything that needs to be done for nginx?

integration/class-litespeed-cache.php
pfefferle: I would love to! I'm always torn between "as short as possible" and "as descriptive as possible".

pfefferle:
I will check for an NGINX version, but this is still quite early in the progress... sadly `insert_with_markers` always "appends" the rules but the litespeed ones have to be at the top of the file.

integration/class-litespeed-cache.php
pfefferle: this is how core does it so I thought I go with the flow ;) !?!

integration/class-litespeed-cache.php
```diff
@@ -0,0 +1,204 @@
...
+	 * @var string
+	 */
+	public static $marker = 'ActivityPub LiteSpeed Cache';
+
+	/**
+	 * Initialize the integration.
+	 */
+	public static function init() {
+		// Add rules if LiteSpeed Cache is active and rules aren't set.
+		if ( is_plugin_active( 'litespeed-cache/litespeed-cache.php' ) && ! \get_option( self::$option_name ) ) {
```
obenland: For other integrations we determine whether a plugin is active by testing for a defined constant, is that not possible here?

integration/class-litespeed-cache.php
```diff
@@ -0,0 +1,204 @@
...
+			\update_option( self::$option_name, '1' );
+		} else {
+			\update_option( self::$option_name, '0' );
+		}
+	}
+
+	/**
+	 * Remove the LiteSpeed Cache htaccess rules.
+	 */
+	public static function remove_htaccess_rules() {
```
obenland, line 99: This doesn't seem to be used outside of tests?

integration/class-litespeed-cache.php
pfefferle: this is what we use in surge too

integration/class-litespeed-cache.php
pfefferle: Claude deleted my activation/deactivation hooks :(

integration/class-litespeed-cache.php
obenland: It didn't need a custom function for it. Maybe we could switch them both?

pfefferle: what do you mean?

this will throw an error because the plugin.php file is not loaded otherwise.

obenland: I mean that we're using Core's `is_plugin_active()` function directly in Surge.
Either way, could we avoid defining our own if we switched both to checking for defined constants?

pfefferle: I changed Surge too. Consts might be fragile, because they might be changed/removed!?

pfefferle: Surge does not add any consts and I am not sure if checking a function is reliable enough, given how often we change functions!

pfefferle: And the lightspeed folks are not even confident that their consts might be defined somewhere else: https://plugins.trac.wordpress.org/browser/litespeed-cache/trunk/litespeed-cache.php#L34

pfefferle: I went through the Surge code a bit to search for a good alternative we could check against... >95% of all the code is hooks that call closures ☺️

integration/load.php
```diff
@@ -146,13 +146,27 @@ function ( $transformer, $data, $object_class ) {
...
 }
 \add_action( 'plugins_loaded', __NAMESPACE__ . '\plugin_init' );
 
 // Register activation and deactivation hooks for Surge integration.
 \register_activation_hook( ACTIVITYPUB_PLUGIN_FILE, array( __NAMESPACE__ . '\Surge', 'add_cache_config' ) );
 \register_deactivation_hook( ACTIVITYPUB_PLUGIN_FILE, array( __NAMESPACE__ . '\Surge', 'remove_cache_config' ) );
 
+// Register activation and deactivation hooks for LiteSpeed Cache integration.
+\register_activation_hook( ACTIVITYPUB_PLUGIN_FILE, array( __NAMESPACE__ . '\LiteSpeed_Cache', 'add_htaccess_rules' ) );
+\register_deactivation_hook( ACTIVITYPUB_PLUGIN_FILE, array( __NAMESPACE__ . '\LiteSpeed_Cache', 'remove_htaccess_rules' ) );
```
obenland, line 167: With this deactivation hook we should be able to remove the changes in `includes/class-activitypub.php`.

pfefferle: The snipped in `includes/class-activitypub.php` made no sense at all 🫣

Promting for runaways!

obenland (approved):

PR #1684, opened 2025-05-09
Improve Mobile-View

pfefferle:
Improve the mobile view of the Settings Screens.

| Before   |      After    |
|----------|:-------------:|
| Screenshot 2025-05-09 at 15 28 58 | Screenshot 2025-05-09 at 15 29 16 |

## Proposed changes:

* Added margins for content area
* Scrollable menu for mobile views

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Settings pages are now more mobile-friendly with more space and easier scrolling.

</details>

### DIFF
--- a/.github/changelog/1684-from-description
+++ b/.github/changelog/1684-from-description
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Settings pages are now more mobile-friendly with more space and easier scrolling.
--- a/assets/css/activitypub-admin.css
+++ b/assets/css/activitypub-admin.css
@@ -401,3 +401,10 @@ input.blog-user-identifier {
 		margin: 0 22px;
 	}
 }
+
+@media screen and (max-width: 782px) {
+	.activitypub-settings .row > div {
+		max-width: calc(100% - 36px);
+		width: 100%;
+	}
+}
--- a/assets/css/activitypub-admin.css
+++ b/assets/css/activitypub-admin.css
@@ -45,6 +45,14 @@
 	padding-left: 0;
 }
 
+.activitypub-settings-tabs-scroller {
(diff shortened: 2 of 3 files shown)


obenland:
It's a bummer that on Core's side these tab-settings styles are not generalized, but specifically namespaced to privacy settings and site health, and there's nothing for us to reuse. This should be a solved problem.

Rather than introducing a tab scroller, I'd suggest recreating the markup and styles Core uses. I guess they never have more than three tabs?

pfefferle:
Form follows function, not the other way around!

pfefferle:
I know that you are no fan of the advanced settings tab, but we currently have it and also the welcome page. And changing the structure of the menu only because core has no css for that case is not a good UX.

Feel free to suggest something different, but for the meantime, this PR makes the admin settings at least usable.

pfefferle:
That is a very general issue of this page design. WordPress has no generic css for that. It prefixes the styles specific to these two sites. And because I wanted to avoid classes prefixed with site-health or privacy-policy in the AP settings, I decided to copy the styling.

I can give this a second look, but my wild guess is that we have not much other options here.

...and limiting/restructuring the menu to only have three items, only to avoid custom CSS is not a good UX in my mind.

We have so many places where we take a different path than core WordPress, like with the restyling of the E-Mails for example.

obenland:
How do privacy and site health manage situations where they have 4-6 tabs?

pfefferle:
No, they do not have more than three. I wanted to explain why there is nothing css wise we could simply reuse.

obenland (approved):

PR #1685, opened 2025-05-09
Send correct status header for Outbox items

pfefferle:
I had issues on several servers that the WordPress 404 page was not loaded properly but a webserver error message instead. This PR fixes the issue and adds a proper 406 code when the Accept type is wrong.

## Proposed changes:

* Send a HTTP "406 Not Acceptable" error code if a server queries an Outbox Item with a missing `application/json+activity` Accept header.
* It also adds the `404` query var, to ensure that the 404 page will be loaded in the frontend.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Better error handling when other servers request Outbox items in the wrong format, and 404 pages now show correctly.

</details>


obenland:
Is this at all unit-testable? It would be nice to have that covered.

includes/class-activitypub.php
```diff
@@ -141,6 +141,10 @@ public static function render_activitypub_template( $template ) {
 		self::add_headers();
 
 		if ( ! is_activitypub_request() || ! should_negotiate_content() ) {
+			if ( \get_query_var( 'p' ) && Outbox::POST_TYPE === \get_post_type( \get_query_var( 'p' ) ) ) {
```
obenland, line 112:
```suggestion
			if ( Outbox::POST_TYPE === \get_query_var( 'post_type' ) ) {
```
Does this work? For custom post types, it should always define the `post_type` var, otherwise it defaults to `post`.

pfefferle: Sadly not! That was what I tried first ☺️

I would assume that it does not work because the post type is not public!?!

obenland (approved):

Issue #426, opened 2023-09-14
Follow Me Block: improve performance / reduce reliance on Core @wordpress packages


jeherve:
The Follow Me block currently relies on different @wordpress packages:

- wp-api-fetch
- wp-components
- wp-compose
- wp-dom-ready
- wp-element
- wp-i18n
- wp-primitives

Those packages have to be loaded on the frontend, potentially on every page of the site if the Follow Me Block is added to a sidebar or the footer that appears on all pages.

Ideally, I think we should aim to have a block that's as light as possible on the frontend. Removing packages that in turn bring in all of React would be a good start, for example.

jeherve:
cc @mattwiebe who recently worked on #1091 for the Reactions block.

It would be nice if we could look into this for all our blocks. Having to load the whole React for the Reactions block, for example, may stop me from using the block on a site where I am careful about performance and try to avoid too many js loaded on the frontend.

mattwiebe:
Thanks @jeherve I definitely want to spend some time investigating the migration to the Interactivity API in the New Year

jeherve:
This also came up in #1455.

obenland:
With the current work on moving the blocks to using the Interactivity API, I think we can achieve almost all of that. They'll continue to rely on `wp-api-fetch` (for now), but remove all other dependencies. This work should be completed with #1747.

obenland:
This was fixed with the move to the Interactivity API.

PR #1691, opened 2025-05-11
Follow Me: Make it interactive!

obenland:
Moves the Follow Me block to using the Interactivity API for front-end rendering. Removes the block's dependency on React in the frontend.

Requires a major version release, updates minimum required WordPress version to 6.5.

## Proposed changes:

* Replaces React frontend of Follow ME block with Interactivity API.

### Other information:

## Testing instructions:

* Go to the Editor and add a Follow Me block.
* Test it in the Editor and frontend and make sure it still works as expected.
* It will no-longer open the modal, when in the Editor.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [x] Major

#### Type

#### Message

The Follow Me block now uses the latest Block Editor technology for display on the frontend.

</details>
Server-side rendering what we canServer-side rendering what we can


obenland:
@pfefferle This should be ready for testing

includes/class-blocks.php
```diff
@@ -469,4 +432,37 @@ function ( $paragraph ) {
...
+	 */
+	public static function add_directions_to_button( $content, $attributes ) {
+		$tags = new \WP_HTML_Tag_Processor( $content );
+
+		$button_class = '';
+		if ( isset( $attributes['buttonSize'] ) && 'small' === $attributes['buttonSize'] ) {
+			$button_class = 'is-small';
+		} elseif ( isset( $attributes['buttonSize'] ) && 'compact' === $attributes['buttonSize'] ) {
+			$button_class = 'is-compact';
+		}
```
obenland: I'm leaning towards removing this functionality and just keep customization options to whatever the core button block offers. What do you think?

pfefferle:
Dark mode is not properly supported by the new overlay:

Screenshot 2025-05-19 at 17 46 03

You can test it for example with: https://github.com/pfefferle/autonomie

pfefferle:
The button design will not be re-used in classic themes:

Screenshot 2025-05-19 at 17 49 15

pfefferle:
If you migrate from old "only button mode" to new button mode, it falls back to the full block

build/follow-me/render.php
```diff
@@ -0,0 +1,195 @@
...
+$block_id = 'activitypub-follow-me-block-' . wp_unique_id();
+
+// Get block style information.
+$style            = wp_get_global_styles();
+$background_color = $attributes['backgroundColor'] ?? $style['color']['background'] ?? '';
+
+// Get button style from block attributes.
+$button_style = $attributes['style'] ?? array();
+
+$actor = \Activitypub\Collection\Actors::get_by_id( $user_id );
```
pfefferle: we should stick with either full path (including Namespace) or define it as `use`, but not mix it.

See `use Activitypub\Blocks;`

build/follow-me/view.js.map
```diff
@@ -0,0 +1 @@
...es","base","style","backgroundColor","buttonTextColor","buttonColor","elements","link","buttonHoverColor","getPopupStyles","store","getContext","apiFetch","wp","trapFocus","element","focusableElements","querySelectorAll","firstFocusableElement","lastFocusableElement","length","focus","addEventListener","event","key","keyCode","shiftKey","activeElement","preventDefault","state","actions","openModal","context","isModalOpen","body","classList","add","setTimeout","blockWrapper","getElementById","blockId","modalFrame","querySelector","closeModal","remove","openButton","toggleModal","copyToClipboard","navigator","clipboard","writeText","webfinger","then","copyButtonText","i18n","copied","copy","error","console","updateRemoteProfile","remoteProfile","target","isError","errorMessage","handleKeyDown","submitRemoteProfile","namespace","input","emptyProfileError","test","invalidProfileError","isLoading","path","userId","encodeURIComponent","response","open","url","message","genericError","callbacks","initButtonStyles","buttonStyle","styleElement","createElement","textContent","head","appendChild","popupStyleElement","documentKeydown","documentClick","toggleButton","contains"],"sourceRoot":""}
```
pfefferle: where does the map files come from? is that a config change?

includes/class-blocks.php
pfefferle: I think it is important to have a "button" only mode, but let's remove the sizing for now.

build/follow-me/render.php
obenland: Fixed in f86bc55b4a794930d687c3520c05de9515b41ae8

includes/class-blocks.php
obenland: Fixed in f86bc55b4a794930d687c3520c05de9515b41ae8

obenland:
> Dark mode is not properly supported by the new overlay:

I added styles for dark mode in [15c0875](https://github.com/Automattic/wordpress-activitypub/pull/1691/commits/15c0875967d75d11820b0f6d73b37dee4fce4598). FWIW, it looks quite out of place and unexpected in any theme that doesn't have separate dark mode styles, which I assume is the vast majority. Not sure what we want to do there. Would you mind testing it out and see what you think?

build/follow-me/view.js.map
obenland: I think they're just left-overs from early dev mode. Good eye!
Removed in [c6b54f2](https://github.com/Automattic/wordpress-activitypub/pull/1691/commits/c6b54f247e171b4b2087b6bf04f9fdb7bb3fc980).

obenland:
> If you migrate from old "only button mode" to new button mode, it falls back to the full block

Fixed in dfee0d2b2c9ba06e2eb7f6672156889d48274f81.

obenland:
@pfefferle Thanks for giving this a look, I think all your feedback should be addressed. Could you give it another run, when you have a chance?

pfefferle:
The Button on the archive page looks different to the button on the single view (Twenty Twenty-Five)

| Archive | Single |
| --- | --- |
| Screenshot 2025-05-20 at 09 56 02 | Screenshot 2025-05-20 at 09 56 07 |

pfefferle:
Buttons still look weird in FSE themes (Twenty Twenty-Five and Twenty Twenty-Four):

Screenshot 2025-05-20 at 09 56 26

pfefferle:
Dark-Mode on FSE themes does not properly switch colors (see the headlines):

Screenshot 2025-05-20 at 10 01 46

src/follow-me/style.scss
```diff
@@ -213,6 +213,81 @@ body.modal-open {
 	overflow: hidden;
 }
 
+/* Dark mode styles. */
+@media (prefers-color-scheme: dark) {
```
pfefferle: I think the current code simply switches the global foreground color with the background and vice verse. This way we would not have to be so specific?

obenland:
> The Button on the archive page looks different to the button on the single view (Twenty Twenty-Five)

That seems to be a result of how Core enqueues inline styles for the button block. They're enqueued in Single view, but not in Archive.

obenland:
@pfefferle If you agree, I'd love to move on with this one. I think the styles are in a pretty good place now, thank you for all your feedback on that. Any additional nitpicks we have we could address in follow-up PRs?

pfefferle (approved):

jeherve:
@obenland Now that this is merged, do you think we could close #426? I haven't tested the block on `trunk` yet.

obenland:
It's getting close, yeah! It's fixed for the Follow Me block and I think #1747 removes the React dependencies for the final block. Thanks for the nudge!

PR #1694, opened 2025-05-13
Change: Move `delete_options` to `Options::class`

pfefferle:
Move the `delete_options` to the Options::class. This PR also fixes a namespace type in the Options::class.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


includes/class-options.php
```diff
@@ -30,6 +28,45 @@ public static function init() {
 		\add_filter( 'option_activitypub_max_image_attachments', array( self::class, 'default_max_image_attachments' ) );
 	}
 
+	/**
+	 * Delete all options.
+	 */
+	public static function delete() {
+		\delete_option( 'activitypub_actor_mode' );
```
obenland: Now that we touch this, I wonder if we should switch it to a `$wpdb->delete()` call that removes all `activitypub_*` options. That way we wouldn't have to keep updating it.

pfefferle: but we would bypass possible hooks!?!

pfefferle: But let's give it a try... I will change it to use $wpdb->delete()

obenland: It would not trigger those hooks, yeah. Is there anything depending on them?

pfefferle: not from the core AP plugin

obenland: If it comes up in the future, I'd recommend other plugins hook into the uninstall hook since that is where this function runs. Good catch.

includes/class-options.php
```diff
@@ -30,6 +28,16 @@ public static function init() {
...
 
+	/**
+	 * Delete all options.
+	 */
+	public static function delete() {
+		global $wpdb;
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
+		$wpdb->query( // phpcs:ignore WordPress.DB.DirectDatabaseQuery
+			"DELETE FROM $wpdb->options WHERE option_name LIKE 'activitypub_%'"
+		);
```
obenland:
```suggestion

		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
		$wpdb->query( "DELETE FROM $wpdb->options WHERE option_name LIKE 'activitypub_%'" );
```
One ignore should probably be sufficient?

obenland (approved):

PR #1696, opened 2025-05-13
Surge: Don't use AP constant in wp-config.

obenland:
Follow-up to #1648.

Fixes a fatal error after plugin update, when Surge cache config uses an Activitypub constant that has not been defined by the time `wp-config.php` executes.

```
Fatal error: Uncaught Error: Undefined constant "ACTIVITYPUB_PLUGIN_DIR" in /…/…/wp-config.php:78
```

Moving the cache config definition into a utility function allows us to reuse it when either the Activitypub plugin or the Surge plugin gets activated.

## Proposed changes:

* Moves cache config into a utility function, so we can format the plugin dir into it.

### Other information:

## Testing instructions:

* Download a zip file of this branch.
* Install and activate the Surge plugin on a test site.
* Upload and activate the plugin zip on the same test site.
* Make sure the site continues to work and doesn't fall into a 500.
* Double-check that the cache config constant definition was set.


integration/class-surge.php
```diff
@@ -177,11 +171,20 @@ public static function test_surge_integration() {
...
 
 		return $result;
 	}
+
+	/**
+	 * Get the cache config.
+	 *
+	 * @return string The cache config.
+	 */
+	public static function cache_config() {
```
pfefferle:
```suggestion
	public static function get_cache_config() {
```

what about using a getter to be consistent with `get_config_file_path` and more descriptive?

integration/class-surge.php
```diff
@@ -177,11 +171,20 @@ public static function test_surge_integration() {
...
+		return sprintf( "define( 'WP_CACHE_CONFIG', '%s/integration/surge-cache-config.php' );", rtrim( ACTIVITYPUB_PLUGIN_DIR, '/' ) );
```
pfefferle:
```suggestion
		return \sprintf( "define( 'WP_CACHE_CONFIG', '%s/integration/surge-cache-config.php' );", \rtrim( ACTIVITYPUB_PLUGIN_DIR, '/' ) );
```

integration/class-surge.php
obenland: Fixed in [4a3020b](https://github.com/Automattic/wordpress-activitypub/pull/1696/commits/4a3020bedce313b8a1efa7d084d9993ebbed6b6b).

pfefferle (approved):

PR #1705, opened 2025-05-14
Reactions: Fix fallback for blocks with default title

### LINKED ISSUE #1704: Did 5.9.0 break Fediverse reactions block?
### Quick summary

Fediverse Reactions are no longer displayed on blog index or individual posts, and the formatting in the template previews look strange (screenshot attached)

### Steps to reproduce

Image

### Site owner impact

### Severity

Critical

_No response_

_No response_

obenland:
Fixes #1704.

## Proposed changes:

* Changes the markup for Reaction blocks without inner content.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fixed a bug where Reaction blocks without modified titles did not get displayed correctly.

</details>


pfefferle (dismissed):
untested but lgtm!

jeherve (commented):
This fixed it for me!

One thing I've noticed, though, is that we now always display the heading on the frontend, even when we have no reactions. That seems to be a regression.

obenland:
What a nightmare. Let's hope this can be remedied when we convert it to using the Interactivity API.

includes/class-blocks.php
```diff
@@ -186,8 +186,21 @@ public static function render_post_reactions_block( $attrs, $content ) {
 		}
 
 		$args = array( 'data-attrs' => wp_json_encode( $attrs ) );
-		if ( isset( $attrs['title'] ) ) {
-			$args['class'] = 'activitypub-reactions-block';
+		if ( empty( $content ) ) {
+			if ( isset( $attrs['title'] ) ) {
+				$content = '<div class="activitypub-reactions"><div class="activitypub-reactions-block" ' . \get_block_wrapper_attributes( $args ) . '></div></div>';
+			} else {
+				$has_comments = \get_comments(
```
jeherve, line 193: Should this be checked even if you have a custom reactions block title?

obenland: In the case of a custom title `view.js` handles displaying that. However there's no good way to determine whether it uses a heading block or not from in there. At least not in this version of the block.

pfefferle (approved):
I have no idea if that is the best way to go, but it seems to work and we should fix that asap.

PR #1711, opened 2025-05-16
Add `in_reply_to` Actor to Activity audience

### LINKED ISSUE #1712: When creating a reply, add user that's being replied to in `cc`.

pfefferle:
Part of #1698 fixes #1712

## Proposed changes:

* Add the Actor of the replied post in `cc` or `to`, depending on the visibility.

### Other information:

## Testing instructions:

* Publish a new post with a reply-block
* Load the post-url with `?activitypub`
* Check the JSON output for the Actor of the replied URL

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

The actor of the replied-to post is now included in cc or to based on the post's visibility.

</details>


includes/class-http.php
```diff
@@ -235,6 +235,17 @@ public static function generate_cache_key( $url ) {
 	 * @return array|WP_Error The Object data as array or WP_Error on failure.
 	 */
 	public static function get_remote_object( $url_or_object, $cached = true ) {
+		/**
+		 * Filters the URL before making a GET request.
```
obenland:
```suggestion
		 * Filters the preemptive return value of a remote object request.
```

includes/class-http.php
```diff
@@ -235,6 +235,17 @@ public static function generate_cache_key( $url ) {
...
+		 *
+		 * @param array|string|null $response      The response.
+		 * @param array|string|null $url_or_object The Object or the Object URL.
+		 */
+		$response = apply_filters( 'activitypub_pre_http_get_remote_object', null, $url_or_object );
+		if ( $response && ! \is_wp_error( $response ) ) {
```
obenland:
```suggestion
		if ( null !== $response ) {
```
If this is filtered, `WP_Error` should probably be as valid a response as any

includes/transformer/class-base.php
```diff
@@ -7,10 +7,13 @@
 
 namespace Activitypub\Transformer;
 
+use Activitypub\Http;
 use Activitypub\Activity\Activity;
 use Activitypub\Collection\Actors;
 use Activitypub\Activity\Base_Object;
```
obenland, line 13:
```suggestion
use Activitypub\Activity\Activity;
use Activitypub\Activity\Base_Object;
use Activitypub\Collection\Actors;
use Activitypub\Http;

```

includes/transformer/class-base.php
```diff
@@ -162,28 +165,40 @@ public function set_content_visibility( $content_visibility ) {
...
-		$actor  = Actors::get_by_resource( $this->get_attributed_to() );
-		if ( ! $actor || is_wp_error( $actor ) ) {
-			$followers = null;
-		} else {
+		$public     = 'https://www.w3.org/ns/activitystreams#Public';
+		$followers  = null;
+		$replied_to = null;
+
+		$actor = Actors::get_by_resource( $this->get_attributed_to() );
+		if ( $actor && ! \is_wp_error( $actor ) ) {
```
```suggestion
		if ( ! \is_wp_error( $actor ) ) {
```
Is there a chance `Actors::get_by_resource()` returns a falsy value?

includes/transformer/class-base.php
```diff
@@ -162,28 +165,40 @@ public function set_content_visibility( $content_visibility ) {
...
 			$followers = $actor->get_followers();
 		}
+
 		$mentions = array_values( $this->get_mentions() );
 
+		if ( \method_exists( $this, 'get_in_reply_to' ) && $this->get_in_reply_to() ) {
```
obenland: With `get_in_reply_to()` defined in this class, can we drop that `method_exists()` check?

includes/transformer/class-base.php
```diff
@@ -162,28 +165,40 @@ public function set_content_visibility( $content_visibility ) {
...
+			$object = Http::get_remote_object( $this->get_in_reply_to() );
+			if ( $object && ! \is_wp_error( $object ) && isset( $object['attributedTo'] ) ) {
```
obenland, line 181:
```suggestion
			if ( isset( $object['attributedTo'] ) ) {
```
We should be able to shorten it to the `isset()` check. It'll return false for WP_Errors and falsy values.

pfefferle: because WP_Error is an object, it fails when you try to handle it like an array... should be the same for other "falsy" values.

obenland (approved):

PR #1713, opened 2025-05-16
Validate `$type` integrity for published posts

pfefferle:
Ensure that a newly created Blog-Post will always we sent to the fediverse as a `Create`.

## Proposed changes:

* Check the `$update` flag of the `wp_after_insert_post` action
* Check for `activitypub_status` of the post

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Ensured that publishing a new blog post always sends a Create to the Fediverse.

</details>


includes/scheduler/class-post.php
```diff
@@ -75,6 +79,11 @@ public static function schedule_post_activity( $post_id, $post, $update, $post_b
 			return;
 		}
 
+		// If the post was not federated before but is an Update activity, it should be a Create activity.
+		if ( get_wp_object_state( $post ) === 'federated' && 'Update' === $type ) {
```
obenland:
```suggestion
		if ( get_wp_object_state( $post ) !== 'federated' && 'Update' === $type ) {
```
_Not_ federated?

pfefferle:
@obenland Is this something we should do?

includes/scheduler/class-post.php
pfefferle: oops... fixed with https://github.com/Automattic/wordpress-activitypub/pull/1713/commits/43582a72c45963845a58138c30bea199e8db06b0

obenland:
Sure, worth a try. 

We could also revisit [this early return](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/functions.php#L1516-L1518), maybe add logging to it.

pfefferle:
Good point, I will do an action and add a logger!

includes/scheduler/class-post.php
```diff
@@ -55,7 +55,11 @@ public static function schedule_post_activity( $post_id, $post, $update, $post_b
 
 		switch ( $new_status ) {
 			case 'publish':
-				$type = ( 'publish' === $old_status ) ? 'Update' : 'Create';
+				if ( $update ) {
+					$type = ( 'publish' === $old_status ) ? 'Update' : 'Create';
+				} else {
+					$type = 'Create';
+				}
```
obenland, line 62: Could you expand on what's being changed here? Is there a case where this would be called with `$update` being false and `$old_status` being `'publish'`?

pfefferle: I decided to not use it, because it was not reliably enough, but mostly because of the other way around (`false === $update` for creates).

pfefferle: but I thought if it is properly set, maybe we can use it and if not (maybe because of block editor AJAX interference) we fall back to the status check.

includes/functions.php
```diff
@@ -1508,12 +1508,14 @@ function add_to_outbox( $data, $activity_type = null, $user_id = 0, $content_vis
 	}
 
 	if ( ! $activity || \is_wp_error( $activity ) ) {
+		\do_action( 'activitypub_add_to_outbox_failed', $data, $activity_type, $user_id, $content_visibility );
```
obenland: I think it would be helpful to pass the WP_Error if we have one (and add hook docs)

includes/scheduler/class-post.php
obenland: "It" being `$old_status`?

If there are unexpected values we should debug them and fix them. Can you help me understand what you're seeing?

pfefferle: This was quite a while ago when I worked on that, but I think `$updated === false` is not necessarily published, but can also be the first draft that was created! But I would have to dig deeper into that again!

pfefferle: I is as I said: `$updated` is true if it is a second save on a post (if published or not). So a draft, that is saved for the second time has `$published === true`.

We could remove the `$updated` check, because it would never be `true` if someone uses the WordPress editor, but it might be a slightly faster check if you directly use the `wp_insert_post` function (might be the same with the classic editor) to directly publish a post and bypassing the drafts.

obenland (approved):
I added hook docs for the new actions and moved the error to the first arg position since that's what the hook is about.

PR #1717, opened 2025-05-19
Add .map files to gitignore

obenland:
@pfefferle noted a stray `.map` file in #1691. This change makes sure we'll never accidentally commit `.map` files again.

See https://github.com/Automattic/wordpress-activitypub/pull/1691#discussion_r2096125668

## Proposed changes:

* Adds `/build/**/*.map` to `.gitignore`.
* Sorts `.gitignore` alphabetically.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,22 @@
-_site
-.sass-cache
-.jekyll-cache
-.jekyll-metadata
+# Directories
+/build/**/*.map
 /coverage/
 /node_modules/
 /vendor/
-package-lock.json
-composer.lock
-.DS_Store
-.cursor
+_site
 .idea/
+
+# Files
+.cursor
+.DS_Store
+.jekyll-cache
+.jekyll-metadata
 .php_cs.cache
 .phpunit.result.cache
+.sass-cache
 .vscode/settings.json
 .windsurf
 .windsurfrules
 .wp-env.override.json
+composer.lock
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,11 @@
-# Directories
 /build/**/*.map
 /coverage/
 /node_modules/
 /vendor/
 _site
 .idea/
-
-# Files
 .cursor
 .DS_Store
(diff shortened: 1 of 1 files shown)


pfefferle:
I understand the idea behind the groups, but this forces us to be precise and that might cause failures. I would remove the differentiation between files and folders.

.gitignore
```diff
@@ -1,18 +1,22 @@
...
-.DS_Store
-.cursor
+_site
 .idea/
+
+# Files
+.cursor
+.DS_Store
+.jekyll-cache
+.jekyll-metadata
```
pfefferle:
```suggestion
```

We removed jekyll and decided to link directly to the docs folder.

.gitignore
```diff
@@ -1,18 +1,22 @@
...
 /vendor/
-package-lock.json
-composer.lock
-.DS_Store
-.cursor
+_site
 .idea/
+
+# Files
+.cursor
```
pfefferle, line 7: this is a folder

.gitignore
```diff
@@ -1,18 +1,22 @@
...
+.DS_Store
+.jekyll-cache
+.jekyll-metadata
 .php_cs.cache
 .phpunit.result.cache
+.sass-cache
```
pfefferle, line 11: I think this is a folder too

.gitignore
```diff
@@ -1,18 +1,22 @@
...
 .vscode/settings.json
 .windsurf
```
pfefferle, line 13: I am not totally sure about windsurf, but if it follows how vscode and cursor is handling configs, this is also a folder.

obenland:
> I understand the idea behind the groups, but this forces us to be precise and that might cause failures. I would remove the differentiation between files and folders.

Agreed, [f1d0e1a](https://github.com/Automattic/wordpress-activitypub/pull/1717/commits/f1d0e1a4a37bc5dab130ca275d47d46643da4eda) should take care of it

pfefferle (approved):

PR #1722, opened 2025-05-20
Reactions: Use Interactivity API

obenland:
Moves the Follow Me block to using the Interactivity API for front-end rendering. Removes the block's dependency on React in the frontend.

## Proposed changes:

* Replaces React frontend of Reactions block with Interactivity API. 
* Shows actual Reactions in the Editor when there are any available.

### Other information:

## Testing instructions:

* Go to the Editor and add a Reactions block.
* Test it in the Editor and frontend and make sure it still works as expected.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [x] Major

#### Type

#### Message

The Reactions block now uses the latest Block Editor technology for display on the frontend.

</details>


obenland:
@pfefferle This will likely need some more detail work on the Edit side, but should be good to test and get feedback on! Curious what you think of it!

src/reactions/style.scss
```diff
@@ -1,13 +1,31 @@
...
 		display: flex;
 		align-items: center;
 		margin: 0.5em 0;
 		position: relative;
 		width: 100%;
-		gap: 0.75em;
+		gap: 0.5rem;
 		justify-content: flex-start;
 
+		&[hidden] {
```
pfefferle: 🧐

pfefferle:
I've not looked too deep into the code yet, but it works quite well on my dev system.

src/reactions/style.scss
obenland: Removed in [f395833](https://github.com/Automattic/wordpress-activitypub/pull/1722/commits/f3958339a04e532a04ef65d88a6f7374a6a35326).

src/reactions/reactions.js
```diff
@@ -100,56 +91,9 @@ const ReactionList = ( { reactions, type } ) => {
...
-		const resizeObserver = new ResizeObserver( calculateVisibleAvatars );
-		resizeObserver.observe( containerRef.current );
-
-		return () => {
-			resizeObserver.disconnect();
-		};
-	}, [ buttonRef, items.length ] );
-
-	const visibleItems = items.slice( 0, visibleCount );
+	const visibleItems = items.slice( 0, 20 );
```
pfefferle, line 96: shouldn't the be `visibleCount` be configurable somehow?

src/reactions/view.js
```diff
@@ -1,14 +1,279 @@
...
+			const context = getContext();
+			const button = target.closest( '[data-reaction-type]' );
+			const reactionType = button.getAttribute( 'data-reaction-type' );
+
+			// Set modal properties.
+			context.isModalOpen = true;
+			context.modal.items = state.reactions[ context.postId ][ reactionType ].items;
+
+			// Position the compact modal relative to the button.
+			setTimeout( callbacks.positionModal, 0 );
```
pfefferle, line 47: `setTimeout`?

src/reactions/reactions.js
obenland: I don't think it needs to be. It was never really configurable, the above code would just adjust it to the amount of space it has available in the viewport. [That is still the case on the frontend](https://github.com/Automattic/wordpress-activitypub/pull/1722/files/2bd0907b616012b1f4f631504f167e3380c49676#diff-12bce84babea4bdfbae0bbaf952e9e9736a08be28910c8301bac64a9f5a8b110R83) but won't be the case in the Editor, which this code is limited to now.

On the frontend, avatars will be server-side rendered, so without a limit it would create a flash of overflowing avatars on posts with a lot of reactions. 20 is an arbitrary number but seemed to work reasonably well in testing.

src/reactions/view.js
obenland: `setTimeout` with a 0 timeout defers execution until after the current JavaScript call stack has cleared, making sure that DOM updates (like the modal being rendered) are complete before it attempts to position it

pfefferle: javascript 🫣

pfefferle (approved):

PR #1726, opened 2025-05-23
Improve how the plugin stores Followers

pfefferle:
This is the first PR to unify the way we store (remote) Actors. This allows us to store Followers and Followings in the same way.

The idea is, to have one Post-Type called `ap_actor` and mappings (through `post_meta`) to the local User:

* `_activitypub_following` stores the User-ID of the user that is followed by the stored Actor
* `_activitypub_followed_by` could store the User-ID of the user, the remote user is following

Using variations og these `post_meta`s we could also store pending `Follow` requests: `_activitypub_pending_following`

## Proposed changes:

* Changes the CPT and Post-Metas
* Migrate current data

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

New system to manage followers and followings more consistently using a unified actor type.

</details>


obenland:
I worry about scalability when tracking follow relationships through post meta. Like, what happens when the same actor follows the blog user and an author? Or multiple authors. Or in the future, what happens when multiple authors from the same site want to follow you?

### Scenario

Local Users: Alice (ID: 123) and Frank (ID: 456)
Both Follow: Bob (`@bob@mastodon.social`) - Same remote actor

`wp_posts` Table
ID | post_title | post_type | post_status | post_author
-- | -- | -- | -- | --
1001 | bob@mastodon.social | ap_actor | publish | 0

`wp_postmeta` for Bob's actor post:
| meta_id | post_id | meta_key | meta_value |
|---------|---------|----------|------------|
| 5006 | 1001 | _activitypub_followed_by | 123 |
| 5025 | 1001 | _activitypub_followed_by | 456 |  // Same key, different value

`get_post_meta($post_id, '_activitypub_followed_by', true)` would only return the FIRST value (123), not both users.

# Possible Alternative

I think we may have touched on it before, but I wonder if a Follow CPT could help with scalability? Something along the lines of:

### wp_posts Table
| ID | post_title | post_type | post_status | post_author | post_parent | post_content |
|----|------------|-----------|-------------|-------------|-------------|---------------|
| 1001 | bob@mastodon.social | ap_actor | publish | 0 | 0 | {"actor_id":"https://...","inbox":"...","username":"bob"} |
| 1002 | carol@pixelfed.social | ap_actor | publish | 0 | 0 | {"actor_id":"https://...","inbox":"..."} |
| 2001 | user_123_follows_actor_1001 | ap_follow | publish | 123 | 1001 | |
| 2002 | user_456_follows_actor_1001 | ap_follow | publish | 456 | 1001 | |
| 2003 | user_123_follows_actor_1002 | ap_follow | publish | 123 | 1002 | |


### Code examples

#### Get Actor Followers
```php
/**
 * Get all followers of a specific actor.
 *
 * @param int $actor_post_id The actor post ID.
 * @return array Array of user IDs who follow this actor.
 */
function activitypub_get_actor_followers( $actor_post_id ) {
	$followers = get_posts( array(
		'post_type'      => 'ap_follow',
		'post_parent'    => $actor_post_id,
		'post_status'    => 'publish', // Accepted follows only
		'posts_per_page' => -1,
	) );

	return wp_list_pluck( $followers, 'post_author' );
}

// Usage:
$bob_followers = activitypub_get_actor_followers( 1001 ); // Returns: array( 123, 456 )
```

#### Get User's Following List
```php
/**
 * Get list of actors that a user follows.
 *
 * @param int $user_id The WordPress user ID.
 * @return array Array of actor post objects.
 */
function activitypub_get_user_following( $user_id ) {
	// First get the follow relationship posts.
	$follow_posts = get_posts( array(
		'post_type'      => 'ap_follow',
		'author'         => $user_id,
		'post_status'    => 'publish',
		'posts_per_page' => -1,
	) );

	$follow_posts = wp_list_pluck( $get_posts, 'post_parent' );

	if ( empty( $follow_posts ) ) {
		return array();
	}

	// Get actor details.
	$actor_args = array(
		'post_type'      => 'ap_actor',
		'post__in'       => $follow_posts,
		'posts_per_page' => -1,
		'post_status'    => 'publish',
	);

	return get_posts( $actor_args );
}
```

#### Check Follow Relationship
```php
/**
 * Check if a user follows a specific actor.
 *
 * @param int $user_id       The WordPress user ID.
 * @param int $actor_post_id The actor post ID.
 * @return bool True if user follows actor, false otherwise.
 */
function activitypub_user_follows_actor( $user_id, $actor_post_id ) {
	$args = array(
		'post_type'      => 'ap_follow',
		'author'         => $user_id,
		'post_parent'    => $actor_post_id,
		'post_status'    => array( 'publish', 'pending' ),
		'posts_per_page' => 1,
		'fields'         => 'ids',
	);

	$existing = get_posts( $args );

	return ! empty( $existing );
}
```

#### Create Follow Relationship
```php
/**
 * Create a follow relationship between a user and an actor.
 *
 * @param int    $user_id       The WordPress user ID.
 * @param int    $actor_post_id The actor post ID.
 * @param string $status        The follow status: 'pending', 'publish', 'private'.
 * @return int|false The follow post ID on success, false on failure.
 */
function activitypub_create_follow_relationship( $user_id, $actor_post_id, $status = 'pending' ) {
	// Check if relationship already exists.
	if ( activitypub_user_follows_actor( $user_id, $actor_post_id ) ) {
		return false;
	}

	$post_data = array(
		'post_type'    => 'ap_follow',
		'post_title'   => sprintf( 'user_%d_follows_actor_%d', $user_id, $actor_post_id ),
		'post_author'  => $user_id,
		'post_parent'  => $actor_post_id,
		'post_status'  => $status,
		'post_content' => '',
	);

	return wp_insert_post( $post_data );
```

pfefferle:
> get_post_meta($post_id, '_activitypub_followed_by', true) would only return the FIRST value (123), not both users.

This is because you set `$single` to `true` (last param). If you set it `false` it returns all.

pfefferle:
> Like, what happens when the same actor follows the blog user and an author? Or multiple authors. Or in the future, what happens when multiple authors from the same site want to follow you?

I am not sure why this might be an issue. Every user has a unique id and therefore I see no difference to the current follower system. It allows multiple local users to follow one remote user, so it is easily possible to model the other way around in the same way.

pfefferle:
I will look at your proposal on Friday. Looks interesting, but might make it even more complicated for contributors.

pfefferle:
The problem with CPTs as follow/following persistence is, that we have a user with id `0` and one with `-1` so you can not simply use the `post_author`.

obenland:
> The problem with CPTs as follow/following persistence is, that we have a user with id 0 and one with -1 so you can not simply use the post_author.

I think that may have been what led to us wanting to go forward with the the post meta approach in the first place? Yeah, that's a good point

includes/constants.php
```diff
@@ -68,6 +68,8 @@
 define( 'ACTIVITYPUB_BLOG_MODE', 'blog' );
 define( 'ACTIVITYPUB_ACTOR_AND_BLOG_MODE', 'actor_blog' );
 
+define( 'ACTIVITYPUB_ACTOR_POST_TYPE', 'ap_actor' );
```
obenland: So far we've used class constants for cpt slugs, is there a way to continue with that? Maybe in `Activitypub\Activity\Actor`?

pfefferle:
...this is such a hen/egg problem! Do we have to reprioritize again, to talk about actors first?

🤕🫣😱

includes/constants.php
pfefferle: To be consistent we would have to use the collection, but yes, valid point... also thought about that. Would love to get rid of the models too. Should be possible to use the actor object.

obenland:
I think I'm fine going forward with post meta. We won't get around trying _something_, might as well be this. We can always write another upgrade routine :)

obenland (dismissed):
Let's give this a shot!

pfefferle:
@obenland the objects in the `Activity` folder should be db-agnostic! `Actor` is a general object that is already used to represent any user remote and local, so adding the POST_TYPE there is quite misleading. On the other hand it is inconsistent, because we use the collections to store the post type (see outbox and extra-fields) https://github.com/Automattic/wordpress-activitypub/pull/1726#discussion_r2112620375

obenland:
My bad, looks like I misunderstood "[Should be possible to use the actor object.](https://github.com/Automattic/wordpress-activitypub/pull/1726#discussion_r2112620375)".

pfefferle:
Can you recheck @obenland ?

obenland (approved):

PR #1730, opened 2025-05-27
Strip shortcodes from the excerpt

pfefferle:
This PR removes the shortcodes from the excerpt, because it seems to be possible to sneak them into the summary... shomehow...

Fixes https://c.im/@michael_martinez/114576417875641281

Update: While working on the unit-tests I found some glitches that bypassed the sanitization. This PR will also fix that issue.

## Proposed changes:

* Remove shortcodes from the excerpt
* Remove the early bailing of the excerpt, to not have to duplicate the `remove_shortcode` and the other checks

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Improved excerpt handling and improved sanitization of summaries.

</details>

### DIFF
--- a/.github/changelog/1730-from-description
+++ b/.github/changelog/1730-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Improved excerpt handling by removing shortcodes from summaries.
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1210,9 +1210,9 @@ function generate_post_summary( $post, $length = 500 ) {
 		$length  = null;
 	}
 
+	$content = \strip_shortcodes( $content );
 	$content = \html_entity_decode( $content );
 	$content = \wp_strip_all_tags( $content );
-	$content = \strip_shortcodes( $content );
 	$content = \trim( $content );
 	$content = \preg_replace( '/\R+/m', "\n\n", $content );
 	$content = \preg_replace( '/[\r\t]/', '', $content );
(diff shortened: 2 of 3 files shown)


obenland:
This feels like a good change to add a unit test for?

pfefferle:
The thing is, that this is not the normal behaviour... must be some plugin that adds the shortcodes directly to the excerpt...

pfefferle:
But sure, will add some tests...

pfefferle:
@obenland I optimized the code a bit and added tests for all known cases.

obenland:
Would you mind updating the "Proposed Changes" list in the PR description with your changes? It looks like existing excerpts no longer short-circuit the function but now get processed fully, is that correct?

pfefferle:
yes, because that bypassed all the following sanitization

pfefferle:
@obenland done.

obenland:
> yes, because that bypassed all the following sanitization

My assumption would be that Core's sanitization for excerpts is enough—is it not? If not, what's missing?

> Optimize excerpt/summary generation and sanitization

What's the change in that optimization?

pfefferle:
> My assumption would be that Core's sanitization for excerpts is enough—is it not? If not, what's missing?

This has some history: Sadly other plugins are doing too much weird things with the excerpt, so that is was needed. Here is some background: https://github.com/Automattic/wordpress-activitypub/issues?q=is%3Aissue%20excerpt

> What's the change in that optimization?

I changed it to a more descriptive one... I am sorry for the confusion.

obenland (approved):
Thank you!

PR #1733, opened 2025-05-27
Fix: Site-Health WebFinger check

pfefferle:
Fixes a bug in the Site-Health that causes a PHP error.

Bug Report:

> The site health check shows a critial issue “WebFinger endpoint is not accessible” but there are no details when opening the accordion item.
>
> Beside that, the Query Monitor Plugin shows two PHP warnings “Array to string conversion” at these locations:
>
> >   wp-content/plugins/activitypub/includes/wp-admin/class-health-check.php:245
>    sprintf()
>    wp-content/plugins/activitypub/includes/wp-admin/class-health-check.php:245
>    Activitypub\WP_Admin\Health_Check::is_webfinger_endpoint_accessible()
>    wp-content/plugins/activitypub/includes/wp-admin/class-health-check.php:169
>    Activitypub\WP_Admin\Health_Check::test_webfinger()
>    wp-admin/includes/class-wp-site-health.php:194
>    WP_Site_Health->perform_test()
>    wp-admin/includes/class-wp-site-health.php:145
>    WP_Site_Health->enqueue_scripts()
>    wp-includes/class-wp-hook.php:324
>    do_action('admin_enqueue_scripts')
>    wp-admin/admin-header.php:123
>
> How can I debug this? Webhoster Netcup can’t help.

https://wordpress.org/support/topic/webfinger-endpoint-is-not-accessible-php-warning/

## Proposed changes:

* Check WP_Error `data` before processing it

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fixed a bug in Site Health that caused a PHP warning and missing details for the WebFinger check.

</details>

### DIFF
--- a/includes/wp-admin/class-health-check.php
+++ b/includes/wp-admin/class-health-check.php
@@ -241,15 +241,21 @@ class Health_Check {
 				$allowed
 			);
 
+			$data       = $url->get_error_data();
+			$author_url = $resource;
+			if ( \is_string( $data['data'] ) ) {
+				$author_url = $data['data'];
+			}
+
 			$health_messages = array(
 				'webfinger_url_not_accessible'   => \sprintf(
 					$not_accessible,
-					$url->get_error_data()['data']
+					$author_url
 				),
 				'webfinger_url_invalid_response' => \sprintf(
 					// translators: %s: Author URL.
 					$invalid_response,
-					$url->get_error_data()['data']
+					$author_url
 				),
 			);
 			$message         = null;
--- a/includes/wp-admin/class-health-check.php
+++ b/includes/wp-admin/class-health-check.php
@@ -243,7 +243,7 @@ class Health_Check {
 
 			$data       = $url->get_error_data();
 			$author_url = $resource;
-			if ( \is_string( $data['data'] ) ) {
+			if ( isset( $data['data'] ) && \is_string( $data['data'] ) ) {
 				$author_url = $data['data'];
 			}
 
--- a/.github/changelog/1733-from-description
+++ b/.github/changelog/1733-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed a bug in Site Health that caused a PHP warning and missing details for the WebFinger check.
--- a/includes/class-webfinger.php
+++ b/includes/class-webfinger.php
@@ -211,17 +211,17 @@ class Webfinger {
 		$webfinger_url = sprintf(
 			'https://%s/.well-known/webfinger?resource=%s',
 			$host,
-			rawurlencode( $identifier )
+			\rawurlencode( $identifier )
 		);
 
-		$response = wp_safe_remote_get(
+		$response = \wp_safe_remote_get(
 			$webfinger_url,
 			array(
 				'headers' => array( 'Accept' => 'application/jrd+json' ),
 			)
 		);
 
-		if ( is_wp_error( $response ) ) {
+		if ( \is_wp_error( $response ) || \wp_remote_retrieve_response_code( $response ) >= 400 ) {
 			return new WP_Error(
 				'webfinger_url_not_accessible',
 				__( 'The WebFinger Resource is not accessible.', 'activitypub' ),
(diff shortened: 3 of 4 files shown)


obenland:
Do we know what caused the bug in the first place? Is there a way to add a unit test for this bug fix so we can make sure it won't happen again?

pfefferle:
It was a bad idea to rely on a properly formatted `WP_Error` throughout multiple layers in the first place, so I would say that this is the best solution to something we can not really predict.

obenland:
Could you help me understand what it is that makes it hard to predict?

pfefferle:
we take the data from a `WP_Error` that could be triggered in so many layers of the whole resolve process. every change in one of these layers regarding the data attribute would cause a simlar error. changing the data attribute in one of these layers might cause errors in a different tree of the code that also rely on that information.

obenland:
Is it possible there was an error in the webfinger endpoint response and the array is [the new error transformation](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/rest/class-server.php#L201-L213)?

pfefferle:
The code is not working with the response data on an error, so I do not think that this might be the issue.

obenland:
I think that used to be the case when the endpoint returned an error, with the error transformation [it now returns an array](https://obietester.blog/.well-known/webfinger?resource=acct%3Aapplications%40obietester.blog).

pfefferle:
But there is no place in the code where the response of the request were taken as data `WP_Error`: https://github.com/Automattic/wordpress-activitypub/blob/170bab3af3ae1e87c8367c480401afdec87d7a4e/includes/class-webfinger.php#L225

pfefferle:
The only place in the code that might return an array are in `get_remote_follow_endpoint` but the code also only work with the response if there were no error before https://github.com/Automattic/wordpress-activitypub/blob/170bab3af3ae1e87c8367c480401afdec87d7a4e/includes/class-webfinger.php#L250

pfefferle:
ok, debugging says you are right...

pfefferle:
Oh, I think I made a false asumption and the API change made that visible! When is `wp_safe_remote_get` returning an `WP_Error`? Does it translate `> 400` response codes to errors??

pfefferle:
That one fixes the problem: https://github.com/Automattic/wordpress-activitypub/pull/1733/commits/3ed335ad45ad890cc0eeeb4f5627754585c1ba4b

That should also improveme the remote requests, which might have ended up with the same issue!

pfefferle:
@obenland Thanks for being so persistent!

includes/wp-admin/class-health-check.php
```diff
@@ -241,15 +241,21 @@ public static function is_webfinger_endpoint_accessible() {
 				$allowed
 			);
 
+			$data       = $url->get_error_data();
+			$author_url = $resource;
+			if ( isset( $data['data'] ) && \is_string( $data['data'] ) ) {
+				$author_url = $data['data'];
+			}
```
obenland, line 248: With the changes in `class-webfinger.php`, are these still needed?

obenland (changes requested):
Let's still add a unit test for the case where the API call returns an error

includes/wp-admin/class-health-check.php
pfefferle: I would keep them, because they also improve stability.

pfefferle:
> Let's still add a unit test for the case where the API call returns an error

I think I have to spend some time tomorrow to add tests for the whole WebFinger class! It is nearly untested so far :(

includes/wp-admin/class-health-check.php
obenland: We should fix whatever is introducing instability, could you tell me more about the concern?

pfefferle: Same argument I mentioned above! The `WP_Error`s are not optimized (and should not be) to work with the Health-Check. We might need some other informations in `data` on other places, so any new change might cause issues.

obenland (approved):
Thank you!

PR #1747, opened 2025-05-29
Followers: Use Interactivity API

obenland:
Before | After
--- | ---
Screenshot 2025-05-29 at 10 38 16 AM | image

#### Block Variations

Default | Compact | Card
--- | --- | ---
image | Screenshot 2025-05-29 at 10 39 25 AM | Screenshot 2025-05-29 at 10 38 55 AM

## Proposed changes:

* Updates design to align closer with Follow-Me and Reactions blocks.
* Updates block variations with a Compact and a Card version.
* Uses Interactivity API for pagination.
* Updates title to use the Heading block for better customization.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [x] Major

#### Type

#### Message

Follwoers block has an updated design, new block variations, and uses the Interactivity API for display on the frontend.

</details>


pfefferle:
Really nice redesign!

src/followers/render.php
```diff
@@ -0,0 +1,153 @@
...
+
+			<button
+				class="pagination-next wp-block-button__link wp-element-button"
+				data-wp-on--click="actions.nextPage"
+				data-wp-bind--disabled="state.disableNextButton"
+				aria-label="<?php esc_attr_e( 'Next page', 'activitypub' ); ?>"
+			>
+				<?php esc_html_e( 'Next', 'activitypub' ); ?>
+			</button>
+		</nav>
```
pfefferle: this part is now defined in several places and makes it hard to update.

pfefferle:
The prev/next buttons are very dominant and draw attention away from the actual block. I would stick with the links.

src/followers/render.php
obenland: Yeah, I'm not quite sure how to reconcile that, to be honest. The JavaScript and Interactivity versions are different enough from one-another, that it's not super obvious to me how to make them reusable. Do you have a suggestion?

pfefferle: anything that might help with these different implementations will not run "out of sync"? (that we only update one of the different related sites?)

obenland: Documentation might be the best thing we can do for now?

pfefferle: my inner monk cries "redundancies"! ;)

obenland: I mean, I agree. I thought the same during development, actually. I'm open to suggestions

obenland:
Block after the update to links:

image

pfefferle:
The editor view seems to be stuck with the follower list of the user, when the site user has no followers.

Precondition: Site user has no follower, user has one (or more followers).

Steps to reproduce:

* Add block to post or page
* Defaults to "Site user" and shows no followers in editor
* Change to a specific user
* Editor shows the one follower of the user
* Switch back to "Site user" and editor will still show the users follower

Update:

Changing to a user with no followers produces the same result, so this is not a problem of the "Site user", but maybe something broken with the update mechanism.

pfefferle:
I can't select my second user from the dropdown.

Second user has subscriber role, but activitypub cap.

obenland:
> I can't select my second user from the dropdown.
> Second user has subscriber role, but activitypub cap.

@pfefferle [The query looks for authors and above](https://github.com/Automattic/wordpress-activitypub/pull/395/files#diff-8325b0b59564280ee473e191d9889e729b0883ddc4b1965a13d40557374054d0R7), introduced with the Follow Me block. My assumption is that it's because Subscribers can't creates Posts and there would be nothing to follow? I'd probably keep it the way it is?

obenland:
> The editor view seems to be stuck with the follower list of the user, when the site user has no followers.

Fixed in [3595768](https://github.com/Automattic/wordpress-activitypub/pull/1747/commits/3595768d3bba40bdc29343faf3d895096a20e1c4).

obenland:
For the case where there are no followers, the block doesn't render in the frontend and I added a notification in the Editor:
image

obenland:
@pfefferle This should be ready for another look

pfefferle:
> [The query looks for authors and above](https://github.com/Automattic/wordpress-activitypub/pull/395/files#diff-8325b0b59564280ee473e191d9889e729b0883ddc4b1965a13d40557374054d0R7), introduced with the Follow Me block. My assumption is that it's because Subscribers can't creates Posts and there would be nothing to follow? I'd probably keep it the way it is?

the idea of the cap independent from the role was, that big news sites could write articles in the name of guest authors.

pfefferle (changes requested):
> [The query looks for authors and above](https://github.com/Automattic/wordpress-activitypub/pull/395/files#diff-8325b0b59564280ee473e191d9889e729b0883ddc4b1965a13d40557374054d0R7), introduced with the Follow Me block. My assumption is that it's because Subscribers can't creates Posts and there would be nothing to follow? I'd probably keep it the way it is?

the idea of the cap independent from the role was, that big news sites could write articles in the name of guest authors.

the current implementation ignores the `activitypub` cap completely and shows users that have no capabilities and hides users with it!

we should not use different rules throughout the code! if we decide to have the `publish_post` cap, we should change the whole codebase, so that it is not possible to add the `activitypub` cap to subscribers.

for now I would suggest to fully rely on the `activitypub` cap, because that is used anywhere else!

pfefferle:
> Fixed in [3595768](https://github.com/Automattic/wordpress-activitypub/pull/1747/commits/3595768d3bba40bdc29343faf3d895096a20e1c4).

Now the editor always show the `No followers found.` even if the user has followers!

obenland:
> for now I would suggest to fully rely on the activitypub cap, because that is used anywhere else!

Thanks, I created #1754 to address that since it's unrelated to this PR.


> Now the editor always show the No followers found. even if the user has followers!

Hm, I can't reproduce that. Is it possible it's a conflict with how Followers are stored after the merge of #1726?


https://github.com/user-attachments/assets/6b763208-ce24-431e-add6-b804b49d9caa

src/followers/edit.js
```diff
@@ -1,40 +1,43 @@
...
+/**
+ * Edit component.
+ *
+ * @param {Object} props Component props.
+ * @param {Object} props.attributes Block attributes.
+ * @param {Function} props.setAttributes Set block attributes.
+ * @param {Object} props.context Block context.
+ * @param {string} props.context.postType Post type.
+ * @param {number} props.context.postId Post ID.
+ * @return {JSX.Element} Edit component.
```
pfefferle:
```suggestion
 * @param {number} props.context.postId Post ID.
 *
 * @return {JSX.Element} Edit component.
```

src/followers/edit.js
```diff
@@ -52,48 +55,244 @@ export default function Edit( {
 		}
 	}, [ selectedUser, usersOptions ] );
 
+	// Template for InnerBlocks - allows only a heading block.
+	const TEMPLATE = [
```
pfefferle, line 60: why caps? we seem to use camelcase for the rest.

src/followers/render.php
```diff
@@ -0,0 +1,152 @@
...
+	unset( $attributes['title'], $attributes['className'] );
+} else {
+	// Remove the outer div if it exists, to avoid nesting issues with block variation classes.
+	$pattern = '/<div\s+class="wp-block-activitypub-followers[^"]*"[^>]*>(.*?)<\/div>/s';
+	if ( preg_match( $pattern, $content, $matches ) ) {
+		$content = $matches[1];
+	}
+}
+
+$followee_user_id = Blocks::get_user_id( $attributes['selectedUser'] );
```
pfefferle: should we call it actor or user? followee might mislead... even more because we are working on followees

src/followers/render.php
```diff
@@ -0,0 +1,152 @@
...
+use Activitypub\Blocks;
+use Activitypub\Collection\Actors;
+use Activitypub\Collection\Followers;
+
+/* @var array $attributes Block attributes. */
+$attributes = wp_parse_args( $attributes );
+
+/* @var string $content Inner blocks content. */
+$content = $content ?? '';
+if ( empty( $content ) ) {
```
pfefferle:
```suggestion
if ( empty( $content ) ) {
```

why setting to empty and then check for empty?

src/followers/render.php
```diff
@@ -0,0 +1,152 @@
...
+$_per_page     = absint( $attributes['per_page'] );
+$follower_data = Followers::get_followers_with_count( $followee_user_id, $_per_page );
+
+// Prepare Followers data for the Interactivity API context.
+$followers = array_map(
+	function ( $follower ) {
+		return array(
+			'handle' => '@' . $follower->get_preferred_username(),
+			'icon'   => $follower->get_icon(),
+			'name'   => $follower->get_name(),
```
pfefferle, line 49: this might be empty... should we fall back to get_preferred_username?

pfefferle:
```suggestion
			'name'   => $follower->get_name() ?? $follower->get_preferred_username(),
```
this might be empty... should we fall back to get_preferred_username?

src/followers/edit.js
obenland: It's [a Block Editor convention to capitalize the innerBlocks template](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/#template). I assume it's meant to convey that it's a constant.

src/followers/render.php
```diff
@@ -0,0 +1,152 @@
...
+			'url'    => $follower->get_url(),
```
pfefferle:
```suggestion
			'url'    => object_to_uri( $follower->get_url() ) ?? $follower->get_id(),
```

this might be empty or an array! we should normalize it and fall back to id!?

src/followers/render.php
obenland: `$content` on its own is an undefined global in the context of `render.php`. The ternary provides a fallback in case it actually is undefined.

src/followers/render.php
pfefferle: needs also to be added in the `use` section

src/followers/render.php
pfefferle: but that the same for `empty` it also checks implicitly for `isset`.

> No warning is generated if the variable does not exist. That means empty() is essentially the concise equivalent to !isset($var) || $var == false. This also applies to nested structures, such as a multidimensional array or chained properties. 

https://www.php.net/manual/en/function.empty.php

src/followers/render.php
```diff
@@ -0,0 +1,152 @@
...
+ * Server-side rendering of the followers block.
+ *
+ * @package Activitypub
+ */
+
+use Activitypub\Blocks;
+use Activitypub\Collection\Actors;
+use Activitypub\Collection\Followers;
+
+/* @var array $attributes Block attributes. */
```
pfefferle:
```suggestion
use Activitypub\Collection\Followers;

use function Activitypub\object_to_uri;

/* @var array $attributes Block attributes. */
```

src/followers/render.php
obenland: There is an else case below, that deals with innerBlocks.

src/followers/render.php
obenland: [5c1adc0](https://github.com/Automattic/wordpress-activitypub/pull/1747/commits/5c1adc0a718a708ada9cf4f3e4128714b78e263b)

src/followers/render.php
obenland: [5c1adc0](https://github.com/Automattic/wordpress-activitypub/pull/1747/commits/5c1adc0a718a708ada9cf4f3e4128714b78e263b)

src/followers/render.php
pfefferle: sure, but the `else` case is only called if `$content` is not empty?

obenland:
@pfefferle I think I got all the changes in

src/followers/render.php
```diff
@@ -13,19 +13,19 @@
...
 	$_title  = $attributes['title'] ?? __( 'Fediverse Followers', 'activitypub' );
 	$content = '<h3 class="wp-block-heading">' . esc_html( $_title ) . '</h3>';
 	unset( $attributes['title'], $attributes['className'] );
 } else {
-	// Remove the outer div if it exists, to avoid nesting issues with block variation classes.
-	$pattern = '/<div\s+class="wp-block-activitypub-followers[^"]*"[^>]*>(.*?)<\/div>/s';
-	if ( preg_match( $pattern, $content, $matches ) ) {
-		$content = $matches[1];
-	}
+	$content = implode( PHP_EOL, wp_list_pluck( $block->parsed_block['innerBlocks'], 'innerHTML' ) );
```
pfefferle, line 28: is it possible that $content is set, but $block not?! In that case it still results in a PHP error/warning!?

src/followers/render.php
```diff
@@ -13,19 +13,19 @@
...
+
+/* @var string $content Block content. */
 $content = $content ?? '';
+
 if ( empty( $content ) ) {
 	// Fallback for v1.0.0 blocks.
 	$_title  = $attributes['title'] ?? __( 'Fediverse Followers', 'activitypub' );
 	$content = '<h3 class="wp-block-heading">' . esc_html( $_title ) . '</h3>';
 	unset( $attributes['title'], $attributes['className'] );
 } else {
```
pfefferle, line 27:
```suggestion
} elseif( $block ) {
```

???

src/followers/render.php
obenland: I don't think so, [they both get passed to the render callback](https://github.com/WordPress/wordpress-develop/blob/1e04fc0864544f5a224dbe4718979367ab2a208c/src/wp-includes/blocks.php#L538-L553).

src/followers/render.php
obenland: It'll always be truthy, see above

pfefferle (approved):

PR #1750, opened 2025-05-29
Blocks: Cleanup after Interactivity moves

obenland:
With the move to the Interactivity API, there's a bit more cleanup we can do, once #1747 is merged:

## Proposed changes:
* Limit `_activityPubOptions` to being enqueued in the Editor only.
* Rearrange `register_block_type_from_metadata()` calls to be in order of folder definition.
* Use `useOptions()` in editor-plugin instead of accessing `_activityPubOptions` directly.
* Move `wp_block` post type check in editor-plugin to the top of the function.

## Testing instructions:

* Once #1747 is merged, all blocks should still work as expected.


pfefferle:
@obenland there are some conflicts!

pfefferle (approved):

PR #1753, opened 2025-05-31
Follow Me: Introduce Author Card style

### LINKED ISSUE #1746: Feature Request: Show follower count (and avatars) in the “Follow on Fediverse” block
### What

Add support for displaying follower-related information in the “Follow on Fediverse” block — such as the number of followers and, optionally, a few follower avatars.

### Why

The current block is static and doesn’t reflect any social activity. Displaying followers would:
	•	Add credibility and social proof
	•	Make the block more visually engaging
	•	Encourage visitors to follow the account
	•	Better reflect the actual fediverse presence of the blog/author

Image

### How

Ideally, this could be implemented as optional settings within the block:
	•	A toggle to show/hide follower count
	•	A toggle to show/hide a few follower avatars

Image

obenland:
I've had this improvement in the back of my mind since working on #1691, #1746 just had good timing :)
This PR changes the Button Only mode from being a toggle attribute to becoming a Block Style and adds an Author Card block style, similar to the author card in the New Follower email.

Fixes #1746.
Follow-up to #1691.

Default | Button | Profile
--- | --- | ---
Screenshot 2025-06-02 at 2 54 32 PM | Screenshot 2025-06-02 at 2 54 37 PM | Screenshot 2025-06-02 at 3 00 43 PM

| Frontend |
| --- |
| Screenshot 2025-06-02 at 3 02 52 PM |

## Proposed changes:

* Adds rest api field with activitypub context for the user endpoint to make it possible to query post count. The `activitypub` context makes it the only value returned from these requests and doesn't add it to any other user API requests.
* Adds setup for block styles in block.json.
* Adds block deprecation for `buttonOnly` attribute to convert to block style. Extends it to v1 deprecation.
* Update Edit component to display all styles and query for additional data, remove Button Only control.
* Updates `render.php` to render the full block. Adds button-only style for blocks that have not been upgraded.
* Updates styles to hide block information based on block style selected.

## Testing instructions:

* Before checking out this branch, create a post with various Follow Me blocks, each with a different configuration.
* Safe the post and checkout this branch.
* Check the editor and frontend view of these blocks and make sure backwards compatibility and block deprecations work as expected.
* Select the profile style for a block and make sure it looks and behaves as expected.
* Select various users and make sure the block updates correctly.


src/follow-me/render.php
```diff
@@ -79,7 +79,12 @@
 if ( empty( $content ) ) {
 	$button_text = $attributes['buttonText'] ?? __( 'Follow', 'activitypub' );
 	$content     = '<div class="wp-block-button"><button class="wp-block-button__link wp-element-button">' . esc_html( $button_text ) . '</button></div>';
+} else {
+	/* @var \WP_Block $block Parsed block.*/
+	$block   = $block ?? null;
+	$content = implode( PHP_EOL, wp_list_pluck( $block->parsed_block['innerBlocks'], 'innerHTML' ) );
```
pfefferle, line 84: isn't this (`$block->parsed_block['innerBlocks']`) failing if `$block` is `null`?

src/follow-me/render.php
```diff
@@ -92,25 +97,83 @@
...
 				alt="<?php echo esc_attr( $actor->get_name() ); ?>"
 			/>
+
 			<div class="activitypub-profile__content">
-				<div class="activitypub-profile__name"><?php echo esc_html( $actor->get_name() ); ?></div>
-				<div class="activitypub-profile__handle"><?php echo esc_html( '@' . $actor->get_webfinger() ); ?></div>
-			</div>
-		<?php endif; ?>
+				<div class="activitypub-profile__info">
+					<div class="activitypub-profile__name"><?php echo esc_html( $actor->get_name() ); ?></div>
```
pfefferle, line 126:
```suggestion
					<div class="activitypub-profile__name"><?php echo esc_html( $actor->get_name() ?? $actor->get_preferred_username() ); ?></div>
```

src/follow-me/render.php
```diff
@@ -92,25 +97,83 @@
...
 	<div class="activitypub-profile">
-		<?php if ( ! $button_only ) : ?>
+		<?php if ( $has_header ) : ?>
+			<div class="activitypub-profile__header" style="background-image: url('<?php echo esc_url( $header_image['url'] ); ?>');"></div>
+		<?php endif; ?>
+
+		<div class="activitypub-profile__body">
 			<img
 				class="activitypub-profile__avatar"
 				src="<?php echo esc_url( $actor->get_icon()['url'] ); ?>"
```
pfefferle, line 120: what if $actor->get_icon() is returning null? (I know that you have not added that, but we might want to handle it anyways!?!)

src/follow-me/render.php
```diff
@@ -92,25 +97,83 @@
 	)
 );
 
+$header_image = $actor->get_image();
+$has_header   = ! empty( $header_image['url'] ) && str_contains( $attributes['className'] ?? '', 'is-style-profile' );
+
+$stats = array(
+	'posts'     => count_user_posts( $user_id, 'post', true ),
+	'followers' => Followers::count_followers( $user_id ),
+	'following' => null,
```
pfefferle:
```suggestion
	'following' => 0,
```

why not always show it?

pfefferle:
@obenland working with JS ;) this has also merge conflicts

src/follow-me/render.php
obenland: Then I'd consider it a bug that should be fixed.

src/follow-me/render.php
obenland: I'm not aware of a case where `get_name()` returns null?

src/follow-me/render.php
obenland: I figured we leave it off while users can't follow others. It would always be 0.

src/follow-me/render.php
obenland: `$block` [should never be undefined or null](https://github.com/WordPress/wordpress-develop/blob/1e04fc0864544f5a224dbe4718979367ab2a208c/src/wp-includes/blocks.php#L538-L553).

src/follow-me/render.php
pfefferle: It is an optional field, so if you do not set it on a platform, it might be empty. See my pixelfed account for example: https://pixelfed.social/users/pfefferle

```
curl "https://pixelfed.social/users/pfefferle" \
     -H 'Accept: application/activity+json'
```

https://browser.pub/https%3A%2F%2Fpixelfed.social%2Fpfefferle

src/follow-me/render.php
pfefferle: why not removing it completely then?

src/follow-me/render.php
obenland: These should all be internal users, no?

pfefferle: oh man!!! too many PRs!!!

src/follow-me/render.php
obenland: Fixed in [767e94a](https://github.com/Automattic/wordpress-activitypub/pull/1753/commits/767e94ad28ae9e99e2e635d22498e8c2c564caf2)

src/follow-me/render.php
```diff
@@ -92,25 +96,82 @@
...
+						<div>
+							<?php
+							printf(
+							/* translators: %s: Number of following */
+								esc_html( _n( '%s following', '%s following', (int) $stats['following'], 'activitypub' ) ),
+								'<strong>' . esc_html( number_format_i18n( $stats['following'] ) ) . '</strong>'
+							);
+							?>
+						</div>
+					<?php endif; ?>
```
pfefferle:
```suggestion
```

I think we have to remove this too then?

obenland: Yup, good catch. Fixed in [3c736e3](https://github.com/Automattic/wordpress-activitypub/pull/1753/commits/3c736e391d3532a28fc455cfa2cea02e91b5edec)

build/follow-me/render.php
```diff
@@ -7,29 +7,31 @@
...
+$block = $block ?? null;
+
+/* @var string $content Inner blocks content. */
+$content = $content ?? '';
+
 // Get the user ID from the selected user attribute.
-$selected_user = $attributes['selectedUser'] ?? 'site';
-$user_id       = Blocks::get_user_id( $selected_user );
-$button_only   = $attributes['buttonOnly'] ?? false;
+$user_id = Blocks::get_user_id( $attributes['selectedUser'] ?? 'site' );
```
pfefferle, line 22: I would prefer to be consistant with the rest of the code and use `0` for the site user. We have so many different wordings, that it gets quite complicated to see if it is the same or not. Aside from that, in PHP we call it blog-user. No strong opinion about the name, but it should be consistant.

src/follow-me/render.php
pfefferle: this is also the local actor and not a remote one, so there is always a `url`.

build/follow-me/render.php
pfefferle: Does not have to be fixed as part of this PR though, because it was introduced before.

Created an issue for a follow-up: https://github.com/Automattic/wordpress-activitypub/issues/1762

pfefferle:
The style breaks if you have no description set.

Screenshot 2025-06-04 at 09 18 55

~Update: It seems to look good on some themes, the broken one is "Pulitzer".~
Update: The PHP warning (see screenshot below) seems to fix it on some themes 😂. After regenerating the sources, it looks broken on all tested themes.

~Aside from that, I see a warning when adding the block to the template.~

Screenshot 2025-06-04 at 09 20 42

~Might be a missing re-build of the sources??~

Update: Fixed with re-bilding sources!

pfefferle:
This is not directly something specific for this PR (maybe more a general question for all blocks), but because this block can be very dominant, this is where I noticed it the most: The "rounded corner" and "drop-shadow" look is very specific and while testing with several themes that mostly use a "flat design" it feels a lot like a "foreign object".

Can we maybe re-use the styling of the theme or at least use a more decent style (no shadows or rounded corners)?

I think this also mainly applies to the reply-context/embed.

obenland:
> Can we maybe re-use the styling of the theme or at least use a more decent style (no shadows or rounded corners)?

Hm, it's pretty much what this PR is about, to introduce this style of an author card that can be embedded anywhere in the site. It's an extension of the Card style in the Followers block, the onboarding checklist, the author card in the New Follower email, the post preview in the DM and mentions email, and the oembed fallback for Fediverse content.

Dropping the border styling makes the Profile style look quite poor IMO, particularly for profiles with header images.

pfefferle:
we can add a thin border!?

there is no way to adapt theme styles?

pfefferle:
I see the same issue for the preview/embed block. I think the email is different, because it does not integrate in a broader context there.

obenland:
[6da39fe](https://github.com/Automattic/wordpress-activitypub/pull/1753/commits/6da39fe5d14f9854a288163f21c9fe14b1b4a8c8) might be the way to have our cake and eat it, too. With the border and shadow controls it'll be possible to completely remove or replace the box shadow and rounded corners, providing a consistent default that's completely customizable:

https://github.com/user-attachments/assets/734f5688-0097-40d3-aaae-14fd7232153c

pfefferle (approved):
NICE! ♥️

openforfuture:
Awesome! 😻 Pretty close to my mockup (https://github.com/Automattic/wordpress-activitypub/issues/1746)!

image

It would be even better to have options to:
	•	Hide the background and description, to make the card more compact.
	•	Show avatars, similar to the Fediverse reactions block.

PR #1755, opened 2025-06-02
Reactions: Don't attempt to load data in Site Editor

obenland:
Follow-up to #1722.

## Proposed changes:

* Avoids querying for reactions in the Site Editor when we're not in a post context.

## Testing instructions:

* Go to a post with the Reactions block and make sure it either shows the reactions to the post or the fallback data.
* Open the Site Editor and select the Single Post template.
* In the Dev Console, make sure there are no API requests to the Reactions endpoint.


pfefferle (approved):
seems to work 🙄

PR #1759, opened 2025-06-03
Rework Follower Structure

pfefferle:
Rework follower structure to enable reuse of the pattern for the following mechanism.

This is the first major change to the follower structure, enabling the unified storage of "Remote Actors" — such as followers, followings, and mentioned actors. The first step is to migrate all generic functionality to the Actors collection and use the Followers collection solely to map an actor to a local user.

The next steps will be (further PRs):

* Deprecate the Follower model in favor of the generic Actor model
* Establish class structure for Followings
* Add an internal API for managing Followings

## Proposed changes:

* Move Actor handling to the Actors collection to enable a unified structure for followers and followings
* Avoid transforming post-type objects into Actor objects for internal operations to reduce resource usage and simplify handling
* Update tests

### Other information:

## Testing instructions:

* No user facing changes
* See tests

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Reworked follower structure to simplify handling and enable reuse for following mechanism.

</details>


includes/class-scheduler.php
```diff
@@ -142,9 +142,9 @@ public static function unschedule_events_for_item( $outbox_item_id ) {
 	}
 
 	/**
-	 * Update followers.
+	 * Update imported actors.
 	 */
-	public static function update_followers() {
+	public static function update_imported_actors() {
```
obenland: It looks like the action in line 45 of this file needs to be updated to use the new function name.

It's probably fine to replace `update_followers()` outright, unless you'd want to deprecate it first?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+
+		return \get_post( $post_id );
+	}
+
+	/**
+	 * Get a remote Actor object by actor URL (guid).
+	 *
+	 * @param int $actor_id The actor ID.
+	 *
+	 * @return \WP_Post|\WP_Error
```
```suggestion
	 * @return \WP_Post
```
Doesn't look like there's a WP_Error here

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+
+	/**
+	 * Get all Followers that have not been updated for a given time.
+	 *
+	 * @param int $number     Optional. Limits the result. Default 50.
+	 * @param int $older_than Optional. The time in seconds. Default 86400 (1 day).
+	 *
+	 * @return \WP_Post[] The Term list of Actors.
+	 */
+	public static function get_outdated( $number = 50, $older_than = 86400 ) {
```
```suggestion
	 * @param int $older_than Optional. The time in seconds. Default DAY_IN_SECONDS.
	 *
	 * @return \WP_Post[] The list of Actors.
	 */
	public static function get_outdated( $number = 50, $older_than = DAY_IN_SECONDS ) {
```
Term might be confusing? It confused me since I could see anything related to terms in the function :)

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+		return $posts->get_posts();
+	}
+
+	/**
+	 * Convert a Custom-Post-Type input to an Activitypub\Model\Actor.
+	 *
+	 * @param \WP_Post $post The post object.
+	 * @return Actor|false The Actor object or false on failure.
+	 */
+	public static function wp_post_to_actor( $post ) {
```
obenland: What about `init_from_post` or `init_from_wp_post` to keep with `init_from_array`, `init_from_json` etc?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+		/* @var Actor $object Actor object. */
+		$object = Actor::init_from_json( $post->post_content );
+
+		if ( \is_wp_error( $object ) ) {
+			return false;
```
obenland: Returning that error might help with providing context down the line, as to why creating the actor failed

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+	}
+
+	/**
+	 * This function is used to store errors that occur when
+	 * sending an ActivityPub message to a Follower.
+	 *
+	 * The error will be stored in post meta.
+	 *
+	 * @param int   $post_id The ID of the WordPress Custom-Post-Type.
+	 * @param mixed $error   The error message. Can be a string or a WP_Error.
```
```suggestion
	 * @param int              $post_id The ID of the WordPress Custom-Post-Type.
	 * @param string|\WP_Error $error   The error message.
```

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+	 *
+	 * @return int|false The meta ID on success, false on failure.
+	 */
+	public static function track_error( $post_id, $error ) {
```
obenland: How about `add_error()` to stay consistent with other naming patterns?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+
+		global $wpdb;
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
+		$post_id = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT ID FROM $wpdb->posts WHERE guid=%s",
+				esc_sql( $actor_data->get_id() )
+			)
+		);
```
obenland: Could this be replaced with `self::get_remote_by_uri( $actor_data->get_id() )`?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+		$inbox = null;
+		if ( ! empty( $actor_data->get_endpoints()['sharedInbox'] ) ) {
+			$inbox = $actor_data->get_endpoints()['sharedInbox'];
+		} elseif ( ! empty( $actor_data->get_inbox() ) ) {
+			$inbox = $actor_data->get_inbox();
+		}
+
+		$args = array(
+			'guid'         => \esc_url_raw( $actor_data->get_id() ),
+			'post_title'   => \wp_strip_all_tags( \sanitize_text_field( $actor_data->get_name() ?? $actor_data->get_preferred_username() ) ),
```
obenland: Is this level of sanitization necessary? I'm curious if Core's built-in `wp_kses` would be enough here?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+			'post_author'  => 0,
+			'post_type'    => self::POST_TYPE,
+			'post_name'    => \esc_url_raw( $actor_data->get_id() ),
```
obenland: I didn't expect the slug to be set to be a URL. Is this something that needs to be unique in some way? If not, it's probably fine to go with the default?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+			'post_content' => \wp_slash( $actor_data->to_json() ),
+			'post_excerpt' => \sanitize_text_field( \wp_kses( \wp_slash( $actor_data->get_summary() ), 'user_description' ) ),
```
obenland: My understanding for slashing post_content was to prevent bugs with json decoding, why is it necessary for the excerpt?
I think `sanitize_text_field` will strip all tags that `wp_kses` preserved. `wp_kses` is probably enough?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+	/**
+	 * Get a remote Actor object by actor URL (guid).
+	 *
+	 * @param int $actor_id The actor ID.
+	 *
+	 * @return \WP_Post|\WP_Error
+	 */
+	public static function get_remote_by_id( $actor_id ) {
+		return \get_post( $actor_id );
+	}
```
obenland: Is that abstraction necessary? Feels like a detour when I see it used in `Followers::add_follower()` and  `Followers::get_follower()`

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
+		$post_id = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT ID FROM $wpdb->posts WHERE guid=%s AND post_type=%s",
+				esc_sql( $actor_uri ),
+				esc_sql( self::POST_TYPE )
+			)
+		);
+
+		if ( ! $post_id ) {
+			return null;
```
obenland: Let's take the opportunity with this new function and be specific about what we return, so we can avoid conditionals like `if ( ! $actor || is_wp_error( $actor ) )`, and return the WP_Error mentioned in function docs

includes/collection/class-followers.php
```diff
@@ -385,63 +371,25 @@ public static function maybe_add_inboxes_of_blog_user( $json, $actor_id ) {
...
-			),
-		);
-
-		$posts = new \WP_Query( $args );
-		$items = array_map( array( Follower::class, 'init_from_cpt' ), $posts->get_posts() );
-
-		return array_filter( $items );
+		_deprecated_function( __METHOD__, 'unreleased', 'Activitypub\Collection\Actors::get_outdated' );
+		$posts = Actors::get_outdated( $number, $older_than );
+		return \array_map( array( Actors::class, 'wp_post_to_actor' ), $posts );
```
```suggestion

		return \array_map( array( Actors::class, 'wp_post_to_actor' ), $posts );
```

includes/collection/class-followers.php
```diff
@@ -385,63 +371,25 @@ public static function maybe_add_inboxes_of_blog_user( $json, $actor_id ) {
...
-			),
-		);
-
-		$posts = new \WP_Query( $args );
-		$items = array_map( array( Follower::class, 'init_from_cpt' ), $posts->get_posts() );
-
-		return array_filter( $items );
+		_deprecated_function( __METHOD__, 'unreleased', 'Activitypub\Collection\Actors::get_faulty' );
+		$posts = Actors::get_faulty( $number );
+		return \array_map( array( Actors::class, 'wp_post_to_actor' ), $posts );
```
```suggestion

		return \array_map( array( Actors::class, 'wp_post_to_actor' ), $posts );
```

includes/collection/class-followers.php
```diff
@@ -456,22 +404,8 @@ public static function get_faulty_followers( $number = 20 ) {
...
-			);
-		}
-
-		return add_post_meta(
-			$post_id,
-			'_activitypub_errors',
-			$error_message
-		);
+		_deprecated_function( __METHOD__, 'unreleased', 'Activitypub\Collection\Actors::track_error' );
+		return Actors::track_error( $post_id, $error );
```
```suggestion

		return Actors::track_error( $post_id, $error );
```

includes/collection/class-followers.php
```diff
@@ -482,6 +416,7 @@ public static function add_error( $post_id, $error ) {
 	 * @return bool True on success, false on failure.
 	 */
 	public static function clear_errors( $post_id ) {
-		return \delete_post_meta( $post_id, '_activitypub_errors' );
+		_deprecated_function( __METHOD__, 'unreleased', 'Activitypub\Collection\Actors::clear_errors' );
+		return Actors::clear_errors( $post_id );
```
obenland, line 418:
```suggestion

		return Actors::clear_errors( $post_id );
```

includes/functions.php
```diff
@@ -1621,3 +1621,44 @@ function _is_type_of( $data, $types ) {
...
+
+/**
+ * Infer a shortname from the Actor ID or URL. Used only for fallbacks,
+ * we will try to use what's supplied.
+ *
+ * @param string $uri The URI.
+ *
+ * @return string Hopefully the name of the Follower.
+ */
+function extract_name_from_uri( $uri ) {
```
obenland, line 1633: This looks to only be used in `Follower`, any reason it can't stay there?

includes/handler/class-move.php
```diff
@@ -41,64 +42,64 @@ public static function handle_move( $activity ) {
...
+		$verified = self::verify_move( $target_actor_object, $origin_actor_object );
 
 		if ( ! $verified ) {
 			return;
 		}
 
-		$target_follower = Followers::get_follower_by_actor( $target );
-		$origin_follower = Followers::get_follower_by_actor( $origin );
+		$target_actor_cpt = Actors::get_remote_by_uri( $target );
+		$origin_actor_cpt = Actors::get_remote_by_uri( $origin );
```
obenland:
```suggestion
		$target_actor = Actors::get_remote_by_uri( $target );
		$origin_actor = Actors::get_remote_by_uri( $origin );
```
Removing the suffix might help with making it visually easier to read without losing meaning

includes/collection/class-actors.php
```diff
@@ -389,4 +389,255 @@ public static function get_type_by_id( $user_id ) {
...
 	}
+
+	/**
+	 * Create or update a remote Actor (e.g., a follower) as a custom post type.
+	 *
+	 * @param array|Actor $actor_data The ActivityPub actor object as associative array (must include 'id').
+	 *
+	 * @return int|\WP_Error The post ID or WP_Error.
+	 */
+	public static function add( $actor_data ) {
```
obenland: I wonder if it would be clearer to split this function into `create()` (or `insert()`) and `update()`? 

`add()`, to me, would be used more in the context of adding to something already existing, like [`WP_Error::add()`](https://developer.wordpress.org/reference/classes/wp_error/add/).

obenland (commented):
Not sure how much you'd like to file on getting the API right vs merging and iterating, but these are some things I noticed, or I'd be curious to learn more about:

obenland:
Not quite sure what's going on with this Repo Gardening action, [it seems to be stuck on this run](https://github.com/Automattic/wordpress-activitypub/actions/runs/15445784937). I'm unable to cancel the workflow, but [was able to delete the run from this page](https://github.com/Automattic/wordpress-activitypub/actions/workflows/gardening.yml), so when we're ready to merge this, removing it from there will turn this PR green.

includes/class-scheduler.php
pfefferle: addressed in https://github.com/Automattic/wordpress-activitypub/pull/1759/commits/239e6750d1500e52518fb71f08d1fea7a73408d1

includes/collection/class-actors.php
pfefferle: If we want to change it, I would prefer something like `get_actor`, similar to what we use in the Outbox Collection: https://github.com/Automattic/wordpress-activitypub/blob/fb7045cc4f815c18968ced74c14a8fa751a1ea37/includes/collection/class-outbox.php#L200

The `init_from_*` functions are used to by objects to initialize itself... so this might be confusing if used here!?!

includes/functions.php
pfefferle: This was meant as a first step to deprecate the Follower.

includes/handler/class-move.php
pfefferle: I thought to be consistent with `target_actor_object`!?!

includes/collection/class-actors.php
pfefferle: Addressed in https://github.com/Automattic/wordpress-activitypub/pull/1759/commits/37d061078f14927bc892ad5edbe35573369556ff

includes/collection/class-actors.php
pfefferle: addressed in https://github.com/Automattic/wordpress-activitypub/pull/1759/commits/7743f4aba70ff54cdaed363207fe8f69bd6e674a

includes/collection/class-actors.php
pfefferle: I would prefer to strip all the tags

includes/collection/class-actors.php
pfefferle: It seems to be that core does not properly slash strings. removing `wp_slash` throws phpunit errors because of missing slashes.

includes/collection/class-actors.php
obenland: `get_actor` sounds good!

includes/functions.php
pfefferle: replace with `Webfinger::uri_to_acct`

includes/class-scheduler.php
```diff
@@ -156,27 +156,28 @@ public static function update_followers() {
 		 *
 		 * @param int $number The number of followers to update.
 		 */
-		$number    = apply_filters( 'activitypub_update_followers_number', $number );
-		$followers = Followers::get_outdated_followers( $number );
+		$number = apply_filters( 'activitypub_update_followers_number', $number );
```
obenland: Should we deprecate this filter in favor of `activitypub_update_remote_actors_number`?

includes/class-scheduler.php
```diff
@@ -188,27 +189,27 @@ public static function cleanup_followers() {
 		 *
 		 * @param int $number The number of followers to clean up.
 		 */
-		$number    = apply_filters( 'activitypub_update_followers_number', $number );
-		$followers = Followers::get_faulty_followers( $number );
+		$number = apply_filters( 'activitypub_update_followers_number', $number );
```
obenland: If we decide to deprecate the filter above, we should update this one, too

includes/collection/class-actors.php
```diff
@@ -389,4 +389,247 @@ public static function get_type_by_id( $user_id ) {
...
+				array( 'status' => 400 )
+			);
+		}
+
+		$inbox = null;
+		if ( ! empty( $actor_data->get_endpoints()['sharedInbox'] ) ) {
+			$inbox = $actor_data->get_endpoints()['sharedInbox'];
+		} elseif ( ! empty( $actor_data->get_inbox() ) ) {
+			$inbox = $actor_data->get_inbox();
+		}
```
obenland, line 423: Is there a chance `$inbox ` would still be `null` at this point, and if so, should we bail?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,247 @@ public static function get_type_by_id( $user_id ) {
...
+		/* @var Actor $object Actor object. */
+		$object = Actor::init_from_json( $post->post_content );
+
+		if ( \is_wp_error( $object ) ) {
+			return $object;
+		}
+
+		$object->set_id( $post->guid );
+		$object->set_name( $post->post_title );
+		$object->set_summary( $post->post_excerpt );
```
obenland: Are ID, name, and summary not already part of the actor json?

pfefferle: yes, but that makes the post at least a bit readable (when debugging) and we would not have to always JSON decode the content for simple UIs.

obenland: I guess my question is, aren't these things already set through `$object = Actor::init_from_json( $post->post_content );` a few lines earlier?

includes/collection/class-actors.php
obenland: @pfefferle Any objections to removing that? It feels slightly unexpected to save a url in what it traditionally used as slug.

includes/collection/class-actors.php
```diff
@@ -389,4 +389,292 @@ public static function get_type_by_id( $user_id ) {
...
+	 *
+	 * The error will be stored in post meta.
+	 *
+	 * @param \WP_Post|int     $post  The post object or ID of the WordPress Custom-Post-Type.
+	 * @param string|\WP_Error $error The error message.
+	 *
+	 * @return int|false The meta ID on success, false on failure.
+	 */
+	public static function add_error( $post, $error ) {
+		$post = \get_post( $post );
```
obenland: Ah, I'm sorry, I wasn't specific enough in my communication when we discussed accepting posts and post_ids.

I think it makes sense for functions that expect a full post object or need to work with a full post object, to accept both `WP_Post` and a post_id.

For cases like `add_error()`, `count_errors()`, and `clear_errors()`, where they only need the post_id, it should not be necessary to grab the whole post. Similar to how `add_post_meta()` only accepts a post_id.

Again, I should have been clearer, apologies.

includes/functions.php
obenland: Looks like this is still a todo?

includes/handler/class-delete.php
```diff
@@ -101,7 +101,7 @@ public static function handle_delete( $activity ) {
 	 * @param array $activity The delete activity.
 	 */
 	public static function maybe_delete_follower( $activity ) {
-		$follower = Followers::get_follower_by_actor( $activity['actor'] );
+		$follower = Actors::get_remote_by_uri( $activity['actor'] );
```
obenland, line 104: There's a `$follower->delete();` call a few lines below that should be replaced with a `wp_delete_post()`.

includes/functions.php
pfefferle: Webfinger::uri_to_acct is not really working like it is

includes/handler/class-follow.php
```diff
@@ -51,25 +51,25 @@ public static function handle_follow( $activity ) {
 		$user_id = $user->get__id();
 
 		// Save follower.
-		$follower = Followers::add_follower(
+		$remote_actor = Followers::add_follower(
 			$user_id,
 			$activity['actor']
 		);
```
obenland, line 57: `Followers::add_follower` returns a post_id now, so we should either run it through get_post() here, or have Followers::add_follower return a `WP_Post` object so `$remote_actor->guid` further down still works.

While we're here, make we could add  some WP_Error handling to avoid sending Notifications for followers that couldn't be saved.

includes/handler/class-move.php
```diff
@@ -34,71 +35,71 @@ public static function init() {
...
+				array( 'ID' => sanitize_key( $origin_object->ID ) )
 			);
 
+			\wp_cache_delete( $origin_object->ID, 'posts' );
+
+			Actors::upsert( $target_json );
+
 			// Clear the cache.
-			wp_cache_delete( $origin_id, 'posts' );
+			\wp_cache_delete( $origin_object->ID, 'posts' );
```
obenland: Looks like we're clearing cache twice now, is that on purpose?

includes/collection/class-actors.php
pfefferle: Ah, thought this was a different place in the code... sorry you are right!

includes/model/class-follower.php
```diff
@@ -250,7 +202,7 @@ public function upsert() {
 	 * @see \Activitypub\Rest\Followers::remove_follower()
 	 */
 	public function delete() {
-		wp_delete_post( $this->_id );
+		Actors::delete( $this->_id );
```
obenland: Should this stay being a `wp_delete_post()` call? I get a `Method 'delete' not found in \Activitypub\Collection\Actors` error.

includes/model/class-follower.php
```diff
@@ -42,13 +44,22 @@ class Follower extends Actor {
 	 */
 	protected $_id; // phpcs:ignore PSR2.Classes.PropertyDeclaration.Underscore
 
+	/**
+	 * Constructor.
+	 *
+	 * @deprecated Use Actor instead.
+	 */
+	public function __construct() {
+		\_deprecated_class( __CLASS__, 'unreleased', 'Actor' );
```
obenland: Any objections to passing the full namespace here?

obenland (changes requested):
There are a few undefined method and properties calls that would benefit from another iteration before this gets merged

includes/model/class-follower.php
pfefferle: for the `Actor`? sure!

includes/functions.php
obenland: I worry that moving it here will make it harder to refactor and remove.

obenland (approved):
Let's merge and iterate if needed. Nice work!

PR #1763, opened 2025-06-04
Block editor: ensure meta data is saved before publishing.

jeherve:
## Proposed changes:

This solves federation issues for some posts, on sites using other plugins that may update post meta right before post publication, like Jetpack and its newsletter feature.

In this scenario, we would run into the following error that was blocking federation:

`Publishing failed. Could not update the meta value of activitypub_content_warning in database.`

In order to fix this, we ensure that the content warning meta is properly saved upon publication.

### Other information:

Internal reference: p1749025385229489-slack-C04TJ8P900J

## Testing instructions:

This is better tested on a site where the Jetpack plugin is installed, connected to WordPress.com, and where you've toggled the Newsletter feature on. All WordPress.com simple sites satisfy those requirements.

1. Go to Posts > Add New
2. Write a new post with some ActivityPub content, such as a Federated reply block with any ActivityPub-compatible content like `https://mastodon.social/@pfefferle/114502331086451476`. 
3. Hit Publish
4. Before to confirm publication, check the Newsletter panel, and change the value to "Post only".
5. Hit Publish.
    * The post should be published with no errors.
    * You should see a `Create` event in the outbox.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


src/editor-plugin/plugin.js
```diff
@@ -55,6 +57,11 @@ const EditorPlugin = () => {
 		</Tooltip>
 	);
 
+	const updateContentWarning = ( value ) => {
+		setMeta( { ...meta, activitypub_content_warning: value } );
+		saveEditedEntityRecord( 'postType', postType, postId );
```
jeherve: This is the main change in this PR.

obenland: Looks like it's now saving the post immediately after updating the content warning setting, rather than waiting until the next autosave or publish, where it gets saved together with other changes?

src/editor-plugin/plugin.js
```diff
@@ -21,20 +21,54 @@ const notAllowed = (
...
 
+/**
+ * Custom hook to update metadata in the post editor.
+ *
+ * @param {string} metaKey The key of the metadata to update.
+ * @param {string} postType The type of post to update the metadata for.
+ * @returns {[string, (value: string) => void]} The current value of the metadata and a function to update it.
+ */
+function useSetMeta( metaKey, postType ) {
+	const [ meta, setMeta ] = useEntityProp( 'postType', postType || 'default', 'meta' );
```
obenland: I think the `default` fallback was an AI suggestion that slipped through in #1750, let's not carry that forward?

src/editor-plugin/plugin.js
```diff
@@ -21,20 +21,54 @@ const notAllowed = (
...
+
 /**
  * Editor plugin for ActivityPub settings in the block editor.
  *
  * @returns {JSX.Element|null} The settings panel for ActivityPub or null for sync blocks.
  */
 const EditorPlugin = () => {
-	const postType = useSelect( ( select ) => select( 'core/editor' ).getCurrentPostType(), [] );
-	const [ meta, setMeta ] = useEntityProp( 'postType', postType || 'default', 'meta' );
+	const postType = useSelect( ( select ) => select( editorStore ).getCurrentPostType(), [] );
```
obenland, line 47: There's a `'wp_block' === postType` check on line 66, could that be moves up here to bail early?

src/editor-plugin/plugin.js
```diff
@@ -21,20 +21,54 @@ const notAllowed = (
...
+		setContentWarning( value );
+	};
+
+	const handleMaxImageAttachmentsChange = ( value ) => {
+		setMaxImageAttachments( value );
+	};
+
+	const handleVisibilityChange = ( value ) => {
+		setContentVisibility( value );
+	};
```
obenland, line 67: Can we use the setter functions in the onChange handlers directly? I'm not sure the wrappers are needed? The setters' function names are also less ambiguous.

src/editor-plugin/plugin.js
jeherve: Done in 0e23a29fd1e86fc83c38f5e537d562cdb50abffd

src/editor-plugin/plugin.js
jeherve: Done in df411f0f2f3a1a900ad0fe8858ed62b0fa55c965

obenland:
I did some more debugging last night, after I was finally able to reproduce the bug. It felt odd to me that it was only `activitypub_content_warning` that failed to update, and not all the other meta values that get saved in a similar way, too. I looked at the meta registration and noticed [the sanitation callback for `activitypub_content_warning`](https://github.com/Automattic/wordpress-activitypub/blob/57c0823220d80146bcdd323c91973353594315d4/includes/class-blocks.php#L46-L52) returns two data types, a string when there is a value, otherwise `null`, while `update_metadata` includes [this bit](https://github.com/WordPress/wordpress-develop/blob/6.8.1/src/wp-includes/meta.php#L246-L254) that compares incoming values with existing ones with a strict comparison.
I wasn't able to debug further, but changing the sanitization callback to `sanitize_text_field` as attempted in #1766 fixes that bug for me. Should we try that first?

jeherve:
> changing the sanitization callback to sanitize_text_field as attempted in https://github.com/Automattic/wordpress-activitypub/pull/1766 fixes that bug for me. Should we try that first?

It fixed it for me too. Let's ship it!

I'll close this PR for now.

PR #1765, opened 2025-06-04
Fix #1764

### LINKED ISSUE #1764: Number of attached images
### Quick summary

I saw the news (https://github.com/Automattic/wordpress-activitypub/pull/1607) that from now on, a maximum of four images could be attached to a Mastodon post.
However, since this update, I haven't seen any changes: I still have three images attached, even though my posts contain at least ten. When I preview my post, I also only see three images.
Is the limit set in another configuration parameter ?

Thank you for your work.

### Steps to reproduce

Add 4 or 5 images to a post :-)

An example :

Blog : https://san.heraut.eu/2025/05/31/chroniques-du-gard-12/
Mastodon : 

Image

### Site owner impact

### Severity

Minor

_No response_

_No response_

pfefferle:
The problem with the current implementation of the "default attachment number" is, that the default will not be changed if the user has saved the old default (`3`). ~This PR checks if the plugin runs in `block` mode and ignores the setting then.~ This PR re-adds the "Media attachments" setting, to be able to change the default.

Fixes #1764

## Proposed changes:

* ~Check for blog mode in `get_option` hook (`classic editor` or `block editor`)~
* ~Only use option if blog uses `classic editor` otherwise fall back to default~
* Re-added "Media attachments" to be able to set the default value also for block-editor users

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Allow users of the block editor to set the default number of media attachments.

</details>


tests/includes/class-test-options.php
```diff
@@ -36,4 +36,20 @@ public function test_delete_removes_all_activitypub_options() {
 		$this->assertFalse( \get_option( 'activitypub_test_option_3', false ) );
 		$this->assertEquals( 'value4', \get_option( 'no_activitypub_test_option' ) );
 	}
+
+	/**
+	 * Test that default_max_image_attachments() returns the default value.
+	 */
+	public function test_default_max_image_attachments() {
+		$this->assertEquals( ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS, \get_option( 'activitypub_max_image_attachments', ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS ) );
```
obenland: I'd recommend removing all default parameter to `get_option()` in this test to make it clearer that whatever is tested is not the default.

tests/includes/class-test-options.php
```diff
@@ -36,4 +36,20 @@ public function test_delete_removes_all_activitypub_options() {
...
+
+		\update_option( 'activitypub_max_image_attachments', '10' );
+		$this->assertEquals( '4', \get_option( 'activitypub_max_image_attachments', ACTIVITYPUB_MAX_IMAGE_ATTACHMENTS ) );
```
obenland: I understand how we got here, but reading those two lines is very confusing.

obenland (commented):
The change has the potential to create a similar confusion on sites where the option value is greater than 4, where it'll be reduced back to the default.

Overall, I think we should re-consider the removal of the setting for sites using the Block Editor. While these sites do have a way of specifying the number of attachments on a per-post basis, the setting would enable it across posts and define a general preference.

pfefferle:
> Overall, I think we should re-consider the removal of the setting for sites using the Block Editor. While these sites do have a way of specifying the number of attachments on a per-post basis, the setting would enable it across posts and define a general preference.

Makes sense! I re-enabled the option for all users. I think this is the simplest, most consistent, and most straightforward solution so far.

obenland (approved):
Thank you!

PR #1766, opened 2025-06-04
Blocks: Don't return null for string value

obenland:
Fixes a bug where updating meta values ends up comparing different data types and failing.

## Proposed changes:

* Always save a string for `activitypub_content_warning`, regardless if it's empty or not.

### Other information:

## Testing instructions:
This is better tested on a site where the Jetpack plugin is installed, connected to WordPress.com, and where you've toggled the Newsletter feature on. All WordPress.com simple sites satisfy those requirements.

1. Go to Posts > Add New
2. Write a new post with some ActivityPub content, such as a Federated reply block with any ActivityPub-compatible content like `https://mastodon.social/@pfefferle/114502331086451476`. 
3. Hit Publish
4. Before to confirm publication, check the Newsletter panel, and change the value to "Post only".
5. Hit Publish.
    * The post should be published with no errors.
    * You should see a `Create` event in the outbox.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Avoid type mismatch when updating `activitypub_content_warning` meta values.

</details>


jeherve (approved):
This fixes the issue for me. 🚢

PR #1767, opened 2025-06-04
Blocks: Standardize deprecation files

obenland:
Minor cleanup to how deprecation files are made up across blocks.

## Proposed changes:

* No functional changes.
* Aligns implementation of deprecation files across blocks.


pfefferle (dismissed):
not sure how to test, but seems legit

pfefferle (approved):

PR #1754, opened 2025-06-02
User Option: Limit to users with Activitypub cap

obenland:
See https://github.com/Automattic/wordpress-activitypub/pull/1747#pullrequestreview-2889384710

## Proposed changes:

* Changes user query from authors to anyone with the `activitypub` capability.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Go to the Editor and insert a Follow Me or Followers block.
* In the sidebar, make sure that all users with the `activitypub` capability are listed, even if they have the Subscriber role.

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

-   [ ] Added - for new features
-   [x] Changed - for changes in existing functionality
-   [ ] Deprecated - for soon-to-be removed features
-   [ ] Removed - for now removed features
-   [ ] Fixed - for any bug fixes
-   [ ] Security - in case of vulnerabilities

#### Message

Follow Me and Followers blocks can now list any user that is Activitypub-enabled, even if they have the Subscriber role.

</details>

PR #1777, opened 2025-06-06
Follow Me: Improve MU compatibility

### LINKED ISSUE #1785: Follow me on the Fediverse: missing data in frontend, missing options and 404 error in console in backend
### Quick summary

The public block has missing informations (frontend):
Image

Weirdly in edit mode, it works fine (backend):
Image

...but block config options are missing (backend):
Image

...and in the browser console I’m encountering this error (backend):
`GET https://fedimedia.it/wp-json/wp/v2/users/0/?context=activitypub&_locale=user 404 (Not Found)`
`{"code":"rest_user_invalid_id","message":"ID utente non valido.","data":{"status":404}}`

### Steps to reproduce

1. Put the **Follow me on the Fediverse** block in a page/post
2. Open your browser’s developer console
3. Observe a 404 error from a REST API call to /wp-json/wp/v2/users/0/
4. Publish and check the frontend view
5. Notice that the block may appear empty or incomplete due to this failed request

### Site owner impact

### Severity

Minor

_No response_

_No response_

obenland:
Fixes #1785.

## Proposed changes:

* Adds a deprecation handler to update broken Button blocks in Follow Me blocks when users don't have the `unfiltered_html` capability.
* Updates variable name for post and follower count to avoid a flash of the follower URL when loading followers.
* Removes button `tagName` attribute in favor of the default anchor to account for users that don't have the `unfiltered_html` capability.
* Adds call to nodeInfo endpoint to populate post count for blog user.
* Moves up `userOptions` check to avoid showing an empty options panel when there is only one user.

### Other information:

## Testing instructions:

* Insert a Follow Me block in the editor and save.
* In the frontend, make sure the Follow button still opens the modal.
* Switch to Blog only mode on your site.
* Update the Follow Me block and make sure there is no user selector in the sidebar.
* Double check that it shows the correct amount of posts in the editor and the front end.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Avoids the button block breaking for users that don't have the `unfiltered_html` capability.
Blog users now get their correct post count displayed in the Editor and the front-end.

</details>


obenland:
This is still not working quite right

obenland:
Back in business! The deprecation now converts all buttons to anchors, even if they don't error, and they behave more like buttons in the front-end, including keyboard navigation

pfefferle:
Is this the way to go in terms of accessibility? Do we have to at least add `role="button"`??

obenland:
I did [add it here](https://github.com/Automattic/wordpress-activitypub/pull/1777/files#diff-4265836156b9d4a4d7207eb77bbf63901d8943a5fe3beaffa1a62ebaf68e2c88R95). 

And no, it is not. But it's either turning the button into an anchor, or the block breaking for everyone who doesn't have the `unfiltered_html` cap. Which I'd also be fine with, if that's what we want to do. We could just remove the block from the inserter for everyone who doesn't have the cap.

obenland:
We could also save it as an anchor, then render it as a button in the front-end.

obenland:
> We could also save it as an anchor, then render it as a button in the front-end.

I was hoping this would be an easy task with `WP_HTML_Tag_Processor`, but it turns out it doesn't have a way to replace tag names. Before we do something with regex, maybe the easiest thing is to just always have it be an anchor and keep it predicatable.

pfefferle:
Is it on purpose that users with the role `author` can only add the blog-user and have no possibility to change the user-mode ("Select user") in the editor?

Author Role:

Screenshot 2025-06-09 at 09 52 18

Admin Role:

Screenshot 2025-06-09 at 09 52 33

[Update] I see the following error in the editor when using an "Author":

Screenshot 2025-06-09 at 09 56 14

src/follow-me/edit.js
```diff
@@ -182,23 +183,28 @@ export default function Edit( { attributes, setAttributes, context: { postType,
...
-					} )
-					.catch( () => {} );
+				if ( effectiveUserId ) {
+					apiFetch( { path: `/wp/v2/users/${ effectiveUserId }/?context=activitypub` } )
+						.then( ( { post_count } ) => {
+							setProfile( ( prevProfile ) => ( { ...prevProfile, postsCount: post_count } ) );
+						} )
+						.catch( () => {} );
+				} else {
+					apiFetch( { path: `/${ namespace }/nodeinfo/2.0` } )
```
pfefferle: this is not a relyable endpoint!

the NodeInfo API depends on the "Search engine visibility" setting. If a site-owner decides to hide their site from search engines, the NodeInfo endpoint is not accessible.

Screenshot 2025-06-09 at 10 14 16

obenland: Fixed in [ab7d298](https://github.com/Automattic/wordpress-activitypub/pull/1777/commits/ab7d29836a7efdb8243a3ff2822c3be1ae8edaae). I'm not sure it'll work in Dotcom, but we can figure that out in a follow-up

pfefferle:
@obenland https://github.com/Automattic/wordpress-activitypub/pull/1777#issuecomment-2954985338 ??

obenland:
> @obenland https://github.com/Automattic/wordpress-activitypub/pull/1777#issuecomment-2954985338 ??

It's not a regression of the Follow Me block, but likely of #1754.

pfefferle:
Is it? The user has the `activitypub` capability but can't add the block properly!

obenland:
It doesn't seem to be a regression at all:

image

That's not to say it's ideal or how we want it to work, but it wasn't introduced in 6.0.0. My assumption is that it has to do with authors not being allowed to edit other people's posts.

pfefferle:
I created the post as the Author!

pfefferle:
It is indeed an issue with the query! An author seems to be not allowed to run the capability filter (use the capability query param)!

pfefferle:
So we might need different checks for:

* The user that is allowed to add the blocks
* The users that are allowed to show in the dropdowns

The first one should be editors and above (how it seems). The second one, the users with the `activitypub` capability.

obenland:
I'm happy to look at it once 6.0.1 is out the door

pfefferle (approved):

PR #1799, opened 2025-06-10
Reply: Normalize for Pleroma

obenland:
Webfingers frequently are prefixed with `acct:`. While account for that in siz other places in the plugin, the reply block callback did not. This PR changes that.

## Proposed changes:

* Removes `acct:` prefix from webfinger before using it to create the reply link.
* Adds unit test.

### Other information:

## Testing instructions:

* Create a post with a reply block and reply to a Pleroma post like https://pleroma.lord.re/notice/AuzeVQZGDbAxW5zd0y
* Publish the post and inspect the activitypub version.
* Make sure the mention uses the webfinger correctly.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Webfinger data from Pleroma instances no longer creates unexpected mention markup.

</details>

### DIFF
--- a/tests/includes/transformer/class-test-post.php
+++ b/tests/includes/transformer/class-test-post.php
@@ -657,21 +657,15 @@ class Test_Post extends \WP_UnitTestCase {
 	/**
 	 * Test reply link generation.
 	 *
+	 * Pleroma prepends `acct:` to the webfinger identifier, which we'd want to normalize.
+	 *
+	 * @see https://a8c.slack.com/archives/C04TJ8P900J/p1749577489043549
 	 * @covers ::generate_reply_link
 	 */
 	public function test_generate_reply_link() {
-		// Create a test post of type "Article".
-		$post = self::factory()->post->create_and_get(
-			array(
-				'post_title'   => 'Test Article',
-				'post_content' => '<!-- wp:activitypub/reply {"url":"https://devs.live/notice/AQ8N0Xl57y8bUQAb6e"} /-->',
-				'post_status'  => 'publish',
-			)
-		);
-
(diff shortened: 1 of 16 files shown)


pfefferle:
Hmmm... I am not sure about that... technically WebFinger requires a scheme and if it is the email like identifier (WebFinger also supports URLs and URIs) this is the `acct:`... So normally `acct:` should be the default...

That is why a normalization that removed it feels semantically incorrect!?

pfefferle:
I would prefer a function to remove the scheme or leading `@` instead of filtering and therefore  manipulate the WebFinger response globally!?

pfefferle:
Or at least a sanitizer that is called in that specific usecase only?

obenland:
> Or at least a sanitizer that is called in that specific usecase only?

Let's go with that one for now

jeherve (dismissed):
That fixes things for me, thank you!

tests/includes/transformer/class-test-post.php
```diff
@@ -654,4 +653,51 @@ public function test_preview_property() {
 		// Check if the preview for a Note is null.
 		$this->assertNull( $note_preview );
 	}
+
+	/**
+	 * Test reply link generation.
+	 *
+	 * Pleroma prepends `acct:` to the webfinger identifier, which we'd want to normalize.
+	 *
+	 * @see https://a8c.slack.com/archives/C04TJ8P900J/p1749577489043549
```
pfefferle: Should we reference internal chats?

pfefferle (dismissed):

PR #1812, opened 2025-06-13
Avatars: Update filters, fallbacks, and API use

obenland:
I was looking for a way to photonize avatar urls on WP.com for the Reactions block and realized that there's no way to filter the avatar url. This PR makes it possible to filter the URL and updates all other direct calls to comment meta to get the avatar url.

## Proposed changes:

* Uses `get_avatar_url()` instead of getting `avatar_url` comment meta directly.
* Adds filter docs for `get_avatar_comment_types` filter.
* Adds fallback for empty comment types instead of failing them.
* Removes `get_avatar_data` filter call to avoid double-filtering the value.
* Adds `get_avatar_url` filter to `avatar_url` comment meta.

### Other information:

## Testing instructions:

* Find a post with Reactions and make sure avatars are the same before and after this change.
* Edit that post and insert a reactions block if it doesn't already have one. Make sure avatars look right.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Unified retrieval of comment avatars and re-used core filters to give access to third-part plugins.

</details>


includes/class-activitypub.php
```diff
@@ -332,18 +332,15 @@ public static function pre_get_avatar_data( $args, $id_or_email ) {
...
-		if (
-			! empty( $id_or_email->comment_type ) &&
-			! \in_array(
-				$id_or_email->comment_type,
-				(array) $allowed_comment_types,
-				true
-			)
-		) {
+		if ( ! \in_array( $id_or_email->comment_type ?: 'comment', $allowed_comment_types, true ) ) { // phpcs:ignore Universal.Operators.DisallowShortTernary
 			$args['url'] = false;
```
obenland: @pfefferle I think I'm missing historic context, hoping for you to be able to fill in my blanks.

My understanding is that setting the `url` will [short-circuit](https://github.com/WordPress/wordpress-develop/blob/a49176e653d53c273ae7a94c0791da2befabb439/src/wp-includes/link-template.php#L4471-L4473) `get_avatar_data()` for all cases where the comment type is not part of the `$allowed_comment_types` array, so things like pings and trackbacks and such. Is that on purpose?

includes/class-activitypub.php
```diff
@@ -332,18 +332,15 @@ public static function pre_get_avatar_data( $args, $id_or_email ) {
...
 		$allowed_comment_types = \apply_filters( 'get_avatar_comment_types', array( 'comment' ) );
-		if (
-			! empty( $id_or_email->comment_type ) &&
-			! \in_array(
-				$id_or_email->comment_type,
-				(array) $allowed_comment_types,
-				true
-			)
-		) {
+		if ( ! \in_array( $id_or_email->comment_type ?: 'comment', $allowed_comment_types, true ) ) { // phpcs:ignore Universal.Operators.DisallowShortTernary
```
obenland, line 341: For context: Previously this conditional failed for an empty `comment_type`, which used to be the default for comments before WP 5.5. We no longer support WP versions that old, but I wonder if we should still just fallback to `comment`, instead of bailing and disabling avatars.

includes/class-activitypub.php
pfefferle: exactly this

pfefferle (dismissed):

includes/class-activitypub.php
obenland: Could you help me unpack that a bit more? What's the purpose for disabling avatars for all other comment types?

pfefferle: they simply do not have avatars!?

includes/class-activitypub.php
pfefferle: happy to remove the fallback!

includes/class-activitypub.php
pfefferle: but I have no strong feelings about that... happy to remove it!

pfefferle (approved):

PR #1814, opened 2025-06-13
Signature: Accept signatures with hs2019 algorithm

obenland:
Follow-up to #1803 and depends on it for all tests to pass.

## Proposed changes:

* Update `get_signature_algorithm` to check for `hs2019` to identify algorithm.
* Add tests.

### Other information:

## Testing instructions:

* The unit test should have it covered? 🤞 

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

HTTP signatures using the hs2019 algorithm now get accepted without error.

</details>


obenland:
@mediaformat It looks like I can't add you to the list of reviewers—would you still be able to give this a look and share any feedback you have?

includes/class-signature.php
```diff
@@ -356,8 +356,8 @@ public static function get_remote_key( $key_id ) {
 	public static function get_signature_algorithm( $signature_block ) {
 		if ( $signature_block['algorithm'] ) {
 			switch ( $signature_block['algorithm'] ) {
-				case 'rsa-sha-512':
-					return 'sha512'; // hs2019 https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12.
+				case 'hs2019':
```
pfefferle, line 359:
```suggestion
				case 'hs2019':
				case 'rsa-sha512':
```

let's keep it for backwards compatibility!

obenland: I'm not sure it's compatible with anything. Is that string used anywhere?

pfefferle: I have no idea, but I would fear so, because there was no error report since nearly two years. Aside from that: technically it is a major change if we remove it, because even if it might be a bug, we break backwards compatibility.

mediaformat: it looks like the original implementation made some falsey assumptions. As @obenland said, I'm not sure any implementations declare `rsa-sha-512` as an algorithm.

pfefferle (approved):

PR #1820, opened 2025-06-16
Welcome Page: Fix missing CSS when used as fallback

pfefferle:
While testing #1802 and #1806 I experienced a bug, where the CSS was not shown for the Welcome page when used as fallback.

When the plugin is configured to use only `Actors` (no blog user), the plugin loads the Welcome page as fallback if you try to load `/wp-admin/options-general.php?page=activitypub&tab=blog-profile` (same with other potentially deactivated settings pages). In this case the `tab` query param is `blog-profile` but it loads the welcome screen and because we have this check in place:

`if ( 'welcome' === ( isset( $_GET['tab'] ) ? \sanitize_key( $_GET['tab'] ) : 'welcome' ) )`

the welcome screen will be loaded without CSS.

Screenshot 2025-06-16 at 10 18 26

## Proposed changes:

* Check for loaded template file instead of tab, to show the CSS

### Other information:

## Testing instructions:

* Checkout trunk
* Change User-Mode to "Actors only"
* Load `/wp-admin/options-general.php?page=activitypub&tab=blog-profile`
* You will see the page with the missing CSS
* Checkout this PR and re-do all the steps above

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

The Welcome screen now loads with proper styling when shown as a fallback.

</details>

### DIFF
--- a/.github/changelog/1820-from-description
+++ b/.github/changelog/1820-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+The Welcome screen now loads with proper styling when shown as a fallback.
--- a/includes/compat.php
+++ b/includes/compat.php
@@ -25,6 +25,23 @@ if ( ! function_exists( 'str_starts_with' ) ) {
 	}
 }
 
+if ( ! function_exists( 'str_ends_with' ) ) {
+	/**
+	 * Polyfill for `str_ends_with()` function added in PHP 8.0.
+	 *
+	 * Performs a case-sensitive check indicating if
+	 * the haystack ends with needle.
+	 *
+	 * @param string $haystack The string to search in.
+	 * @param string $needle   The substring to search for in the `$haystack`.
+	 *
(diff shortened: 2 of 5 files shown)


obenland:
Oh, nice catch! I saw that once in my testing and couldn't figure out what happened.

How would you feel about enqueueing it from the template itself, [like we do with embeds](https://github.com/Automattic/wordpress-activitypub/blob/trunk/templates/embed.php#L28)? Could help avoid a bit of logic

pfefferle:
no real preference, so happy to move it to the template.

pfefferle:
@obenland done

obenland (approved):
Do you want to keep the `str_ends_with()` stuff? It doesn't seem to be needed any longer

pfefferle:
I would love to keep it and simplify other parts of the code to use that instead!

PR #1821, opened 2025-06-16
Do not store `activitypub_max_image_attachments` if it uses the default value

pfefferle:
Do not store the `activitypub_max_image_attachments` Post-Meta, if it is the same as the default, to not spam the DB.

## Proposed changes:

* Checks if the `activitypub_max_image_attachments` setting from the block editor has the same value as the default, and remove it if so.

### Other information:

## Testing instructions:

* See unittest

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

The image attachment setting is no longer saved to the database if it matches the default value.

</details>

### DIFF
--- a/.github/changelog/1821-from-description
+++ b/.github/changelog/1821-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+The image attachment setting is no longer saved to the database if it matches the default value.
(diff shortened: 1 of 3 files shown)


obenland (approved):
This should be good to go as-is.

The function doc is currently specific to `activitypub_content_visibility`, so that could be generalized, but I'm not sure I'd consider it a blocker.

Potential follow-ups:
* We could consider giving `activitypub_content_warning` a similar treatment.
* I don't know if it's possible to move this to the `update_post_metadata` hook and avoid the step of updating the value?
* Should all of this also happen when a meta value gets added?

pfefferle:
I will work on a follow up! I think it should work for both!

PR #1826, opened 2025-06-17
Reactions: Unlock customization options

obenland:
While prepping a post about the block updates in 6.0.0 I found a bug and then dove down a short customization rabbit hole.

image

### The bug

https://github.com/user-attachments/assets/5d6b06d4-7692-44a4-a4ef-e46067b7874c

While editing a post on a site with the Twenty Twenty-Five theme active, I noticed that the alignment options did not work and there was an additional setting in the sidebar that was not available in other themes I tested it. The setting's default seems to disable the alignment feature as it is.

To fix it, I set the `allowEditing` flag in `block.json` to false.

### The Additions

While working in 6.0.0, we had discussed giving all blocks the same customization tools, so while I was in there I added the same customization options as the Follow Me block already has and updated the styles to work with them.

## Proposed changes:

* Fixes a bug where Layout editing UI was shown in block themes that support them.
* Adds customization support to bring en par with Follow Me block.

## Testing instructions:

* Insert a Reactions block into a post.
* Play with the customization options in the sidebar.
* Make sure there are no unexpected effects and the front-end reflects the settings.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Reactions block now supports customizing colors, borders, box-shadows, and typography.

</details>


pfefferle:
@obenland and you really thought **that** was the best color choice to showcase this new feature??? 😂

pfefferle:
It is hard to understand what the problem was, how it was fixed and what was added.

obenland:
Thanks, that's great feedback! I updated the issue description with more details and a video showing the bug.

pfefferle (approved):

Issue #1808, opened 2025-06-12
Feature Request: Upgrade HTTP Message signatures to implement RFC-9421


mediaformat:
### What

Following a survey of HTTP Signature support, the SWICG is recommending an upgrading path from the previous [draft-cavage-http-signatures-12](https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12) to [RFC9421](https://www.rfc-editor.org/rfc/rfc9421) 

### Why

Compatibility
> The HTTP Signatures standard has made a few backward-incompatible changes on its path to becoming a full Proposed Standard, [[RFC9421](https://swicg.github.io/activitypub-http-signature/#bib-rfc9421)]. Many fediverse servers currently handle older versions of the standard and aren't yet compatible with the final version. Here's advice on how to implement HTTP Signatures so as to be compatible with as many different servers as possible.

### How

Use the double knock technique
> The primary technique we recommend is double-knocking. First, try generating or verifying an HTTP Signature with one version, ideally (but not necessarily) the latest. If the remote server rejects that signature, eg with an HTTP 401 response, or the incoming signature doesn't verify, try with another version. Repeat until a signature passes or you've tried all supported versions. 
— [How to upgrade supported versions](https://swicg.github.io/activitypub-http-signature/#how-to-upgrade-supported-versions)

An existing [PHP implementation](https://github.com/macgirvin/HTTP-Message-Signer) has been developed by Mike MacGirvin, which could be included or used as inspiration (since there is a Guzzle dependency)

obenland:
Interesting! Yes, the PHP implementation might need some adjustment, WordPress still supports PHP 7.2.
Does the library have double-knock built-in or would that need to be built on top of it?

I'm afraid I'm far from being an expert on signatures—would this work be something you'd be interested in leading?

pfefferle:
Aside from the  is the min PHP requirement of 8.1, I see the PSR dependency as a quite bigger issue. There are quite a few tools to downgrade PHP dependencies, but it is not that easy to rewrite the code to support the WordPress request object and the plain PHP one. This is quite a similar problem that spoke against using the ActivityPub php library from @landrok

pfefferle:
My thoughts from 2022 https://notiz.blog/2022/05/15/wpsr/ (in German)

Jiwoon-Kim:
[...]

PR #1832, opened 2025-06-18
Signature: Make key management Actor-specific

obenland:
Minor house keeping item in preparation for #1808, separating the key management bits out from the signature functionality itself. I'm hoping it'll make it easier to update the Signature class in the future.

## Proposed changes:

* Moves key management functions to `Actors` class.
* Updates all references to those functions.
* Deprecates public functions in Signature class.
* Moves unit tests.

### Other information:

## Testing instructions:

* Subscribe to your test site from the Fediverse.
* Publish a new post.
* Make sure it still gets delivered to your subscriber and signature management still works.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Management of public/private keys for Actors now lives in the Actors collection, in preparation for Signature improvements down the line.

</details>


obenland:
@pfefferle Is `Actors` the best place for this?

includes/collection/class-actors.php
```diff
@@ -389,4 +389,162 @@ public static function get_type_by_id( $user_id ) {
...
+			);
+		}
+
+		$key_pair = array(
+			'private_key' => $private_key,
+			'public_key'  => $detail['key'],
+		);
+
+		// Persist keys.
+		\add_option( $option_key, $key_pair );
```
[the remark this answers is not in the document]
obenland: I was actually wondering about that, too. Is there ever a time where we'd want to change the keys?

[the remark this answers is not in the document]
pfefferle: There is a recommendation to do so, but I am not sure how well supported it is yet https://swicg.github.io/activitypub-http-signature/#key-rotation

pfefferle:
> Is Actors the best place for this?

I would think so!

Maybe we can unify the `get_public_key` functions a bit to allow a similar functionality for remote users?

Either in the same function:
* if `$id` is an URI, check for the public key in the `ap_actor` `post_content`
* if it is an ID, use the current function

Or a dedicated `get_remote_public_key`!?

includes/class-signature.php
```diff
@@ -194,7 +81,7 @@ protected static function check_legacy_key_pair_for( $user_id ) {
 	 */
 	public static function generate_signature( $user_id, $http_method, $url, $date, $digest = null ) {
```
pfefferle: it would be awesome to pass the private key as a param instead of the `$user_id`, to be completely user agnostic.

obenland: Agreed!

obenland:
> Maybe we can unify the get_public_key functions a bit to allow a similar functionality for remote users?

I was wondering about that, too, but couldn't think of a use case. To verify signatures on incoming requests I think we should always make remote requests to the `keyId` provided in the signature itself, even if it's from an actor in the db. I wouldn't want to rely on a cached version.

includes/class-signature.php
obenland: Fixed in [2a0373c](https://github.com/Automattic/wordpress-activitypub/pull/1832/commits/2a0373cf689d6275a510547547b71038bb54fe5a). This turned out to be a bit more invasive than I hoped, since there are two data points dependent on the user.

Long term, I wonder if we should move Signature from being static to being an object that we'd instantiate with the user stuff and then call `generate_signature()` with just the request information

includes/class-signature.php
```diff
@@ -24,178 +24,63 @@ class Signature {
...
 	 * @param string $date        The date the request is sent.
 	 * @param string $digest      Optional. The digest of the request body. Default null.
 	 *
 	 * @return string The signature.
 	 */
-	public static function generate_signature( $user_id, $http_method, $url, $date, $digest = null ) {
-		$user = Actors::get_by_id( $user_id );
-		$key  = self::get_private_key_for( $user->get__id() );
-
+	public static function generate_signature( $key_id, $private_key, $http_method, $url, $date, $digest = null ) {
```
obenland: Heads up that replacing `$user_id` with the required values changes the function signature.

includes/class-signature.php
```diff
@@ -220,15 +105,13 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
...
 		$signature = null;
-		\openssl_sign( $signed_string, $signature, $key, \OPENSSL_ALGO_SHA256 );
+		\openssl_sign( $signed_string, $signature, $private_key, \OPENSSL_ALGO_SHA256 );
 		$signature = \base64_encode( $signature ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
 
-		$key_id = $user->get_id() . '#main-key';
-
 		if ( ! empty( $digest ) ) {
-			return \sprintf( 'keyId="%s",algorithm="rsa-sha256",headers="(request-target) host date digest",signature="%s"', $key_id, $signature );
+			return \sprintf( 'keyId="%s",algorithm="rsa-sha256",headers="(request-target) host date digest",signature="%s"', $key_id . '#main-key', $signature );
```
pfefferle: I think the `#main-key` hash should be part of the key-id

obenland: Fixed in [487f284](https://github.com/Automattic/wordpress-activitypub/pull/1832/commits/487f28436fb7af0ba31275547dc5074157297c22).

pfefferle (dismissed):

includes/class-signature.php
pfefferle: I have indeed also thought about that... I think the http class would also benefit from beeing an object instead of a static class...

obenland:
@pfefferle Third time's the charm!

includes/model/class-application.php
```diff
@@ -219,7 +217,7 @@ public function get_public_key() {
 		return array(
 			'id'           => $this->get_id() . '#main-key',
```
pfefferle, line 218: shouldn't we use a function that generates the signature ID, to be sure that it is consistent?

pfefferle: (but this might be a different PR)

pfefferle (approved):

includes/model/class-application.php
obenland: Honestly, I'm not even sure we need a "separate" signature ID from the actor ID. The hash doesn't seem to serve a purpose for as long as we only support one public key.

PR #1835, opened 2025-06-18
Reactions: Fix avatar display bugs

obenland:
Fixes margins between avatars and details button, when there are hidden avatars:
Before | After
--- | ---
Screenshot 2025-06-18 at 2 56 47 PM | Screenshot 2025-06-18 at 2 56 34 PM

Fixes hidden avatars when there are fewer Likes than reposts:
Before | After
--- | ---
Screenshot 2025-06-18 at 2 28 21 PM | Screenshot 2025-06-18 at 3 01 37 PM

## Proposed changes:

* Adjust styles to maintain consistent margin between avatars and detail button.
* Add comment type context to decide how many avatars to show.

## Testing instructions:

* On a test post with a different amount of Likes and Reposts, make sure all avatars are displayed as expected.
* Make sure the distance between avatars and button is consistent.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

The amount of avatars shown in the Reactions block no longer depends on the amount of likes, but is comment type agnostic.

</details>


src/reactions/view.js
```diff
@@ -87,7 +87,7 @@ const { callbacks, state } = store( 'activitypub/reactions', {
 				}
 
 				getElement()
-					.ref.querySelectorAll( '.reaction-group' )
+					.ref.querySelectorAll( `.reaction-group[data-reaction-type="${ reactionType }"]` )
```
pfefferle, line 90: do these quotation marks work properly 😳

src/reactions/style.scss
```diff
@@ -49,7 +49,8 @@
 				margin: 0 -10px 0 0;
 				transition: transform 0.2s ease;
 
-				&:last-child {
+				// Remove margin on avatar that doesn't have a hidden sibling.
+				&:not([hidden]):not(:has(~ li:not([hidden]))) {
```
pfefferle, line 53: ...and they say html and css are not programming languages 😄

src/reactions/view.js
obenland: In my testing they did. What's the concern?

src/reactions/style.scss
obenland: Haha, I know, right?! That's why I added the comment, there's no way I remember what that's supposed to do in a month's time.

src/reactions/view.js
pfefferle: that they won't work 😂

(simply haven't seen them in use for JS, CSS or HTML)

pfefferle (approved):

PR #1837, opened 2025-06-19
Signature: Enable digest check for REST Requests

obenland:
The digest check was essentially skipped, even if a digest header was available, since `$body` was never set.
I suppose there's a good chance that requests might fail now that didn't use to fail? Not sure.

## Proposed changes:

* Enables Digest check for `WP_REST_Request`s.
* Uses signed headers directly instead of in a separate variable.
* Breaks down digest in separate variables.
* Adds tests.

### Other information:

## Testing instructions:

* Apply this PR to a test site that can receive requests from the internet.
* Send a request that includes a digest, like a like from Mastodon maybe? 
* Make sure that still works.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

When verifying signatures on incoming requests, the digest header now gets checked as expected.

</details>


includes/class-signature.php
```diff
@@ -287,17 +285,15 @@ public static function verify_http_signature( $request ) {
 			return new WP_Error( 'activitypub_signature', __( 'Unsupported signature algorithm (only rsa-sha256 and hs2019 are supported)', 'activitypub' ), array( 'status' => 401 ) );
 		}
 
-		if ( \in_array( 'digest', $signed_headers, true ) && isset( $body ) ) {
+		if ( \in_array( 'digest', $signature_block['headers'], true ) && $request instanceof \WP_REST_Request ) {
```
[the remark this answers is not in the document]
pfefferle: with `AUTHORIZED_FETCH` we also have to verify signatures on profile pages and posts, so we should keep the vanilly request handling.

as an alternative, we could try to generate a `WP_REST_Request` for the requests mentioned above!?

includes/class-signature.php
```diff
@@ -287,17 +287,15 @@ public static function verify_http_signature( $request ) {
 			return new WP_Error( 'activitypub_signature', __( 'Unsupported signature algorithm (only rsa-sha256 and hs2019 are supported)', 'activitypub' ), array( 'status' => 401 ) );
 		}
 
-		if ( \in_array( 'digest', $signed_headers, true ) && isset( $body ) ) {
+		if ( \in_array( 'digest', $signature_block['headers'], true ) && isset( $body ) ) {
```
pfefferle, line 290:
```suggestion
		if ( \in_array( 'digest', $signature_block['headers'], true ) && ! empty( $body ) ) {
```

pfefferle: No need to proceed if the body is empty!?

obenland: If there's a digest in the signature we should always process it, even if the body is empty

pfefferle (approved):

PR #1838, opened 2025-06-19
Signature: Add created and expires headers

obenland:
Not sure yet whether that is a good idea. It's possible we'd have to add double-knocking to even have this work?

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -212,11 +212,15 @@ class Signature {
 		}
 
 		$http_method = \strtolower( $http_method );
+		$created     = \strtotime( $date );
+		$expires     = $created + HOUR_IN_SECONDS;
 
 		if ( ! empty( $digest ) ) {
-			$signed_string = "(request-target): $http_method $path\nhost: $host\ndate: $date\ndigest: $digest";
+			$signed_string = "(request-target): $http_method $path\nhost: $host\ndate: $date\ndigest: $digest\n(created): $created\n(expires): $expires";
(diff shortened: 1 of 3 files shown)


includes/class-signature.php
```diff
@@ -225,11 +229,14 @@ public static function generate_signature( $user_id, $http_method, $url, $date,
 
 		$key_id = $user->get_id() . '#main-key';
 
-		if ( ! empty( $digest ) ) {
-			return \sprintf( 'keyId="%s",algorithm="rsa-sha256",headers="(request-target) host date digest",signature="%s"', $key_id, $signature );
```
pfefferle, line 229: why removing the digest?

pfefferle:
I am not sure about the benefit!?!

pfefferle:
Is it a good idea to use `hs2019`? What I understand in the discussions to the signature PRs is, that this is not properly supported and differently implemented yet?

The whole signature part is very crucial and might break interoperability, so I would be very conservative with it!

Otherwise we would have to test it with a lot of platforms!

obenland:
Agreed, like I said in the PR description, Not sure yet whether that is a good idea.

I think there's an opportunity to simplify `generate_signature`, but that can be done when we work on that part for RFC-9421.

Issue #1036, opened 2024-12-03
Feature Request: Follow users on the Fediverse


hayo:
### What

Follow users on mastodon.

### Why

makes relationships reciprocal, adding a layer of interaction.

### How

This is not as easy, as shown by how badly the friends plugin works.  nevertheless i envisage a simple implementation where you follow a person, and this leads to an overview page showing the relationships in admin, with a rss feed link next to each relationship so you can add them to your feedreader. this negates the need for complicated/expensive operations to keep pushing new posts somewhere. Until someone finds a clever way to do that.

pfefferle:
keep!

pfefferle:
This is now possible: https://activitypub.blog/2025/10/01/7-5-0-follow-the-feed-quote-the-lead/

PR #1839, opened 2025-06-20
Build `Following` structure

pfefferle:
This PR is the next step of the Followers/Following milestone. It wraps up the Followers v2 initiative and implements the Following v1 logic. 

## Proposed changes:

* Actors CRUD: `Create`, `read`, `update`, and `delete` operations for remote Actor entities.
* Follow collection support: Basic `Accept` handling when a follow request is accepted by the target actor.
* `follow()` function: Helper to initiate a Follow request to a remote actor.
* Outbox handling:
  * `Follow` activities are now added to the outbox.
  * `Follow` activity generation and delivery.
* Status updates: Follower relationship status transitions from pending to accepted upon receiving an `Accept`.

## Notes:

* Only `Accept` is handled at this stage; other responses (e.g. `Reject`) are not implemented with this PR. See #1865

First step towards: #1036

### Other information:

## Testing instructions:

* See tests.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Support for sending follow requests to remote actors is now in place, including outbox delivery and status updates—UI integration will follow later.

</details>


includes/collection/class-actors.php
```diff
@@ -203,6 +204,14 @@ public static function get_by_resource( $uri ) {
 			// Check for http(s) URIs.
 			case 'http':
 			case 'https':
+				// Check locally stored remote Actor.
+				$post = self::get_remote_by_uri( $uri, true );
```
```suggestion
				$post = self::get_remote_by_uri( $uri );
```
Maybe `get_remote_by_uri()` used to accept a second arg? Looks like it no longer does.

includes/collection/class-actors.php
```diff
@@ -391,70 +399,98 @@ public static function get_type_by_id( $user_id ) {
...
+			$actor = Actor::init_from_array( $actor );
 		}
 
-		if ( ! $actor_data instanceof Actor ) {
-			return new \WP_Error(
-				'activitypub_invalid_actor_data',
-				\__( 'Invalid actor data', 'activitypub' ),
-				array( 'status' => 400 )
-			);
+		$post = self::get_remote_by_uri( $actor->get_id(), true );
```
```suggestion
		$post = self::get_remote_by_uri( $actor->get_id() );
```

includes/collection/class-actors.php
```diff
@@ -505,8 +541,36 @@ public static function get_remote_by_uri( $actor_uri ) {
...
 	/**
-	 * This function is used to store errors that occur when
-	 * sending an ActivityPub message to a Follower.
+	 * Lookup a remote actor post by actor URI (guid), fetching from remote if not found locally.
+	 *
+	 * @param string $actor_uri The actor URI.
+	 *
+	 * @return \WP_Post|\WP_Error Post object or WP_Error if not found.
+	 */
+	public static function lookup_remote_by_uri( $actor_uri ) {
```
obenland: I'd love to find an even more semantic name for this, if you're open to it? Happy to try and come up with some options

includes/collection/class-following.php
```diff
@@ -0,0 +1,86 @@
...
+	 * @param \WP_Post|int $post    The ID of the remote Actor.
+	 * @param int          $user_id The ID of the WordPress User.
+	 *
+	 * @return \WP_Post|\WP_Error The ID of the Actor or a WP_Error.
+	 */
+	public static function follow( $post, $user_id ) {
+		$post = \get_post( $post );
+
+		if ( ! $post ) {
+			return new \WP_Error( 'activitypub_remote_actor_not_found', 'Remote actor not found' );
```
obenland, line 42: With `Actors::lookup_remote_by_uri()` available, I think it would be great to accept a URI and call that here instead of bailing, following the idea of APIs being permissible with inputs.

includes/collection/class-following.php
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		$following = \get_post_meta( $post->ID, self::FOLLOWING_META_KEY, false );
+
+		if ( ! \is_array( $following ) ) {
+			$following = array();
+		}
+
+		$pending = \get_post_meta( $post->ID, self::PENDING_META_KEY, false );
```
obenland:
```suggestion
		$all_meta  = get_post_meta( $post->ID );
		$following = $all_meta[ self::FOLLOWING_META_KEY ] ?? array();
		$pending   = $all_meta[ self::PENDING_META_KEY ] ?? array();
```
I don't know if this is helpful, performance-wise it probably doesn't make a difference, with meta caching and all, but might help with adding a fallback for empty pending key?

obenland (commented):
This feels really close, nice work!

includes/collection/class-actors.php
pfefferle: Sure, I am happy about new ideas! AI was not really helpful here 😏

includes/collection/class-following.php
pfefferle: This is implemented very similar to the `Outbox`. There is a global function `follow` that accepts an `Actor` URL and does all the heavy lifting and "normalization", like the lookup you mentioned and it also generates the Outbox entry, to send the follow.

This function is "only" to generate the post-metas consistent with the following `accept` function.

includes/collection/class-following.php
pfefferle: let's try it!

pfefferle:
@obenland I changed the `Accept` handling, so maybe you need to have a second look at: https://github.com/Automattic/wordpress-activitypub/pull/1839/commits/2b65e48782aa11b661d345998b5a13d2d4784f56

includes/collection/class-following.php
obenland: I guess my expectation would be for this method to work with an external user ID, as the purpose is to follow an external user. I think it's fine for the global function to connect the new follow with adding it to the Outbox, but I'd expect a follow method to fully own what it does, if that makes sense?

obenland:
> I changed the Accept handling, so maybe you need to have a second look at: https://github.com/Automattic/wordpress-activitypub/commit/2b65e48782aa11b661d345998b5a13d2d4784f56

Looking good! ✨

includes/collection/class-actors.php
obenland: A couple of options come to mind for me:
* Merge this method with `get_remote_by_uri` and add a second parameter that enables the fallback to fetching it:
```php
	public static function get_remote_by_uri( $actor_uri, $fetch_if_missing = false ) {
		global $wpdb;
		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
		$post_id = $wpdb->get_var(
			$wpdb->prepare(
				"SELECT ID FROM $wpdb->posts WHERE guid=%s AND post_type=%s",
				esc_sql( $actor_uri ),
				esc_sql( self::POST_TYPE )
			)
		);

		if ( ! $post_id ) {
			if ( $fetch_if_missing ) {
				$object = Http::get_remote_object( $actor_uri, false );
				if ( \is_wp_error( $object ) ) {
					return $object;
				}

				$post_id = self::create( $object );
				if ( \is_wp_error( $post_id ) ) {
					return $post_id;
				}
			} else {
				return new \WP_Error(
					'activitypub_actor_not_found',
					\__( 'Actor not found', 'activitypub' ),
					array( 'status' => 404 )
				);
			}
		}

		return \get_post( $post_id );
	}
```
* If there's a preference for separate methods, maybe something like `get_or_fetch_remote_actor`, `get_or_fetch_remote_by_uri`, or `maybe_fetch_remote`.

pfefferle: I thought it might be good to only check the db and I am not sure I am a fan of a param. Will think bout that over the weekend ☺️

obenland: I don't have a strong preference for the combined method, fwiw, just thought I'd bring it up as a way for simplification. I do think there's an opportunity for a better name for `lookup_remote_by_uri` though

includes/collection/class-following.php
pfefferle: As I said, it is exactly the same mechanism as with the Outbox. This function was meant to be a helper function to "only" add the follow connection between an already stored `Actor` and a WordPress User. Maybe we would have to rename it then?

includes/collection/class-actors.php
obenland: > maybe something like `get_or_fetch_remote_actor`, `get_or_fetch_remote_by_uri`, or `maybe_fetch_remote`.

Not sure if that got lost below the combine function example.

pfefferle: What about `fetch_remote_actor`? I don’t think we need to mention ‘get’ here. The purpose of this function is to retrieve a remote Actor, regardless of whether they’re already present in the database. The database is essentially just a cache and, in my view, doesn’t need to be referenced explicitly.

obenland: `fetch_remote_actor` sounds good!

obenland (approved):
Let's get it in!

PR #1841, opened 2025-06-20
Add fallback for Follower post-meta

pfefferle:
We recently migrated the followers post-type to use `_activitypub_follower` instead of `_activitypub_user_id` to also support followings using the same mechanism.

It seems that the migration wasn't properly run on every installation, so this PR adds fallback support for `_activitypub_user_id` until we found and fixed the issue.

Fight the symptoms ✊

😔

## Proposed changes:

* Add fallback support for `_activitypub_user_id`

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Improved compatibility with older follower data.

</details>


obenland:
Maybe this could be paired with a site health check to see if there are remaining meta keys. And if so, give users a button to run an upgrade routine?

I do wonder if there are certain sites where these wholesale replacement DB calls are capped to a certain number of rows, to prevent updating a large amount of rows in one go. Maybe we need to batch them too, to sets of 50 or so?

pfefferle:
@obenland I agree on your points, but I think these are different features and should be separate PRs.

obenland (approved):

PR #1848, opened 2025-06-23
Signatures: Improve support for hs2019 algorithm

obenland:
Improves handling of hs2019 signatures by deriving the signature algorithm from the actor's public key.

## Proposed changes:

* Moves public key retrieval to earlier in the function, so it can be passed to `get_signature_algorithm()` and inspected.
* Extends `get_signature_algorithm()` to accept a public key and return a WP_Error.
* Updates error return after `get_signature_algorithm()` to return the generated error rather than a generic one.
* Updates `hs2019` case to inspect the public key and derive the algorithm used.

### Other information:

## Testing instructions:

* Beyond unit tests, it would be good to find a Fediverse instance that signs their outgoing requests with hs2019 to test against.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

hs2019 signatures for incoming REST API requests now have their algorithm determined based on their public key.

</details>


includes/class-signature.php
```diff
@@ -347,21 +346,57 @@ public static function get_remote_key( $key_id ) {
...
+									return \OPENSSL_ALGO_SHA256;
+								case 'secp384r1':
+									return \OPENSSL_ALGO_SHA384;
+								case 'secp521r1':
+									return \OPENSSL_ALGO_SHA512;
+							}
+					}
+
+					return new \WP_Error( 'unsupported_key_type', 'Unsupported key type (only RSA and EC keys are supported).', array( 'status' => 401 ) );
+
```
obenland: @vpzomtrrfrt I'm curious if you'd have any advice on how this could be improved to [switch `algorithm="hs2019"` support for wordpress-activitypub to "Yes"](https://arewehs2019yet.vpzom.click/)?

obenland:
@mediaformat Curious if you had any thoughts on this change?

includes/class-signature.php
vpzomtrrfrt: I haven't followed this space as closely lately, the last I heard was that there was no consensus on how keys should indicate their supported algorithms

pfefferle (dismissed):

pfefferle (approved):

PR #1849, opened 2025-06-23
Signatures: Add basic support for RFC-9421

### LINKED ISSUE #1808: Feature Request: Upgrade HTTP Message signatures to implement RFC-9421

obenland:
Fixes #1808.

## Proposed changes:

* Refactored the signature verification logic in `Signature::verify_http_signature` to delegate to new standard-specific classes. 
* Added `Draft_Cavage_Signature` and `Http_Message_Signature` classes implementing a new `Signature_Standard` interface, supporting both the legacy draft-cavage and new HTTP Message Signature (RFC 9421) standards. 
* Deprecated and stubbed out legacy parsing and verification helpers in `class-signature.php`. 

### Other information:

## Testing instructions:

* No one is using the new standard yet, so unit tests probably suffice.
* Apply the PR to a test site that's accessible from the internet.
* Make sure receiving activities still works as expected.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Added basic support for RFC-9421 style signatures for incoming activities.

</details>


obenland:
Done:
* Parsing and verifying Signature-Input and Signature headers.
*  Dynamic label support (e.g., @method, @target-uri, etc.).
*  Parameter support (alg=, created=, expires=, nonce=, keyid=).
*  Canonicalization of headers (whitespace folding and trimming).
*  Digest validation.
*  Multiple signature label support (e.g., sig1=, sig2=).
*  Signature string composition separated out.
* Algorithm resolution based on alg and key details (with PHP version check).

Skipped for now:
* Structured field parsing
* Nonce uniqueness tracking

What do you think @mediaformat?

.github/workflows/gardening.yml
```diff
@@ -59,6 +59,7 @@ jobs:
             {"path": "includes/class-blocks.php", "label": "[Focus] Editor"},
             {"path": "includes/collection", "label": "[Feature] Collections"},
             {"path": "includes/rest", "label": "[Feature] REST API"},
+            {"path": "includes/signature", "label": "[Feature] Signature"},
```
pfefferle, line 62:
```suggestion
            {"path": "includes/signature", "label": "[Feature] Signature"},
            {"path": "includes/class-signature.php", "label": "[Feature] Signature"},
```

obenland: I was planning on moving `class-signature.php` into `/includes/signature` in a follow-up PR, would that work for you?

pfefferle: I think best practice is to have the main class (or factory or however you want to call it) in the root and only the special cases in the subfolder, so I quite like the structure as is. that said, we havent't done that with transformers, so it would be at least consistent to move it 🫣

mediaformat:
> * Parsing and verifying Signature-Input and Signature headers.
> * Dynamic label support (e.g., @method, @target-uri, etc.).
> * Parameter support (alg=, created=, expires=, nonce=, keyid=).
> * Canonicalization of headers (whitespace folding and trimming).
> * Digest validation.
> * Multiple signature label support (e.g., sig1=, sig2=).
> * Signature string composition separated out
> * Algorithm resolution based on alg and key details (with PHP version check).

Only looked at the code so far, but looks great!

obenland:
@pfefferle I think this is ready for another look. Unit tests look firm now and #1858 helped polish out some remaining bugs. As long as legacy signatures still work this should be safe to merge.

pfefferle (approved):

PR #1855, opened 2025-06-25
Fix: Remove unnecessary check for Actor-ID

pfefferle:
Removed a duplicate check that was already done the lines before.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/includes/collection/class-interactions.php
+++ b/includes/collection/class-interactions.php
@@ -248,10 +248,6 @@ class Interactions {
 
 		$url = object_to_uri( $actor['url'] ?? $actor['id'] );
 
-		if ( ! $url ) {
-			$url = object_to_uri( $actor['id'] );
-		}
-
 		if ( isset( $activity['object']['content'] ) ) {
 			$comment_content = \addslashes( $activity['object']['content'] );
 		}


includes/collection/class-interactions.php
```diff
@@ -248,10 +248,6 @@ public static function activity_to_comment( $activity ) {
 
 		$url = object_to_uri( $actor['url'] ?? $actor['id'] );
 
-		if ( ! $url ) {
-			$url = object_to_uri( $actor['id'] );
-		}
```
obenland, line 253: I reckon it is probably fine to keep?

The fallback in line 249 is for the case where `url` is null or undefined, the fallback here for the case when `$actor['url']` doesn't resolve to a url, is how I'm reading it.

pfefferle: that is not what `object_to_uri` tests!

`object_to_uri` only checks if it is a string, object or array, to then react, but there is no `is_url` validation (yet).

obenland (approved):

PR #1857, opened 2025-06-25
Only replace `comment_link` for remote comments

pfefferle:
The plugin sets the comment_link for all types of remote objects—likes, replies, and shares. However, only replies have actual URLs. Likes and boosts are merely Activities without direct links, so the code defaults to using their ID. As a result, users might end up viewing raw JSON when clicking such links.

## Proposed changes:

* Only replace the `comment_link` for replies with a URL attribute

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fixed an issue where links to remote likes and boosts could open raw JSON instead of a proper page.

</details>


includes/class-comment.php
```diff
@@ -389,13 +389,17 @@ public static function get_source_url( $wp_comment_id, $fallback = true ) {
 	 * @return string $url
 	 */
 	public static function remote_comment_link( $comment_link, $comment ) {
+		$remote_comment_link = null;
+
 		if ( ! $comment || is_admin() ) {
 			return $comment_link;
 		}
```
```suggestion
		if ( ! $comment || is_admin() ) {
			return $comment_link;
		}

		$remote_comment_link = null;
```
Let's wait with defining it until we need it?

obenland (approved):

PR #1858, opened 2025-06-25
Signature: Add option to use RFC-9421 signatures

obenland:
Follow-up to #1849.

## Proposed changes:

* Updated tests to expect deprecation warnings and use the new signature class.
* Added registration for the new RFC‑9421 signature option.
* Added a new settings field to toggle the modern signature format.
* Extended the interface to include RFC‑9421 signature and digest generation.
* Introduced RFC‑9421 compliant signing with updated signature base string.
* Updated draft signature implementation with its own digest generation.
* Refactored request signing to choose between the two signature methods.
* Reworked HTTP request creation to integrate the new signing approach.

### Other information:

## Testing instructions:

* Apply this PR to a site with followers.
* Publish a post or make a change to an actor.
* Make sure the Outbox item gets processed and sent successfully.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Optional support for RFC-9421 style signatures for outgoing activities, including retry with Draft-Cavage-style signature.

</details>


obenland:
~~Does not do double-knocking yet~~ Fixed!

includes/signature/class-draft-cavage-signature.php
```diff
@@ -22,6 +22,56 @@
  */
 class Draft_Cavage_Signature implements Signature_Standard {
 
+	/**
+	 * Generate Signature headers for an outgoing HTTP request.
+	 *
+	 * @param array  $args The request arguments.
+	 * @param string $url  The request URL.
```
pfefferle, line 29:
```suggestion
	 * @param string $url  The request URL.
	 *
```

includes/signature/class-http-message-signature.php
```diff
@@ -22,6 +22,50 @@
  */
 class Http_Message_Signature implements Signature_Standard {
 
+	/**
+	 * Generate RFC-9421 compliant Signature-Input and Signature headers for an outgoing HTTP request.
+	 *
+	 * @param array  $args The request arguments.
+	 * @param string $url  The request URL.
```
pfefferle, line 29:
```suggestion
	 * @param string $url  The request URL.
	 *
```

includes/signature/class-http-message-signature.php
```diff
@@ -22,6 +22,50 @@
...
+	 * @return array Request arguments with signature headers.
+	 */
+	public function sign( $args, $url ) {
```
pfefferle, line 33: do we need some kind of `$args` validation? or at least set defaults with `wp_parse_args` to not produce any "array key does not exist" errors?

includes/signature/interface-signature-standard.php
```diff
@@ -14,6 +14,16 @@
  * Interface Signature_Standard.
  */
 interface Signature_Standard {
+
+	/**
+	 * Generate Signature headers for an outgoing HTTP request.
+	 *
+	 * @param array  $args The request arguments.
+	 * @param string $url  The request URL.
```
pfefferle, line 22:
```suggestion
	 * @param string $url  The request URL.
	 *
```

includes/class-signature.php
```diff
@@ -191,6 +212,7 @@ protected static function check_legacy_key_pair_for( $user_id ) {
 	 * @return string The signature.
 	 */
 	public static function generate_signature( $user_id, $http_method, $url, $date, $digest = null ) {
+		_deprecated_function( __METHOD__, 'unreleased', self::class . '::sign_request()' );
```
pfefferle:
```suggestion
		\_deprecated_function( __METHOD__, 'unreleased', self::class . '::sign_request()' );
```

would slashing the `_deprecated_function` break the version replacement script?

includes/class-signature.php
```diff
@@ -328,11 +373,15 @@ public static function get_signed_data( $signed_headers, $signature_block, $head
...
 	 * Generates the digest for an HTTP Request.
 	 *
+	 * @deprecated unreleased Use {@see Signature::sign_request()}.
+	 *
 	 * @param string $body The body of the request.
 	 *
 	 * @return string The digest.
 	 */
 	public static function generate_digest( $body ) {
+		_deprecated_function( __METHOD__, 'unreleased', self::class . '::sign_request' );
```
pfefferle:
```suggestion
		\_deprecated_function( __METHOD__, 'unreleased', self::class . '::sign_request' );
```

includes/signature/class-http-message-signature.php
obenland: `Signature::sign_request()` is almost like a filter callback that runs before the request args get passed to `wp_remote_request()`. Actually, now that key_id and private_key are passed as part of the request args, we could turn it into a filter callback for `http_request_args`.

With that in mind, I think it would be good to not validate and return errors, but maybe just bail if we find that key_id and private_key are not available?
The date header we could add as a default parameter, and potentially also the request method. Although if we do end up making it a filter, the method will always be set.

Long story short: Yes, that's a good idea!

obenland: Fixed in [5573da5](https://github.com/Automattic/wordpress-activitypub/pull/1858/commits/5573da5f9fa88c16bb41ffee576e4476e13f574e).

pfefferle (approved):
Nice!

PR #1863, opened 2025-06-27
Signature: Improved naming consistency

pfefferle:
While reviewing all the Signature changes, I reconsidered the file naming. Since all Signature-related files now in the Signature namespace, I think it's reasonable to drop signature from the file and class names—similar to how we handle Transformers.

However, I'm still unsure about the best name for the interface. The term standard doesn’t quite convey that it's an interface—it sounds more like a usable implementation for a "standard" signature, perhaps even a base class.

## Proposed changes:

* Removed `signature` from class and filenames
* Renamed interface

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


obenland:
> I'm still unsure about the best name for the interface. The term standard doesn’t quite convey that it's an interface—it sounds more like a usable implementation for a "standard" signature, perhaps even a base class.

The name is based on the idea that it represents a Signature standard that gets implemented. "standard" as in RFC 9421 being a "proposed standard" (rather than standard vs enhanced).

pfefferle:
But the draft also implements the interface!?!

What do you think about the other renamings?

obenland:
> But the draft also implements the interface!?!

Right, it's a draft standard. I was just trying to point out the difference between "standard" as in "a formal specification or protocol" and standard as on "the default or basic option", in case that is what's at issue.

> What do you think about the other renamings?

I like them, I wasn't 100% happy with the `_Signature` suffix either. I also don't mind `Http_Signature` for the interface, just thought I'd add context for the naming since you didn't seem sure about a good name for it.

pfefferle:
@obenland I also would love to find a different name for `Draft_Cavage`. `Cavage` is the name of the editor and we should not use that for the naming. Maybe something like `Draft_Http_Messages` or simply `Draft_Spec`?

pfefferle:
I will wait for cleaning up the PR until after we merged all missing signature PRs.

pfefferle:
@obenland what about `Class Draft` and `Class Rfc9421`?

obenland:
How would you feel about `Draft` and `HTTP_Message`?

pfefferle:
In the end both are HTTP Message Signatures, so this would be misleading, I think:

* Draft -> Signing HTTP Messages
* RFC9421 -> HTTP Message Signatures

obenland:
So `Signing_HTTP_Messages` and `HTTP_Message_Signatures` then? 

RFC9421 is a pretty poor class name IMO

pfefferle:
what about `activitypub_rfc9421_signature` and `self::rfc9421_is_unsupported` then?

obenland:
I'd be happy for us to update them

pfefferle:
Mastodon is calling them:

* HttpMessageSignature
* HttpSignatureDraft

Signify is calling them `Draft` and `RFC` https://github.com/fedify-dev/fedify/blob/506023b5d2d2cdbd0a5b25a6397b695c15c60150/fedify/sig/http.ts

obenland:
> Mastodon is calling them:
>
> * HttpMessageSignature
> * HttpSignatureDraft

I'd be happy with snake cased versions of those, too.

pfefferle:
I would love to have the generic `HttpMessageSignature` one as interface... so how to call the RFC one?

pfefferle:
I will see if AI might be helpful...

obenland (approved):
✨

PR #1865, opened 2025-06-27
Handle `Follow` rejects

pfefferle:
Implement basic `Reject` handling.

The focus of this PR is on rejecting `Follow` requests, but we will add support for other `Reject`s in the future: https://www.w3.org/wiki/ActivityPub/Primer/Reject_activity#Additional_uses_of_Reject

## Proposed changes:

* Handle `Follow` rejects

### Other information:

## Testing instructions:

* Follow an Account that supports `Accept`ing and `Reject`ing `Follow` requests
* Reject `Follow` request

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added basic support for handling remote rejections of follow requests.

</details>


includes/handler/class-reject.php
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		$required_attributes = array(
+			'actor',
+			'object',
+		);
+
+		if ( ! empty( \array_diff( $required_attributes, \array_keys( $json_params ) ) ) ) {
+			return false;
+		}
```
obenland, line 120:
```suggestion
		if ( ! isset( $json_params['actor'], $json_params['object'] ) ) {
			return false;
		}
```
For instances where we only check for a couple of attributes, we could consider using `isset` as a micro-optimization

includes/handler/class-reject.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	/**
+	 * Handles "Reject" requests.
+	 *
+	 * @param array $reject  The activity-object.
+	 * @param int   $user_id The id of the local blog-user.
+	 */
+	public static function handle_reject( $reject, $user_id ) {
+		// Validate that there is a Follow Activity.
```
obenland:
```suggestion
		// Validate that there is a preceding Activity.
```
To make it more generally applicable

includes/handler/class-reject.php
obenland:
```suggestion
		if ( ! isset( $json_params['actor'], $json_params['object'] ) ) {
			return false;
		}
```
For instances where we only check for a couple of attributes, we could consider using `isset` as a micro-optimization

obenland (commented):
This looks really good. I think it'd be optimal to wait with merging until #1839 is in, to avoid making that PR even bigger.

includes/handler/class-reject.php
pfefferle: it is not enough if it is only set, it does not have to be `null`

obenland (dismissed):

obenland (approved):

PR #1866, opened 2025-06-27
Following v1: Add basic following UI

pfefferle:
This is the first step to add a Following UI for Author-Actors. For now the feature is hidden behind a filter, so it could be enabled when plugins like Friends, Enable Mastodon Apps, or the Event Bridge will add support to their plugins.

Screenshot 2025-07-07 at 15 08 18

The next steps will be to also add a simple follow dialogue so that users can add new followers.

## Proposed changes:

* Add menu entry for Following
* Add basic UI that matches the look and feel of the Followers list

### Other information:

## Testing instructions:

* Enable "Author Profiles Only" or "Both author and blog profiles"
* Go to `/wp-admin/users.php?page=activitypub-following-list`

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added initial Following support for Author-Actors, hidden for now until plugins add support.

</details>


includes/table/class-following.php
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Followers Table-Class file.
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\Table;
+
+use Activitypub\Collection\Actors;
+use Activitypub\Collection\Following as FollowingCollection;
```
obenland:
```suggestion
use Activitypub\Collection\Following as Following_Collection;
```
Let's stick with snake case here?

includes/table/class-following.php
```diff
@@ -0,0 +1,236 @@
...
+		$nonce = \sanitize_text_field( \wp_unslash( $_REQUEST['_wpnonce'] ) );
+		if ( ! \wp_verify_nonce( $nonce, 'bulk-' . $this->_args['plural'] ) ) {
+			return;
+		}
+
+		if ( ! \current_user_can( 'edit_user', $this->user_id ) ) {
+			return;
+		}
+
+		$following = $_REQUEST['following']; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput
```
obenland:
```suggestion
		$following = \sanitize_url( \wp_unslash( $_REQUEST['following'] ) );
```
Would this be a good way to sanitize this value?

includes/table/class-following.php
```diff
@@ -0,0 +1,236 @@
...
+		if ( $this->current_action() === 'delete' ) {
+			if ( ! is_array( $following ) ) {
+				$following = array( $following );
+			}
+			foreach ( $following as $following ) {
+				$actor = Actors::get_remote_by_uri( $following );
```
obenland:
```suggestion
			foreach ( $following as $actor_id ) {
				$actor = Actors::get_remote_by_uri( $actor_id );
```
This will need a separated variable to not overwrite `$following` on the first iteration of the `foreach`.

includes/wp-admin/class-menu.php
```diff
@@ -41,6 +41,18 @@ public static function admin_menu() {
 
 			\add_action( 'load-' . $followers_list_page, array( Admin::class, 'add_followers_list_help_tab' ) );
 
+			if ( \apply_filters( 'activitypub_show_following_ui', false ) ) {
```
obenland, line 49:
```suggestion
			/**
			 * Filter to show the following UI.
			 *
			 * @param bool $show Show following UI.
			 */
			if ( \apply_filters( 'activitypub_show_following_ui', false ) ) {
```

obenland (changes requested):

includes/table/class-following.php
pfefferle: Fixed with https://github.com/Automattic/wordpress-activitypub/pull/1866/commits/ce80f5f70e2d9eafb5ff021aeb454a48dc12d493

includes/table/class-following.php
```diff
@@ -0,0 +1,236 @@
...
+		$nonce = \sanitize_text_field( \wp_unslash( $_REQUEST['_wpnonce'] ) );
+		if ( ! \wp_verify_nonce( $nonce, 'bulk-' . $this->_args['plural'] ) ) {
+			return;
+		}
+
+		if ( ! \current_user_can( 'edit_user', $this->user_id ) ) {
+			return;
+		}
+
+		$following = \sanitize_url( \wp_unslash( $_REQUEST['following'] ) );
```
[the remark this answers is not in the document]
obenland: Oh, that's actually a good point, `$_REQUEST['following']` can be an array

pfefferle:
@obenland added the changes reuqested by copilot

obenland (approved):
I still can't unfollow someone from the UI, but that might be a dependency on the way I added these followings and #1905.

There's probably more opportunity to improve the way `$_REQUEST['following']` gets sanitized in the table class, but that can be a follow up.

Maybe we could review the follower and following template files to see if they can be combined for blog and user types?

None of this I'd consider merge-blocking, though.

pfefferle:
🤔

Just tested unfollow and it worked fine on my end!?

PR #1868, opened 2025-06-27
Add support for capability discovery

obenland:
I came across FEP 844e during my research on RFC 9421. It's pretty fresh and I'm happy for us to determine this would not be a good fit for the plugin, but I thought supporting it seems pretty trivial and could help with making it clearer that we support RFC 9421 signatures.

See https://codeberg.org/fediverse/fep/src/branch/main/fep/844e/fep-844e.md

## Proposed changes:

* Adds support for FEP 844e.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Documented support for FEP-844e.

</details>

### DIFF
--- a/.github/changelog/1868-from-description
+++ b/.github/changelog/1868-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Documented support for FEP-844e.
--- a/FEDERATION.md
+++ b/FEDERATION.md
@@ -19,6 +19,7 @@ The WordPress plugin largely follows ActivityPub's server-to-server specificatio
 - [FEP-fb2a: Actor metadata](https://codeberg.org/fediverse/fep/src/branch/main/fep/fb2a/fep-fb2a.md)
 - [FEP-b2b8: Long-form Text](https://codeberg.org/fediverse/fep/src/branch/main/fep/b2b8/fep-b2b8.md)
 - [FEP-7888: Demystifying the context property](https://codeberg.org/fediverse/fep/src/branch/main/fep/7888/fep-7888.md)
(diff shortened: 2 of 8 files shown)


Jiwoon-Kim:
The only practically meaningful use case for this is outbox proxy posting.

For example, a proxy service like Jetpack Social could federate content on behalf of a user, while still displaying the original author's identity.

I believe this might be useful for wp.com and Jetpack-based services in the future, but outside of that context, it probably has limited utility.

Jiwoon-Kim:
**Possible Use Case for Capability Discovery in Jetpack Reader**

Actually, one of the only truly practical scenarios for capability discovery would be **outbox proxy posting** — similar to how Jetpack Social works: the Jetpack account federates on behalf of the real author, with attribution.

This could be especially useful within WordPress.com’s ecosystem.

Moreover, if **Jetpack Reader** ever extends to integrate with ActivityPub, capability discovery could serve multiple purposes:

- Check which content types an instance supports (e.g. `Article`, `Note`, `Poll`, `Event`)
- Determine if the instance allows proxy posting
- Dynamically adjust UI based on supported capabilities
- Handle interactions like likes, comments, and polls directly from the Reader when possible

In this sense, capability discovery could become the foundational layer for enabling Jetpack Reader to properly federate, interact, and display content from external ActivityPub instances.

**Summary:**  
- Outbox proxy posting is realistically the most useful scenario for now.  
- Jetpack Reader could leverage capability discovery when integrating with ActivityPub.  
- This would make it possible to optimize UX and federation reliability across different instances.

pfefferle:
Added:

* Context URI
* Item Schema
* "Discovery through an actor"

pfefferle (dismissed):

pfefferle (approved):

PR #1869, opened 2025-06-27
Signature: Cache instances that don't support RFC 9421

obenland:
Follow-up to #1858.

## Proposed changes:

* Adds a check for known instances that don't support RFC 9421 signatures, before attempting to send them.
* Adds hosts who returned a 401 for an RFC 9421 signature to the list of unsupported hosts.
* Adds tests.

### Other information:

### DIFF
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -90,7 +90,7 @@ class Signature {
 			)
 		);
 
-		if ( '1' === \get_option( 'activitypub_rfc9421_signature' ) ) {
+		if ( '1' === \get_option( 'activitypub_rfc9421_signature' ) && ! self::rfc9421_is_unsupported( $url ) ) {
 			$signature = new Http_Message_Signature();
 			\add_filter( 'http_response', array( self::class, 'maybe_double_knock' ), 10, 3 );
 		} else {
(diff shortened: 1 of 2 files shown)


includes/class-signature.php
```diff
@@ -208,6 +208,7 @@ public static function maybe_double_knock( $response, $parsed_args, $url ) {
 
 		if ( 401 === wp_remote_retrieve_response_code( $response ) ) {
 			unset( $parsed_args['headers']['Signature'], $parsed_args['headers']['Signature-Input'], $parsed_args['headers']['Content-Digest'] );
+			self::set_rfc9421_unsupported( $url );
```
pfefferle: to be consistent with `rfc9421_is_unsupported` I would call this function `rfc9421_set_unsupported` or vice versa.

pfefferle:
The current PR uses the inbox URL to store the information... can this lead to undesirable side effects?

For WordPress.com, the inbox host is the same for all blogs... The current caching mechanism would set the whole WordPress.com infrastructure to "unsupported" if one single blog does not support it...

Is there a case where only a subset of users/sites of a domain/host support `rfc9421` but not all?

obenland:
> Is there a case where only a subset of users/sites of a domain/host support rfc9421 but not all?

Yeah, good question. It's possible, I don't know how likely that is. With our cache expiring monthly, it would require multiple instances on the same host with varying implementations, where at least one doesn't update to support RFC 9421. I feel like that's a remote enough edge case for us to take up individually, if needed.

Even on Dotcom there shouldn't be one blog that doesn't support incoming RFC 9421 signatures.

includes/class-signature.php
obenland: Fixed in [0f2db4a](https://github.com/Automattic/wordpress-activitypub/pull/1869/commits/0f2db4aa964ca9c619b3f5b3a789f02c50266eac)

pfefferle (approved):
even better 😍

PR #1875, opened 2025-06-30
Signature: Double-knock on any 400 response

obenland:
Lemmy processes content digest first and fails it with a `400` response. We'll need to expand the status code check to also send a second knock in that case.

Follow-up to #1858.

## Proposed changes:

* Sends a second knock on any 4xx response code.

### DIFF
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -206,7 +206,8 @@ class Signature {
 		// Remove this filter to prevent infinite recursion.
 		\remove_filter( 'http_response', array( self::class, 'maybe_double_knock' ) );
 
-		if ( 401 === wp_remote_retrieve_response_code( $response ) ) {
+		$response_code = \wp_remote_retrieve_response_code( $response );
+		if ( 400 <= $response_code && 500 > $response_code ) {
 			unset( $parsed_args['headers']['Signature'], $parsed_args['headers']['Signature-Input'], $parsed_args['headers']['Content-Digest'] );
 			self::rfc9421_add_unsupported_host( $url );
 
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -207,7 +207,9 @@ class Signature {
 		\remove_filter( 'http_response', array( self::class, 'maybe_double_knock' ) );
 
 		$response_code = \wp_remote_retrieve_response_code( $response );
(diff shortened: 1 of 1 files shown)


includes/class-signature.php
```diff
@@ -206,7 +206,8 @@ public static function maybe_double_knock( $response, $parsed_args, $url ) {
 		// Remove this filter to prevent infinite recursion.
 		\remove_filter( 'http_response', array( self::class, 'maybe_double_knock' ) );
 
-		if ( 401 === wp_remote_retrieve_response_code( $response ) ) {
+		$response_code = \wp_remote_retrieve_response_code( $response );
+		if ( 400 <= $response_code && 500 > $response_code ) {
```
pfefferle: I think WordPress does not require yoda conditions for greater/less than comparisons!?! Can we use the normal comparison here? It is so hard to read otherwise!

obenland: Updated in [b79ca17](https://github.com/Automattic/wordpress-activitypub/pull/1875/commits/b79ca170c6d52d01a5534e2dfcfd5a8a9f6d487a).

includes/class-signature.php
```diff
@@ -206,7 +206,10 @@ public static function maybe_double_knock( $response, $parsed_args, $url ) {
 		// Remove this filter to prevent infinite recursion.
 		\remove_filter( 'http_response', array( self::class, 'maybe_double_knock' ) );
 
-		if ( 401 === wp_remote_retrieve_response_code( $response ) ) {
+		$response_code = \wp_remote_retrieve_response_code( $response );
+
+		// Fall back to Draft Cavage signature for any 4xx responses.
+		if ( 400 <= $response_code && $response_code < 500 ) {
```
obenland: @pfefferle Are we being to broad with this? So far I've only observed 401s and 400s. And a 500 for Ghost, but I think that may be a one-off.

pfefferle:
```suggestion
		if ( $response_code >= 400 && $response_code < 500 ) {
```

pfefferle: I think I am fine with checking all 4xx codes for now. Otherwise we could limit to 400 and 401 and start logging error codes!?

pfefferle (approved):

PR #1876, opened 2025-06-30
Signature: Always add Host header

obenland:
Follow-up to https://github.com/Automattic/wordpress-activitypub/pull/1858.

[Mitra looks for the Host header](https://codeberg.org/silverpill/mitra/src/branch/main/apx_core/src/http_signatures/verify.rs#L303-L308) when processing the `@authority` parameter.

## Proposed changes:

* The Host header is now explicitly set in both `Draft_Cavage_Signature` and `Http_Message_Signature` classes when signing requests.

### DIFF
--- a/includes/signature/class-draft-cavage-signature.php
+++ b/includes/signature/class-draft-cavage-signature.php
@@ -63,6 +63,7 @@ class Draft_Cavage_Signature implements Signature_Standard {
 		\openssl_sign( $signed_string, $signature, $args['private_key'], \OPENSSL_ALGO_SHA256 );
 		$signature = \base64_encode( $signature );
 
+		$args['headers']['Host']      = $host;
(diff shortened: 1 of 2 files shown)


pfefferle:
The host should be part of the default headers!

It is a required header by http 1.1 https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Host

There should be no need to add that manually!

obenland:
Yeah, I was confused why it wasn't part of my debug output, but I think I was just too early. Something is not working with Mitra when sending RFC9421 signatures, not sure what it is.

pfefferle:
Maybe you (only) have to add the `'"@authority"'  => $host,` part?

obenland:
That just moves the `wp_parse_url()` call to be shared. I wonder if it's the way the signature base is put together, that doesn't match

PR #1877, opened 2025-07-01
Remove reactions from comment feed

pfefferle:
The comment feed contains all reactions (like, repost), which can be very distracting.

Screenshot 2025-07-01 at 11 41 01

## Proposed changes:

* Exclude all comment type except `comment`.
* Add `type` query string, to have different feeds for different types (`/feed?type=like`).
* Add `type=all` to list all reactions to a post

### Other information:

## Testing instructions:

* Go to a posts feed and change the `type` query

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Comment feeds now show only comments by default, with a new `type` filter (e.g., `like`, `all`) to customize which reactions appear.

</details>

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -25,6 +25,7 @@ class Comment {
 
 		\add_filter( 'comment_reply_link', array( self::class, 'comment_reply_link' ), 10, 3 );
 		\add_filter( 'comment_class', array( self::class, 'comment_class' ), 10, 3 );
+		\add_filter( 'comment_feed_where', array( static::class, 'comment_feed_where' ) );
 		\add_filter( 'get_comment_link', array( self::class, 'remote_comment_link' ), 11, 2 );
 		\add_action( 'pre_get_comments', array( static::class, 'comment_query' ) );
 		\add_filter( 'pre_comment_approved', array( static::class, 'pre_comment_approved' ), 10, 2 );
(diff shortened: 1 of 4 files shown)


Menrath:
> Exclude all comment type except comment.

This is not what is happening, is it? Isn't this PR just excluding all comment types "registered" by the ActivityPub plugin?!

pfefferle:
@Menrath it checks if the `?type` is set and if it is part of the registered types and filters by type then (`?type=like` shows only likes) and otherwise falls back to `comment` as default.

pfefferle:
Copilot captures it perfectly!

includes/class-comment.php
```diff
@@ -340,6 +341,31 @@ public static function comment_class( $classes, $css_class, $comment_id ) {
 		return $classes;
 	}
 
+	/**
+	 * Filters the comment feed where clause to exclude ActivityPub comment types.
```
```suggestion
	 * Makes the comment feed filterable by comment type.
	 *
	 * Also excludes ActivityPub comment types from the feed when no type is specified.
```
Not sure this is suggestion is the best way to convey that information, but it would be nice to reflect both aspects of what this function does.

includes/class-comment.php
```diff
@@ -340,6 +341,31 @@ public static function comment_class( $classes, $css_class, $comment_id ) {
...
+		if ( 'all' === $comment_type ) {
+			return $where;
+		}
+
+		$comment_types = self::get_comment_type_slugs();
+
+		if ( in_array( $comment_type, $comment_types, true ) ) {
+			$where .= " AND comment_type = '" . esc_sql( $comment_type ) . "'";
+		} else {
+			$where .= " AND comment_type = 'comment'";
```
obenland: Rather than making decisions for comment types no registered by us, can we use a NOT IN and only exclude `$comment_types`?

tests/includes/class-test-comment.php
```diff
@@ -607,4 +607,41 @@ public function test_pre_comment_approved_filter() {
 		\wp_delete_comment( $comment_id, true );
 		\wp_delete_post( $post_id, true );
 	}
+
+	/**
+	 * Test comment_feed_where.
+	 *
+	 * @covers ::comment_feed_where
+	 */
+	public function test_comment_feed_where() {
```
obenland, line 616: These tests are probably fine, more interesting might be whether the results of a comments feed query reflect the changes that we're looking for. For a post with all of these different comment types, do we actually get not get like and repost comments?

includes/class-comment.php
pfefferle: In order not to potentially interfere with the behavior of other plugins? Makes sense!

obenland (dismissed):

obenland (approved):

PR #1893, opened 2025-07-03
Introduce general namespace support, using `dcterms` as an example

pfefferle:
This implements general support for namespaced attributes and implements the `dcterms:subject` field, defined in [FEP-b2b8](https://codeberg.org/fediverse/fep/src/branch/main/fep/b2b8/fep-b2b8.md).

This is a first step to deprecate the usage of Summary as Content-Warning description. https://codeberg.org/fediverse/fep/src/branch/main/fep/b2b8/fep-b2b8.md#sensitive

## Proposed changes:

* Add support for namespaced attributes
* Implement `dcterms:subject` support as described in FEP-b2b8

### Other information:

## Testing instructions:

* Publish a post with a Content-Warning
* Check the JSON output

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added support for namespaced attributes and the dcterms:subject field (FEP-b2b8), as a first step toward phasing out summary-based content warnings.

</details>

### DIFF
--- a/includes/activity/class-generic-object.php
+++ b/includes/activity/class-generic-object.php
@@ -297,8 +297,11 @@ class Generic_Object {
 				$value = $value->to_array( false );
 			}
 
-			// If value is still empty, ignore it for the array and continue.
-			if ( isset( $value ) ) {
+			if ( $this->is_namespaced( $key ) && is_array( $value ) ) {
+				foreach ( $value as $sub_key => $sub_value ) {
+					$array[ snake_to_camel_case( $key ) . ':' . snake_to_camel_case( $sub_key ) ] = $sub_value;
+				}
+			} elseif ( isset( $value ) ) {
 				$array[ snake_to_camel_case( $key ) ] = $value;
 			}
 		}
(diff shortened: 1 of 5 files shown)


includes/activity/class-generic-object.php
```diff
@@ -297,8 +297,11 @@ public function to_array( $include_json_ld_context = true ) {
 				$value = $value->to_array( false );
 			}
 
-			// If value is still empty, ignore it for the array and continue.
-			if ( isset( $value ) ) {
+			if ( $this->is_namespaced( $key ) && is_array( $value ) ) {
```
obenland: I wonder if swapping the two checks could be a micro-optimization? We wouldn't have to check keys where the value is not an array.

includes/transformer/class-post.php
```diff
@@ -57,6 +57,7 @@ public function to_object() {
 			$object->set_sensitive( true );
 			$object->set_summary( $content_warning );
 			$object->set_summary_map( null );
+			$object->set_dcterms( array( 'subject' => $content_warning ) );
```
obenland, line 60: Optional: I think it would be cool if there was a unit test that covered this showing up in a post with content warning. It doesn't look like the ones in `Test_Outbox` cover that scenario

obenland (approved):
I think this is good to go!

obenland:
Oh, should we add the FEP to `FEDERATED.md` or is it too early for that?

pfefferle:
@obenland it is part of the Long-Form FEP and that is already added.

PR #1900, opened 2025-07-07
Change default `Content Visibility` for unfederated posts older than a month

### LINKED ISSUE #1216: Feature Request: Option to set the "Federate" setting to Do Not Federate by default when editing existing unfederated posts
### What

I'd like an option to set the "Federate" setting when editing posts to be set to "Do Not Federate" by default if that post hasn't already been federated before.

### Why

I have over 20 years of blog posts, a lot of them imported from my old LiveJournal, and to this day I still occasionally come across things like broken links that need fixing, egregious spelling errors that I didn't notice at the time, and so on. On multiple occasions since installing the ActivityPub plugin I've edited these old posts and forgotten that the "Federate" option is turned on by default, and so end up publishing a post from like 15 years ago to my blog's fediverse account.

It would be nice if I didn't constantly need to remember this and I could just have the Federate option turned off by default.

### How

For existing posts that have already been federated and are being edited, those would want to default to the "Federate" option being on, this feature request is just for existing posts being edited that have never been federated before.

An additional thought I had, I don't know much about the internals of how Wordpress works so I'm not sure if you can distinguish between a _new_ post versus editing an existing one, but if so it'd be neat if some sort of "This post will be federated" prepublish check could be shown for new posts.

pfefferle:
Fixes #1216

## Proposed changes:

* Set default "Content Visibility" to `local` for unfederated posts older than 1 month.

### Other information:

## Testing instructions:

* change the date of a post to a date older than a year
* delete content visibility post_meta
* the default should be `local` now

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Set older unfederated posts to local visibility by default.

</details>


includes/class-activitypub.php
```diff
@@ -703,6 +704,36 @@ public static function prevent_empty_post_meta( $check, $object_id, $meta_key, $
...
+	 */
+	public static function default_post_metadata( $meta_value, $object_id, $meta_key ) {
+		// Check if the meta key is `activitypub_content_visibility`.
+		if ( 'activitypub_content_visibility' !== $meta_key ) {
+			return $meta_value;
+		}
+
+		// If the post is federated, return the default visibility.
+		$is_federated = \get_post_meta( $object_id, 'activitypub_status', true );
+		if ( 'federated' === $is_federated ) {
```
```suggestion
		if ( 'federated' === \get_post_meta( $object_id, 'activitypub_status', true ) ) {
```
`$is_federated` is not used again.

includes/class-activitypub.php
```diff
@@ -703,6 +704,36 @@ public static function prevent_empty_post_meta( $check, $object_id, $meta_key, $
...
+			return $meta_value;
+		}
+
+		// If the post is not federated and older than a year, return local visibility.
+		$date = \get_the_date( 'U', $object_id );
+		if ( $date < \strtotime( '-1 year' ) ) {
```
obenland: What's the rationale for the 1 year window? Why would we not want to have this applied generally?

includes/class-activitypub.php
```diff
@@ -703,6 +704,36 @@ public static function prevent_empty_post_meta( $check, $object_id, $meta_key, $
 		return $check;
 	}
 
+	/**
+	 * Get post meta.
```
```suggestion
	 * Adjusts default post meta values.
```
Maybe something like that? Could probably still be more descriptive

includes/class-activitypub.php
pfefferle: Because we can simply change the default to `local` then. The idea is, to not federate older posts when they are edited. Happy to discuss the timeslot though.

obenland: Please bear with me as I'm still trying to wrap my head around this:

If the post is not federated, would the post status give us an indication what to go with? Return `local` when the post is published? Would we ever want to change the status for an already published post?

pfefferle: As described in the isse, the idea was to use `local` as default for older posts (posts that were published before installing the plugin), so that they will not be published when edited.

If we would not add a timeframe, we would have to check for the post status, because otherwise it would always show `local` even for new posts (because they are not federated yet).

obenland: It feels like a post status check might be a bit cleaner, less arbitrary? It would help solve the issue for Mastodon imports as well. What do you think?

pfefferle: I have a huge problem with using the post status for anything! We had so much issues (and still have) to properly detect publishes and updates!

pfefferle: Aside from that, I think it is not a good indicator for an "old" blog post!?

pfefferle: @obenland I am not sure if your proposed change is worth the risk! Changing the code, the way you suggested, results in a ton of failed unit-tests.

Screenshot 2025-07-10 at 12 35 36

that means that there is high potential that this changes something of the default behaviour.

obenland: I assume the failures and errors are due to incomplete test setup, where we don't set the visibility meta on every post we create :/

pfefferle: yes, but I still see a risk that this will change too much of the default behaviour!?

obenland: Yeah, it's super brittle. When we determine whether to create an Outbox object for a newly published post, it checks for the visibility meta. For some reason, even if I set the `meta_input` to visibility => public, it doesn't save that value and returns the default.

pfefferle: @obenland oh! A whole new issue! There is no `public` stored in the meta! `public` is the lack of a meta, to not flood the db with default entries.
So we do not even have a change to do it without a time-frame!

obenland (approved):

PR #1903, opened 2025-07-07
Signature: Examine public key when no algorithm passed

obenland:
The `alg` param is optional in RFC 9421, and when it's not provided we should examine the public key for its algorithm.

## Proposed changes:

* Improved handling of empty algorithm strings by inferring the algorithm from the public key details.
* Moved supported algorithm definitions to a class property and updated verify_algorithm() to use this property instead of a local map.
* Added test.

### Other information:

### DIFF
--- a/includes/signature/class-http-message-signature.php
+++ b/includes/signature/class-http-message-signature.php
@@ -23,7 +23,7 @@ use Activitypub\Collection\Actors;
 class Http_Message_Signature implements Signature_Standard {
 
 	/**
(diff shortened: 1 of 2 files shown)


pfefferle:
@obenland you have to walk me through all your signature reworks at some time soon!

obenland:
Deal! 🙂

pfefferle (dismissed):

pfefferle (approved):

PR #1904, opened 2025-07-07
Follow: Make sure Actor doesn't already exist before saving

obenland:
This allows for webfingers and actor URLs to be passed to `ActivityPub\follow()` and handles resolving them to actor IDs through `Actors::fetch_remote_by_uri()`

## Proposed changes:

* Use `upsert()` to add in another `get_remote_by_uri()` check before deciding whether to insert or update.
* Enhances the CLI documentation to show how to specify a different user with the `--user` flag.


pfefferle (approved):
thanks @obenland

PR #1905, opened 2025-07-07
Following: Undo the correct Follow activity

obenland:
When creating the Undo activity for unfollowing someone, it was possible to accidentally undo someone else's follow, or another remote-actor-specific Activity.
With these changes we're now querying for Follows specifically, generated by the passed user.

## Proposed changes:

* Updated the unfollow method to filter by author and activity type when querying outbox items. 
* Improved PHPDoc comments for return types.
* Added comprehensive unit tests for the unfollow functionality, including checks for correct Undo post creation.

### Other information:


includes/collection/class-following.php
```diff
@@ -133,6 +136,14 @@ public static function unfollow( $post, $user_id ) {
 						'key'   => '_activitypub_object_id',
 						'value' => $post->guid,
 					),
+					array(
+						'key'   => '_activitypub_activity_type',
+						'value' => 'Follow',
+					),
+					array(
+						'key'   => '_activitypub_activity_actor',
+						'value' => $actor_type,
```
obenland, line 145: @pfefferle What's the benefit of adding actor type? Would we expect that to ever be different from `author`?

pfefferle: For blog actor and application actor the user id is `0` (because we can not use `-1` as user id) so we need the actor type to address the "virtual" users properly.

pfefferle (approved):

PR #1907, opened 2025-07-08
Query: Don't double-query posts

obenland:
While testing Followers and Following I noticed that we run queries twice, between passing args to `WP_Query` and calling `get_posts()` on it. In reality it's probably cached, but still worth removing the overhead of running `get_posts()` twice.

## Proposed changes:

* Replaces `get_posts()` with direct access to the posts property or the `query()` method in collection classes.
* Updates related test to remove unnecessary get_posts() call.

### Other information:

## Testing instructions:

* Make sure lists of Followers and Following still show up as expected.
* When unfollowing someone, the Undo Outbox item should still be created.
* When processing followers for an outbox item, all inboxes should still be retrieved.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Minor performance improvement when querying posts of various types, by avoiding double queries.

</details>

### DIFF
--- a/.github/changelog/1907-from-description
+++ b/.github/changelog/1907-from-description
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Minor performance improvement when querying posts of various types, by avoiding double queries.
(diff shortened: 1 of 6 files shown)


obenland:
FWIW, I don't think this needs to get in before the next release, if we feel unsure about it.

pfefferle (approved):

PR #1909, opened 2025-07-09
Following v1: Make table look more like Core and add "pending" support

pfefferle:
I followed the recent changes done by @obenland and added styles and filters to show/hide pending follow requests.

Screenshot 2025-07-10 at 10 31 15

## Proposed changes:

* Re-arranges Following table to closer resemble users table, with avatar and username in one column.
* Date columns use better readable format.
* Fixes a bug where instead of creation and last modification of the following relationship, the creation and last modification of the actor was shown.
* Removed nonce check from search check as it doesn't make changes to following. It's just another query arg.
* Added initial views support in place of custom string with the number of followings.
* Updated Delete Followings logic to always assume a list of followings.
* Added search context in list table template.
* Made search submit button translatable and more contextual.
* Made hidden inputs dynamic in preparation for users and blog user to use the same template file.
* Split search form from follower form and made follower form use POST method.
* Added filter to see `All`, `Accepted` and `Pending` Followings

### Other information:

## Testing instructions:

* Go to Users > Following
* Search for a user and make sure search still works.
* Delete a follower and make sure it works.
* Sorting columns should also work.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

The following tables now more closely match the appearance of other WordPress tables and can be filtered by status.

</details>


includes/table/class-following.php
```diff
@@ -116,20 +121,99 @@ public function prepare_items() {
...
+
+	/**
+	 * Returns views.
+	 *
+	 * @return string[]
+	 */
+	public function get_views() {
+		$count_all      = Following_Collection::count_following( $this->user_id );
+		$count_pending  = Following_Collection::count_following( $this->user_id, 'pending' );
+		$count_accepted = Following_Collection::count_following( $this->user_id, 'accepted' );
```
obenland: I think it would be good to avoid individual queries to get this data. Maybe it's worth creating a separate function for this use-case, that returns followers grouped by status, similar to [`wp_count_comments()`](https://developer.wordpress.org/reference/functions/wp_count_comments/). It would also avoid introducing the status flag to these functions.

includes/table/class-following.php
```diff
@@ -116,20 +121,99 @@ public function prepare_items() {
...
+				'id'         => $following->ID,
 				'icon'       => \esc_attr( $actor->get_icon()['url'] ?? '' ),
 				'post_title' => \esc_attr( $actor->get_name() ),
 				'username'   => \esc_attr( $actor->get_preferred_username() ),
 				'url'        => \esc_attr( object_to_uri( $actor->get_url() ) ),
 				'identifier' => \esc_attr( $actor->get_id() ),
-				'published'  => \esc_attr( $actor->get_published() ),
-				'modified'   => \esc_attr( $actor->get_updated() ),
+				'published'  => \esc_attr( $following->post_date_gmt ),
+				'modified'   => \esc_attr( $following->post_modified_gmt ),
```
obenland: These dates don't line up with the time of the Following for me. I believe that's the actor's "birthday" and latest update?  With following relationships now tracked in post meta, we don't really have that information available to us anymore. These two columns should probably just be removed.

image

templates/following-list.php
```diff
@@ -0,0 +1,46 @@
...
+		<input type="hidden" name="tab" value="<?php echo esc_attr( $_tab ); ?>" />
+		<?php $table->search_box( esc_html__( 'Search Followings', 'activitypub' ), 'search' ); ?>
+	</form>
+
+	<form method="post">
+		<input type="hidden" name="page" value="<?php echo esc_attr( $_page ); ?>" />
+		<input type="hidden" name="tab" value="<?php echo esc_attr( $_tab ); ?>" />
+		<?php wp_nonce_field( 'bulk-' . $table->_args['plural'] ); ?>
+		<?php $table->display(); ?>
+	</form>
```
obenland, line 45: Taxonomy screens (/wp-admin/edit-tags.php?taxonomy=category) have some meta information below the term table. Maybe that is a convention we could re-use to explain how pending Follows work, and that it's up to the other person to accept them?

includes/table/class-following.php
```diff
@@ -176,31 +257,56 @@ public function column_avatar( $item ) {
...
+	public function column_username( $item ) {
 		return sprintf(
-			'<a href="%s" target="_blank">%s</a>',
-			esc_url( $item['url'] ),
-			$item['url']
+			'<img src="%1$s" width="32" height="32" alt="%2$s" loading="lazy"/> <strong><a href="%3$s">%4$s</a></strong><br />',
+			\esc_url( $item['icon'] ),
+			\esc_attr( $item['username'] ),
+			\esc_url( $item['url'] ),
+			\esc_html( $item['post_title'] )
```
obenland: As a follow-up, I think it would be good to add an "Undo" action link here.

obenland (dismissed):
I understand it was me who suggested to make pending followings look like pending comments—seeing it in person it feels odd to be unable to approve these followings.

Maybe we should do the same as in the Outbox and use a label instead?

image

None of this I'd consider merge-blocking however, this is a solid step forward.

includes/table/class-following.php
pfefferle: is the bulk not enough?

obenland: I think it would be a nice addition for all the cases where you don't want to unfollow in bulk.

pfefferle:
@obenland I added some changes. I think the pending list (and showing pending Actors) is a nice feature but it should be a separate list. So I changed the view to have the Accpted Followings as the default and a pending list to show only the pending requests.

I think there is no need to have a mixed list at all. A nice addition to the `pending` list might be to add a "re-send follow request" in a follow up PR.

I also added extra functions for querying the pending followers and a function to get all counts by type.

| Accepted | Pending |
| -------- | ------- |
| Screenshot 2025-07-11 at 09 46 10  | Screenshot 2025-07-11 at 09 46 14    |

obenland:
>  I think the pending list […] should be a separate list. […] I think there is no need to have a mixed list at all.

Could you expand on why it should be separate? The rest of WordPress doesn't separate lists by status, so this feels rather inconsistent. Are there examples outside of WordPress that do that?

pfefferle:
Because you can't do something to get a pending Follow Request to become Accepted. Most other platform do not even list them somewhere.

I think it is nice to have that list, but it is only a FYI thing and to maybe re-send the follow request.

pfefferle:
> The rest of WordPress doesn't separate lists by status

This is not completely true btw. A comment with the status `spam` or `trash` will not be listed in the comments overview page!

obenland:
Would you be willing to take another look at this decision? I think it would be a jarring experience to send a new following request and then not see it in the Following list because it's hidden in the Pending view. To me Views are a filtering mechanism, which would not be the experience this creates.

I agree with the notion of not being able to do anything about the Pending status, which led me to [suggest using a label](https://github.com/Automattic/wordpress-activitypub/pull/1909#pullrequestreview-3007044013), like the posts/pages table does, to give it more of an informational feel. What do you think about that?

> A comment with the status spam or trash will not be listed in the comments overview page!

Both of these require an action to become spam/trash. Initially we even tried to emulate the Comments list.

pfefferle:
Re-added `All` tab. Thanks to the feedback @obenland and @davemart-in !

Screenshot 2025-07-14 at 18 55 38

obenland (approved):

Jiwoon-Kim:
In Misskey, the Privacy tab includes a toggle labeled "Manually approve follow requests."

Mastodon also has a checkbox labeled "Automatically accept new followers" (or similar wording), allowing users to control whether new follow requests are approved automatically or manually.

PR #1913, opened 2025-07-09
Followers: Make table look more like Core

obenland:
Before | After
--- | ---
Screenshot 2025-07-09 at 3 21 40 PM | Screenshot 2025-07-09 at 3 21 29 PM
Screenshot 2025-07-09 at 3 40 02 PM | Screenshot 2025-07-09 at 3 39 45 PM

## Proposed changes:

* Re-arranges Followers table to closer resemble users table, with avatar and username in one column.
* Date columns use better readable format.
* Fixes a bug where instead of creation and last modification of the follow relationship, the creation and last modification of the actor was shown.
* Removed nonce check from search check as it doesn't make changes to followers. It's just another query arg.
* Added initial views support in place of custom string with the number of followers.
* Updated Delete Followers logic to always assume a list of followers.
* Added search context in list table template.
* Made search submit button translatable and more contextual.
* Made hidden inputs dynamic in preparation for users and blog user to use the same template file.
* Split search form from follower form and made follower form use POST method.

## Testing instructions:

* Go to Users > Followers
* Search for a user and make sure search still works.
* Delete a follower and make sure it works.
* Sorting columns should also work.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Follower tables now look closer to what other tables in WordPress look like.

</details>


obenland:
There might be other/better information we'd want to show in this table, than the timestamps. Happy to improve all that.

I thought I'd propose this as a first step and get your feedback, before combining it with blog user template and updating Following tables.

pfefferle:
Have tested it on mobile? That was the issue why I rebuilt it the way it looks now.

obenland:
I hadn't, good callout. I added screenshots of the mobile view to the PR description.

Do you remember what types of issues you ran into?

pfefferle:
It showed only the avatar without the name!

obenland:
Ah, maybe because it was in a column of its own?

The changes in this PR pretty much do the same as the Users list.

pfefferle:
Really nice change! I also updated the [`Following` PR](https://github.com/Automattic/wordpress-activitypub/pull/1909) to follow your improvements! Props!!

Btw. I updated your PR to have the same template for both follower lists (blog-user and authors)!

pfefferle (dismissed):

pfefferle (approved):

PR #1918, opened 2025-07-11
Fix Followers Count: `Followers::count_followers()`

pfefferle:
`Followers::get_followers_with_count()` and `Followers::count_followers()` have used different queries, so the list of followers was not always consistant with the counter.

Screenshot 2025-07-11 at 09 25 57

Also deprecate `get_all_followers` in favour of `Activitypub\Collection\Actors::get_all`, because based on the used query, it only returns Actors with no check for the following relation.

## Proposed changes:

* Use the same query for both functions

### Other information:

## Testing instructions:

* Add a follower with no inbox post-meta
* Check the result of the list-followers page

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fixed an issue where the number of followers shown didn’t always match the actual follower list.

</details>

### DIFF
--- a/.github/changelog/1918-from-description
+++ b/.github/changelog/1918-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed an issue where the number of followers shown didn’t always match the actual follower list.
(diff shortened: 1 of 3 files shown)


obenland (approved):
I did some experimentation with different query adjustments to see if they would improve performance, but it looks like there's more value in re-using the same query (and having it cached) than optimizing individual queries but having to run more of them.

PR #1919, opened 2025-07-11
Importer: Fediverse Starter Kits

pfefferle:
This is a first draft to support Starter Kits.

Screenshot 2025-07-11 at 12 28 37

The code supports the format defined by @dansup for pixelfed: https://github.com/pixelfed/starter-kits/issues/1

With this feature, we might be able to supersede the implementation of the event plugin: https://codeberg.org/Event-Federation/wordpress-event-bridge-for-activitypub/commit/f231ece788574937a5653853d6bb4816432bd65c

It currently only supports following Actors and ignores everything else, but we can work on supporting other features with the further development of the plugin.

I worked on the feature, because it also helps to test the following implementation.

Here are two demo files:

[starter-kit-2.json](https://github.com/user-attachments/files/21314307/starter-kit-2.json)
[starter-kit.json](https://github.com/user-attachments/files/21314308/starter-kit.json)

## Proposed changes:

* Add an importer for Fediverse Starter Kits

### Other information:

## Testing instructions:

* Go to the importer and try to import one of the linked kits

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added initial support for Fediverse Starter Kits, allowing users to follow recommended accounts from a predefined list.

</details>


includes/wp-admin/import/class-starter-kit.php
```diff
@@ -0,0 +1,295 @@
...
+				++$skipped;
+				continue;
+			}
+
+			$result = follow( object_to_uri( $item ), self::$author );
+
+			if ( \is_wp_error( $result ) ) {
+				++$skipped;
+			} else {
+				printf( '<p>%s%s</p>', \esc_html__( 'Followed ', 'activitypub' ), \esc_html( $item['id'] ) );
```
obenland:
```suggestion
				/* translators: %s: Account ID */
				printf( '<p>' . \esc_html__( 'Followed %s', 'activitypub' ) . '</p>', \esc_html( $item['id'] ) );
```

obenland (commented):
Neat! I'd probably give it a namespace, like Fediverse Starter Kit or ActivityPub Starter Kit to make it clearer what it does, but this looks pretty good!

Does it work? 🙂

pfefferle:
Sure it does :)

I added two demo files!

obenland (approved):
This is probably good enough to merge already.

I noticed it doesn't skip actors where we have a pending follow request, that should probably added.

PR #1928, opened 2025-07-14
Remove "Followed" date

### LINKED ISSUE #1927: ActivityPub Followers
### Quick summary

Unfortunately, I can't tell, if this happened with Release 7.0.1 or already in 7.0.0. But it seems the page

wp-admin/users.php?page=activitypub-followers-list

is confusing the different date attributes of a following account.

The "followed" date seems to display the original account creation date instead of the date this user followed the blog. As well, the "last updated" column appears to be empty.

This looks like a bug to me, as it has been working nicely before. 

### Steps to reproduce

Visit the activitypub-followers-list:

Image

### Site owner impact

### Severity

Minor

_No response_

_No response_

pfefferle:
Now that we use the persistence for following and followers, we can not say for sure if that date is the date of following or followed or simply the date when they were imported through an @-reply.

This PR removes it as quick fix. If we want to have that information, we would need to have it in a separate PR, using a post-meta for example.

Fixes #1927 

## Proposed changes:

* Remove "Followed" date

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Removed follower dates to avoid confusion, as they may not have accurately reflected the actual follow time.

</details>

### DIFF
--- a/templates/followers-list.php
+++ b/templates/followers-list.php
@@ -16,7 +16,7 @@ $table->prepare_items();
 ?>
 <div class="wrap">
 	<?php if ( ! $_tab ) : ?>
-	<h1 class="wp-heading-inline"><?php esc_html_e( 'Author Followers', 'activitypub' ); ?></h1>
+	<h1 class="wp-heading-inline"><?php esc_html_e( 'Followers', 'activitypub' ); ?></h1>
 	<?php endif; ?>
 
 	<?php
--- a/.github/changelog/1928-from-description
+++ b/.github/changelog/1928-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Removed follower dates to avoid confusion, as they may not have accurately reflected the actual follow time.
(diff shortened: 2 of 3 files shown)


templates/followers-list.php
```diff
@@ -16,7 +16,7 @@
 ?>
 <div class="wrap">
 	<?php if ( ! $_tab ) : ?>
-	<h1 class="wp-heading-inline"><?php esc_html_e( 'Author Followers', 'activitypub' ); ?></h1>
+	<h1 class="wp-heading-inline"><?php esc_html_e( 'Followers', 'activitypub' ); ?></h1>
```
obenland, line 19: ✨

obenland (approved):
This is good to go as-is. 

Would you want to remove "Modified" as well? It doesn't feel super useful.

Jiwoon-Kim:
As I understand it, we need to separate the account registration date from the date when a follow action occurred. It feels somewhat similar to how Instagram handles account metadata. In the case of Mastodon, you can refer to how relationships are managed here: https://mastodon.social/relationships.

https://misskey.io/@pfefferle@mastodon.social/raw

PR #1934, opened 2025-07-15
Following: Add info section about follow requests

obenland:
image

## Proposed changes:

* Added an informational section explaining the meaning of 'Pending' follow requests and the approval process in the ActivityPub protocol.


obenland:
@pfefferle Happy to drop this if you don't think that's helpful. It adds some context, similar to category and tags lists in Posts > Category.

pfefferle:
Should we mention something like "be patient and simply wait until it gets approved, don't be pushy"?

pfefferle (approved):

PR #1935, opened 2025-07-15
Followers: First pass at blocking accounts

obenland:
image

Screenshot 2025-07-15 at 3 13 20 PM

Depends on #2027

## Proposed changes:
This PR proposes a two-step block process, that allows users to block accounts individually or in bulk. It comes with an "Are you sure?" step, similar to how User deletes work in Core.

This is probably something we might want to add to the Following list as well, in which case we probably should think about a parent list class that follower and following can inherit from.

### Other information:

## Testing instructions:

* Go to Users > Followers ⁂ in wp-admin
* Click "Block" link next to any follower OR select multiple followers and choose "Block" from bulk actions
* Verify confirmation screen shows correct account details and blocking options
* Test blocking with both individual and site-wide options checked
* Confirm blocked actors appear in Users > Blocked Actors ⁂ page
* Test unblocking functionality from the Blocked Actors page
* Verify appropriate permissions (only users with 'manage_options' can do site-wide blocks)

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add follower blocking functionality with confirmation screens and Blocked Actors management

</details>


Jiwoon-Kim:
Hmm… I think this could become a more general-purpose implementation if blocking were possible not only between remote actors but also between local actors. While this may not be very meaningful for personal WordPress use cases (as opposed to forums or social networks), it could help maintain logical consistency.

Also, since blog actors are not the instance themselves, I believe we should distinguish between object types when handling blocking. In that sense, separating the ability to block individual users and to block at the instance level (remote user/instance) seems like a more objective approach.

Additionally, at the instance level, it might be more useful to allow for moderation-level actions such as soft-hiding or de-prioritizing content rather than enforcing a strict block. This would provide more nuanced control over federation behavior, especially in cases where full blocking may be too aggressive or unnecessary.

includes/table/class-followers.php
```diff
@@ -104,6 +104,90 @@ public function process_action() {
 				}
 				break;
 
+			case 'block':
+				$redirect_to = \remove_query_arg( array( 'follower', 'followers' ), $redirect_to );
+
+				// Handle single follower block.
+				if ( isset( $_GET['follower'], $_GET['_wpnonce'] ) ) {
+					$follower = \esc_url_raw( \wp_unslash( $_GET['follower'] ) );
```
pfefferle: why not also use the ID? it makes the code easier to maintain and allow a more generic handling of the block mechanism below.

includes/table/class-followers.php
```diff
@@ -104,6 +104,90 @@ public function process_action() {
...
+							// Add actor ID to disallowed keys.
+							$disallowed_keys .= "\n" . $actor->get_id();
+
+							// Remove follower relationship.
+							Follower_Collection::remove( $follower, $this->user_id );
+
+							// If user is admin for blog actor, also remove relationships for blog actor.
+							if ( \user_can( $this->user_id, 'manage_options' ) ) {
+								Follower_Collection::remove( $follower, Actors::BLOG_USER_ID );
+							}
```
pfefferle: what about encapsulating that part, to be able to use it for single and bulk handling?

includes/table/class-followers.php
obenland: Fixed in [dc4ecc7](https://github.com/Automattic/wordpress-activitypub/pull/1935/commits/dc4ecc7e838e0fa8b89410db6dd57d650b04449b).

pfefferle:
I gave that case some more thought. it's a tricky one because it mixes up user-specific actions with blog-wide behavior.

Scenario:

* User A blocks a profile. As a result, that user is removed from their personal followers list, but also added to the global block list.
* Now, if User B also had that same remote user as a follower, they might be confused when they suddenly stop receiving messages from them.

Jiwoon-Kim:
I think it would be helpful to separate user-level block lists (similar to following/follower fields) from instance-level moderation tools. For example, soft-hiding or excluding flagged accounts from search results. It’s not clear how blocking is currently implemented, but UI-based filtering using per-user blocklists might be necessary.

I originally thought implementing a site-level superuser role would help manage this better — not for enforcing instance-wide blocks, but to allow individual users to block or unblock others on their own terms. This would enable more nuanced control without resorting to blanket moderation.

obenland:
> I gave that case some more thought. it's a tricky one because it mixes up user-specific actions with blog-wide behavior.

Agreed, that's generally the case with how blocking currently works.

We could limit the inline action to the site owner, to make it essentially an instance-level tool?

pfefferle:
> We could limit the inline action to the site owner, to make it essentially an instance-level tool?

might be a good compromise to start with. we could limit the block action to the blog-user UI, but we should be consistent with the follow up process. after a block, we should:

* remove all connections to all local users, so that no users will see the remote-actor in a local followers/following list.
* if someone follows the blockerd user, we should maybe also send an email notification, that the site admin blocked this account!?

pfefferle:
we should also block outgoing follow requests for accounts on the block list!?

pfefferle:
So maybe it makes sense to once again evaluate a dedicated block list aside the comment block list.

Jiwoon-Kim:
I realize this is a separate issue, but just to mention — I personally dislike email notifications since I rarely check my inbox. I think a PWA-integrated notification block would be much more useful, though I understand it's currently a low priority.

Jiwoon-Kim:
Mute(user) / Silence(admin)

1. https://misskey.io/settings/mute-block
2. https://mastodon.social/severed_relationships
 `https://mastodon.social/severed_relationships`

The `https://mastodon.social/severed_relationships` page lists **publicly severed federation relationships** between the Mastodon instance and other remote instances.

## 🔍 What It Shows

Each entry on the page typically includes:

- **Domain Name**: The remote instance that is blocked or defederated.
- **Reason (optional)**: Such as spam, abuse, hate speech, or user safety.
- **Block Type**:
  - `suspend`: Full block, removes all content and blocks all federation.
  - `silence`: Federation remains, but content does not appear in public timelines.
  - `limit`: Partial feature restrictions.
- **Date of Action**: When the block or limitation was enforced.

## 🎯 Purpose

This list serves to maintain **transparency** and support **community-level moderation** in the decentralized Fediverse environment. Other instance administrators can:

- Reference it when forming their own moderation policies.
- Understand the context of federation behavior with Mastodon.social.

## 🪟 Relevance to WordPress + ActivityPub

If you're using a federated WordPress site via the ActivityPub plugin, a similar feature could enhance moderation capabilities:

- Ability to **block or silence remote actors or instances** at the site level.
- Maintain a **public block list** similar to `severed_relationships`.
- Support nuanced moderation strategies like:
  - Soft-hiding actors
  - Deprioritizing content in feeds or search results
- Provide **transparency and discoverability** regarding federation policies.

## ✅ Suggestions for WordPress ActivityPub Plugin

- Introduce a `severed_relationships`-style endpoint (e.g., `/severed_relationships.json`)  
- Add admin panel tools for managing:
  - Instance-level blocks
  - Actor-level blocks
  - Block reason and metadata
- Make moderation actions visible in a public or semi-public way, as an opt-in feature

pfefferle:
@obenland there is a merge conflict and what do you think about my latest comment:

> So maybe it makes sense to once again evaluate a dedicated block list aside the comment block list.

(I am not ignoring this PR)

obenland:
Yeah, based on that comment I've deprioritized this a bit. Can pick it up again later today maybe or next week. I switched the PR back to draft to better reflect its actual status

includes/wp-admin/table/class-following.php
```diff
@@ -123,11 +124,34 @@ public function process_action() {
...
+				// Check if user has blocked the account.
+				if ( \in_array( $profile, Moderation::get_user_blocks( $this->user_id )['actors'], true ) ) {
+					/* translators: %s: Account profile that could not be followed */
+					\add_settings_error( 'activitypub', 'followed', \sprintf( \__( 'Unable to follow account &#8220;%s&#8221;. The account is blocked.', 'activitypub' ), \esc_html( $profile ) ) );
+					$redirect_to = \add_query_arg( 'resource', $profile, $redirect_to );
+					break;
+				}
+
+				// Check if site has blocked the account.
+				if ( \in_array( $profile, Moderation::get_site_blocks()['actors'], true ) ) {
```
pfefferle, line 148: we should have a dedicated function for that check, something like `Moderation::is_actor_blocked( $uri )` to not have too much array comparisons in the places we want to have this check.

pfefferle: This way, we could also easily support any kind of Actor representation (Object, Array or URI).

templates/bulk-block-confirmation.php
```diff
@@ -0,0 +1,88 @@
...
+			<p>
+				<label>
+					<input type="checkbox" name="site_wide" value="1" />
+					<?php esc_html_e( 'Also block these accounts site-wide (affects all users and the blog actor).', 'activitypub' ); ?>
+				</label>
+			</p>
+		<?php endif; ?>
+
+		<p class="submit">
+			<?php submit_button( __( 'Confirm Block', 'activitypub' ), 'primary', 'submit', false ); ?>
```
pfefferle: KISS.

```suggestion
			<?php submit_button( __( 'Confirm', 'activitypub' ), 'primary', 'submit', false ); ?>
```

templates/block-confirmation.php
```diff
@@ -0,0 +1,71 @@
...
+			<p>
+				<label>
+					<input type="checkbox" name="site_wide" value="1" />
+					<?php esc_html_e( 'Also block this account site-wide (affects all users and the blog actor).', 'activitypub' ); ?>
+				</label>
+			</p>
+		<?php endif; ?>
+
+		<p class="submit">
+			<?php submit_button( __( 'Confirm Block', 'activitypub' ), 'primary', 'submit', false ); ?>
```
pfefferle:
```suggestion
			<?php submit_button( __( 'Confirm', 'activitypub' ), 'primary', 'submit', false ); ?>
```

includes/wp-admin/table/class-following.php
obenland: Sure! Okay to do in a follow-up?

pfefferle:
let's merge this one into the feature branch. it feels easier to review the PR this way.

PR #1938, opened 2025-07-16
CLI: Add command to add follower for testing

obenland:
## Proposed changes:

* Introduces the `add_follower` WP-CLI command to add a follower to a user's followers list for testing purposes. The command accepts an actor URL and an optional user argument, defaulting to the blog actor if not specified.


pfefferle:
I see this in the same category as https://github.com/Automattic/wordpress-activitypub/pull/1075

We should not add test/debug/dev cli commands, because this could confuse users.

pfefferle:
Oh nice, you also brought over my comment generator 😍

pfefferle:
How do I best test this PR?

When running `wp-env run cli wp activitypub add_follower pfefferle@mastodon.social --user=1` I see an error:

Screenshot 2025-07-24 at 13 37 18

obenland:
Ah, I should have added that to the testing steps. I use `npm run env -- run cli wp activitypub add_follower pfefferle@mastodon.social --user=1`, but I think your version with the global wp-env should work, too. Did you have docker running when you tried it?

pfefferle:
yes, docker was running. you can see the error message above. something because of security concerns.

...but it works with the command you provided! 👍

pfefferle (approved):

PR #1939, opened 2025-07-16
Tables: Combine actor search term sanitization

obenland:
## Proposed changes:

* Moved and unified actor search term sanitization into a new trait.
* Updated followers and following tables to use this method.

### DIFF
--- a/includes/table/trait-actor-list-table.php
+++ b/includes/table/trait-actor-list-table.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * Actor Table Trait file.
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\Table;
+
+/**
+ * Actor Table Trait.
+ */
+trait Actor_List_Table {
+
+	/**
+	 * Sanitize an actor search term.
+	 *
+	 * @param string $search The search term.
+	 * @return string The sanitized search term.
+	 */
+	public static function sanitize_search_term( $search ) {
+		$search = \sanitize_text_field( $search );
+		$search = \str_replace( array( 'acct:', 'http://', 'https://', 'www.' ), '', $search );
+		$search = \str_replace( '@', ' ', $search );
+
+		return \trim( $search );
+	}
+}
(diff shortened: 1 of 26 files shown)


pfefferle:
I agree that we should encapsulate the functionality, but I am not sure about the sanitization class. The only real sanitization is `sanitize_text_field` the rest is more a preparation thing and not even a good one. What about a trait? Maybe we can also encapsulate the admin-notices then?

While working on "Add Follower" I am quite unhappy with it.

For example: If you search for `@pfefferle@notiz.blog` I would expect to only find my blog profile, but it also shows my Mastodon profile, because I link to my blog from Mastodon. Maybe we have to rethink simply removing the `@` and replace the WebFinger handle with the URL/ID instead.

includes/table/trait-actor-list-table.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+trait Actor_List_Table {
+
+	/**
+	 * Sanitize an actor search term.
+	 *
+	 * @param string $search The search term.
+	 * @return string The sanitized search term.
+	 */
+	public static function sanitize_search_term( $search ) {
```
pfefferle: 😩

pfefferle: if we still call it "sanitize", then it can stay in the `Sanitize` Class!? my argument for moving it, was that it is not (only) about sanitizing, but does a full prepare/optimize for a better result.

obenland: Any suggestions on what else it should be called?

I guess we're sanitizing the input for the purpose that we're using it for, not just security vulnerabilities?

pfefferle: is that technically still `sanitizing`? (you are the native speaker (now))

obenland: I guess it depends on how loosely we want to define it. Maybe it should be `normalize_search_term()`?

> If clarity is important, it’s often better to distinguish between:
	•	Sanitizing → making safe (security-focused)
	•	Validating → checking if it meets required rules
	•	Normalizing → transforming to a consistent, usable form

pfefferle: I think I can live with `normalize_search_term()` ☺️

pfefferle (approved):

pfefferle:
thanks ♥️

PR #1940, opened 2025-07-16
Fix: readme of modal module

Menrath:


obenland (approved):
Nice catch, thank you!

PR #1953, opened 2025-07-17
Tables: Use Settings Errors to pass admin notices

obenland:
## Proposed changes:

* Switches from query parameter-based admin notices to using WordPress settings errors for follower and following deletion actions.

### DIFF
--- a/includes/table/class-followers.php
+++ b/includes/table/class-followers.php
@@ -57,8 +57,21 @@ class Followers extends \WP_List_Table {
 			return;
 		}
 
+		if ( ! $this->current_action() ) {
+			return;
+		}
+
+		$redirect_to = \add_query_arg(
+			array(
+				'settings-updated' => true,  // Tell WordPress to load settings errors transient.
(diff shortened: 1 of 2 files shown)


includes/table/class-followers.php
```diff
@@ -57,8 +57,21 @@ public function process_action() {
...
+		$redirect_to = \add_query_arg(
+			array(
+				'settings-updated' => true,  // Tell WordPress to load settings errors transient.
+				'action'           => false, // Remove action parameter to prevent redirect loop.
+			)
+		);
+
 		switch ( $this->current_action() ) {
 			case 'delete':
+				$redirect_to = \remove_query_arg( array( 'follower', 'followers' ), $redirect_to );
```
obenland, line 73: We talked about not removing query args in a central place—this introduces that handling on a per-action basis to avoid having to do it individually in each individual/bulk handling case. 

Not quite central, but also not too repetitive? Let me know what you think!

pfefferle (approved):

PR #1930, opened 2025-07-14
Following v1: Follow Dialogue

pfefferle:
Add a first version of a Follow form, that allows Users to follow other Actors by Username or Profile link!

follow

## Proposed changes:

* Add Follow form
* Hook into search to suggest following Actors if they were not found
* Improve search handling

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Go to following admin page and try to follow Actor-Accounts

### Changelog entry

-   [X] Automatically create a changelog entry from the details below.

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

-   [X] Added - for new features
-   [ ] Changed - for changes in existing functionality
-   [ ] Deprecated - for soon-to-be removed features
-   [ ] Removed - for now removed features
-   [ ] Fixed - for any bug fixes
-   [ ] Security - in case of vulnerabilities

#### Message

Added a first version of the Follow form, allowing users to follow other Actors by username or profile link.

</details>

PR #1955, opened 2025-07-17
Following: Search in All when no status is set

obenland:
Fixes a bug where searching from the "All" view would switch views to "Accepted" and not return all results.

## Proposed changes:

* Removes the default 'accepted' status filter and only includes the status parameter in the form if it is set. This allows the following list to be searched without always defaulting to 'accepted' status.

### DIFF
--- a/templates/following-list.php
+++ b/templates/following-list.php
@@ -17,7 +17,7 @@ global $following_list_table;
 $_search = \sanitize_text_field( \wp_unslash( $_REQUEST['s'] ?? '' ) );
 $_page   = \sanitize_text_field( \wp_unslash( $_REQUEST['page'] ?? '' ) );
 $_tab    = \sanitize_text_field( \wp_unslash( $_REQUEST['tab'] ?? '' ) );
-$_status = \sanitize_text_field( \wp_unslash( $_REQUEST['status'] ?? 'accepted' ) );
+$_status = \sanitize_text_field( \wp_unslash( $_REQUEST['status'] ?? '' ) );
 
 $following_list_table->prepare_items();
 ?>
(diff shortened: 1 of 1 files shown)


templates/following-list.php
```diff
@@ -17,7 +17,7 @@
 $_search = \sanitize_text_field( \wp_unslash( $_REQUEST['s'] ?? '' ) );
 $_page   = \sanitize_text_field( \wp_unslash( $_REQUEST['page'] ?? '' ) );
 $_tab    = \sanitize_text_field( \wp_unslash( $_REQUEST['tab'] ?? '' ) );
-$_status = \sanitize_text_field( \wp_unslash( $_REQUEST['status'] ?? 'accepted' ) );
+$_status = \sanitize_text_field( \wp_unslash( $_REQUEST['status'] ?? '' ) );
```
pfefferle, line 20: I also have the change in my follow pr

templates/following-list.php
```diff
@@ -39,7 +39,9 @@
 	<form method="get">
 		<input type="hidden" name="page" value="<?php echo esc_attr( $_page ); ?>" />
 		<input type="hidden" name="tab" value="<?php echo esc_attr( $_tab ); ?>" />
-		<input type="hidden" name="status" value="<?php echo esc_attr( $_status ); ?>" />
+		<?php if ( ! empty( $_status ) ) : ?>
```
pfefferle, line 42: is this necessary if stats defaults to empty string?

obenland: Apparently an empty string constitutes an 'unknown' view and "All" loses its `current` status.

pfefferle: this is fixed in my PR without the check...

obenland:
Closing in favor of #1930.

PR #1959, opened 2025-07-18
Ensure that all schedulers are registered

pfefferle:
Ensure that all schedulers are registered during every new migration process. This guarantees they are refreshed at least every one to two weeks.

This also protects against accidental cleanup or deletion of schedules, as previously experienced by @jeherve.

## Proposed changes:

* Run `Scheduler::register_schedules()` on every migration process.

### Other information:

## Testing instructions:

* Remove one of the schedules
* Run the migration
* Check if all schedules are properly set

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Ensure that all schedulers are registered during every plugin update.

</details>

### DIFF
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -752,8 +752,6 @@ class Migration {
 	public static function add_default_settings() {
 		self::add_activitypub_capability();
 		self::add_default_extra_field();
-
-		Scheduler::register_schedules();
 	}
 
 	/**
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -199,8 +199,7 @@ class Migration {
 			}
 		}
 
-		// Always register schedules, to ensure they are registered even
-		// if the plugin is activated after the first activation.
+		// Ensure all required cron schedules are registered.
 		Scheduler::register_schedules();
 
 		/*
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -160,13 +160,9 @@ class Migration {
 			add_action( 'init', 'flush_rewrite_rules', 20 );
 		}
 		if ( \version_compare( $version_from_db, '5.0.0', '<' ) ) {
-			Scheduler::register_schedules();
 			\wp_schedule_single_event( \time(), 'activitypub_upgrade', array( 'create_post_outbox_items' ) );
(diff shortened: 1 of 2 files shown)


includes/class-migration.php
```diff
@@ -203,6 +199,10 @@ public static function maybe_migrate() {
 			}
 		}
 
+		// Always register schedules, to ensure they are registered even
+		// if the plugin is activated after the first activation.
```
obenland:
```suggestion
		// Ensure all required cron schedules are registered.
```
This might be all that's needed in terms of context?

Reading "ensure they are registered even if the plugin is activated after the first activation" makes me wonder what's happening there, what's going wrong, what needs fixing, if that makes sense?

It's probably fine for us to just keep this as a fallback. If it's to fix a bug that we've not investigated, we should investigate it and fix it 🙂

pfefferle: That case can have so many reasons! I also accidentally removed too many schedules, because of some were stuck! I would not assume a bug here! The logic is quite simple and I see nothing else that might delete a schedule.

pfefferle: @obenland because of the context: haven't you seen the slack about that issue? I hoped so and didn't want to reference private chats!

pfefferle: I see this similar to the default value of an option/meta! It can always happen that a db entry will be deleted (by cleanup plugins, manual db editing, backup/restore issues), so you have to build a system to compensate these cases.

obenland: It didn't mention deactivations/re-activations from what I could tell?

Like I said, I don't mind the code change and having this as a fallback/insurance. It's more about the context for it if that makes sense?

obenland: Let's just say we want it to make sure cron schedules are what they should be

pfefferle: Ok, maybe I have misunderstood 'context' 😉

obenland (approved):

includes/class-migration.php
pfefferle: I understand your point, but in this case we only have the context that one schedule was not existent!

PR #1963, opened 2025-07-18
Add support for development toolings

pfefferle:
This PR builds a foundation to have development tools that will be only included in a local dev environment and will not part of the release packages.

This allows us to have Dev-CLI commands like #1938 or #1075 that will not confuse end users.

## Proposed changes:

* Set wp-env to `development`
* Remove old Docker setup
* Move current debug/dev files to development folder
* Add development folder to .distignore to not be part of the release

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


.wp-env.json
```diff
@@ -2,6 +2,11 @@
 	"core": null,
 	"plugins": [ "." ],
 	"env": {
+		"development": {
+			"config": {
+				"WP_ENVIRONMENT_TYPE": "development"
+			}
+		},
```
obenland: It looks like it's `local` out of the box, in wp env, so we might not even need this

```
% npm run env -- run cli wp eval 'echo wp_get_environment_type();'

> env
> wp-env run cli wp eval echo wp_get_environment_type();

ℹ Starting 'wp eval echo wp_get_environment_type();' on the cli container. 

local
```

activitypub.php
```diff
@@ -84,10 +84,13 @@ function plugin_init() {
...
-	$debug_file = __DIR__ . '/includes/debug.php';
-	if ( \WP_DEBUG && file_exists( $debug_file ) && is_readable( $debug_file ) ) {
-		require_once $debug_file;
-		Debug::init();
+	// Load development tools.
+	if ( 'development' === wp_get_environment_type() ) {
+		$debug_file = __DIR__ . '/development/debug.php';
+		if ( \WP_DEBUG && file_exists( $debug_file ) && is_readable( $debug_file ) ) {
+			require_once $debug_file;
+			Debug::init();
```
obenland: Do we need the `\WP_DEBUG && file_exists( $debug_file ) && is_readable( $debug_file )` line in dev? Or can we just assume it to be true?

This would remove debug from other environments where `WP_DEBUG` is true. Not sure if that's what we want?

pfefferle: We can assume WP_DEBUG as true...  maybe we should also add a load.php file, to handle all files in there!

obenland: Alternatively, we could check if a folder exists that corresponds to the environment and load all files that it finds in there

```php
$env_dir  = __DIR__ . '/' . wp_get_environment_type();

if ( is_dir( $env_dir ) ) {
	foreach ( glob( $env_dir . '/*.php' ) as $file ) {
		require_once $file;
	}
}

pfefferle: 🤔

can you verify that the env is development when you restart?

pfefferle: should we move it to a ~dotfile~ dotfolder?

`/.development/` ???

obenland: > should we move it to a dotfolder?

I think it's fine to keep in a regular folder, like `bin`, `docs`, and others that don't get distributed.

obenland:
I like the direction of having a place for development-only code, I'm just not sure moving our debug stuff in there is the best way forward.

In the current state of the PR, all the debug stuff that's currently shipping with the plugin would no longer be included and I worry it would make it harder for us to help users debug their sites.

There's a benefit in allowing `debug.php` to be enabled just by setting `WP_DEBUG`, regardless of the environment a site is in. I use it on WP.com dev sites and I'm sure others do too. So having that separate, or independent from one-another, would probably be good.

pfefferle:
Are we talking about debug.php or class-debug.php? Because debug.php is not even deployed to Dotcom.

obenland:
> debug.php is not even deployed to Dotcom.

Maybe that changed with the updated sync scripts?

I use it primarily on my WoA site and started including it on my sandbox through a callback in `0-sandbox.php`.

pfefferle:
Oh 😳! Then we maybe should talk on Monday!!!

pfefferle:
reverted all debug classes/files. the PR now only provides the env, to be used by other implementations like the CLI extensions in the future.

@obenland can we please work on unifying the development/debugging initiatives? the debug file was never meant to be part of a DotCom release! Let's move all the code from there to the custom integration code and move things around!

obenland (approved):

PR #1967, opened 2025-07-18
HTTP: Make it easier to search for users to follow

obenland:
We should support submitting just a domain, without schema, when searching for an account to follow.
Most of the time folks will probably just copy/paste a user ID or URL and won't encounter this. For the cases where they type it in manually, I think we should be "smart" enough to add a schema and try that URL instead of returning a `The "object" is/has no valid URL` error.

## Proposed changes:

* Ensures that URLs without a scheme are prefixed with 'https://' before further processing.

### Other information:

## Testing instructions:

* Go to Users > Following.
* Type in `activitypub.blog` and hit Enter.
* Don't be greeted by an error message.


pfefferle:
Can you explain why you added that to the HTTP class instead of the follower/following classes? I fear that there might be unexpected side effects when we add this change to every possible HTTP request!?

obenland:
Up until this point in the code, `$actor_uri` could be a webfinger or URL, so this is the first time we should have a full URL with schema.

> I fear that there might be unexpected side effects when we add this change to every possible HTTP request!?

I wondered about that too. Is there a scenario where we wouldn't want to heal a URL with missing schema? The check is pretty specific just to that.

pfefferle:
One possible szenario: We store an invalid URL in the DB, because the final remote-check will be done with a different, fixed and normalized one?

pfefferle:
> validated and sanitized as early as possible

-- https://docs.wpvip.com/security/validating-sanitizing-and-escaping/

obenland:
> We store an invalid URL in the DB, because the final remote-check will be done with a different, fixed and normalized one?

Are there places where we store the URL we get, rather than the URL that gets returned in the remote object?
A URL without schema won't get past the `wp_http_validate_url()` call a few lines below that, so the expectation would be that that invalid, stored URL keeps erroring?

pfefferle:
> Are there places where we store the URL we get, rather than the URL that gets returned in the remote object?

This might not really be an argument for or against. We shouldn't do things only because it does not break actual code.

pfefferle:
> A URL without schema won't get past the wp_http_validate_url() call a few lines below that, so the expectation would be that that invalid, stored URL keeps erroring?

I'm not sure I fully understand the question. My concern is that the normalization happens too late, and that we, or other implementers, might want to rely on the result of that call to determine if the URL is valid. It's mainly about managing expectations, and I would prefer the normalization to happen right after or during the sanitization of the FORM request.

includes/table/class-following.php
```diff
@@ -123,7 +123,12 @@ public function process_action() {
 				}
 
 				$profile = \sanitize_text_field( \wp_unslash( $_REQUEST['activitypub-profile'] ) );
-				$result  = follow( $profile, $this->user_id );
+				if ( false === \strpos( $profile, '@' ) && empty( \wp_parse_url( $profile, PHP_URL_SCHEME ) ) ) {
```
pfefferle: The `\strpos( $profile, '@' )` is tricky, because the `@` is also used for profile URLs by Mastodon and a lot of other platforms.

One example `mastodon.social/@pfefferle`

obenland: I guess these URLs don't get the benefit of auto-healing schemes then. Do you have a suggestion?

pfefferle: I would assume that these are the most used URLs, so if we want to have the auto-schema feature, we should support these! (see my comment about IDs aboce).

We could check for slashes and/or missing dots in the host part of a potential WebFinger id.

pfefferle: What about

```suggestion
				if ( ! \is_email( \ltrim( $profile, '@' ) ) && empty( \wp_parse_url( $profile, PHP_URL_SCHEME ) ) ) {
```

obenland: Oh I like it!

pfefferle (approved):
thanks ♥️

PR #1969, opened 2025-07-19
Following: Add missing Outbox Activity

pfefferle:
The current code only adds the local Actor and the relation to the local User, but it does not send the `Follow` Activity.

This PR changes the code to use the correct function for that.

## Proposed changes:

* use `follow` instead of `Following::follow`
* proper separation of concerns

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1529,12 +1529,16 @@ function add_to_outbox( $data, $activity_type = null, $user_id = 0, $content_vis
 /**
  * Follow a user.
  *
- * @param string $remote_actor The Actor URL or WebFinger Resource.
- * @param int    $user_id      The ID of the WordPress User.
+ * @param string|int $remote_actor The Actor URL, WebFinger Resource or Post-ID of the remote Actor.
+ * @param int        $user_id      The ID of the WordPress User.
  *
  * @return \WP_Post|\WP_Error The ID of the Outbox item or a WP_Error.
  */
 function follow( $remote_actor, $user_id ) {
+	if ( \is_numeric( $remote_actor ) ) {
+		return Following::follow( $remote_actor, $user_id );
+	}
+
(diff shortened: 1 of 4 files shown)


pfefferle:
I think you addressed the problem with the inconsistent implementation of `follow` and `Following::follow` @obenland . I hope I have addressed them now.

includes/functions.php
```diff
@@ -1549,24 +1549,33 @@ function follow( $remote_actor, $user_id ) {
...
-		return $actor;
+/**
+ * Unfollow a user.
+ *
+ * @param string $remote_actor The Actor URL or WebFinger Resource.
+ * @param int    $user_id      The ID of the WordPress User.
+ *
+ * @return \WP_Post|\WP_Error The ID of the Outbox item or a WP_Error.
+ */
+function unfollow( $remote_actor, $user_id ) {
```
obenland, line 1567: How would you feel about waiting to add this function until it's necessary and has a use case? This would be the fifth unused function in `functions.php`.

obenland (commented):
> I think you addressed the problem with the inconsistent implementation of follow and Following::follow @obenland.

Yeah, I was advocating for moving the `$remote_actor` sanitization into `Following::follow` to follow the idea of being as lenient as possible with inputs when building APIs.

But relly I'm in favor of anything that moves stuff out of `functions.php` and into namespaced and feature-specific class methods 🙂

includes/functions.php
pfefferle: sure, but this is not meant to be used by us in the first place!

I think I mentioned that some time ago: I would love to point implementers to use these global functions instead of the Class functions, because Classes are still very fragile and we changed them a lot over the past few month. The global functions instead were pretty much untouched.

obenland (approved):

PR #1973, opened 2025-07-21
Fix #1970

### LINKED ISSUE #1970: Uncaught Error: Call to undefined method WP_Error::get_icon()
### Quick summary

Overview of the Follower in the Plugin Options page is broken.
Following PHP error comes up.

Uncaught Error: Call to undefined method WP_Error::get_icon()
in /www/htdocs/.../wp-content/plugins/activitypub/includes/table/class-followers.php on line 127

See enclosed screenshot.

Image

### Steps to reproduce

1. Open Plugin Options
2. Open Follower Section

### Site owner impact

### Severity

Moderate

_No response_

_No response_

pfefferle:
Fixes #1970 

## Proposed changes:

* Check if Actor transformation was properly executed and skip Actor element if not
* Add error to Actor object
* Improve cleanup method, to update Actor if remote request returns valid JSON

### Other information:

## Testing instructions:

* Add a follower/following
* Edit DB entry to a broken JSON
* Load follower/following table

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fixed a PHP error that prevented the Follower overview from loading.

</details>

### DIFF
--- a/includes/collection/class-actors.php
+++ b/includes/collection/class-actors.php
@@ -702,9 +702,11 @@ class Actors {
 		$post = \get_post( $post );
 
 		if ( ! $post ) {
+			$text = \__( 'Actor not found', 'activitypub' );
+			self::add_error( $post->ID, $text );
 			return new \WP_Error(
 				'activitypub_actor_not_found',
-				\__( 'Actor not found', 'activitypub' ),
+				$text,
 				array( 'status' => 404 )
 			);
 		}
@@ -715,7 +717,13 @@ class Actors {
 			$json = \get_post_meta( $post->ID, '_activitypub_actor_json', true );
 		}
 
-		return Actor::init_from_json( $json );
+		$actor = Actor::init_from_json( $json );
+
+		if ( \is_wp_error( $actor ) ) {
+			self::add_error( $post->ID, $actor );
+		}
+
+		return $actor;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


includes/table/class-followers.php
```diff
@@ -199,7 +199,13 @@ public function prepare_items() {
 		);
 
 		foreach ( $followers as $follower ) {
-			$actor     = Actors::get_actor( $follower );
+			$actor = Actors::get_actor( $follower );
+
+			if ( \is_wp_error( $actor ) ) {
+				Actors::add_error( $follower->ID, $actor );
```
obenland: Should we add the error inside of `Actors::get_actor()` when we return an error, rather than having calling code be responsible for it?

pfefferle: good point!

includes/collection/class-actors.php
```diff
@@ -702,9 +702,11 @@ public static function get_actor( $post ) {
 		$post = \get_post( $post );
 
 		if ( ! $post ) {
+			$text = \__( 'Actor not found', 'activitypub' );
+			self::add_error( $post->ID, $text );
```
obenland: yeah, without a `$post`, this won't work. Let's just limit it to the `init_from_json()` case you added below.

pfefferle: oh :D

obenland (approved):

PR #1978, opened 2025-07-21
Following: Improve follow error handling

obenland:
`The WebFinger Resource is not accessible.` or `The "object" is/has no valid URL` are not terribly great error messages. This PR updates the error message when we can't follow an account to something more user-friendly, albeit slightly less detailed.
It also improves user experience by re-supplying the attempted username, instead of forcing folks to type it back in again.

## Proposed changes:

* Removes resource query arg to not pre-fill the username again after submitting the follow form.
* Updated error message to be a bit more user-friendly.
* Re-fills the username after a failed follow attempt.
* Also updates missing html entities in form string (sorry, Claude went above and beyond)

### DIFF
--- a/includes/table/class-following.php
+++ b/includes/table/class-following.php
@@ -111,7 +111,7 @@ class Following extends \WP_List_Table {
 				}
 				break;
 			case 'follow':
-				$redirect_to = \remove_query_arg( array( 's' ), $redirect_to );
+				$redirect_to = \remove_query_arg( array( 'resource', 's' ), $redirect_to );
 
 				if ( ! isset( $_REQUEST['activitypub-profile'], $_REQUEST['_wpnonce'] ) ) {
 					return;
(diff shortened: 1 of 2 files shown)


includes/table/class-following.php
```diff
@@ -125,7 +125,9 @@ public function process_action() {
 				$profile = \sanitize_text_field( \wp_unslash( $_REQUEST['activitypub-profile'] ) );
 				$result  = follow( $profile, $this->user_id );
 				if ( \is_wp_error( $result ) ) {
-					\add_settings_error( 'activitypub', 'followed', $result->get_error_message() );
+					/* translators: %s: Account profile that could not be followed */
+					\add_settings_error( 'activitypub', 'followed', \sprintf( \__( 'Unable to follow account &#8220;%s&#8221;. Please verify the account exists and try again.', 'activitypub' ), \esc_html( $profile ) ) );
```
pfefferle, line 129: should we maybe (also) change the origin error message to be more desciptive and enduser friendly?

pfefferle (approved):

includes/table/class-following.php
obenland: Yes, I think that would be good

PR #1979, opened 2025-07-22
Improve WebFinger lookups

pfefferle:
This PR enhances the WebFinger fallback mechanism. The plugin now attempts to infer the WebFinger handle using details from the Actor JSON and by analyzing the URL structure.

Previously, the code tried to extract the username by parsing the URL's path component. However, URLs without a path (e.g., `https://fed.brid.gy/` or `https://john.onolan.org/`) weren’t handled correctly. In such cases, the full URL was mistakenly used as the WebFinger ID, resulting in an awkward format.

Screenshot 2025-07-22 at 11 41 00

## Proposed changes:

* Better fallback for empty PATH variables
* Actor::get_preferred_username() fallback

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Improved WebFinger fallback to better guess usernames from profile links.

</details>

### DIFF
--- a/includes/class-webfinger.php
+++ b/includes/class-webfinger.php
@@ -8,6 +8,7 @@
 namespace Activitypub;
 
 use WP_Error;
+use Activitypub\Activity\Actor;
 use Activitypub\Collection\Actors;
 
 /**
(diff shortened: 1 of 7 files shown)


includes/class-webfinger.php
```diff
@@ -302,11 +303,24 @@ public static function generate_cache_key( $uri ) {
...
-	 * @param string $uri The URI.
+	 * @param Actor|string $actor_or_uri The Actor or URI.
 	 *
 	 * @return string Hopefully the name of the Follower.
 	 */
-	public static function guess( $uri ) {
-		return extract_name_from_uri( $uri ) . '@' . \wp_parse_url( $uri, PHP_URL_HOST );
+	public static function guess( $actor_or_uri ) {
+		if ( ! $actor_or_uri instanceof Actor ) {
+			$actor = Actors::fetch_remote_by_uri( $actor_or_uri );
```
obenland, line 312: Making an external request here makes the function much heavier. Are you worried about the function getting in a way where it doesn't get actors passed and inadvertently causes lots of external requests, if used in a loop?

Is the path problem not solved by the change in `extract_name_from_uri()`?

(I'm asking to get more context, not asking for a change quite yet)

obenland (changes requested):
Let's add tests for `guess()` with the URLs that were identified as not being processed correctly.

includes/class-webfinger.php
pfefferle: this is only a fallback! normally we have all the infos or get the Actor passed!

pfefferle: > Is the path problem not solved by the change in extract_name_from_uri()?

It depends what you mean by solved. It does return something that looks like a WebFinger address, but it is not necessarily the WebFinger ID (at least in the cases I tested)!

obenland (approved):

PR #1982, opened 2025-07-22
Trigger Actor updates on (un)setting a post as sticky.

pfefferle:
Send Actor updates if the list of sticky posts change, to force remote platforms to re-read the "featured" endpoint.

Fixes: https://wordpress.com/blog/2025/07/18/3-fresh-features-activitypub-experience-for-wordpress/#comment-448212

## Proposed changes:

* Send Actor update when sticky posts list changes.

### Other information:

## Testing instructions:

* Add a sticky post or remove it
* Check the outbox for a new Actor Update Activity

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Trigger Actor updates on (un)setting a post as sticky.

</details>


Jiwoon-Kim:
I’m not entirely convinced that the current “Sticky post” feature in WordPress — which is typically toggled via the Quick Edit or post editor — can fully capture the intended semantics in a fediverse context.

From experience, the concept of “pinning” content on federated profiles (e.g. Mastodon or Misskey) is better represented and understood when explicitly configured within the actor’s profile feed. It’s a social signal, not just a display tweak.

That makes me wonder — should this feature be considered more of a presentation layer element (handled locally), or should it be mapped to an Update activity for the Actor object itself across the fediverse?

In either case, it seems crucial to define:

How pinned/sticky state is stored at the AP level (in actor metadata? as a featured collection?),

Whether this state syncs with existing sticky_posts logic or diverges for ActivityPub purposes.


Would love to hear your thoughts on how to bridge these models — especially regarding backend implementation and compatibility with other platforms.

Jiwoon-Kim:
While I agree that for single-profile "blog" actors this feature could be implemented relatively simply, I can foresee a potential side effect: the sticky status might also affect the site's native post feed (e.g., /posts or homepage query), which could unintentionally distort the intended layout or content flow.

In the ActivityPub context, pinning a post is usually more about visibility on the remote profile — it's a social affordance, not necessarily a site-wide editorial choice. If sticky posts start overriding local feed logic (e.g., query_posts() ordering), that might not be desirable for all users.

Perhaps it would make sense to differentiate "fediverse pinning" from traditional WordPress sticky_post behavior, or at least offer a toggle to prevent site-level layout interference when pinning posts for federation.

Jiwoon-Kim:
The existing WordPress `sticky_post` feature doesn't align well with how pinned posts are used in the Fediverse, where they are typically tied to the actor's profile feed rather than the site's main post loop. A better approach might be to handle this separately using a custom `meta_key` like `activitypub_sticky`, allowing pinned posts to be managed specifically for ActivityPub without affecting the site's layout or behavior.

includes/scheduler/class-actor.php
```diff
@@ -48,6 +48,8 @@ public static function init() {
 		\add_action( 'update_option_activitypub_actor_mode', array( self::class, 'blog_user_update' ) );
 
 		\add_action( 'transition_post_status', array( self::class, 'schedule_post_activity' ), 33, 3 );
+
+		\add_action( 'update_option_sticky_posts', array( self::class, 'detect_sticky_posts_change' ), 10, 2 );
```
obenland: Can we use the [`post_stuck` hook](https://developer.wordpress.org/reference/functions/stick_post/#hooks)? (and, [post_unstuck](https://developer.wordpress.org/reference/functions/unstick_post/#hooks))

pfefferle: Is there a possibility that we miss changes? Bulk edits or direct option changes?

obenland: Bulk edits should go through `edit_post()`. Direct option changes maybe, but I think that's fine. That would be true for all options.

pfefferle: That is why we use the options hooks for most options ☺️

But sure, I will change it!

obenland (approved):

PR #1994, opened 2025-07-23
Signature: Move HTTP request signing to a filter

obenland:
Follow up to #1986 and #1991.

Moves double-knock into a filter that's always set.
With `Signature` getting an init callback, we can now also move the signing logic into a hook callback.

## Proposed changes:

* Moved HTTP request signing to a filter in Signature::init and removed direct calls from Http class. 
* Updated Signature class to add filters for signing and double knock logic. 
* Adjusted tests to use filter-based approach for signature verification.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Moved HTTP request signing to a filter instead of calling it directly.

</details>


pfefferle:
You remove the `method` from the `$args` but it seems that `Http_Signature_Draft` is still using it in line 49.

I wonder why the tests are green 🤔

pfefferle:
Ah, the test is still adding the `method` in the `$args` array!

obenland:
Yes, that would be expected. Tests create signatures manually to sign "incoming" requests and they need the method provided.

When signing outgoing requests, `http_request_args` has a method set in those request args, [`GET` by default](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-includes/class-wp-http.php#L171), and any specific method based on the type of request is sent ([post](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-includes/class-wp-http.php#L636), [head](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-includes/class-wp-http.php#L672) ).

pfefferle (dismissed):

pfefferle:
nice job!

pfefferle (approved):

PR #2009, opened 2025-07-25
Create a persistent Inbox

pfefferle:
This PR is primarily intended to log all incoming requests that start with Update and Create, to help debug whether all supported Activity types are being correctly processed.

## Proposed changes:

* Introduce a global Inbox handler to manage all incoming requests, currently limited to Create and Update types.
* Add an Inbox Collection to persist the received entries.

### Other information:

## Testing instructions:

* Follow one of your Mastodon Accounts
* Publish a new status update
* Check the Inbox for entries

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

A global Inbox handler and persistence layer to log incoming Create and Update requests for debugging and verifying Activity handling.

</details>


includes/collection/class-inbox.php
```diff
@@ -0,0 +1,146 @@
...
+			\kses_init_filters();
+		}
+
+		if ( \is_wp_error( $id ) ) {
+			return $id;
+		}
+
+		if ( ! $id ) {
+			return false;
+		}
```
obenland: Can this be removed? With the wp_error args set to true in the `wp_insert_post()` call, it should never return a falsy value

pfefferle:
@obenland This should be ready to review now.

includes/handler/class-inbox.php
```diff
@@ -0,0 +1,76 @@
...
+	 * @param string             $type     The type of the activity.
+	 * @param Activity|\WP_Error $activity The Activity object.
+	 */
+	public static function handle_inbox_requests( $data, $user_id, $type, $activity ) {
+		/**
+		 * Filters the activity types to persist in the inbox.
+		 *
+		 * @param array $activity_types The activity types to persist in the inbox.
+		 */
+		$activity_types = \apply_filters( 'activitypub_persist_inbox_activity_types', array( 'Create', 'Update' ) );
```
obenland, line 42: Could you remind me why we don't use the type-specific handlers for this? I don't remember where we left off on that, if at all.

includes/collection/class-inbox.php
```diff
@@ -0,0 +1,140 @@
...
+
+		return $title;
+	}
+
+	/**
+	 * Get the inbox item by activity id.
+	 *
+	 * @param string $guid The activity id.
+	 *
+	 * @return int|null The inbox item id or null if not found.
```
```suggestion
	 * @returnarray|\WP_Error|\WP_Post The inbox item or an error.
```

includes/collection/class-inbox.php
```diff
@@ -0,0 +1,140 @@
+<?php
+/**
+ * Inbox collection file.
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\Collection;
+
+use Activitypub\Webfinger;
```
```suggestion
```

Seems unused.

includes/collection/class-inbox.php
```diff
@@ -0,0 +1,140 @@
...
+		}
+
+		$title      = self::get_object_title( $activity->get_object() );
+		$visibility = is_activity_public( $activity ) ? ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC : ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE;
+
+		$inbox_item = array(
+			'post_type'    => self::POST_TYPE,
+			'post_title'   => sprintf(
+				/* translators: 1. Activity type, 2. Object Title or Excerpt */
+				__( '[%1$s] %2$s', 'activitypub' ),
```
```suggestion
			'post_title'   => \sprintf(
				/* translators: 1. Activity type, 2. Object Title or Excerpt */
				\__( '[%1$s] %2$s', 'activitypub' ),
```

includes/collection/class-inbox.php
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	private static function get_object_title( $activity_object ) {
+		if ( ! $activity_object ) {
+			return '';
+		}
+
+		if ( is_string( $activity_object ) ) {
+			$post_id = url_to_postid( $activity_object );
+
+			return $post_id ? get_the_title( $post_id ) : '';
```
```suggestion
		if ( \is_string( $activity_object ) ) {
			$post_id = \url_to_postid( $activity_object );

			return $post_id ? \get_the_title( $post_id ) : '';
```

includes/collection/class-inbox.php
```diff
@@ -0,0 +1,140 @@
...
+	 *
+	 * @return int|null The inbox item id or null if not found.
+	 */
+	public static function get_by_guid( $guid ) {
+		global $wpdb;
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+		$post_id = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT ID FROM $wpdb->posts WHERE guid=%s AND post_type=%s",
+				\esc_url( $guid ),
```
```suggestion
				\esc_sql( $guid ),
```
Not sure if `esc_sql()` is more appropriate here? Or `esc_url_raw()` or something?

includes/collection/class-inbox.php
```diff
@@ -0,0 +1,140 @@
...
+	 * @return false|int|\WP_Error The added item or an error.
+	 */
+	public static function add( $activity, $user_id ) {
+		if ( \is_wp_error( $activity ) ) {
+			return $activity;
+		}
+
+		// Check for duplicate activity.
+		if ( ! \is_wp_error( self::get_by_guid( $activity->get_id() ) ) ) {
+			return false;
```
obenland: If there's already a post, should we return that post_id here? Or a WP_Error? Would be nice to not have to account for that `false`.

includes/class-activitypub.php
```diff
@@ -7,12 +7,13 @@
...
 
-use Exception;
-use Activitypub\Options;
+use Activitypub\Activity\Activity;
 use Activitypub\Collection\Actors;
+use Activitypub\Collection\Inbox;
 use Activitypub\Collection\Outbox;
 use Activitypub\Collection\Followers;
 use Activitypub\Collection\Extra_Fields;
+use Activitypub\Options;
```
```suggestion
```

Seems unused.

includes/class-activitypub.php
```diff
@@ -633,8 +634,127 @@ private static function register_post_types() {
 			)
 		);
 
+		// Register Inbox Post-Type.
+		register_post_type(
```
obenland, line 634: 🤔

Image | Image
--- | ---

includes/handler/class-inbox.php
pfefferle: the inbox should persist all incoming activities, this limitation is only for the time beeing to not flood the inbox in the testing phase. If I would use the type-specific ones, I would have to implement the storing in every class and would have to add even more.

includes/collection/class-inbox.php
```diff
@@ -0,0 +1,147 @@
...
+			),
+			'post_content' => wp_slash( $activity->to_json() ),
+			// ensure that user ID is not below 0.
+			'post_author'  => \max( $user_id, 0 ),
+			'post_status'  => 'pending',
+			'guid'         => $activity->get_id(),
+			'meta_input'   => array(
+				'_activitypub_object_id'             => object_to_uri( $activity->get_object() ),
+				'_activitypub_activity_type'         => $activity->get_type(),
+				'_activitypub_activity_actor'        => Actors::get_type_by_id( $user_id ),
```
obenland, line 63: Can you tell me more about the role of `post_author` and `_activitypub_activity_actor` for inbox items?

In my testing I had a user and the blog user follow the same account. When it posted the inbox item would only be saved once (as expected) but author and actor type were set to whoever received it first. Should `get_by_guid()` check for the post author as well, so we save one for each follower?

Image

pfefferle: yes, it should save two inbox entries!

includes/collection/class-inbox.php
```diff
@@ -0,0 +1,147 @@
...
+		// Check for duplicate activity.
+		if ( ! \is_wp_error( $item ) ) {
+			if ( \is_array( $item ) ) {
+				$item = \current( $item );
+			}
+
+			if ( $item instanceof \WP_Post ) {
+				return $item->ID;
+			}
+		}
```
obenland: Still a bit uncertain about this. Could this be simplified to  


```suggestion
		if ( $item instanceof \WP_Post ) {
			return $item->ID;
		}
```
`wpdb::get_var()` should not return an array, did you run into that in testing?

includes/collection/class-inbox.php
pfefferle: Maybe I have to rethink the usage of the GUID completely... Shouldn't the GUID be unique?

pfefferle:
@obenland added the user id to the getter.

obenland (approved):
Let's :shipit: 

We should probably follow it up with a scheduled purge, similar to the Outbox.

PR #2012, opened 2025-07-25
Tables: Refactor action URL generation

obenland:
## Proposed changes:

* Extracted action URL generation logic into a new private `get_action_url()` method in the `Actor_List_Table` trait.
* Updated Followers and Following tables to use this method, reducing code duplication and improving maintainability.


includes/wp-admin/table/trait-actor-list-table.php
```diff
@@ -48,4 +48,23 @@ public static function get_webfinger( $actor ) {
...
+	 */
+	private function get_action_url( $action, $follower ) {
+		return \wp_nonce_url(
+			\add_query_arg(
+				array(
+					'action'   => $action,
+					'follower' => $follower,
+				)
+			),
+			$action . '-follower_' . $follower
```
obenland, line 67: Any mention of `follower` here should probably be `actor` to be more broadly applicable? However, changing this would also mean that `process_action()` needs to have all its parameters updated.

pfefferle (approved):

PR #2013, opened 2025-07-25
Tables: Minor code changes

obenland:
## Proposed changes:

* Converted static method calls to instance method calls for `normalize_search_term` and `get_webfinger` in Followers and Following list tables. 
* Updated `trait-actor-list-table.php` to define these methods as non-static. 
* Also renamed `_is_followable` to `is_followable` as it has a visibility flag.
* Improved translation comment.


pfefferle:
I liked your initial solution to have them as static methods, because you do not really rely on the object-context and this way it might be even more reusable!?

Was it not on purpose when you initially defined them?

obenland:
No, I think it was just muscle memory because we use it almost everywhere else :)

`Actor_List_Table` feels pretty specific, if we want to reuse these functions outside of it, I'd probably suggest moving them into `Sanitize` and `Webfinger`.

pfefferle (approved):

PR #2021, opened 2025-07-29
Change environment for unittesting to `production`

pfefferle:
This PR sets the unittest environment to production.

As we begin adding tools and making changes to our local development setup, for example bypassing signature verification or other critical features, it’s important that tests continue to run in conditions that closely resemble production. To ensure this, we should prevent local tools from being loaded during testing. This can be achieved either by switching the environment or by explicitly unloading the local development extensions.

## Proposed changes:

* Move from `development` to `local` folder to have a consistant wording
* Update code to match these changes
* Set unittest env to `production`

### Other information:

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Set the unittest environment to staging to ensure tests run in production-like conditions, avoiding interference from local development tools.

</details>


tests/bootstrap.php
```diff
@@ -6,6 +6,7 @@
  */
 
 \define( 'ACTIVITYPUB_DISABLE_INCOMING_INTERACTIONS', false );
+\define( 'WP_ENVIRONMENT_TYPE', 'staging' );
```
obenland:
```
PHP Warning:  Constant WP_ENVIRONMENT_TYPE already defined in /wordpress-phpunit/wp-tests-config.php on line 118
```
I think we need to define it in `.wp-env.json` to override the default constant. When testing that, you'll need to stop and then start wp-env again, for the constant to be set correctly.

obenland (changes requested):

tests/bootstrap.php
pfefferle: I have not seen any errors, but I will re-check!

obenland:
```
% npm run env-test

> env-test
> wp-env run tests-cli --env-cwd="wp-content/plugins/activitypub" vendor/bin/phpunit

ℹ Starting 'vendor/bin/phpunit' on the tests-cli container. 

[29-Jul-2025 12:43:50 UTC] PHP Warning:  Constant WP_ENVIRONMENT_TYPE already defined in /wordpress-phpunit/wp-tests-config.php on line 118
Installing...
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 8.5.42 by Sebastian Bergmann and contributors.

```

obenland: I don't see it in phpunit GitHub action output. Either way, I think it should be in `.wp-env.json` to replace the default value for it.

pfefferle: seems to be not that easy :(

obenland: In what way?

pfefferle: the test seem to not take the environment from the .wp-env.json

obenland: try `npm run env-stop && npm run env-start && npm run env-test`

obenland (approved):

PR #2027, opened 2025-07-29
Add actor blocking functionality with list table interface

### LINKED ISSUE #418: Question: How does this project interact with moderation and trust & safety on the fediverse?
Hi! I've had a look through the changelog & through open issues, and can't see anything about this topic.

Will this plugin implement support for handling Flag activities, additionally, will it support (at a bare minimum) a mean through which to specify instances to not federate with? Or a way to block certain actors from interacting with the blog via the plugin?

Would be happy to provide advice either here or via the IFTAS matrix server.

obenland:
Fixes #418.

## Proposed changes:
* Implements a new actor blocking system similar to the existing following/followers functionality
* Uses post meta storage with `_activitypub_blocked_by` key on actor posts for scalable blocking
* Provides list table interface for managing blocked actors with add/unblock functionality
* Includes conditional menu registration that only shows if user has blocked actors
* Adds site-wide blocking support in Settings > ActivityPub tab
* Uses unified blocking API through `add_user_block`/`remove_user_block` methods

### Other information:

## Testing instructions:

* Go to Users > Blocked Actors (only appears if you have blocked actors)
* Try blocking an actor by entering `@username@domain.com` or profile URL
* Verify the actor appears in the blocked actors list
* Test unblocking individual actors and bulk unblock functionality
* Go to Settings > ActivityPub > Blocked Actors tab to test site-wide blocking
* Verify blocked actors cannot interact with your content

### Changelog entry

<details>

#### Significance

- [ ] Patch
- [x] Minor
- [ ] Major

#### Type

- [ ] Fixed - for any bug fixes

#### Message

Add actor blocking functionality with list table interface for managing blocked users and site-wide blocks

</details>


includes/wp-admin/table/class-blocked-actors.php
```diff
@@ -0,0 +1,406 @@
...
+				\printf(
+					/* translators: %s: Actor name. */
+					\esc_html__( 'Would you like to block %s?', 'activitypub' ),
+					\sprintf(
+						'<a href="%s">%s</a>',
+						\esc_url( \add_query_arg( 'resource', $search ) ),
+						\esc_html( $actor->post_title )
+					)
+				);
+			}
```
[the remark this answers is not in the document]
obenland: Not suer what it saw here, it does return a URL.

includes/wp-admin/table/class-blocked-actors.php
```diff
@@ -0,0 +1,406 @@
...
+			$actor = Actors::fetch_remote_by_uri( $search );
+			if ( ! \is_wp_error( $actor ) ) {
+				echo ' ';
+				\printf(
+					/* translators: %s: Actor name. */
+					\esc_html__( 'Would you like to block %s?', 'activitypub' ),
+					\sprintf(
+						'<a href="%s">%s</a>',
+						\esc_url( \add_query_arg( 'resource', $search ) ),
+						\esc_html( $actor->post_title )
```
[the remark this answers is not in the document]
obenland: AI misread the context. It's a WP_Post

templates/blocked-actors-list.php
```diff
@@ -0,0 +1,82 @@
...
+				<h2><?php echo esc_html__( 'Block Actor', 'activitypub' ); ?></h2>
+				<div class="form-wrap">
+					<form method="post" id="activitypub-block-form">
+						<?php wp_nonce_field( 'activitypub-block-nonce' ); ?>
+						<div class="form-field form-required">
+							<label for="activitypub-profile" class="screen-reader-text"><?php echo esc_html__( 'Profile link', 'activitypub' ); ?></label>
+							<input type="hidden" name="action" value="block" />
+							<input name="activitypub-profile" id="activitypub-profile" type="text" value="<?php echo esc_attr( $_resource ); ?>" size="40" aria-required="true" class="<?php echo $_resource ? 'highlight' : ''; ?>" />
+						</div>
+						<?php submit_button( esc_attr__( 'Block Actor', 'activitypub' ) ); ?>
```
pfefferle: I think `Block` is fine and it better aligns with the `Follow` button on the Following site.

```suggestion
						<?php submit_button( esc_attr__( 'Block', 'activitypub' ) ); ?>
```

templates/blocked-actors-list.php
```diff
@@ -0,0 +1,82 @@
...
+		<div id="col-left">
+			<div class="col-wrap">
+				<h2><?php echo esc_html__( 'Block Actor', 'activitypub' ); ?></h2>
+				<div class="form-wrap">
+					<form method="post" id="activitypub-block-form">
+						<?php wp_nonce_field( 'activitypub-block-nonce' ); ?>
+						<div class="form-field form-required">
+							<label for="activitypub-profile" class="screen-reader-text"><?php echo esc_html__( 'Profile link', 'activitypub' ); ?></label>
+							<input type="hidden" name="action" value="block" />
+							<input name="activitypub-profile" id="activitypub-profile" type="text" value="<?php echo esc_attr( $_resource ); ?>" size="40" aria-required="true" class="<?php echo $_resource ? 'highlight' : ''; ?>" />
```
pfefferle, line 53: The Following form has a label for the profile URI. I know you are not a fan of it, but we should be consistent, so either we should remove the label on the following form or add it to the block form.

pfefferle:
If I block someone I follow or who follows me, unblocking them doesn’t restore that previous connection—it only removes them from my block list. If I had been following them, I can choose to follow again, but if they were following me, that connection is lost. It could be helpful to have an option to restore a follow in case a block was made by mistake.

pfefferle:
I see why you’re using the `Moderation` class to manage the Blocked Actors, but it’s not consistent with the rest of the codebase, where we typically use collection classes to query lists. While this isn’t necessarily an issue right now (or for us directly), it could make the code harder to read and understand in the long run, especially for contributors.

This isn’t a blocker—I’m not sure what the best approach here would be—but I thought it was worth pointing out.

pfefferle:
FYI: I may consolidate all feedback into this PR, since #1935 is based on it. That makes it tricky to tell which UI changes come from which PR :)

pfefferle:
There’s an edge case where an incoming Activity might use the WebFinger handle as its Actor-ID. Should we address that in this PR, treat it as an improvement for later, or just wait until the case actually comes up?

includes/class-moderation.php
```diff
@@ -95,11 +101,45 @@ public static function activity_is_blocked_for_user( $activity, $user_id ) {
 	 */
 	public static function add_user_block( $user_id, $type, $value ) {
 		switch ( $type ) {
+			case 'actor':
```
pfefferle: What is the reason to have one function for all blocks instead of a dedicated function for blocking Actors? This function can become quite big and confusing.

Aside from that, I think we should use constants for `$type`.

includes/class-moderation.php
```diff
@@ -119,14 +159,50 @@ public static function add_user_block( $user_id, $type, $value ) {
 	 */
 	public static function remove_user_block( $user_id, $type, $value ) {
 		switch ( $type ) {
+			case 'actor':
```
pfefferle: same as above!

includes/class-moderation.php
```diff
@@ -256,4 +356,57 @@ private static function check_activity_against_blocks( $activity, $blocked_actor
...
+	 * @param array    $args    The WP_Query arguments.
+	 *
+	 * @return array {
+	 *      Data about the blocked actors.
+	 *
+	 *      @type \WP_Post[] $blocked_actors List of blocked Actor WP_Post objects.
+	 *      @type int        $total         Total number of blocked actors.
+	 *  }
+	 */
+	public static function get_blocked_actors_with_count( $user_id, $number = -1, $page = null, $args = array() ) {
```
pfefferle: Shouldn't these live in the Actor class instead?

It also might be an argument for having a dedicated `Actor_Blocks_Collection`!

includes/class-moderation.php
pfefferle: It also might be an argument for having a dedicated `Actor_Blocks_Collection`!

templates/blocked-actors-list.php
obenland: Fixed in [48d3211](https://github.com/Automattic/wordpress-activitypub/pull/2027/commits/48d321141ed6e97d5649f5aa21c1f0cbfa33975b).

obenland:
> There’s an edge case where an incoming Activity might use the WebFinger handle as its Actor-ID. Should we address that in this PR, treat it as an improvement for later, or just wait until the case actually comes up?

Fixed in [3c43505](https://github.com/Automattic/wordpress-activitypub/pull/2027/commits/3c43505462e60d9699192609c5f2d2d7b93acb46).

includes/class-moderation.php
obenland: Fixed in [625e23f](https://github.com/Automattic/wordpress-activitypub/pull/2027/commits/625e23f820dbd0eea1b9c0ad8ae4e2aa423685e8).

includes/class-moderation.php
obenland: > What is the reason to have one function for all blocks

To keep the API consistent and easy. Happy to have a blocked actors collection handle the management under the hood, but `Moderation` should be the top level API for it

includes/class-moderation.php
obenland: Fixed in d2736afdedfc4d5c6d0e321a190b458ddcbb384a.

includes/class-moderation.php
obenland: Constants fixed in d2736afdedfc4d5c6d0e321a190b458ddcbb384a.

obenland:
@pfefferle This should be ready for another look.

tests/includes/class-test-moderation.php
```diff
@@ -31,7 +31,7 @@ public function set_up(): void {
 		parent::set_up();
 
 		// Create a test user.
-		$this->test_user_id = $this->factory->user->create(
+		$this->test_user_id = self::factory()->user->create(
```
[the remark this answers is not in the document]
obenland: `$this->factory` is deprecated.

includes/wp-admin/table/class-blocked-actors.php
```diff
@@ -0,0 +1,407 @@
...
+	 *
+	 * @var int
+	 */
+	private $user_id;
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		if ( \get_current_screen()->id === 'settings_page_activitypub' ) {
```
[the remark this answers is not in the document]
obenland: It only gets loaded within an admin context.

pfefferle:
@obenland what about

> If I block someone I follow or who follows me, unblocking them doesn’t restore that previous connection—it only removes them from my block list. If I had been following them, I can choose to follow again, but if they were following me, that connection is lost. It could be helpful to have an option to restore a follow in case a block was made by mistake.

???

obenland:
Ah, sorry, I missed replying to that. 

Can that be a follow-up? I'm not even sure that's something that's desired or that we'd need to worry too much about.

pfefferle:
We should be at least sure about it, before we release it ;)

obenland:
I think for a v1 it's good enough to not support automatic re-following.

A lot of assumptions here, just as a heads up:
Blocking an actor is a pretty serious action that is probably not done lightly and relatively rare. Unblocking a blocked actor is probably even rarer, since there'll have been strong reasons to block in the first place and they would not have to be in place anymore. And then I'm not sure I'd consider unblocking the same as "everything is fine and normal again" and go back to the way it was.

Blocking is a pretty disruptive and destructive action, I think it's fine for it to be reflected in that way for the time being.

includes/wp-admin/table/class-following.php
```diff
@@ -123,11 +124,34 @@ public function process_action() {
...
 					$profile = \esc_url_raw( 'https://' . \ltrim( $profile, '/' ) );
 				}
 
+				// Check if user has blocked the account.
+				if ( \in_array( $profile, Moderation::get_user_blocks( $this->user_id )['actors'], true ) ) {
+					/* translators: %s: Account profile that could not be followed */
+					\add_settings_error( 'activitypub', 'followed', \sprintf( \__( 'Unable to follow account &#8220;%s&#8221;. The account is blocked.', 'activitypub' ), \esc_html( $profile ) ) );
+					$redirect_to = \add_query_arg( 'resource', $profile, $redirect_to );
+					break;
+				}
```
pfefferle: can you file an issue for this part to be refactored in a `Moderation::is_actor_blocked( $webfinger_or_uri_or_object_or_array )`?

pfefferle: The function should also have a possibility to check either user, site or both.

includes/wp-admin/table/class-following.php
```diff
@@ -123,11 +124,34 @@ public function process_action() {
 				}
 
 				$profile = \sanitize_text_field( \wp_unslash( $_REQUEST['activitypub-profile'] ) );
-				if ( ! \is_email( \ltrim( $profile, '@' ) ) && empty( \wp_parse_url( $profile, PHP_URL_SCHEME ) ) ) {
+				$profile = \trim( $profile, '@' );
+
+				if ( \filter_var( $profile, FILTER_VALIDATE_EMAIL ) ) {
+					$remote_actor = Webfinger::resolve( $profile );
+					if ( ! \is_wp_error( $remote_actor ) ) {
+						$profile = object_to_uri( $remote_actor );
```
pfefferle: With this change, the form input will always be replaced with the ID, even if you have used the URL or WebFinger to follow. Is that fine?

includes/wp-admin/table/class-following.php
obenland: Done! #2095.

includes/wp-admin/table/class-following.php
obenland: Oh man, I forgot about that. Yes, I noticed that, too. It feels a bit jarring but I couldn't think of a much better way. Maybe keep the original input separate?

obenland: Fixed in [6d0db54](https://github.com/Automattic/wordpress-activitypub/pull/2027/commits/6d0db5454184bf76304400246203f581989b88f0).

includes/class-moderation.php
pfefferle: the question is still, why a global function with a type, instead of dedicated functions? there is nearly no shared code between the actor blocks and the keyword blocks.

obenland: oh, because it started out using the same saving mechanism in options and user meta, so they all shared boilerplate code for adding/removing/checking.

In a subsequent step the actor stuff was removed, once it became clear it should get a user list treatment and a different way of storing that information

obenland: See https://github.com/Automattic/wordpress-activitypub/pull/2020/commits/15679ca5ce1e0474307509a62dca2f41403add53

pfefferle (approved):

PR #2034, opened 2025-07-31
Mimic Mastodons search experience

pfefferle:
This PR enables searching for (remote) URLs. If the URL has already been imported, it redirects to the corresponding reply, otherwise, it imports the object first. This allows users to bring in replies that didn’t arrive through the Inbox.

## Proposed changes:

* Lookup URLs and redirect to the comment thread
* Import remote replies

### Other information:

## Testing instructions:

* Delete a remote comment and search for the URL, to reimport it

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Adds support for searching (remote) URLs similar to Mastodon, redirecting to existing replies or importing them if missing.

</details>

### DIFF
--- a/includes/class-search.php
+++ b/includes/class-search.php
@@ -56,6 +56,8 @@ class Search {
 		$imported = self::try_import_activitypub_object( $search_term );
 
 		if ( $imported ) {
+			// Ensure the imported comment is approved/published.
+			\wp_set_comment_status( $imported, 'approve' );
 			$comment_link   = \get_comment_link( $imported );
 			$validated_link = \wp_validate_redirect( $comment_link, \home_url() );
 			if ( $validated_link ) {
@@ -104,6 +106,8 @@ class Search {
 		$imported = self::try_import_activitypub_object( $search_term );
 
 		if ( $imported ) {
+			// Ensure the imported comment is approved/published.
+			\wp_set_comment_status( $imported, 'approve' );
 			$comment_link   = \get_comment_link( $imported );
 			$validated_link = \wp_validate_redirect( $comment_link, \home_url() );
 			if ( $validated_link ) {
--- a/includes/class-search.php
+++ b/includes/class-search.php
@@ -56,7 +56,7 @@ class Search {
 
 		if ( $imported ) {
 			$comment_link   = \get_comment_link( $imported );
-			$validated_link = \wp_validate_redirect( $comment_link, home_url() );
+			$validated_link = \wp_validate_redirect( $comment_link, \home_url() );
 			if ( $validated_link ) {
 				\wp_safe_redirect( $validated_link );
 				exit;
--- a/includes/class-search.php
+++ b/includes/class-search.php
@@ -21,16 +21,17 @@ class Search {
 	 * Initialize the search enhancement.
 	 */
 	public static function init() {
-		\add_filter( 'pre_get_posts', array( self::class, 'enhance_search' ) );
+		\add_filter( 'pre_get_posts', array( self::class, 'enhance_public_search' ) );
+		\add_action( 'admin_init', array( self::class, 'enhance_admin_comment_search' ) );
 	}
 
 	/**
-	 * Enhance search functionality to check for URLs and ActivityPub objects.
+	 * Enhance public search functionality to check for URLs and ActivityPub objects.
 	 *
 	 * @param \WP_Query $query The WP_Query instance.
 	 * @return \WP_Query The modified query.
 	 */
-	public static function enhance_search( $query ) {
+	public static function enhance_public_search( $query ) {
 		// Check for a valid user session.
 		if ( ! is_user_logged_in() ) {
 			return $query;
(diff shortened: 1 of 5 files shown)


obenland:
Oh, neat

includes/class-search.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @return \WP_Query The modified query.
+	 */
+	public static function enhance_search( $query ) {
+		// Check for a valid user session.
+		if ( ! is_user_logged_in() ) {
+			return $query;
+		}
+
+		// Only enhance main search queries on frontend.
+		if ( ! $query->is_main_query() || ! $query->is_search() || is_admin() ) {
```
obenland:
```suggestion
		if ( ! $query->is_main_query() || ! $query->is_search() || \is_admin() ) {
```

includes/class-search.php
```diff
@@ -0,0 +1,104 @@
...
+		if ( ! \wp_http_validate_url( $search_term ) ) {
+			return $query;
+		}
+
+		// Try to import ActivityPub object.
+		$imported = self::try_import_activitypub_object( $search_term );
+
+		if ( $imported ) {
+			$comment_link   = \get_comment_link( $imported );
+			$validated_link = \wp_validate_redirect( $comment_link, home_url() );
```
obenland:
```suggestion
			$validated_link = \wp_validate_redirect( $comment_link, \home_url() );
```

obenland (dismissed):
It worked as expected, so happy to have this merged.

It did feel odd having to go to the front-end to do this. All of our Fediverse handling happens in wp-admin, so that felt a little out of place for me. I wonder if extending the Comment search in wp-admin's Comments list might be a more natural place for this? Not sure.

pfefferle:
> It did feel odd having to go to the front-end to do this. All of our Fediverse handling happens in wp-admin, so that felt a little out of place for me.

Maybe this will change with the reader experience, but for now you are right!

> I wonder if extending the Comment search in wp-admin's Comments list might be a more natural place for this? Not sure.

Good idea, I will give that a look! ~But I think I will have a follow up for that!~

pfefferle:
@obenland I added the comment search too!

includes/class-search.php
```diff
@@ -0,0 +1,154 @@
...
+
+	/**
+	 * Enhance public search functionality to check for URLs and ActivityPub objects.
+	 *
+	 * @param \WP_Query $query The WP_Query instance.
+	 * @return \WP_Query The modified query.
+	 */
+	public static function enhance_public_search( $query ) {
+		// Check for a valid user session.
+		if ( ! is_user_logged_in() ) {
```
obenland: This should probably be a capability check, so not any logged-in user can do that. `moderate_comments` or something? Same for the other callback.

includes/class-search.php
```diff
@@ -0,0 +1,154 @@
...
+ * and attempt to import ActivityPub objects if found, otherwise falls back to classic search.
+ */
+class Search {
+
+	/**
+	 * Initialize the search enhancement.
+	 */
+	public static function init() {
+		\add_filter( 'pre_get_posts', array( self::class, 'enhance_public_search' ) );
+		\add_action( 'admin_init', array( self::class, 'enhance_admin_comment_search' ) );
```
obenland:
```suggestion
		\add_action( 'load-edit-comments.php', array( self::class, 'enhance_admin_comment_search' ) );
```
Changing it to the load hook would allow you to remove most checks from `enhance_admin_comment_search()`. Should this live in wp-admin somewhere?

includes/class-search.php
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		$activity = array(
+			'type'   => 'Create',
+			'actor'  => $object['attributedTo'],
+			'object' => $object,
+		);
+
+		// Import the reply as a comment.
+		return Interactions::add_comment( $activity );
```
obenland, line 141: I noticed that the comment date/time is the time when the comment gets imported, not when it was created originally. I assume this would need updating down the stack.

obenland (commented):
Sorry, noticed a few more things

includes/class-search.php
pfefferle: Because this is not a problem of this PR but a general one, I will fix it in a follow up!

includes/class-search.php
pfefferle: is this importent? The object can only be imported if it is a remote reply, so it doesn't really matter who triggers it!?

obenland: I'd err on the side of being too strict than not enough.

pfefferle: but this is simply the same trigger as an inbox request. it can be seen as re-try of a missing item!

includes/class-search.php
pfefferle: @obenland https://github.com/Automattic/wordpress-activitypub/pull/2038

includes/class-search.php
pfefferle: done!

obenland (approved):

PR #2035, opened 2025-07-31
Send `Delete` Activity on User-Delete

pfefferle:
This PR implements user deletion handling in the ActivityPub system by sending a Delete activity when a user is deleted and making related infrastructure changes.

## Proposed changes:

- Adds user deletion handling to send Delete activities to followers when a user is deleted
- Refactors inbox distribution logic to handle Delete activities differently than other activity types
- Deprecates the `maybe_add_inboxes_of_blog_user` method and simplifies inbox selection

### Other information:

## Testing instructions:

* Delete a user with the `activitypub` capability
* Check Outbox for `Delete` entries

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Adds support for sending Delete activities when a user is removed.

</details>


includes/scheduler/class-actor.php
```diff
@@ -161,4 +167,44 @@ public static function sticky_post_update( $post_id ) {
...
+
+		// Get the actor before deletion to ensure we have the data.
+		$actor = Actors::get_by_id( $user_id );
+
+		Tombstone::bury( $actor->get_id() );
+		Tombstone::bury( $actor->get_url() );
+
+		if ( \is_wp_error( $actor ) ) {
+			return;
+		}
```
obenland:
```suggestion
		$actor = Actors::get_by_id( $user_id );
		if ( \is_wp_error( $actor ) ) {
			return;
		}
		
		Tombstone::bury( $actor->get_id() );
		Tombstone::bury( $actor->get_url() );
```
Let's check for error before calling a method on it.

includes/scheduler/class-actor.php
```diff
@@ -161,4 +167,44 @@ public static function sticky_post_update( $post_id ) {
...
+
+	/**
+	 * Remove outbox from post types to delete with user.
+	 *
+	 * @param array $post_types The post types to delete with user.
+	 *
+	 * @return array The post types to delete with user without outbox.
+	 */
+	public static function post_types_to_delete_with_user( $post_types ) {
+		return \array_diff( $post_types, array( 'ap_outbox' ) );
```
obenland:
```suggestion
		return \array_diff( $post_types, array( \Activitypub\Collection\Outbox::POST_TYPE ) );
```

includes/class-http.php
```diff
@@ -52,7 +52,7 @@ public static function post( $url, $body, $user_id ) {
 				'Date'         => \gmdate( 'D, d M Y H:i:s T' ),
 			),
 			'body'                => $body,
-			'key_id'              => Actors::get_by_id( $user_id )->get_id() . '#main-key',
+			'key_id'              => \json_decode( $body )->actor . '#main-key',
```
obenland, line 55: You're positive that this is safe?

includes/collection/class-followers.php
```diff
@@ -331,6 +331,8 @@ public static function get_inboxes_for_activity( $json, $actor_id, $batch_size =
 	 * @return bool True if the Inboxes of the Blog User should be added, false otherwise.
 	 */
 	public static function maybe_add_inboxes_of_blog_user( $json, $actor_id ) {
+		\_deprecated_function( __METHOD__, 'unreleased' );
```
obenland, line 337: Let's add a `@deprecated unreleased` annotation to function docs.

includes/collection/class-followers.php
```diff
@@ -310,14 +312,12 @@ public static function get_inboxes( $user_id ) {
...
-		if ( self::maybe_add_inboxes_of_blog_user( $json, $actor_id ) ) {
-			$inboxes = \array_fill_keys( $inboxes, 1 );
-			foreach ( self::get_inboxes( Actors::BLOG_USER_ID ) as $inbox ) {
-				$inboxes[ $inbox ] = 1;
-			}
-			$inboxes = \array_keys( $inboxes );
+		$activity = \json_decode( $json, true );
+		// Only if this is an Update or Delete. Create handles its own "Announce" in dual user mode.
+		if ( \in_array( $activity['type'] ?? null, array( 'Update', 'Delete' ), true ) ) {
+			$inboxes = Actors::get_inboxes();
```
obenland, line 318: The way I'm reading this, it would send a post update to all inboxes we have on file, even if they're not following the actor or blog actor. Is that accurate?

includes/class-http.php
pfefferle: What do you fear?

pfefferle: If the body is not formated properly it will fail anyways... so yes, I am pretty positive in that case!

includes/scheduler/class-actor.php
```diff
@@ -161,4 +168,45 @@ public static function sticky_post_update( $post_id ) {
...
+	 *
+	 * @deprecated unreleased
+	 *
+	 * @param array $post_types The post types to delete with user.
+	 *
+	 * @return array The post types to delete with user without outbox.
+	 */
+	public static function post_types_to_delete_with_user( $post_types ) {
+		return \array_diff( $post_types, array( Outbox::POST_TYPE ) );
+	}
```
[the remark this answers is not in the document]
pfefferle: Can you propose a code change??

includes/collection/class-followers.php
obenland: Should it do that?

I guess I'd understand if we wanted to cast a wide net for actor deletions, but this is much broader.

includes/scheduler/class-actor.php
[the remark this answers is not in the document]
obenland: I'm not sure Copilot is smart enough for conversations yet

includes/collection/class-followers.php
pfefferle: As far as I know, Mastodon is pinging all known instances with the deletes, to take care of delted users that might be known to instances through boosts and @-replies aside from direct follows/followings.

obenland: maybe we could limit this to Deletes where the activity -> object -> type is in Actor::TYPES?

pfefferle: But what is the difference between a user and a post? A post could also be boosted to instances that might not be the instance of a follower or it might be an @-mention of a user for example.

To get sure an Object will be deleted from the fediverse it should be shared with as much instances as possible. Not ideal, but I see no real other solution right now.

Btw. The idea of the old code was the same, we simply had no central place to store Actors back in the days when it was implemented.

obenland: I see. Do you remember why we included Updates in that?

pfefferle: I think Matt wanted to be sure that updates would also be spread as broad as possible, but I am not sure we should keep doing that. But I am happy to re-add `Update`.

obenland: Yeah, let's limit it to only Deletes for now

pfefferle:
@obenland ready to merge?

obenland:
[It still sends updates to all inboxes](https://github.com/Automattic/wordpress-activitypub/pull/2035/files#diff-de86b44bd1f5493579d94614f1eb268672e37ebd8429a4f426ba4455d99a7230R317), did I misunderstand you wanted to limit that to Deletes?

pfefferle:
oh no... it was a long day! thought that I removed it alrady with the initial PR 🫣

pfefferle:
@obenland done!

includes/collection/class-followers.php
```diff
@@ -310,14 +312,12 @@ public static function get_inboxes( $user_id ) {
...
-		$inboxes = self::get_inboxes( $actor_id );
-
-		if ( self::maybe_add_inboxes_of_blog_user( $json, $actor_id ) ) {
-			$inboxes = \array_fill_keys( $inboxes, 1 );
-			foreach ( self::get_inboxes( Actors::BLOG_USER_ID ) as $inbox ) {
-				$inboxes[ $inbox ] = 1;
-			}
-			$inboxes = \array_keys( $inboxes );
+		$activity = \json_decode( $json, true );
+		// Only if this is an Update or Delete. Create handles its own "Announce" in dual user mode.
```
```suggestion
		// Only if this is a Delete. Create handles its own "Announce" in dual user mode.
```

obenland (approved):

PR #2039, opened 2025-07-31
Starter Kits: Add support for list of IDs

### LINKED ISSUE #2037: Feature Request: Support for importing non-inlined starter kits
### What

Heya. I'm currently helping out fedidevs.com with a little bit of ActivityPub support for their starter kits. There's ongoing work in anze3db/fedidevs/pull/177 if you want to watch (or partake).

Since this plugin is the only publicly existing consumer of starter kits that I know of, I figure it makes sense to check that we're compatible. 🙂

My current plan (which is still being discussed) is to publish fedidevs.com starter packs with the `items` listed as IDs instead of as inlined objects, like this:

```json
{
    "type": "Collection",
    "id": "https://example.org/s/abcde/",
    "name": "Test Starter Pack",
    "totalItems": 2,
    "items": [
        "https://example.org/users/test",
        "https://example.social/p/a487c0ae-5218-4632-b1a3-59053a79494f"
    ]
}
```

As I assume you know, this is equivalent JSON-LD to having the objects inlined, and an ActivityPub recipient accepting one should also accept the other.

Now, it's been well over a decade since I did any PHP programming to speak of, so I'm not 100% confident. But if I'm reading [this](https://github.com/Automattic/wordpress-activitypub/blob/47d1d9d030f93629f5a71784132d31e088929295/includes/wp-admin/import/class-starter-kit.php#L239) and [this](https://github.com/Automattic/wordpress-activitypub/blob/47d1d9d030f93629f5a71784132d31e088929295/includes/functions.php#L1648) correctly, then it looks like your importer doesn't yet support this kind of object reference. Since it looks like you're transforming the objects back into IDs right afterwards anyway, maybe it'd be a quick fix. But if you need to do any checks on the object (looks like you're validating the types?), maybe it's not trivial.

I'm filing this as a feature request since it doesn't currently break anything, but if your goal is to be able to parse all valid ActivityPub starter kits conforming to Dan's schema, this would technically be a bug.

### Why

tl;dr: Readiness for compatibility with fedidevs.com's ActivityPub starter packs whenever we launch them.

### How

_No response_

obenland:
Fixes #2037.

## Proposed changes:

* Updated the import logic to consistently treat items as actor IDs, ensuring they are strings and valid URLs before attempting to follow.

Demo files:

[starter-kit.json](https://github.com/user-attachments/files/21314308/starter-kit.json) (actor objects)
[starter-kit-2.json](https://github.com/user-attachments/files/21535677/starter-kit-2.json) (list of actor ids and webfingers)

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Added support for list of Actor IDs in Starter Kits.

</details>

### DIFF
--- a/includes/wp-admin/import/class-starter-kit.php
+++ b/includes/wp-admin/import/class-starter-kit.php
@@ -243,10 +243,7 @@ class Starter_Kit {
 		$items = self::$starter_kit['items'] ?? self::$starter_kit['orderedItems'] ?? array();
 
 		foreach ( $items as $actor_id ) {
-			if ( ! \is_string( $actor_id ) ) {
-				$actor_id = object_to_uri( $actor_id );
-			}
-
+			$actor_id = object_to_uri( $actor_id );
 			$actor_id = \ltrim( $actor_id, '@' );
 
 			if ( ! filter_var( $actor_id, FILTER_VALIDATE_URL ) && ! filter_var( $actor_id, FILTER_VALIDATE_EMAIL ) ) {
--- a/includes/wp-admin/import/class-starter-kit.php
+++ b/includes/wp-admin/import/class-starter-kit.php
@@ -242,19 +242,23 @@ class Starter_Kit {
 
 		$items = self::$starter_kit['items'] ?? self::$starter_kit['orderedItems'] ?? array();
 
-		foreach ( $items as $item ) {
-			if ( ! is_actor( $item ) ) {
+		foreach ( $items as $actor_id ) {
+			if ( ! \is_string( $actor_id ) ) {
+				$actor_id = object_to_uri( $actor_id );
+			}
+
+			if ( ! filter_var( $actor_id, FILTER_VALIDATE_URL ) ) {
 				++$skipped;
 				continue;
 			}
 
-			$result = follow( object_to_uri( $item ), self::$author );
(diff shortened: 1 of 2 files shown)


includes/wp-admin/import/class-starter-kit.php
```diff
@@ -242,19 +242,23 @@ public static function follow() {
 
 		$items = self::$starter_kit['items'] ?? self::$starter_kit['orderedItems'] ?? array();
 
-		foreach ( $items as $item ) {
-			if ( ! is_actor( $item ) ) {
+		foreach ( $items as $actor_id ) {
+			if ( ! \is_string( $actor_id ) ) {
+				$actor_id = object_to_uri( $actor_id );
+			}
```
pfefferle:
```suggestion
			$actor_id = object_to_uri( $actor_id );
```

This check is already done by `object_to_uri`.

obenland: Indeed! Fixed in [9fcb478](https://github.com/Automattic/wordpress-activitypub/pull/2039/commits/9fcb4781547db6facdbae1b4975fb2e97447448e).

pfefferle (approved):

PR #2042, opened 2025-08-01
Simplify Code and remove usage of `get__id` where possible

pfefferle:
We currently check for an Actor too frequently and often generate Actors even when only the User-ID is needed. This PR cleans up most `get__id()` cases, paving the way to deprecate the Actor models in favor of a more generic Actor.

While working on Delete, I noticed that we rely too heavily on unnecessary Actor generation, which complicates the process when the User is missing.

## Proposed changes:

* Reduce usage of `get__id()` calls
* Unified usage of `validate_user_id` in REST API
* Deprecated usage of `username` as user-id in REST API URLs
* Add option to return user-id from `Actors::get_by_*` functions

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -70,7 +70,7 @@ class Dispatcher {
 		$activity = Outbox::get_activity( $outbox_item );
 
 		// Send to mentioned and replied-to users. Everyone other than followers.
-		self::send_to_additional_inboxes( $activity, $actor->get__id(), $outbox_item );
+		self::send_to_additional_inboxes( $activity, $outbox_item->post_author, $outbox_item );
 
 		if ( self::should_send_to_followers( $activity, $actor, $outbox_item ) ) {
 			\do_action(
@@ -96,9 +96,9 @@ class Dispatcher {
 	 * @return array|void The next batch of followers to process, or void if done.
 	 */
 	public static function send_to_followers( $outbox_item_id, $batch_size = ACTIVITYPUB_OUTBOX_PROCESSING_BATCH_SIZE, $offset = 0 ) {
-		$json    = Outbox::get_activity( $outbox_item_id )->to_json();
-		$actor   = Outbox::get_actor( \get_post( $outbox_item_id ) );
-		$inboxes = Followers::get_inboxes_for_activity( $json, $actor->get__id(), $batch_size, $offset );
+		$outbox_item = \get_post( $outbox_item_id );
+		$json        = Outbox::get_activity( $outbox_item_id )->to_json();
+		$inboxes     = Followers::get_inboxes_for_activity( $json, $outbox_item->post_author, $batch_size, $offset );
 
 		$retries = self::send_to_inboxes( $inboxes, $outbox_item_id );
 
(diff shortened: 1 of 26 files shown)


pfefferle:
@akirk maybe you can also have a look! it seems that the EMA integration was broken since quite some releases!

akirk:
@pfefferle what could have been broken? I didn't notice it and my unit tests were fine, maybe I need to add more.

pfefferle:
Only in the integration part of the AP plugin.

obenland:
`% npm run env-test -- --filter test_create_item_with_multiple_recipients_and_inactive_recipient` fails for me locally, but not here for some reason…

obenland:
ah, I was missing commits

includes/class-dispatcher.php
```diff
@@ -96,9 +96,9 @@ public static function process_outbox( $id ) {
 	 * @return array|void The next batch of followers to process, or void if done.
 	 */
 	public static function send_to_followers( $outbox_item_id, $batch_size = ACTIVITYPUB_OUTBOX_PROCESSING_BATCH_SIZE, $offset = 0 ) {
-		$json    = Outbox::get_activity( $outbox_item_id )->to_json();
-		$actor   = Outbox::get_actor( \get_post( $outbox_item_id ) );
-		$inboxes = Followers::get_inboxes_for_activity( $json, $actor->get__id(), $batch_size, $offset );
+		$outbox_item = \get_post( $outbox_item_id );
+		$json        = Outbox::get_activity( $outbox_item_id )->to_json();
```
obenland, line 100:
```suggestion
		$json        = $outbox_item->post_content;
```
Would this work, too?

includes/class-migration.php
```diff
@@ -955,11 +955,17 @@ public static function update_notification_options() {
 		\add_option( 'activitypub_blog_user_mailer_new_follower', $new_follower );
 		\add_option( 'activitypub_blog_user_mailer_new_mention', '1' );
 
+		$users = \get_users(
+			array(
+				'capability__in' => array( 'activitypub' ),
```
obenland, line 960:
```suggestion
				'capability__in' => array( 'activitypub' ),
				'fields'         => 'id',
```
If we only need IDs, this should work, too

includes/rest/class-actors-controller.php
```diff
@@ -42,10 +42,11 @@ public function register_routes() {
...
 				'args'   => array(
 					'user_id' => array(
-						'description' => 'The ID or username of the actor.',
-						'type'        => 'string',
-						'required'    => true,
-						'pattern'     => '[\w\-\.]+',
+						'description'       => 'The ID of the actor.',
+						'type'              => 'string',
+						'required'          => true,
+						'pattern'           => '-?\d+',
```
obenland: Should the type be changed to `integer`? Not sure if that works with `-1`.

I don't know why I added the pattern here initially, it shouldn't matter. If it doesn't match the rest_base pattern we never even get here.

obenland (commented):
I think this is close. Feel free to merge once you've had a chance to consider these comments.

includes/class-dispatcher.php
pfefferle: Oh nice! Will check next week!

PR #2046, opened 2025-08-04
Delete: Self-Destruct

### LINKED ISSUE #2099: Separate Instance Actor and Blog Profile for Delete Actor Implementation
**Problem**
Currently, the handling of `Delete` Activities is ambiguous when it comes to Actor removal. It is not always clear whether the `Delete` should apply only to the Actor profile or also cascade to all related Objects (posts, notes, media, etc.).

**Proposal**
Introduce a clearer separation between:

* **Instance-level Actor** (`Application`): `@yourdomain.com@yourdomain.com`
* **Blog/Profile-level Actor** (`Organization` or `Group`): `@blog@yourdomain.com`

This separation ensures that when a `Delete` Activity targets an Actor:

* We can remove just the Actor profile, while leaving the handling of their historical Objects (`Note`, `Article`, `Image`, etc.) up to the server.
* This avoids the problem of generating a `Delete` for every individual Note or Object created by that Actor.

**Key Considerations**

* A `Delete` for an Actor still needs to be federated to **all remote servers** where that Actor has followers.
* To ensure delivery efficiency, the **shared inbox** should always be used for Actor deletions.
* Actor deletion can be treated as a **special case** in the federation logic, distinct from content deletion.

**Benefits**

* Cleaner separation of concerns between application-level identity and profile-level identity.
* Reduced network overhead (no per-object `Delete`).
* Consistency across federation when a user or blog is fully removed.

pfefferle:
Fixes #2099.

This PR implements a self-destruct feature for ActivityPub that allows users to remove their blog from the Fediverse by sending Delete activities to all followers. The implementation includes a CLI command, completion tracking, and admin notifications.

Screenshot 2025-08-06 at 12 35 39

## Proposed changes:

* Implements the `wp activitypub self_destruct` CLI command with confirmation prompts and progress tracking
* Adds completion detection logic that monitors for pending Delete activities and notifies when the process finishes
* Provides admin notice functionality to inform users when the self-destruct process is complete

### Other information:

## Testing instructions:

* Run `wp activitypub self_destruct` in the CLI
* Check the Outbox for `Delete` Activities

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Adds a self-destruct feature to remove a blog from the Fediverse by sending Delete activities to followers.

</details>


includes/wp-admin/class-admin.php
```diff
@@ -81,6 +81,24 @@ public static function admin_notices() {
...
+			<div class="notice notice-success is-dismissible">
+				<p>
+					<strong><?php esc_html_e( 'ActivityPub Self-Destruct Complete!', 'activitypub' ); ?></strong>
+				</p>
+				<p>
+					<?php esc_html_e( 'All Delete activities have been successfully sent to the Fediverse. Your blog is no longer discoverable via ActivityPub and all followers have been notified of the deletion.', 'activitypub' ); ?>
+				</p>
+			</div>
+			<?php
+		}
```
obenland, line 100: With this being triggered from CLI, how about sending an email on completion instead?

includes/class-cli.php
```diff
@@ -23,15 +25,88 @@ class Cli extends \WP_CLI_Command {
...
+		$user_count = \count( $user_ids );
+		\WP_CLI::line( \WP_CLI::colorize( '%GStarting Fediverse deletion process...%n' ) );
+		\WP_CLI::line( \WP_CLI::colorize( "%BFound {$user_count} ActivityPub user(s) to process:%n" ) );
+		\WP_CLI::line( '' );
+
+		// Set the self-destruct flag.
+		\add_option( 'activitypub_self_destruct', true );
+
+		// Hook into outbox processing completion to notify when self-destruct is done.
+		\add_action( 'activitypub_outbox_processing_complete', array( Scheduler::class, 'check_self_destruct_completion' ), 10, 2 );
```
obenland: Does this work? Any outbox processing should happen in separate threads where this action would not be added.

pfefferle: good question!

pfefferle: any idea where to put this instead? does it cost too much to add the hook, to be called on every `activitypub_outbox_processing_complete` call?

obenland: It's probably not too expensive to put into the Dispatcher, but kind of a bummer given that this will almost never run.

Do we need the notification? Could we give them a CLI command to keep track of the scheduled events manually? Like, tell them to check `wp cron event list XYZ`?

pfefferle: This felt like the best UX. We should get sure that this delete is properly processed, because otherwise this could end up in a big inconsistancy that can never be properly cleaned up!?

obenland:
```
%  wp activitypub self_destruct --status

🔍 Self-Destruct Status Check

⏳ Status: IN PROGRESS
The self-destruct process is currently running.

📊 Progress: 1 Delete Activities still pending

💡 Note: The process may take several minutes to complete.
```
AI sure loves their emoji

pfefferle:
not good?

obenland:
I don't care too much but others might. Having one on every line arguably makes their effects moot.

includes/class-cli.php
```diff
@@ -21,17 +23,289 @@ class Cli extends \WP_CLI_Command {
...
+			\WP_CLI::line( \WP_CLI::colorize( '%Y⏳ Status: IN PROGRESS%n' ) );
+			\WP_CLI::line( \WP_CLI::colorize( '%YThe self-destruct process is currently running.%n' ) );
+			\WP_CLI::line( '' );
+
+			if ( $pending_count > 0 ) {
+				\WP_CLI::line( \WP_CLI::colorize( "%Y📊 Progress: {$pending_count} Delete Activities still pending%n" ) );
+			} else {
+				\WP_CLI::line( \WP_CLI::colorize( '%Y📊 Progress: All Delete Activities have been processed%n' ) );
+				\WP_CLI::line( \WP_CLI::colorize( '%Y⚠️  Completion notification may be pending%n' ) );
+			}
```
obenland:
```suggestion
			\WP_CLI::line( \WP_CLI::colorize( "%Y📊 Progress: {$pending_count} Delete Activities still pending%n" ) );
```
If `$pending_count` is not more than 0, the `if` clause on line 284 will catch it.

includes/class-cli.php
```diff
@@ -21,17 +23,289 @@ class Cli extends \WP_CLI_Command {
...
+				'post_type'      => Outbox::POST_TYPE,
+				'post_status'    => 'pending',
+				'posts_per_page' => -1,
+				'fields'         => 'ids',
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+				'meta_query'     => array(
+					array(
+						'key'     => '_activitypub_activity_type',
+						'value'   => 'Delete',
+						'compare' => '=',
```
obenland:
```suggestion
```
Should be the default

obenland (approved):

PR #2086, opened 2025-08-21
Improve Following class documentation and method efficiency

obenland:
Fixes #

## Proposed changes:

This PR improves the Following class with better documentation and more efficient method implementations:

* **Remove unused import**: Removed unused `Activitypub\Http` import
* **Fix @see references**: Updated `@see` tags to use proper namespace format (`\Activitypub\follow` instead of `Activitypub\follow`)
* **Improve return type documentation**: Updated `follow()` method documentation to accurately reflect all possible return types
* **Optimize count methods**: Modified count methods (`count_following`, `count_pending`, `count_all`) to use minimal parameters (limit=1) instead of fetching all data when only the total count is needed
* **Fix documentation typo**: Corrected return type documentation in `get_all_with_count()` to reference 'following' instead of 'followers'

### Other information:

- [x] Have you written new tests for your changes, if applicable? - No new functionality, only optimizations and documentation improvements

## Testing instructions:

* Go to WordPress admin and navigate to ActivityPub settings
* Test following/unfollowing functionality to ensure it still works correctly
* Verify that follower counts are displayed correctly in the UI
* Check that performance is maintained or improved when viewing large follower lists

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Improve Following class documentation and optimize count methods for better performance

</details>

### DIFF
--- a/includes/collection/class-following.php
+++ b/includes/collection/class-following.php
@@ -8,7 +8,6 @@
 namespace Activitypub\Collection;
 
 use Activitypub\Activity\Activity;
-use Activitypub\Http;
 
 use function Activitypub\add_to_outbox;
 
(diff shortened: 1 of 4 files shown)


includes/collection/class-following.php
```diff
@@ -54,14 +53,14 @@ class Following {
...
+	 * Please do not use this method directly, use `\Activitypub\follow` instead.
 	 *
-	 * @see Activitypub\follow
+	 * @see \Activitypub\follow
 	 *
 	 * @param \WP_Post|int $post    The ID of the remote Actor.
 	 * @param int          $user_id The ID of the WordPress User.
 	 *
-	 * @return \WP_Post|\WP_Error The ID of the Actor or a WP_Error.
+	 * @return int|false|\WP_Post|\WP_Error The Outbox ID or false on failure, the Actor post or a WP_Error.
```
obenland, line 63: @pfefferle The possible return values of this method are a bit confusing. It's a WP_Error with an invalid actor, the outbox item ID on success and false on failure, and if we didn't get to adding it to the outbox, the actor post. Would it be possible to limit it to two types?

pfefferle: should we do this as part of this PR or in a follow up?

can you file an issue if it should be a follow up?

obenland: Created issue #2092 to track the return type simplification for the `Following::follow()` method

pfefferle (approved):

PR #2097, opened 2025-08-26
Refactor actor blocking with unified API

### LINKED ISSUE #2095: Refactor: Create centralized Moderation::is_actor_blocked() method
## Background

From the discussion in https://github.com/Automattic/wordpress-activitypub/pull/2027#discussion_r2301770071, there's a need to refactor the blocking logic to be more centralized and flexible.

## Current State

The blocking logic is currently spread across different parts of the codebase, with specific implementations in places like the Following class.

## Proposed Refactoring

Create a new method `Moderation::is_actor_blocked()` that:

### Method Signature
- Accepts multiple input types: webfinger, URI, object, or array
- Provides flexibility for different blocking levels

### Functionality
- User-specific blocks
- Site-wide blocks  
- Option to check both user and site blocks

### Benefits
- Centralizes blocking logic
- Provides consistent API across the plugin
- Reduces code duplication
- Makes blocking checks more maintainable

## Implementation Notes

This would replace the current scattered blocking checks throughout the codebase with a single, flexible method that can handle various input formats and blocking scenarios.

obenland:
Fixes #2095

## Proposed changes:
* Add `Actors::normalize_identifier()` method for centralizing actor identifier normalization logic
* Simplify `Moderation::is_actor_blocked()` to accept pre-normalized URIs with consistent int parameter
* Refactor Following and Blocked Actors table classes to use unified normalization approach
* Improve separation of concerns between actor normalization and blocking logic

### Other information:

## Testing instructions:

* Go to 'Users > Following' or 'Settings > ActivityPub > Following' 
* Try following an account using different formats:
  - Webfinger format: `@user@example.com`
  - Email format: `user@example.com` 
  - URL format: `https://example.com/@user`
  - URL without scheme: `example.com/@user`
* Verify that blocked actors are properly prevented from being followed
* Go to 'Users > Blocked Actors' or 'Settings > ActivityPub > Blocked Actors'
* Try blocking accounts using the same various formats
* Verify all formats work correctly and error handling is appropriate

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Refactor actor blocking with unified API for better maintainability

</details>


includes/class-moderation.php
```diff
@@ -271,6 +271,35 @@ public static function get_site_blocks() {
...
 	}
 
+	/**
+	 * Check if an actor is blocked by user or site-wide.
+	 *
+	 * @param string $actor_uri Actor URI to check.
+	 * @param int    $user_id   Optional. User ID to check user blocks for. Defaults to 0 (site-wide only).
+	 * @return bool True if blocked, false otherwise.
+	 */
+	public static function actor_is_blocked( $actor_uri, $user_id = 0 ) {
```
pfefferle: This sounds more like a statement than a question. What about

```suggestion
	public static function is_actor_blocked( $actor_uri, $user_id = 0 ) {
```

includes/wp-admin/table/class-following.php
```diff
@@ -123,41 +123,28 @@ public function process_action() {
...
-					}
-				} elseif ( empty( \wp_parse_url( $profile, PHP_URL_SCHEME ) ) ) {
-					// Add scheme if missing.
-					$profile = \esc_url_raw( 'https://' . \ltrim( $profile, '/' ) );
-				}
-
-				// Check if user has blocked the account.
-				if ( \in_array( $profile, Moderation::get_user_blocks( $this->user_id )['actors'], true ) ) {
+				$original = \sanitize_text_field( \wp_unslash( $_REQUEST['activitypub-profile'] ) );
+				$profile  = Actors::normalize_identifier( $original );
```
pfefferle, line 127: I would move the normalizer into the `is_actor_blocked` function. It reduces the preconditions for the URL you want to check and simplifies the code.

includes/class-moderation.php
obenland: For it to read like a statement is on purpose, the idea is for if statements to be read like a sentence. That said, it looks like Core prioritized consistency around the `is_` prefix over if statement sentences (`is_post_type_hierarchical()`), so I'm happy to revert back to `is_actor_blocked()`.

includes/wp-admin/table/class-following.php
obenland: It includes prepending `https` for host-only queries, so the table functionality depends on it, too.

includes/class-moderation.php
obenland: Fixed in [1e6a1ee](https://github.com/Automattic/wordpress-activitypub/pull/2097/commits/1e6a1eedf1bd04bebbaa39ec8035d7850b6afc35).

includes/wp-admin/table/class-following.php
pfefferle: I am not sure I understand?

pfefferle: How does it speak against calling it directly inside of the blocked check?

obenland: If it was moved into `is_actor_blocked` we'd break the functionality that allows users to put in a host only and still follow the blog actor. Like `activitypub.blog`.

The normalized identifier is a service to both `is_actor_blocked()` and `follow()` and prevents a webfinger from having to be resolved twice.

pfefferle: FYI: It will not be resolved twice, we cache the request. so yes, it might be a bit of overload, but might simplify the code. but I am fine with improving over time.

includes/wp-admin/table/class-following.php
```diff
@@ -123,41 +123,28 @@ public function process_action() {
...
-					$redirect_to = \add_query_arg( 'resource', $original ?? $profile, $redirect_to );
+					\add_settings_error( 'activitypub', 'followed', \sprintf( \__( 'Unable to follow account &#8220;%s&#8221;. Please verify the account exists and try again.', 'activitypub' ), \esc_html( $profile ) ) );
+					$redirect_to = \add_query_arg( 'resource', $original, $redirect_to );
 					break;
 				}
 
-				// Check if site has blocked the account.
-				if ( \in_array( $profile, Moderation::get_site_blocks()['actors'], true ) ) {
+				// Check if actor is blocked.
+				if ( Moderation::is_actor_blocked( $profile, $this->user_id ) ) {
```
pfefferle, line 136: Do we also have to check the Actor against the blocked domains?

includes/wp-admin/table/class-following.php
obenland: Dang, that's impressive. I'm looking forward to having the same detailed grasp of the plugin as you do, I did not know that off the top of my head

includes/wp-admin/table/class-following.php
obenland: Hm, good question. Yeah probably. Maybe? What do you think?

pfefferle: I would say we should.

If you block a domain, it does not make sense to follow someone from that host, because you would not receive any Activities from them.

pfefferle: Should we encapsulate this in a specific function? I think I need something similar for the Starter Kits import.

pfefferle: We can directly add it to `is_actor_blocked` because it seems that `check_activity_against_blocks` does not rely on that function!

obenland: That makes sense, yes

obenland: Fixed in [d19866b](https://github.com/Automattic/wordpress-activitypub/pull/2097/commits/d19866b3bf1b3297bca58abb857f3572dbfc61cb).

pfefferle (approved):

PR #2101, opened 2025-08-27
Add `Reject` handling for Application user follow requests

pfefferle:
Prevents application users from being followed by automatically sending a 'Reject' response when a follow request targets the application user.

For reference: https://github.com/Automattic/wordpress-activitypub/issues/2099#issuecomment-3228716657

## Proposed changes:

* Adds automatic rejection of follow requests for application users
* Introduces a new queue_reject method to handle rejection responses

### Other information:

## Testing instructions:

* Try to follow the Application Actor
* You should receive a `Reject`

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Prevents Application users from being followed.

</details>

### DIFF
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -199,6 +199,10 @@ class Migration {
 			}
 		}
 
+		if ( \version_compare( $version_from_db, 'unreleased', '<' ) ) {
+			self::remove_pending_application_user_follow_requests();
+		}
+
 		// Ensure all required cron schedules are registered.
 		Scheduler::register_schedules();
 
@@ -1054,5 +1058,20 @@ class Migration {
 				'batch_size' => $batch_size,
 			);
 		}
+	}
+
+	/**
+	 * Removes pending follow requests for the application user.
+	 */
+	public static function remove_pending_application_user_follow_requests() {
+		global $wpdb;
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
+		$wpdb->query(
+			$wpdb->prepare(
+				"DELETE FROM {$wpdb->postmeta} WHERE meta_key = '_activitypub_following' AND meta_value = %s",
+				Actors::APPLICATION_USER_ID
+			)
+		);
 	}
 }
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -1067,10 +1067,11 @@ class Migration {
 		global $wpdb;
 
 		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
-		$wpdb->query(
-			$wpdb->prepare(
-				"DELETE FROM {$wpdb->postmeta} WHERE meta_key = '_activitypub_following' AND meta_value = %s",
-				Actors::APPLICATION_USER_ID
+		$wpdb->delete(
+			$wpdb->postmeta,
+			array(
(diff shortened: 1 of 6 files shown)


includes/handler/class-follow.php
```diff
@@ -100,4 +105,33 @@ public static function queue_accept( $actor, $activity_object, $user_id, $remote
...
+		$origin_activity = array_intersect_key(
+			$activity,
+			array_flip(
+				array(
+					'id',
+					'type',
+					'actor',
+					'object',
+				)
+			)
```
obenland:
```suggestion
			array(
				'id'     => 1,
				'type'.  => 1,
				'actor'  => 1,
				'object' => 1,
			)
```

includes/model/class-application.php
```diff
@@ -241,11 +241,10 @@ public function get_public_key() {
...
 	 */
 	public function get_summary() {
-		return \wpautop(
-			\wp_kses(
-				\get_bloginfo( 'description' ),
-				'default'
-			)
+		return sprintf(
+			/* translators: %s: Domain of the site */
+			__( 'This is the Application Actor for %s. It&#8217;s a tool account only, please don&#8217;t follow or interact with it.', 'activitypub' ),
```
obenland: Feels odd to say that, no? Maybe just return an empty summary instead?

includes/class-migration.php
```diff
@@ -1055,4 +1059,19 @@ public static function update_actor_json_storage( $batch_size = 100 ) {
...
+	public static function remove_pending_application_user_follow_requests() {
+		global $wpdb;
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
+		$wpdb->query(
+			$wpdb->prepare(
+				"DELETE FROM {$wpdb->postmeta} WHERE meta_key = '_activitypub_following' AND meta_value = %s",
+				Actors::APPLICATION_USER_ID
+			)
+		);
```
obenland: Would this be a candidate for `$wpdb->delete()`?

includes/model/class-application.php
pfefferle: Why? This user should simply not be followed at all.

includes/class-migration.php
pfefferle: Fixed in https://github.com/Automattic/wordpress-activitypub/pull/2101/commits/84d362e3437bb56fcb74bec3a67ba3a5e1a829ef

includes/handler/class-follow.php
pfefferle: ???

obenland: we can save the array_flip call by defining the array with the right key value pairs

includes/model/class-application.php
obenland: "please don't follow or interact with it" just sounds odd. It shouldn't be up to others, we shouldn't put the onus on others to do or not do something. If anything I bet it'll make people want to interact with it more :)

pfefferle: the problem is, that people do, becaue I have follow requests for that user. we can remove the please do not follow and only add the first part.

includes/handler/class-follow.php
pfefferle: I see! Context wins :)

obenland (approved):

PR #2116, opened 2025-08-29
Tombstone: Add `remove` method

pfefferle:
~Renames the `Tombstone::exists_local` method to `Tombstone::was_buried` for clarity and updates all references accordingly. Adds a new `Tombstone::exhume` method to remove URLs from the local tombstone registry. Updates and expands tests to cover the new method and renamed functionality.~

~The renaming aligns with the `bury` function, since `exists_local` literally checks whether something is buried.~

This PR lays the groundwork for virtual deletes and undeletes. In practice, that means an object will be removed from the fediverse but remain locally. A potential use case is removing the `activitypub` capability from a user.

## Proposed changes:

* ~Renames Tombstone::exists_local() to Tombstone::was_buried() for better semantic alignment with the bury() method~
* Adds new Tombstone::remove() method to remove URLs from the tombstone registry
* ~Updates all references to use the renamed method and adds comprehensive test coverage~

### Other information:

## Testing instructions:

* See unit-tests

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Adds support for virtual deletes and restores, allowing objects to be removed from the fediverse without being deleted locally.

</details>

### DIFF
--- a/includes/class-tombstone.php
+++ b/includes/class-tombstone.php
@@ -219,12 +219,7 @@ class Tombstone {
 	 */
 	public static function exhume( $url ) {
 		$urls = \get_option( 'activitypub_tombstone_urls', array() );
-		$urls = \array_filter(
-			$urls,
-			function ( $_url ) use ( $url ) {
-				return normalize_url( $_url ) !== normalize_url( $url );
-			}
-		);
+		$urls = \array_diff( $urls, array( normalize_url( $url ) ) );
 		\update_option( 'activitypub_tombstone_urls', $urls );
 	}
 }
--- a/includes/class-tombstone.php
+++ b/includes/class-tombstone.php
@@ -211,13 +211,13 @@ class Tombstone {
 	/**
 	 * Remove a URL from the local tombstone registry.
 	 *
-	 * "Exhumes" a URL by removing it from the local tombstone URL registry.
+	 * Removes a URL from the local tombstone URL registry.
 	 * The URL is normalized before comparison to ensure consistent matching.
 	 * This marks the URL as no longer tombstoned for future local checks.
 	 *
 	 * @param string $url The URL to remove from the tombstone registry.
 	 */
-	public static function exhume( $url ) {
+	public static function remove( $url ) {
 		$urls = \get_option( 'activitypub_tombstone_urls', array() );
 		$urls = \array_diff( $urls, array( normalize_url( $url ) ) );
 		\update_option( 'activitypub_tombstone_urls', $urls );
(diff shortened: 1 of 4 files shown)


includes/class-tombstone.php
```diff
@@ -111,7 +111,7 @@ public static function exists_remote( $url ) {
 	 *
 	 * @return bool True if the local URL is in the tombstone registry, false otherwise.
 	 */
-	public static function exists_local( $url ) {
+	public static function was_buried( $url ) {
```
obenland: I first thought "why is it not `is_buried()`?", then "well then it's the same as `exists()`", so I guess this doesn't really do a great job conveying the context that it's meant for local URLs only?

Maybe this should not be part of the public API and all checks should go through `exists()`?

I understand where you're coming from with the namings of these methods (at least I think I do 🙂), but maybe we could keep it around Tombstones? Otherwise we'd have to rename the class `Corpse` 😱

includes/class-tombstone.php
```diff
@@ -207,4 +207,24 @@ public static function bury( $url ) {
...
+	/**
+	 * Remove a URL from the local tombstone registry.
+	 *
+	 * "Exhumes" a URL by removing it from the local tombstone URL registry.
+	 * The URL is normalized before comparison to ensure consistent matching.
+	 * This marks the URL as no longer tombstoned for future local checks.
+	 *
+	 * @param string $url The URL to remove from the tombstone registry.
+	 */
+	public static function exhume( $url ) {
```
obenland: Is this going to be used anywhere outside of tests? If so, could I suggest `remove()`, as in removing the tombstone?

includes/class-tombstone.php
```diff
@@ -207,4 +207,24 @@ public static function bury( $url ) {
...
+		$urls = \get_option( 'activitypub_tombstone_urls', array() );
+		$urls = \array_filter(
+			$urls,
+			function ( $_url ) use ( $url ) {
+				return normalize_url( $_url ) !== normalize_url( $url );
+			}
+		);
```
obenland:
```suggestion
		$urls = \array_diff( $urls, array( normalize_url( $url ) ) );
```
Would this work?

includes/class-tombstone.php
pfefferle: It will... (I am confused that you assume that I write functions only to write a test for ^^)

I am preparing the "virtual delete", when an object will be deleted from the fediverse but not locally. For the add/remove `activitypub` capability I would need this function.

obenland (approved):

PR #2117, opened 2025-08-29
Prevent notifications to Application user in new_follower

pfefferle:
Added a check in `Mailer::new_follower` to skip sending notifications if the recipient is the Application user. This prevents unnecessary or unintended notifications to system-level users.

It is not possible to follow the Application Actor, so there is no need to send `follow` notifications.

## Proposed changes:
- Added a guard clause in the `new_follower` method to skip notifications for the Application user
- Uses the existing `Actors::APPLICATION_USER_ID` constant for the check

### Other information:

## Testing instructions:

* Follow the Application user on Mastodon and check for follow E-Mails.

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Stopp sending follow notifications to the Application user, since system-level accounts cannot be followed.

</details>

### DIFF
--- a/includes/class-mailer.php
+++ b/includes/class-mailer.php
@@ -126,6 +126,11 @@ class Mailer {
 	 * @param int   $user_id  The id of the local blog-user.
 	 */
 	public static function new_follower( $activity, $user_id ) {
+		// Do not send notifications to the Application user.
+		if ( Actors::APPLICATION_USER_ID === $user_id ) {
+			return;
+		}
+
 		if ( $user_id > Actors::BLOG_USER_ID ) {
 			if ( ! \get_user_option( 'activitypub_mailer_new_follower', $user_id ) ) {
 				return;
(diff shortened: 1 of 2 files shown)


includes/class-mailer.php
```diff
@@ -126,6 +126,11 @@ public static function comment_notification_text( $message, $comment_id ) {
 	 * @param int   $user_id  The id of the local blog-user.
 	 */
 	public static function new_follower( $activity, $user_id ) {
+		// Do not send notifications to the Application user.
+		if ( Actors::APPLICATION_USER_ID === $user_id ) {
```
obenland, line 130: With Application users not being able to be followed, when would this condition be true?

pfefferle: This filter is directly hooked into the inbox action, so it operates independently from the follow process. If someone tries to follow the Application user they will receive a mail, even if the follow is rejected.

We should maybe use a different hook for that, but in the meantime this PR fixes the problem.

obenland (approved):

PR #2118, opened 2025-08-29
Add CLI actor delete/update and improve tombstone handling

pfefferle:
Introduces a new WP-CLI command to delete or update actors, with appropriate checks and activity queuing. Refactors the Actors class to allow bypassing user capability checks in retrieval methods. Moves tombstone creation for deleted objects to the outbox handler, and adds a test for the Tombstone::bury method. Removes redundant tombstone calls from the scheduler.

## Proposed changes:
- Added a new `actor` command to WP-CLI for deleting and updating actors
- Moved tombstone creation from the scheduler to the outbox handler for centralized processing
- Added test coverage for the `Tombstone::bury` method

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added a new WP-CLI command to manage Actors.

</details>

### DIFF
--- a/includes/class-cli.php
+++ b/includes/class-cli.php
@@ -432,6 +432,7 @@ class Cli extends \WP_CLI_Command {
 					\WP_CLI::error( 'Actor not found.' );
 				}
 
+				\WP_CLI::line( \WP_CLI::colorize( '%YThe Actor will not deleted from the database, but only from the Fediverse. Be aware that this action might be irreversible.%n' ) );
 				\WP_CLI::confirm( 'Do you really want to delete the Actor with the ID: ' . $args[1] );
 
 				$activity = new Activity();
--- a/includes/class-cli.php
+++ b/includes/class-cli.php
@@ -396,6 +396,68 @@ class Cli extends \WP_CLI_Command {
 	}
 
 	/**
+	 * Delete or Update an Actor.
+	 *
+	 * ## OPTIONS
+	 *
+	 * <action>
+	 * : The action to perform. Either `delete` or `update`.
+	 * ---
+	 * options:
+	 *   - delete
+	 *   - update
+	 * ---
+	 *
+	 * <id>
+	 * : The id of the Actor.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     $ wp activitypub actor delete 1
+	 *
+	 * @synopsis <action> <id>
+	 *
+	 * @param array $args The arguments.
+	 */
+	public function actor( $args ) {
+		if ( Actors::APPLICATION_USER_ID === (int) $args[1] ) {
+			\WP_CLI::error( 'You cannot delete the application actor.' );
+		}
+
+		switch ( $args[0] ) {
+			case 'delete':
+				$actor = Actors::get_by_id( $args[1], true );
+
+				if ( \is_wp_error( $actor ) ) {
+					\WP_CLI::error( 'Actor not found.' );
+				}
+
+				\WP_CLI::confirm( 'Do you really want to delete the Actor with the ID: ' . $args[1] );
+
+				$activity = new Activity();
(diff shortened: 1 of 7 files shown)


includes/collection/class-actors.php
```diff
@@ -59,16 +59,17 @@ class Actors {
...
 	 * Get the Actor by ID.
 	 *
-	 * @param int $user_id The user ID.
+	 * @param int  $user_id         The user ID.
+	 * @param bool $bypass_user_can Whether to bypass the `user_can_activitypub` check.
 	 *
 	 * @return Actor|User|Blog|Application|\WP_Error Actor object or WP_Error if not found or not permitted.
 	 */
-	public static function get_by_id( $user_id ) {
+	public static function get_by_id( $user_id, $bypass_user_can = false ) {
```
obenland: Can we add an `add_filter( 'activitypub_user_can_activitypub', '__return_true' );` in the CLI call instead?

includes/class-cli.php
```diff
@@ -395,6 +395,69 @@ public function comment( $args ) {
...
+
+				\WP_CLI::line( \WP_CLI::colorize( '%YThe Actor will not deleted from the database, but only from the Fediverse. Be aware that this action might be irreversible.%n' ) );
+				\WP_CLI::confirm( 'Do you really want to delete the Actor with the ID: ' . $args[1] );
+
+				$activity = new Activity();
+				$activity->set_actor( $actor->get_id() );
+				$activity->set_object( $actor->get_id() );
+				$activity->set_type( 'Delete' );
+
+				add_to_outbox( $activity, null, $args[1] );
```
obenland: It feels odd to me that we don't have an API for this. What makes calling `Actor::schedule_user_delete()` not a good option?

includes/functions.php
```diff
@@ -1536,6 +1536,11 @@ function add_to_outbox( $data, $activity_type = null, $user_id = 0, $content_vis
 
 	set_wp_object_state( $data, 'federated' );
 
+	// Set Tombstones for deleted objects.
+	if ( 'Delete' === $activity->get_type() ) {
+		Tombstone::bury( object_to_uri( $activity->get_object() ) );
+	}
```
obenland: I don't know if we have a good place for this, but it feels like type-specific actions should be handled in a callback to `post_activitypub_add_to_outbox`, what do you think?

pfefferle: Thought about that too.

includes/class-cli.php
pfefferle: This is exactly the reason why I created: https://github.com/Automattic/wordpress-activitypub/issues/2119

It is simply getting more and more confusing when a function is for local actor or remote actor handling.

There is a `delete` function for example that handles remote actor deletes.

pfefferle: That is why I started simple and created the issue instead.

obenland: Sounds good. In this case it's a local user, is this not what `Actor::schedule_user_delete()` handles?

pfefferle: true!

includes/functions.php
pfefferle: @obenland where should we put that hook? Is that part of the handler domain?

includes/collection/class-actors.php
pfefferle: totally forgot about that filter! nice!

includes/functions.php
obenland: Yeah no idea. The handlers are all inbox related currently, although I guess it doesn't have to stay that way?

Actually, nvm, Move already has an outbox callback. Let's put it in the handler, yeah.

obenland (approved):
Image

PR #2132, opened 2025-09-03
Update reply block rendering based on position

### LINKED ISSUE #2124: Federated reply block remote representation
### Quick summary

Locally when a post adds a Federated Reply block, the url being replied to is transformed into a link

Image

The remote representation however turns the url into an `@mention` link, I would have expected the same behaviour.

Image 

View from Mastodon

Image

### Steps to reproduce

- Create a new post, 
- Add the Federated Reply block (with the url of an activitypub post you'd like to reply to)
- Publish

### Site owner impact

### Severity

Moderate

_No response_

_No response_

obenland:
Fixes #2124

## Proposed changes:
* Transform reply blocks to @-mention links only when they are the first block in a post for ActivityPub content
* Refactor block rendering logic from Post transformer to Blocks class for better organization
* Add dynamic callback registration during ActivityPub content generation
* Deprecate old methods in Post transformer while maintaining backward compatibility

### Other information:

## Testing instructions:

* Create a new WordPress post with a reply block as the first block
* Publish the post and view it via ActivityPub (e.g., using the ActivityPub REST API endpoint)
* Verify the reply block is rendered as an @-mention link in ActivityPub content
* Create another post with a reply block that is NOT the first block
* Verify this reply block is rendered as a regular clickable link in ActivityPub content
* Test that local rendering remains consistent (shows clickable links in both cases)

### Changelog entry

<details>

#### Significance

- [x] Patch

#### Type

#### Message

Fix reply block rendering inconsistency where blocks were always converted to @-mentions in ActivityPub content. Now only first reply blocks become @-mentions, others remain as regular links.

</details>


pfefferle:
Can you please describe the problem a bit and how this PR fixes it?

obenland:
#2124 pointed out how remote representations of the federated reply block always get rendered as a mention, regardless of whether the block is the first block in a post or somewhere in the middle of post. This can lead to awkward renderings:

Image

This PR [adds a check](https://github.com/Automattic/wordpress-activitypub/pull/2132/files#diff-89d47d0bcb43bf40e33a19f8e947c9942d5f8cc07882ea949a55eeecff35a2b7R396) in a callback that's hooked into the post transformer to only convert the reply block display to a mention-type handle when it's the first block in a post.
If that is not the case, [it will render the URL](https://github.com/Automattic/wordpress-activitypub/pull/2132/files#diff-89d47d0bcb43bf40e33a19f8e947c9942d5f8cc07882ea949a55eeecff35a2b7R236).

pfefferle:
The cases you described, work as expected, but I thought I save some time and test all cases in one post :)

I do not know if that is a usecase, but if you have two reply blocks in one post, one as the first block and one later in the post, both will still show up as @-replies.

Screenshot 2025-09-10 at 10 52 22

obenland:
@pfefferle This is ready for another round

pfefferle (approved):
oops, forget that one!

works like a charm!

Issue #2125, opened 2025-09-01
Allow Federation of Posts with Backdated Publish Dates


Jiwoon-Kim:
I usually write photolog posts using **categories for year/topic classification** and **tags for location-based classification**.
For chronological organization, I rely on the **publish date**, not the actual date of posting.

This means that I often set the publish date to a past time — for example, to match the media file’s storage path or to ensure my posts appear in the correct timeline order.

⚠️ **Important:** WordPress uses the post’s publish date to determine the media upload path (`uploads/YYYY/MM/`).
Therefore, backdating is not just optional but a natural workflow for users who want their media library and timeline to remain consistent.

The problem is that the **WordPress ActivityPub plugin currently forces old posts to "do not federate"**, even if I manually switch the visibility to *public* and save it. It automatically reverts back to *do not federate*.

I understand this restriction was introduced to avoid issues where older posts are unintentionally federated when edited. However:

1. **Latest post display** can be based not only on creation date but also on modification date.
2. **There is a valid demand for federating older posts**, especially since WordPress explicitly allows users to backdate their publish dates.

---

### Suggested Solution

* When a post is set to **public**, the plugin should generate a copy in the outbox.
* When a post is set to **do not federate**, the plugin should delete the copy from the outbox.

This approach would allow users more flexibility while still respecting visibility choices, without automatically blocking federation of all backdated posts.

- https://kimjiwoon96.wordpress.com/2025/04/18/250418-black-ogye-chicken/
- https://mastodon.social/@kimjiwoon96.wordpress.com@kimjiwoon96.wordpress.com

Jiwoon-Kim:
### Title

**Rethinking "Maximum Image Attachments": Support for Multiple Media Types and Order Control**

---

### Description

Currently, the plugin provides a setting for **“Maximum Image Attachments”**.
However, from my understanding, the ActivityPub specification allows not only **images** but also **videos and audio files** to be attached.

This makes the current naming somewhat misleading, since attachments are not limited to images.

Another limitation is that the plugin seems to control the number of attachments by simply restricting the count of media files — without giving the user any control over:

1. **Which media type is prioritized** (image, video, audio), and
2. **In what order attachments are included**.

As it stands, attachments are automatically selected in order from the top of the post content. This means users have no flexibility to determine the presentation of their media.

---

### Problems with the Current Mechanism

* The setting name “Maximum Image Attachments” does not accurately reflect support for multiple media types.
[...]

PR #2138, opened 2025-09-04
Add image attachment selection control for ActivityPub posts

obenland:
Before | After
--- | ---
Screenshot 2025-09-03 at 10 03 25 AM | Screenshot 2025-09-03 at 10 03 03 AM

## Proposed changes:
- **NEW FEATURE**: Complete image attachment selection control for ActivityPub posts in the Gutenberg editor
- Added interactive attachment selector component with drag-and-drop reordering functionality
- Implemented visual selection interface with numbered badges and selection states
- Added seamless integration with WordPress media library and post content parsing
- Enhanced ActivityPub post transformer to properly format selected attachments
- Added comprehensive styling and responsive design for the attachment selector UI

### Other information:

## Testing instructions:
* Go to 'wp-admin/post-new.php'
* Navigate to the **Fediverse** panel in the post sidebar
* Verify help text appears: "Choose which images to share with the fediverse."
* Add multiple images to the post content using the block editor (Image blocks, Cover blocks, etc.)
* Return to the Fediverse panel and confirm images automatically appear in the attachment selector
* Test **selecting/deselecting images** by clicking on them - selected images show blue borders and numbered badges
* Test **drag-and-drop reordering** of selected images using the drag handles
* Test **automatic selection** - first 4 images should be auto-selected on new posts (based on max_image_attachments setting)
* Verify **ActivityPub JSON output** contains proper attachment objects with correct format:
  ```json
  "attachment": [
    {
      "type": "Image",
      "url": "http://localhost:8888/wp-content/uploads/image.jpg",
      "mediaType": "image/jpeg"
    }
  ]
  ```
* Test with **featured images** - they should appear in the attachment list
* Test **deletion behavior** - removing images from post content should update the sidebar

### Changelog entry

<details>

#### Significance

-   [x] Minor

#### Type

#### Message

Add interactive image attachment selection control for ActivityPub posts with drag-and-drop reordering.

</details>


pfefferle:
I understand the general idea of the PR why this was requested.

But the implementation feels a lot like the old crossposting days: Having a special output for social media.
There is still a common misconception, that users think of the blog post on mastodon as a copy instead of a full representation of the blog post. This whole `Attachment` concept is a problem of microblogging platforms not supporting inline images (yet). Before tweaking too much in the WP UI I would prefer working on a better support of `Articles` in the fediverse and use our users to raise the pressure on other platforms to support it.

And we are already on a good way: https://socialwebfoundation.org/2025/05/01/steps-forward-in-long-form-text/

And if we really want to provide a feature to hide images from the fediverse, why not hook into the image block and provide a flag there instead?

Jiwoon-Kim:
@pfefferle I’d personally like to express my support for the improvements around the `attachment` implementation.

I think supporting inline images and attachments should be treated as **separate features with distinct purposes**.

> But the implementation feels a lot like the old crossposting days: Having a special output for social media.

I actually see it the other way around. To me, this is an opportunity to extend WordPress with unique frontend layouts. For example, imagine if there were a block specifically for rendering attachments — without inline images, such an approach could allow WordPress to take on a PixelFed-like theme.

We also have to acknowledge the practical limitations: inline images are hard to implement across microblogging platforms. Mastodon is not the only challenge here; PixelFed and Lemmy have completely different layout approaches. Personally, I don’t expect inline images to ever be properly supported on PixelFed.

That’s why I think **treating attachments as the default template for placing media files is the most broadly compatible solution**. Layout compatibility across platforms cannot be ignored, and ultimately it should be up to the end user how they want to use these features and which platforms’ layout compatibility they want to prioritize.

> And if we really want to provide a feature to hide images from the fediverse, why not hook into the image block and provide a flag there instead?

I’ve actually been thinking about this too — I already drafted an idea for an `Image Object` block:
[https://github.com/Automattic/wordpress-activitypub/issues/2146](https://github.com/Automattic/wordpress-activitypub/issues/2146)

If the Media Library provided a per-image CW (content warning) toggle, that could give users direct control over how each image object is federated.

pfefferle:
> I actually see it the other way around. To me, this is an opportunity to extend WordPress with unique frontend layouts. For example, imagine if there were a block specifically for rendering attachments — without inline images, such an approach could allow WordPress to take on a PixelFed-like theme.

But then this should be handled as part of this block and not in the sidebar!

The sidebar is quite de-synced to data that is already there, so if we have a handling like that, I would (as I wrote) recommend to use the blocks meta data.

Aside from that, the Pixelfed block is very theoretically and we should discuss how to solve the issue when there is such a block.

Jiwoon-Kim:
> The sidebar is quite de-synced to data that is already there, so if we have a handling like that, I would (as I wrote) recommend to use the blocks meta data.

Hmm, I’m curious — what kind of difficulties does the sidebar attachments feature introduce when it comes to implementing the object for media (attachment) posts?

As I mentioned earlier, my position is that supporting inline images and supporting attachments should be treated as separate approaches.

pfefferle:
But this adds a lot of distraction to WordPress’ UI. While everything might be possible through settings or templates, we should stay focused on the core “blogging” and “publishing” use case. Other platforms have different priorities and design their UI specifically for those needs.

This feature has only been requested once, and I’m not sure we should open this box just yet. One of WordPress’ credos is “decisions, not options,” and I think we should take that seriously.

Jiwoon-Kim:
> But this adds a lot of distraction to WordPress’ UI. While everything might be possible through settings or templates, we should stay focused on the core “blogging” and “publishing” use case. Other platforms have different priorities and design their UI specifically for those needs.

Exactly — designing the UI to meet the needs of other platforms is itself one of WordPress’ privileges.
That’s precisely why I don’t see this as simply “going back to cross-posting.”

https://github.com/Automattic/wordpress-activitypub/issues/2125#issuecomment-3241804709

Jiwoon-Kim:
I agree that introducing this feature right now may not be the best timing. In my opinion, it would be better to first improve the handling of the image object and then revisit this feature. However, I don’t fully agree with the nuance that there is almost no demand for it. I believe this feature could potentially have a very high demand once the foundation is ready.

obenland:
Lots to digest, thank you for the feedback!

I agree with prioritizing the advancement of long-from content over catering to short-form content platforms and I wish that perspective occurred to me when I first saw the feedback about the image selector. It's probably not fruitful to dive into specifics of the feedback or my feelings about the UI improvements in this PR—they'll all be superseded by this broader goal.

> One of WordPress’ credos is “decisions, not options,” and I think we should take that seriously.

💯! It's also a double-edged sword, heavy and sharp. I'm looking forward to further explore with you how it can be applied to this plugin.

Jiwoon-Kim:
https://github.com/Automattic/wordpress-activitypub/discussions/2382#discussioncomment-14825939
> > Hmm... it seems possible to include only 4 images in the preview's attachment, while including 50 images in the article's attachment, right?
> 
> Sure, but it seems not too complicated to then only use 4 of the 50 images for a preview!?

Yes, I proposed the FMO (Federated Media Object) to prevent duplicate caching between the `preview`'s `attachment`, inline images in the `content`, and the `article`'s `attachment` images.

Also, it’s important to note that the image sizes in the `preview`, the `article`, and the original source may each differ in version or resolution.

In this case, specifying the ID of the FMO image object in the `source` property of a Link object could be the most robust approach for federating isolated media objects within the WordPress media library.

Jiwoon-Kim:
Since all inline images included in the `content` must be attached to the `article`’s `attachment`, our focus should be on the `preview`’s `attachment`.

Issue #2065, opened 2025-08-08
Federated comments become orphans when the original post is deleted


Jiwoon-Kim:
### **Bug Report: Deleting a WordPress Post Orphans its Federated Comments**

**Issue:** When a parent post is deleted from WordPress, its associated federated comments are not deleted from the Fediverse, causing them to become "orphaned."

#### **1. Issue Description**

When a federated WordPress post with federated comments is moved to the Trash and permanently deleted, the `Delete` activity is not being sent for the associated comment objects. As a result, the comments continue to exist on the Fediverse but their parent post (`inReplyTo` object) is gone. This breaks the conversation thread and leaves the comments without context, appearing as "orphaned" objects.

#### **2. Steps to Reproduce**

1.  Create a new, federated post in WordPress.
2.  Write a comment on that new post from a WordPress user account.
3.  Confirm that the comment is successfully federated to the Fediverse as a `Note` that is `inReplyTo` the original post.
4.  From the WordPress dashboard, move the original post to the Trash.
5.  Permanently delete the post from the Trash.
6.  Observe the federated comment on a platform like Mastodon. The comment still exists, but its parent post is unavailable.

#### **3. Evidence**

* **Orphaned Comment on Mastodon:** [https://mastodon.social/@kowporg.wordpress.com@kowporg.wordpress.com/114991534914451937](https://mastodon.social/@kowporg.wordpress.com@kowporg.wordpress.com/114991534914451937)
* **Original Comment Address (now broken):** `https://kowporg.wordpress.com/?p=561/comment-page-1/#comment-129`
* **Federated Comment's Unique Permalink:** `https://kowporg.wordpress.com/?c=129`

This shows that the comment (`c=129`) has its own ActivityPub identity but is now orphaned from its parent (`p=561`).

#### **4. Expected Behavior**

When a post is **permanently deleted** from WordPress, the plugin should send a `Delete` activity for the parent post's object *and* for all of its federated comment objects.

#### **5. Current Behavior**

When a post is deleted, only the parent object is (or appears to be) deleted. The federated comments remain, breaking conversational context across the Fediverse.

---

#### **Proposal 5.1: Differentiate "Trash" vs. "Permanent Delete"**

The plugin's behavior should align with WordPress's standard content lifecycle.

* **When a post is moved to the Trash:** Do **not** send a `Delete` activity. The post is in a temporary state and can be restored. The federated content should remain active.
[...]

PR #2141, opened 2025-09-04
Add Delete activity support for permanently deleted federated comments

obenland:
See https://github.com/Automattic/wordpress-activitypub/issues/2065.

## Summary

When a federated comment is permanently deleted using `wp_delete_comment()`, the plugin now creates a Delete activity and sends it to the appropriate ActivityPub inboxes. This ensures that remote servers are notified when comments they sent are deleted.

## Changes

- Added new hook `delete_comment` in `Comment` scheduler class
- Added `schedule_comment_delete_activity()` method that:
  - Checks if comment was previously federated using `\Activitypub\Comment::was_sent()`
  - Only creates Delete activities for federated comments
  - Ignores non-federated comments to avoid unnecessary federation traffic
- Added comprehensive tests with both positive and negative test cases
- Tests verify exact outbox post creation and metadata validation

## Testing instructions

1. Create a comment on a post by a registered user
2. Mark the comment as federated: `add_comment_meta( $comment_id, 'activitypub_status', 'federated' )`
3. Permanently delete the comment: `wp_delete_comment( $comment_id, true )`
4. Verify a Delete activity is created in the outbox with correct metadata
5. Test that non-federated comments do not create Delete activities

### Changelog entry

<details>

#### Significance
- [x] Patch

#### Type

#### Message
Add Delete activity support for permanently deleted federated comments.

</details>

### DIFF
--- a/includes/scheduler/class-comment.php
+++ b/includes/scheduler/class-comment.php
@@ -25,6 +25,7 @@ class Comment {
 		// Comment transitions.
 		\add_action( 'transition_comment_status', array( self::class, 'schedule_comment_activity' ), 20, 3 );
 		\add_action( 'wp_insert_comment', array( self::class, 'schedule_comment_activity_on_insert' ), 10, 2 );
+		\add_action( 'delete_comment', array( self::class, 'schedule_comment_delete_activity' ), 10, 2 );
 	}
 
 	/**
@@ -59,8 +60,9 @@ class Comment {
 			$type = 'Update';
 			\update_comment_meta( $comment->comment_ID, 'activitypub_comment_modified', time(), true );
 		} elseif (
-			\in_array( $new_status, array( 'trash', 'delete', 'spam' ), true ) &&
-			\Activitypub\Comment::was_received( $comment->comment_parent )
+			'trash' === $new_status ||
+			( 'delete' === $new_status && '' === $old_status ) || // Went through schedule_comment_delete_activity().
+			'spam' === $new_status
 		) {
 			$type = 'Delete';
 		}
@@ -86,6 +88,19 @@ class Comment {
 	public static function schedule_comment_activity_on_insert( $comment_id, $comment ) {
 		if ( 1 === (int) $comment->comment_approved ) {
 			self::schedule_comment_activity( 'approved', '', $comment );
+		}
+	}
+
+	/**
+	 * Schedule Delete activity when a comment is permanently deleted.
+	 *
+	 * @param int         $comment_id Comment ID.
+	 * @param \WP_Comment $comment    Comment object.
+	 */
+	public static function schedule_comment_delete_activity( $comment_id, $comment ) {
+		// Only send Delete activities for comments that were previously federated.
+		if ( \Activitypub\Comment::was_sent( $comment ) ) {
+			self::schedule_comment_activity( 'delete', '', $comment );
 		}
 	}
 }
--- a/includes/scheduler/class-comment.php
+++ b/includes/scheduler/class-comment.php
@@ -25,6 +25,7 @@ class Comment {
 		// Comment transitions.
 		\add_action( 'transition_comment_status', array( self::class, 'schedule_comment_activity' ), 20, 3 );
 		\add_action( 'wp_insert_comment', array( self::class, 'schedule_comment_activity_on_insert' ), 10, 2 );
+		\add_action( 'delete_comment', array( self::class, 'schedule_comment_delete_activity' ), 10, 2 );
 	}
 
 	/**
@@ -87,5 +88,29 @@ class Comment {
 		if ( 1 === (int) $comment->comment_approved ) {
 			self::schedule_comment_activity( 'approved', '', $comment );
 		}
+	}
+
+	/**
+	 * Schedule Delete activity when a comment is permanently deleted.
+	 *
+	 * @param int         $comment_id Comment ID.
+	 * @param \WP_Comment $comment    Comment object.
+	 */
+	public static function schedule_comment_delete_activity( $comment_id, $comment ) {
+		if ( defined( 'WP_IMPORTING' ) && WP_IMPORTING ) {
+			return;
+		}
+
+		// Only send Delete activities for comments that were previously federated.
+		if ( ! \Activitypub\Comment::was_sent( $comment ) ) {
+			return;
+		}
+
+		// Federate only comments that are written by a registered user.
+		if ( ! $comment || ! $comment->user_id ) {
+			return;
+		}
+
+		add_to_outbox( $comment, 'Delete', $comment->user_id );
 	}
 }
--- a/includes/scheduler/class-comment.php
+++ b/includes/scheduler/class-comment.php
@@ -25,7 +25,6 @@ class Comment {
 		// Comment transitions.
 		\add_action( 'transition_comment_status', array( self::class, 'schedule_comment_activity' ), 20, 3 );
 		\add_action( 'wp_insert_comment', array( self::class, 'schedule_comment_activity_on_insert' ), 10, 2 );
-		\add_action( 'delete_comment', array( self::class, 'schedule_comment_delete_activity' ), 10, 2 );
 	}
 
 	/**
@@ -59,10 +58,7 @@ class Comment {
 		} elseif ( 'approved' === $new_status ) {
 			$type = 'Update';
 			\update_comment_meta( $comment->comment_ID, 'activitypub_comment_modified', time(), true );
-		} elseif (
-			'trash' === $new_status ||
-			'spam' === $new_status
-		) {
+		} elseif ( \in_array( $new_status, array( 'trash', 'delete', 'spam' ), true ) ) {
 			$type = 'Delete';
 		}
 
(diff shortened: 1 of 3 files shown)


tests/includes/scheduler/class-test-comment.php
```diff
@@ -126,4 +126,118 @@ public function test_no_activity_scheduled( $comment_data ) {
...
+					array(
+						'key'   => '_activitypub_activity_type',
+						'value' => 'Delete',
+					),
+				),
+			)
+		);
+
+		$this->assertEmpty( $outbox_posts_after, 'Should not create Delete activity for non-federated comment deletion' );
+		$this->assertCount( count( $outbox_posts_before ), $outbox_posts_after, 'Number of Delete activities should remain unchanged' );
```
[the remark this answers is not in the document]
obenland: fixed in [e9f5378](https://github.com/Automattic/wordpress-activitypub/pull/2141/commits/e9f5378e7f51b57f9dbbb2fe87459e6a0603cb84).

tests/includes/scheduler/class-test-comment.php
```diff
@@ -126,4 +126,133 @@ public function test_no_activity_scheduled( $comment_data ) {
...
+				'user_id'          => self::$user_id,
+				'comment_approved' => 1,
+			)
+		);
+
+		// Mark the comment as sent (federated).
+		\add_comment_meta( $comment_id, 'activitypub_status', 'federated' );
+
+		$comment        = \get_comment( $comment_id );
+		$activitypub_id = \Activitypub\Comment::generate_id( $comment );
```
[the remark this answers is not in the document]
obenland: Fixed in [4552d39](https://github.com/Automattic/wordpress-activitypub/pull/2141/commits/4552d39141af876df88958d7f64b815bdcca7060).

tests/includes/scheduler/class-test-comment.php
```diff
@@ -126,4 +126,133 @@ public function test_no_activity_scheduled( $comment_data ) {
...
+		$comment_id = self::factory()->comment->create(
+			array(
+				'comment_post_ID'  => self::$comment_post_ID,
+				'user_id'          => self::$user_id,
+				'comment_approved' => 1,
+			)
+		);
+
+		$comment        = \get_comment( $comment_id );
+		$activitypub_id = \Activitypub\Comment::generate_id( $comment );
```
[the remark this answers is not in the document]
obenland: Fixed in [4552d39](https://github.com/Automattic/wordpress-activitypub/pull/2141/commits/4552d39141af876df88958d7f64b815bdcca7060).

pfefferle:
Why not simply add `delete` to this check: https://github.com/Automattic/wordpress-activitypub/blob/073afc39df81daba761c2dc2a9cedc126e203448/includes/scheduler/class-comment.php#L62 ???

The hook is called anyways and it would make the code a bit simpler!?

obenland:
Sorry, I wasn't super explicit in the PR description with the context for this change, beyond mentioning #2065.

When a post gets deleted, [all associated comments also get deleted](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-includes/post.php#L3793-L3799), without going through the process of updating comment status.

That said, we could call `schedule_comment_activity()` internally, like in `schedule_comment_activity_on_insert()`.

pfefferle:
But `wp_delete_comment` is triggering [`wp_transition_comment_status`](https://developer.wordpress.org/reference/functions/wp_transition_comment_status/)!?

obenland:
Ha! Now your initial comment makes much more sense :) Yes, let's do that

tests/includes/scheduler/class-test-comment.php
```diff
@@ -7,6 +7,8 @@
 
 namespace Activitypub\Tests\Scheduler;
 
+use Activitypub\Comment;
```
pfefferle, line 11:
```suggestion
use Activitypub\Collection\Outbox;
use Activitypub\Comment;
```

tests/includes/scheduler/class-test-comment.php
```diff
@@ -120,10 +122,106 @@ public function test_no_activity_scheduled( $comment_data ) {
...
+		// Ensure this comment is NOT marked as sent.
+		\delete_comment_meta( $comment_id, 'activitypub_status' );
+
+		// Permanently delete the comment.
+		\wp_delete_comment( $comment_id, true );
+
+		// Check that no Delete activity was created for this specific comment.
+		$outbox_posts = \get_posts(
+			array(
+				'post_type'   => \Activitypub\Collection\Outbox::POST_TYPE,
```
pfefferle:
```suggestion
				'post_type'   => Outbox::POST_TYPE,
```

tests/includes/scheduler/class-test-comment.php
```diff
@@ -120,10 +122,106 @@ public function test_no_activity_scheduled( $comment_data ) {
...
+
+		$activitypub_id = Comment::generate_id( $comment_id );
+
+		// Permanently delete the comment - this should trigger a Delete activity.
+		\wp_delete_comment( $comment_id, true );
+
+		// Check if a Delete activity was created.
+		$outbox_posts = \get_posts(
+			array(
+				'post_type'   => \Activitypub\Collection\Outbox::POST_TYPE,
```
pfefferle:
```suggestion
				'post_type'   => Outbox::POST_TYPE,
```

tests/includes/scheduler/class-test-comment.php
```diff
@@ -179,49 +179,4 @@ public function test_schedule_comment_delete_activity() {
 		$this->assertEquals( $activitypub_id, \get_post_meta( $outbox_post->ID, '_activitypub_object_id', true ) );
 		$this->assertEquals( self::$user_id, $outbox_post->post_author );
 	}
-
```
pfefferle: why not keep this test? it should have the same result!?

obenland: I needed to update the test setup and add another check in the status transition handler. Fixed in [8c6f99a](https://github.com/Automattic/wordpress-activitypub/pull/2141/commits/8c6f99afd00b6a3a6d7463ecd3b505acce98be7b).

includes/scheduler/class-comment.php
```diff
@@ -59,8 +59,8 @@ public static function schedule_comment_activity( $new_status, $old_status, $com
 			$type = 'Update';
 			\update_comment_meta( $comment->comment_ID, 'activitypub_comment_modified', time(), true );
 		} elseif (
-			'trash' === $new_status ||
-			'spam' === $new_status
+			\in_array( $new_status, array( 'trash', 'delete', 'spam' ), true ) &&
+			\Activitypub\Comment::was_received( $comment->comment_parent )
```
pfefferle: We can not get sure that it is only a federated comment if the comment above was received. Comments directly to the post can also be federated for example or comments to federated local comments.

pfefferle: Btw. there is already a check some lines below that is doing the trick: `should_comment_be_federated`.

obenland: When comments get deleted, [all meta data is deleted](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-includes/comment.php#L1529-L1533) before `wp_transition_comment_status()` runs, so `Comment::was_sent()` won't know whether a comment was federated or not.

I guess we need to chose between sending Deletes out for all comments that get deleted or creating a separate hook callback that can check for it?

pfefferle: It is in the same functions, only some lines below! https://github.com/Automattic/wordpress-activitypub/blob/4a1d3ea36c12bdd455f55a006662a44a6f92516c/includes/scheduler/class-comment.php#L73

pfefferle: What I mean: It does not make any sense to have that check there, it has to be in the `should_comment_be_federated` function, otherwise it will break anyways.

pfefferle: But you are right, the transition is set way too late, so maybe we should simply keep your function. Sorry :(

pfefferle: What about reverting your changes, but instead of copying the code over, to simply call the `schedule_comment_activity` like `self::schedule_comment_activity( 'trash', 'publish', $comment )`???

obenland: Maybe [45d4271](https://github.com/Automattic/wordpress-activitypub/pull/2141/commits/45d427160d187303fc544ba3c94f2b68e12c369d) is the charm?

includes/scheduler/class-comment.php
```diff
@@ -88,4 +90,17 @@ public static function schedule_comment_activity_on_insert( $comment_id, $commen
...
+
+	/**
+	 * Schedule Delete activity when a comment is permanently deleted.
+	 *
+	 * @param int         $comment_id Comment ID.
+	 * @param \WP_Comment $comment    Comment object.
+	 */
+	public static function schedule_comment_delete_activity( $comment_id, $comment ) {
+		// Only send Delete activities for comments that were previously federated.
+		if ( \Activitypub\Comment::was_sent( $comment ) ) {
```
pfefferle: shoudln't we add this check to the `Comment::should_be_federated` instead?

obenland: `was_sent` is  already part of `should_be_federated()` through `is_local()`. I decided to go with `was_sent()` here directly since that's what we're interested in and `should_be_federated()` runs as part of `schedule_comment_activity()`.

pfefferle:
```suggestion
		if ( Comment_Utils::was_sent( $comment ) ) {
```

obenland: [01351c1](https://github.com/Automattic/wordpress-activitypub/pull/2141/commits/01351c1ba27c9d69642948f38d0dc00c41c3da60)

pfefferle (approved):

Jiwoon-Kim:
https://core.trac.wordpress.org/ticket/63931#comment:10

PR #2148, opened 2025-09-08
Move `Activity-Object-Type` setting to advanced settings

pfefferle:
The `Activity-Object-Type` option and its field have been relocated from the main settings to the advanced settings section. Registration of the option is now under `activitypub_advanced`, and the UI has been simplified to a single checkbox for backward compatibility. Help tab for template tags is now conditionally shown only when `note` is selected.

I think we should hide this feature for new Users, to push the automated mode a bit more.

## Proposed changes:
- Move `activitypub_object_type` setting registration from main to advanced settings group
- Simplify UI from radio buttons to a single checkbox for enabling template tags
- Add conditional display of template tags help tab based on object type selection
- Show template tags help only if activated

### Other information:

## Testing instructions:

* Check for the settings under the Advanced Tab
* The Template form is still in the main settings, to not confuse existing customers/users.
* Check if checking/unchecking the settings will have the expected results.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

-   [X] Deprecated - for soon-to-be removed features

#### Message

ActivityPub now defaults to automated object type selection, with the old manual option moved to Advanced settings for compatibility.

</details>

### DIFF
--- a/includes/wp-admin/class-advanced-settings-fields.php
+++ b/includes/wp-admin/class-advanced-settings-fields.php
@@ -108,9 +108,9 @@ class Advanced_Settings_Fields {
 			array( 'label_for' => 'activitypub_persist_inbox' )
 		);
 
-		add_settings_field(
+		\add_settings_field(
 			'activitypub_object_type',
-			__( 'Activity-Object-Type', 'activitypub' ),
+			\__( 'Activity-Object-Type', 'activitypub' ),
 			array( self::class, 'render_object_type_field' ),
 			'activitypub_advanced_settings',
 			'activitypub_advanced_settings',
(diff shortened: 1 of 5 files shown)


includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -279,4 +288,22 @@ public static function render_persist_inbox_field() {
...
+		$value = \get_option( 'activitypub_object_type', ACTIVITYPUB_DEFAULT_OBJECT_TYPE );
+		?>
+		<p>
+			<label>
+				<input type="checkbox" name="activitypub_object_type" value="note" <?php \checked( 'note', $value ); ?> />
+				<?php \esc_html_e( 'Use Template Tags instead of letting the plugin choose the best possible format for you.', 'activitypub' ); ?>
+			</label>
+		</p>
+		<p class="description">
+			<?php \esc_html_e( 'This is mainly for backwards compatibility. It is not recommended to use the Template Tags, because it might not be supported in future versions.', 'activitypub' ); ?>
```
obenland, line 305:
```suggestion
			<?php \esc_html_e( 'This is mainly for backwards compatibility. It is not recommended to use the Template Tags, because they will be removed in a future version.', 'activitypub' ); ?>
```
Optional: I think it's fine to be more definitive here.

obenland (approved):

PR #2160, opened 2025-09-10
Support multiple `inReplyTo`s in post transformer

pfefferle:
Updated the reply block extraction logic to collect all reply URLs from post content blocks. Now returns a single URL if only one reply block is present, or an array of unique URLs if multiple reply blocks are found.

Follow up of #2132

## Proposed changes:
- Modified reply block extraction logic to collect all reply URLs from post content
- Added logic to return appropriate format based on number of reply blocks found
- Removed early return behavior that only supported single reply blocks

### Other information:

## Testing instructions:

* Add multiple reply blocks.
* Check the ActivityPub JSON output if there are multiple `inReplyTo` URLs.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

It’s now possible to reply to multiple posts using multiple reply blocks.

</details>


pfefferle:
This can be merged independently from #2132

obenland (approved):

PR #2161, opened 2025-09-11
Remove Mastodon oEmbed provider registration

pfefferle:
~The Mastodon oEmbed provider registration logic was relocated from the Activitypub class to the Embed class for better separation of concerns. This change ensures oEmbed provider registration is handled within the class responsible for embedding functionality.~

Remove oEmbed endpoint registration completely, because Mastodon supports the oEmbed discovery through HTML link header.

## Proposed changes:
- ~Relocated `register_oembed_providers` method from Activitypub to Embed class~
- ~Updated initialization hook to call the method from its new location~
- ~Maintains identical functionality while improving code organization~

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


obenland:
I'm not sure we even need these. Mastodon supports oembed discovery, so we should find out why that doesn't work.

pfefferle:
Happy to change the PR to completely remove them... But when we want to keep them, they should live in the Embed class...

obenland:
`wp_filter_oembed_result()` [invalidates discovered Mastodon embeds because they don't use an iframe](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-includes/embed.php#L925-L953) to display their embed. 

I think I'm fine with removing support since they do fall back to our activitypub embed and still work. We should also petition Mastodon to add an iframe version so auto-discovery is less risky for oembed consumers.

obenland (approved):

Jiwoon-Kim:
**Does oEmbed require an iframe to work properly?**

I was considering whether the ActivityPub plugin should directly implement oEmbed for Mastodon posts. But if we go down that road, wouldn’t we also need to implement separate oEmbed providers for Misskey, Lemmy, Pleroma, and so on?

So the real question is:
**“Should the ActivityPub plugin itself provide oEmbed support for Mastodon (and other Fediverse software) Notes?”**

---

### 📌 Strictly speaking, this isn’t the role of the ActivityPub plugin

* The primary goal of the ActivityPub plugin is **federation of WordPress posts/comments ↔ Fediverse objects**.
* oEmbed, on the other hand, is about **how external content is displayed inside WordPress**, which is not directly related to the ActivityPub spec.
* Maintaining separate oEmbed providers for Mastodon, Misskey, Lemmy, Pleroma, etc. would blow up the scope of the plugin and create a maintenance nightmare.

---

### ✅ Why a JSON-LD approach makes more sense

Since Mastodon/Lemmy/Misskey/Pleroma posts are **all ActivityPub JSON-LD objects**, a better approach would be:

👉 **ActivityPub Object → WordPress Embed Transformer**

1. **Shared Format**

   * All major Fediverse platforms return ActivityStreams/ActivityPub JSON.
   * With `Accept: application/activity+json`, you can fetch a Note object directly, without platform-specific hacks.

2. **Closer to the Standard**

   * oEmbed is essentially a preview API.
   * In the Fediverse, ActivityPub *is* the API. It’s more consistent to generate embeds from JSON-LD than to hard-code multiple oEmbed providers.

3. **Scalability / Future Proofing**

   * Any new Fediverse software that supports ActivityPub would “just work.”
   * By contrast, oEmbed would require per-platform endpoint registration forever.

---

### ⚠️ Practical Challenges

* **Rendering layer**: JSON-LD is machine-readable; we still need templates to display it.

  * Example: `Note` → author avatar, display name, content, media attachments, links.
  * `Article` or `Event` → different templates.
* **WordPress mismatch**: Core oEmbed relies on iframes, but ActivityPub embedding requires server-side fetch → HTML transform.
* **Performance**: Every remote fetch adds latency. Needs caching (e.g. Transients or WP\_Object\_Cache).

---

### 🎯 Conclusion

Instead of separate oEmbed providers like “Mastodon Embed,” “Misskey Embed,” etc., we should aim for a **“Fediverse Embed Handler”**:

1. Request `application/activity+json` from the URL.
2. Parse the JSON-LD object.
3. Render it into a WordPress card template (Note, Article, Image, etc.).

This is more consistent with ActivityPub, scales better, and avoids provider lock-in.

PR #2169, opened 2025-09-12
Refactor actor WebFinger resolution logic

### LINKED ISSUE #2166: Include webfinger info in actor metadata to reduce external requests
## Problem

Currently, we make external requests to every follower in the Follower table to get their webfinger information. While this data is cached for a week, when paginating through the follower table, every page load takes 10+ seconds to load due to these external requests.

## Proposed Solution

Include webfinger information directly in the actor metadata to eliminate the need for external requests during follower table pagination.

## Additional Context

- This affects the follower management interface where users browse through their followers
- The performance issue is most noticeable on sites with many followers
- Each external webfinger request adds latency that compounds across multiple followers per page

pfefferle:
Moved WebFinger resolution from table classes to Remote_Actors::get_acct for consistency and reuse. Removed redundant get_webfinger method from trait-actor-list-table.php and updated usages in blocked, followers, and following tables to use the new static method.

Fixes #2166

## Proposed changes:
- Removed duplicate WebFinger resolution logic from the Actor_List_Table trait
- Added centralized `get_acct` method to Remote_Actors class with caching support
- Updated all table classes to use the new centralized method instead of the trait method

### Other information:

## Testing instructions:

* See unittests

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Optimized WebFinger lookups by centralizing and caching account resolution for faster, more consistent handling across lists.

</details>

### DIFF
--- a/includes/collection/class-remote-actors.php
+++ b/includes/collection/class-remote-actors.php
@@ -9,6 +9,7 @@ namespace Activitypub\Collection;
 
 use Activitypub\Activity\Actor;
 use Activitypub\Http;
+use Activitypub\Webfinger;
 
 use function Activitypub\get_remote_metadata_by_actor;
 use function Activitypub\is_actor;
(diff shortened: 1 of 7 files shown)


includes/collection/class-remote-actors.php
```diff
@@ -485,4 +486,41 @@ public static function get_public_key( $key_id ) {
...
+		}
+
+		$acct = Webfinger::uri_to_acct( $post->guid );
+
+		if ( \is_wp_error( $acct ) ) {
+			$actor = self::get_actor( $post );
+			$acct  = Webfinger::guess( $actor );
+		}
+
+		$acct = \str_replace( 'acct:', '', $acct );
```
obenland: `Sanitize::webfinger()`? Not sure if trimming `@` is necessary here.

obenland (dismissed):

includes/collection/class-remote-actors.php
pfefferle: consistency ftw!

obenland (approved):

PR #2171, opened 2025-09-12
Fix release script to catch more 'unreleased' patterns

obenland:
Fixes missing 'unreleased' strings that were not being updated during releases.

## Proposed changes:
* Enhanced release script to catch more WordPress deprecation function patterns that were previously missed
* Added support for `_deprecated_constructor`, `_deprecated_file`, `_deprecated_argument`, and `_deprecated_hook`
* Added support for `_doing_it_wrong` calls with both `__FUNCTION__` and `__METHOD__`
* Added support for `do_action_deprecated` calls (in addition to existing `apply_filters_deprecated`)
* Consolidated regex patterns from 8 separate patterns to 3 cleaner, more maintainable patterns
* Updated missed 'unreleased' strings with appropriate version numbers

### Other information:

## Testing instructions:

1. Review the updated `bin/release.js` file to see the consolidated regex patterns
2. Check the updated files where 'unreleased' strings were replaced with version numbers
3. The release script will now properly update more deprecation patterns in future releases

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fix release script to catch more 'unreleased' deprecation patterns that were previously missed during version updates.

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -72,7 +72,7 @@ function safe_remote_get( $url ) {
  * @return string The User resource.
  */
 function get_webfinger_resource( $user_id ) {
-	\_deprecated_function( __FUNCTION__, 'unreleased', 'Activitypub\Webfinger::get_user_resource' );
+	\_deprecated_function( __FUNCTION__, '7.1.0', 'Activitypub\Webfinger::get_user_resource' );
 
 	return Webfinger::get_user_resource( $user_id );
 }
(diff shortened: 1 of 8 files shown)


includes/functions.php
```diff
@@ -180,7 +180,7 @@ function url_to_authorid( $url ) {
  * @return int|bool Comment ID or false if not found.
  */
 function is_comment() {
-	\_deprecated_function( __FUNCTION__, ' 7.1.0' );
+	\_deprecated_function( __FUNCTION__, '7.1.0' );
```
pfefferle, line 183: was that a typo by you or an error in the script?

obenland: A typo, I had to update them manually.

pfefferle (approved):
would be nice to have a JS test framework 🤔

pfefferle:
I am not sure if we need to add such PRs to the changelog!?

obenland:
> I am not sure if we need to add such PRs to the changelog!?

Yeah I was on the fence about that, too. I decided to go with it since it updates developer-facing deprecation versions.

PR #2200, opened 2025-09-17
Fix Flag activity object list processing

obenland:
## Proposed changes:
* Fix Flag activity object processing to preserve URL arrays instead of converting them to Generic_Object
* Add `array_is_list()` check in `Activity::set_object()` to detect URL arrays  
* Create `maybe_convert_to_object()` helper method that only converts items with ActivityPub `type` field
* Add comprehensive tests for both simple URL arrays and mixed arrays with typed objects
* Ensures array_is_list() checks work correctly in object_to_uri() function
* Maintains ActivityPub specification compliance for Flag activities with multiple object URLs

### Other information:

## Testing instructions:

* Run the new tests: `npm run env-test -- --filter=test_activity_object_list` and `npm run env-test -- --filter=test_activity_object_mixed_array`
* Both tests should pass after the fix is implemented
* Verify Flag activities with URL arrays are processed correctly through the inbox
* Run all Activity tests to ensure no regressions: `npm run env-test -- --filter=Test_Activity`

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fix Flag activity object list processing to preserve URL arrays

</details>


pfefferle:
What is the usecase for lists as objects? Do we even want to support that?

obenland:
It's how [Mastodon has their Flag activity objects structured](https://docs.joinmastodon.org/spec/activitypub/#Flag) and [it's part of the spec ](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-object-term) (see Example 99). 

With the moderation check sitting in front of inbox handlers, it currently handles everything that comes in, including these Flag activities if a WP site or user gets flagged somewhere.

pfefferle (approved):

Issue #181, opened 2022-11-15
Support commentPolicy 


mediaformat:
This plugin follows Wordpress site options, but a Post's comment options are not advertised in anyway.

The [streams](https://codeberg.org/streams/streams/src/branch/dev/FEDERATION.md) project uses `commentPolicy` with various options:
 
> 'commentPolicy' can be any of
> 'authenticated' - matches the typical ActivityPub permissions
> 'contacts' - matches approved followers
> 'any connections' - matches followers regardless of approval
> 'site: foobar.com' - matches any actor or clone instance from 'foobar.com'
> 'public' - matches anybody at all, may require moderation if the network isn't known
> 'self' - matches the activity author only
> 'until=2001-01-01T00:00Z' - comments are closed after the date given. This can be supplied on its own or appended to any other commentPolicy string by preceding with a space; for example 'contacts until=2001-01-01T00:00Z'.

We could minimally support `until` with `get_option( 'close_comments_days_old' ) + post_publish` date.

https://github.com/w3c/activitypub/issues/319 `replyPreference`
https://github.com/mastodon/mastodon/issues/8565

mediaformat:
A FEP is being proposed that covers most of the use cases.

https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-5624.md

pfefferle:
Keep

Menrath:
[Mobilizon](https://docs.joinmobilizon.org/contribute/activity_pub/#repliesmoderation) uses `repliesmoderation` with:
`allow_all, closed, moderated`.
PeerTube has a `commentsEnabled` field with `true` or `false`.

We are currently investigating this as well.

Issue #1888, opened 2025-07-02
Add Notification Support for Quote Activities


Jiwoon-Kim:
Currently, Mastodon does not support quote functionality, while both Threads and Misskey provide implementations for quoting posts and notifying the original author when their content is quoted.

At this time:

- Threads does not allow quoting external posts.

- Mastodon is not implementing quotes in the near term.

- The only practically applicable case for WordPress would be when a Misskey user quotes a WordPress post.

Even though this may be a relatively rare occurrence, I believe it’s still worthwhile to implement a notification mechanism for quote activities in advance. This will help future-proof the plugin as more platforms may adopt similar features.

Example: A notification like:

> “…quoted this!”

when a Misskey user quotes a federated WordPress post.

Reasoning: Implementing this will help ensure consistent user feedback across different types of activities in the Fediverse and allow us to properly track and display quote interactions where supported.

PS.
It seems that on many Misskey.io instances, profiles from Threads are often hidden or intentionally filtered out. Additionally, when quoting a Threads profile via Mastodon proxy, the quote note often does not appear on other Misskey instances either.

Since WordPress does not implement such instance-level blocks, filters, or visibility restrictions, it arguably makes even more sense for us to properly implement quote notifications. This would allow WordPress users to be reliably informed when their posts are quoted — regardless of quirks or restrictions from specific Fediverse platforms.
![20250703_032434.jpg](https://github.com/user-attachments/assets/586186ab-b0e0-4200-8dd2-a1d2e91cb9e4)

Jiwoon-Kim:
{
  "@context": "https://www.w3.org/ns/activitystreams",
  "type": "Announce",
  "actor": "https://example.com/users/jiwoon",
  "object": {
    "type": "Article",
    "name": "이 글에 대한 제 생각",
    "content": "<p>이 글 정말 좋은데요! 👍<br>여기에 제 의견을 덧붙여 봅니다.</p>",
    "inReplyTo": "https://another.example/users/foo/statuses/12345"
  },
  "published": "2025-06-26T12:00:00Z"
}

---
{
  "@context": "https://www.w3.org/ns/activitystreams",
  "type": "Quote",
  "actor": "https://example.com/users/jiwoon",
  "object": "https://another.example/users/foo/statuses/12345",
  "published": "2025-06-26T12:00:00Z",
  "objectType": "Article",
  "name": "이 글에 대한 제 생각",
[...]

PR #2207, opened 2025-09-18
Add interaction policy and quoting controls to ActivityPub

pfefferle:
Introduces support for interaction policy fields in ActivityPub objects, updates the JSON-LD context to include GotoSocial and Mastodon extensions, and adds a 'Who can quote' select control to the editor plugin. Updates tests and ensures the new fields are handled in activity and base object classes.

Screenshot 2025-09-18 at 15 49 14

Related: #181

See:

* https://docs.gotosocial.org/en/latest/federation/interaction_policy/
* https://blog.joinmastodon.org/2025/09/introducing-quote-posts/
* https://wordpress.org/support/topic/feature-request-support-for-mastodon-quotes/

## Proposed changes:
- Adds interaction policy metadata registration and validation
- Implements quote policy controls in the WordPress editor
- Updates JSON-LD context with GotoSocial extensions for interaction policies

### Other information:

## Testing instructions:

* Go to the editor
* Change the "Who can quote" settings
* Check the JSON output for the corresponding JSON extension 

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added a setting to control who can quote your posts.

</details>


Jiwoon-Kim:
https://github.com/Automattic/wordpress-activitypub/issues/1888

includes/transformer/class-post.php
```diff
@@ -74,6 +74,68 @@ public function get_content_visibility() {
 		return $this->content_visibility;
 	}
 
+	/**
+	 * Get the Interaction Policy.
+	 *
+	 * @see https://docs.gotosocial.org/en/latest/federation/interaction_policy/
+	 *
+	 * @return array The interaction policy.
```
obenland, line 82:
```suggestion
	 * @return array|null The interaction policy.
```

src/editor-plugin/plugin.js
```diff
@@ -150,6 +158,26 @@ const EditorPlugin = () => {
...
 				className="activitypub-visibility"
 			/>
+
+			<BaseControl
+				label={ __( 'Interaction Policy', 'activitypub' ) }
+				help={ __( 'Choose who can interact with this post.', 'activitypub' ) }
+			></BaseControl>
+
+			<SelectControl
+				label={ __( 'Who can quote', 'activitypub' ) }
```
obenland:
```suggestion
			<SelectControl
				label={ __( 'Who can quote this post?', 'activitypub' ) }
				help={ __( 'Quoting allows others to reshare your post while adding their own commentary.', 'activitypub' ) }
```

The two controls attempting to create hierarchy feel a bit awkward. Could we try playing with label and help in the select control to create the necessary context for this to work?

includes/transformer/class-post.php
pfefferle: technically every attribute can be `null`, so we would have to add it to every doc!?

Context: `null` is removing it from the output.

src/editor-plugin/plugin.js
pfefferle: I am also not very happy with it. I tried a lot of different versions, but hoped this one would be the best compromise.

The idea behind that was, that there might be other "Interaction Policy" controls coming, so it might be nice to somehow group the.

includes/constants.php
```diff
@@ -74,6 +74,10 @@
 define( 'ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE', 'private' );
 define( 'ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL', 'local' );
 
+// Interaction policy constants.
+define( 'ACTIVITYPUB_INTERACTION_POLICY_ANYONE', 'anyone' );
+define( 'ACTIVITYPUB_INTERACTION_POLICY_FOLLOWERS', 'followers' );
+define( 'ACTIVITYPUB_INTERACTION_POLICY_ME', 'me' );
```
obenland, line 80: If we expect other interaction policy items, should we make these quote-specific?

includes/class-blocks.php
```diff
@@ -83,6 +83,29 @@ public static function register_postmeta() {
...
+							'enum'    => array( ACTIVITYPUB_INTERACTION_POLICY_ANYONE, ACTIVITYPUB_INTERACTION_POLICY_FOLLOWERS, ACTIVITYPUB_INTERACTION_POLICY_ME ),
+							'default' => ACTIVITYPUB_INTERACTION_POLICY_ANYONE,
+						);
+
+						if ( is_wp_error( rest_validate_enum( $value, $schema, '' ) ) ) {
+							return $schema['default'];
+						}
+
+						return $value;
+					},
```
obenland, line 106: No action needed, just thinking out loud:

For sanitize callbacks like that, shouldn't we be able to define a schema in `show_in_rest` and it does all of that for us? Does it not fallback to the default if the new value doesn't fit the enum? There was _something_ that didn't work, but I don't remember what it was.

includes/constants.php
pfefferle: My idea was to be able to re-use them for every policy! They post_meta is what makes it specific.

obenland: Gotcha, I didn't realize these were applicable to multiple policies

includes/class-blocks.php
pfefferle: I have no idea if that works, but happy to test it in a follow up!

obenland (approved):

Jiwoon-Kim:
https://github.com/misskey-dev/misskey/issues/16590

PR #2210, opened 2025-09-18
Improved recipient handling for clarity and added better inbox support

pfefferle:
Refactored recipient extraction logic by introducing `extract_recipients_from_activity_property()` for improved modularity and clarity. Updated `extract_recipients_from_activity()` to use the new helper. Added comprehensive unit tests for both functions to ensure correct handling of various recipient formats and uniqueness.

## Proposed changes:
- Extracted recipient parsing logic into a reusable helper function
- Updated existing recipient extraction to use the new helper
- Added comprehensive unit tests covering various recipient formats and edge cases
- ~Used in inbox actions to simplify visibility handling also for third party implementations~

### Other information:

## Testing instructions:

* See unittests

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Improved recipient and visibility handling for clarity and added better inbox support.

</details>


includes/functions.php
```diff
@@ -696,7 +688,7 @@ function url_to_commentid( $url ) {
  *
  * @param array|string $data The ActivityPub object.
  *
- * @return string The URI of the ActivityPub object
+ * @return string|false The URI of the ActivityPub object or false if not found.
```
obenland: One of the strengths of `object_to_uri()` is that it reliably returns a string and we don't need to check whether it failed. That is rare(!) and I think we should do what we can to keep it that way.

I assume this change is meant to facilitate running `array_filter()` over the output? Could we use a specific callback in that instance instead?

includes/rest/trait-audience.php
```diff
@@ -0,0 +1,83 @@
...
+use function Activitypub\extract_recipients_from_activity_property;
+use function Activitypub\is_same_domain;
+use function Activitypub\user_can_activitypub;
+
+/**
+ * Audience Trait.
+ *
+ * Provides methods for handling ActivityPub audience and recipient extraction.
+ */
+trait Audience {
```
obenland: Could these be methods in `Collections\Actors` instead? They feel more like utility functions than something that's specific to rest api controllers. A trait doesn't feel like the best place for these.

includes/rest/trait-audience.php
```diff
@@ -0,0 +1,83 @@
...
+	/**
+	 * Extract recipients from the given Activity.
+	 *
+	 * @param array $activity The activity data.
+	 *
+	 * @return array An array of user IDs who are the recipients of the activity.
+	 */
+	public function determine_local_recipients( $activity ) {
```
obenland:
```suggestion
	public function get_local_recipients( $activity ) {
```

includes/rest/trait-audience.php
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	/**
+	 * Determine the visibility of the activity based on its recipients.
+	 *
+	 * @param array $activity The activity data.
+	 *
+	 * @return string The visibility level: 'public', 'private', or 'direct'.
+	 */
+	public function determine_visibility( $activity ) {
```
obenland:
```suggestion
	public function get_visibility( $activity ) {
```

includes/rest/class-actors-inbox-controller.php
```diff
@@ -177,24 +178,28 @@ public function create_item( $request ) {
...
-			 * @param string             $type     The type of the activity.
-			 * @param Activity|\WP_Error $activity The Activity object.
+			 * @param array              $data       The data array.
+			 * @param int|null           $user_id    The user ID.
+			 * @param string             $type       The type of the activity.
+			 * @param Activity|\WP_Error $activity   The Activity object.
+			 * @param string             $visibility The visibility of the activity.
 			 */
-			\do_action( 'activitypub_inbox', $data, $user_id, $type, $activity );
+			\do_action( 'activitypub_inbox', $data, $user_id, $type, $activity, $visibility );
```
obenland: It's not quite clear to me why the visibility stuff needs to be moved up a level, when (so far) it's only `Create` that's concerned about it. Can we not just expand `is_activity_public()` with any improvements this PR aims to make?

obenland (changes requested):

includes/rest/class-actors-inbox-controller.php
pfefferle: Sure, we are currently only use it for `Create`, but the idea is that implementers do not have to care and simply can hook into the inbox to get all needed infos.

Aside from that, we are about to implement the reader, so we might want to have a more detailed visibility handling and not only a "Public or not".

includes/functions.php
pfefferle: That is not completely true! We have way to often an issue, because of missing `id` fields! I am fine to change it to empty string, but simply not checking if an attribute is available is not "we don't need to check whether it failed"!

includes/rest/trait-audience.php
pfefferle: Even if we check the Audience, this is not a function I see in the Actors Collection!

I will see if I can move it to the Activity.

includes/functions.php
pfefferle: From the [spec](https://www.w3.org/TR/activitystreams-core/#object):

> All properties are optional (including the id and type).

ghost:
Thanks for the updates here! I want to make sure I fully understand a couple of points:

**On `object_to_uri()`:**
I see the tradeoff between always returning a string vs. handling missing IDs. Returning an empty string seems like a safe compromise. Just to confirm are there any edge cases where this could lead to silent failures (e.g., code assuming a valid URI when it's actually empty)?

**On the `visibility` parameter:**
I agree that lifting it up to the inbox level makes future extensibility easier. Could you elaborate on how you see this being used beyond `Create` (e.g., for Reader or other activity types)? I’d like to understand the longer-term vision here.

Overall, I think the refactor improves clarity and future-proofs the code. The added unit tests are also great to see!

obenland:
For the failing coverage test, it's currently a bit hard to figure out, both because the warning parsing is broken (fixed in #2215), but also because it's limited by GitHub's annotation API to only displaying messages in changed files.

Unfolding the "Unit Testing with Coverage" step will show you the warnings it encountered and that need fixing.

pfefferle:
@obenland can you maybe have a separated PR for that fix?

includes/functions.php
```diff
@@ -733,10 +750,10 @@ function object_to_uri( $data ) {
 			$data = object_to_uri( $data['url'] );
 			break;
 		case 'Link':
-			$data = $data['href'];
+			$data = $data['href'] ?? '';
 			break;
 		default:
-			$data = $data['id'];
+			$data = $data['id'] ?? $data['url'] ?? '';
```
obenland: Apologies for being a stickler here. These still would require adding `empty()` checks before being able to use the return value. Is this still about making the `array_filter()` call work in `extract_recipients_from_activity_property()`?

includes/rest/class-inbox-controller.php
```diff
@@ -288,4 +278,37 @@ public function get_item_schema() {
...
 	}
+
+	/**
+	 * Extract recipients from the given Activity.
+	 *
+	 * @param array $activity The activity data.
+	 *
+	 * @return array An array of user IDs who are the recipients of the activity.
+	 */
+	private function get_recipients( $activity ) {
```
obenland: How would you feel about calling it `get_local_recipients()` to reflect the work this function does?

includes/rest/class-actors-inbox-controller.php
obenland: > the idea is that implementers do not have to care and simply can hook into the inbox to get all needed infos

I think it's reasonable for implementers to do their own determinations based on the passed activity. I'd be happy for us to provide tools like `extract_recipients_from_activity_property()` and `get_activity_visibility()` to make it easier for them, but I'm not sure it provides a lot of value to make these determinations up front.

* It makes the hooks more cumbersome to use, having to juggle (now) 5 parameters, and who knows how many others there might follow. 
* I worry it'll end up being a duping ground for meta's we'll need in activity handlers down the line. What makes visibility a particularly good meta value to pass? Should there be an `$is_reply` or `$has_attachments` or array of `$tags`? 
* Any one of these meta values will be created regardless of whether they're needed or not (in the case of visibility it's currently just one type and implementors might not even require it for their use-case).

I understand where this is coming from, but rather than providing something prophylactically could we see what the Reader implementation actually needs and make these changes accordingly?

includes/functions.php
pfefferle: I have no problem to return an empty string instead of `false`, to not have to care too much about the output. I simply want to avoid PHP warnings/fatals!

(I hope we are talking about the same issue, because I am not totally sure what you are referring to 🫣)

includes/rest/class-actors-inbox-controller.php
pfefferle: I believe visibility is a crucial property, more so than `has_images` or `is_reply`, because it directly relates to privacy. The current approach in the fediverse, which relies on audience fields (`to`, `cc`, etc.), feels overly complex and fragile.

Since not everyone uses/knows about all the functions provided by this plugin, different implementations may handle Activities in WordPress differently and could ultimately fall back on us. In my view, having a global attribute to define visibility is the simplest and most reliable solution.

includes/functions.php
obenland: Currently we can be reasonably sure that the output of `object_to_uri()` is a URI that we can use right away. I'd consider PHP warnings/fatals a good thing, actually, as a signal that there's something missing in this function or in object validation before this function is called.

Up until now, we don't know of any instance where we need a fallback to an empty string. Let's not introduce one just in case, but rather make fixes as we encounter problems, so `object_to_uri()`stays as reliable as it has been.

pfefferle: I think we see PHP warnings/fatals differently in terms of being a “good thing,” but I’m fine with reverting those changes. How should we handle the `url` fallback?

Reverted with https://github.com/Automattic/wordpress-activitypub/pull/2210/commits/1c6a2744258a91eef55d282cea0d0ac48b4ad989

obenland: > I think we see PHP warnings/fatals differently in terms of being a “good thing,”

Heh, fair 🙂 I think if they're a result of oversight, something avoidable, they're not a good thing. But they can also be an indicator for something unexpected that should be dealt with, rather than avoided. And then they help make code more intentional and robust.

With that in mind, I'd probably also avoid the `url` fallback. It doesn't look like it's needed currently, and it keeps the code intentional.

pfefferle: My issue with the statement "But they can also be an indicator for something unexpected that should be dealt with, rather than avoided" is that the `id` property isn’t required by ActivityPub. So if it’s missing, it doesn’t necessarily indicate that something went wrong, it just means we failed to validate for it.

obenland: Agreed, that's the point I'm trying to make. Wouldn't you want to know if we failed to validate something?

pfefferle: The even worse issue here is, that we use `object_to_uri` as exactly that validation method. Here is one example: https://github.com/Automattic/wordpress-activitypub/blob/dbf0d11631fa74ba00b56c3ed51746d582cb40b6/includes/rest/class-inbox-controller.php#L63

pfefferle: So my point is mainly that we already have that issue and we still wait for the first big chunk of errors/warnings.

pfefferle: But this PR might not the right place to discuss that. I will file a separate issue for it and we will see how and when to handle it.

obenland (approved):

PR #2215, opened 2025-09-19
Consolidate custom post type and meta registrations into Post_Types class

obenland:
This has been on my mental list for a while. Mainly because we register post meta in two places. After that, because the Activitypub class is a junk drawer class and this is something that we can pull out and it makes sense on its own.

## Proposed changes:

This PR consolidates all custom post type and post meta registrations into a single, well-organized `Post_Types` class to improve code maintainability and organization.

* Created new `Post_Types` class with dedicated methods for each post type registration
* Each method handles both post type registration and its related meta fields together
* Moved registrations from scattered locations (`Activitypub::register_post_types()` and `Blocks::register_postmeta()`) into semantic functions
* Maintained user meta registrations in original location as requested
* Updated test to use new class methods
* Removed unused imports after consolidation
* Updated code coverage annotation job to account for multiple warnings

### Other information:

- [x] Have you written new tests for your changes, if applicable? (Updated existing test)

## Testing instructions:

1. Ensure all existing functionality works as expected
2. Run `npm run env-test` to verify all 895 tests pass
3. Check that custom post types (`ap_actor`, `ap_inbox`, `ap_outbox`, extra fields) are properly registered
4. Verify post meta fields are correctly associated with their respective post types
5. Test ActivityPub functionality to ensure no regressions


includes/class-post-types.php
```diff
@@ -0,0 +1,434 @@
...
+			)
+		);
+
+		\register_post_meta(
+			Remote_Actors::POST_TYPE,
+			Followers::FOLLOWER_META_KEY,
+			array(
+				'type'              => 'string',
+				'single'            => false,
+				'sanitize_callback' => 'esc_sql',
```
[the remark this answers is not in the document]
obenland: @pfefferle Should this be `absint`?

includes/class-post-types.php
```diff
@@ -0,0 +1,434 @@
...
+class Post_Types {
+	/**
+	 * Initialize the class, registering all custom post types and post meta.
+	 */
+	public static function init() {
+		self::register_remote_actors_post_type();
+		self::register_inbox_post_type();
+		self::register_outbox_post_type();
+		self::register_extra_fields_post_types();
+		self::register_activitypub_post_meta();
```
pfefferle, line 29: no alphabetic order? 😱

includes/class-activitypub.php
```diff
@@ -949,7 +608,7 @@ public static function register_user_meta() {
 			$blog_prefix . 'activitypub_description',
 			array(
 				'type'              => 'string',
-				'description'       => 'The user’s description.',
+				'description'       => 'The user description.',
```
obenland, line 512: These changes are from Claude copying user meta registration back-and-forth. I left the changes in as these are "texturized" quotes, not translated, and the text still makes sense without them.
Happy to revert if you feel differently.

includes/class-post-types.php
obenland: I KNOW!!!

includes/class-blocks.php
```diff
@@ -32,83 +31,6 @@ public static function init() {
 		\add_filter( 'activitypub_the_content', array( self::class, 'remove_post_transformation_callbacks' ) );
 	}
 
-	/**
```
pfefferle, line 35: Nice, I just wanted to bring this up!!

obenland:
@pfefferle Should `prevent_empty_post_meta()` and `default_post_metadata()` also be moved into `Post_Types`?

pfefferle:
If we also move every meta registration I am not sure if Post_Type is still the right name!?

obenland:
> If we also move every meta registration I am not sure if Post_Type is still the right name!?

I guess I defined it generously in my head, anything post types related, including built-in post types.

includes/class-post-types.php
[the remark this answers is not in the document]
pfefferle: technically it could be `-1` so `absint` might not be the best callback.

pfefferle:
> Should `prevent_empty_post_meta()` and `default_post_metadata()` also be moved into `Post_Types`?

When we talk about these, we would also have to move:

* `prevent_empty_post_meta`
* `register_user_meta`

But then, the class would be very bloated again.

Should we only have the Post_Types defined there and introduce two other classes for `user_meta` and `post_meta`?

pfefferle:
I think I would move all the post_metas to the `Post_Types` class and we should think about having a `User_Metas` class.

includes/class-post-types.php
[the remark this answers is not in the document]
obenland: `sanitize_text_field()` feels good then

obenland:
> When we talk about these, we would also have to move:
>
> * prevent_empty_post_meta
> * register_user_meta

What's the connection between `prevent_empty_post_meta` and `register_user_meta`?

pfefferle:
> What's the connection between prevent_empty_post_meta and register_user_meta?

There is none, except what I wrote above: We might want to also have a dedicated space for user metas. That said, I had a look at the code and the metas and options are all over the place, because it is not always the best solution to have them in one place (options for example). So I think we are fine with your PR for now.

includes/class-post-types.php
```diff
@@ -0,0 +1,476 @@
...
+			'type'              => 'string',
+			'description'       => 'The type of the activity',
+			'single'            => true,
+			'show_in_rest'      => true,
+			'sanitize_callback' => function ( $value ) use ( $default_type ) {
+				$value  = ucfirst( strtolower( $value ) );
+				$schema = array(
+					'type'    => 'string',
+					'enum'    => Activity::TYPES,
+					'default' => $default_type,
```
obenland: @pfefferle There's a difference in default activity type meta between Inbox and Outbox post types, where it's `Create` for Inbox and `Announce` for Outbox. Is that on purpose?

pfefferle: We do not store `Announce`s in the inbox yet, so yes!

includes/class-post-types.php
```diff
@@ -119,61 +119,19 @@ public static function register_inbox_post_type() {
...
-					);
-
-					if ( \is_wp_error( \rest_validate_enum( $value, $schema, '' ) ) ) {
-						return $schema['default'];
-					}
-
-					return $value;
-				},
-			)
+			self::get_activity_actor_meta_args()
```
pfefferle: I understand why you did this, but it makes readability a lot worse! In that case I would prefer to do it the same way we did is with Extra_Fields: https://github.com/Automattic/wordpress-activitypub/blob/dbf0d11631fa74ba00b56c3ed51746d582cb40b6/includes/class-activitypub.php#L776

obenland: Reverted in [36ec110](https://github.com/Automattic/wordpress-activitypub/pull/2215/commits/36ec110d55f757e9aa833deb618dedbe1bbb22cd)

includes/class-activitypub.php
```diff
@@ -45,9 +42,9 @@ public static function init() {
 
 		\add_action( 'user_register', array( self::class, 'user_register' ) );
 
-		\add_filter( 'add_post_metadata', array( self::class, 'prevent_empty_post_meta' ), 10, 4 );
-		\add_filter( 'update_post_metadata', array( self::class, 'prevent_empty_post_meta' ), 10, 4 );
-		\add_filter( 'default_post_metadata', array( self::class, 'default_post_metadata' ), 10, 3 );
+		\add_filter( 'add_post_metadata', array( Post_Types::class, 'prevent_empty_post_meta' ), 10, 4 );
+		\add_filter( 'update_post_metadata', array( Post_Types::class, 'prevent_empty_post_meta' ), 10, 4 );
+		\add_filter( 'default_post_metadata', array( Post_Types::class, 'default_post_meta_data' ), 10, 3 );
```
pfefferle: why keep these here? I would prefer to move them too.

obenland: oh shoot, that's right, we have a `init()` there

includes/class-post-types.php
```diff
@@ -0,0 +1,548 @@
...
+	public static function init() {
+		self::register_remote_actors_post_type();
+		self::register_inbox_post_type();
+		self::register_outbox_post_type();
+		self::register_extra_fields_post_types();
+		self::register_activitypub_post_meta();
+
+		\add_action( 'rest_api_init', array( self::class, 'register_ap_actor_rest_field' ) );
+
+		\add_filter( 'activitypub_get_actor_extra_fields', array( Extra_Fields::class, 'default_actor_extra_fields' ), 10, 2 );
```
pfefferle, line 33: This feels a bit like a hack, we did because we wanted to keep the collections free from `init` hooks!

What do you think about also move the `default_actor_extra_fields` function to this class?

obenland: As an alternative, how about we call `default_actor_extra_fields()` from `get_actor_fields()` when they've not been created yet? That would also avoid the side effect of creating posts in a filter callback. Happy to do that in a follow-up.

pfefferle: I think we should keep the filter version, because not everyone might want to have the default fields. With the filter they can unhook the defaults and/or add their own ones.

obenland: Is that really a common use case? Would you not expect them to just delete these fields in the UI?

pfefferle: It is about site owners. Like https://blogs.vivaldi.net/ for example.

obenland: I'd leave it as is then. Moving it would break any existing unhooks. It's fine to stay in that class.

pfefferle (approved):

PR #2220, opened 2025-09-21
Fix critical security vulnerabilities, improve error handling, and enhance code quality in ActivityPub plugin admin files

ghost:


includes/wp-admin/import/class-mastodon.php
```diff
@@ -95,7 +95,7 @@ public static function handle_upload() {
 
 		\check_admin_referer( 'import-upload' );
 
-		if ( ! isset( $_FILES['import']['name'] ) ) {
+		if ( ! isset( $_FILES['import']['name'] ) || empty( $_FILES['import']['name'] ) ) {
```
pfefferle, line 98: Empty implicitly checks for `isset`, so no need to have both.

```suggestion
		if ( empty( $_FILES['import']['name'] ) ) {
```

includes/class-blocks.php
```diff
@@ -130,7 +142,12 @@ public static function enqueue_editor_assets() {
 			return;
 		}
 
-		$asset_data = include ACTIVITYPUB_PLUGIN_DIR . 'build/editor-plugin/plugin.asset.php';
+		$asset_file = ACTIVITYPUB_PLUGIN_DIR . 'build/editor-plugin/plugin.asset.php';
+		if ( ! file_exists( $asset_file ) ) {
+			return;
+		}
+
+		$asset_data = include $asset_file;
```
pfefferle, line 150: This file is part of the plugin, so there is no need to check if it is available.

includes/class-blocks.php
```diff
@@ -145,7 +162,12 @@ public static function handle_in_reply_to_get_param() {
 			return;
 		}
 
-		$asset_data = include ACTIVITYPUB_PLUGIN_DIR . 'build/reply-intent/plugin.asset.php';
+		$asset_file = ACTIVITYPUB_PLUGIN_DIR . 'build/reply-intent/plugin.asset.php';
+		if ( ! file_exists( $asset_file ) ) {
+			return;
+		}
+
+		$asset_data = include $asset_file;
```
pfefferle, line 170: same as above

includes/class-blocks.php
```diff
@@ -154,16 +176,28 @@ public static function handle_in_reply_to_get_param() {
...
+
+		$reply_block_path = ACTIVITYPUB_PLUGIN_DIR . '/build/reply';
+		if ( file_exists( $reply_block_path ) ) {
+			\register_block_type_from_metadata(
+				$reply_block_path,
+				array(
+					'render_callback' => array( self::class, 'render_reply_block' ),
+				)
+			);
+		}
```
pfefferle, line 200: This looks more complicated than the inital code, so I would not change it. There is also no need to check if the file exists, because the files are part of this plugin.

includes/class-blocks.php
```diff
@@ -240,7 +278,7 @@ public static function get_user_id( $user_string ) {
 
 		// For a single post, use the post author's ID.
 		if ( is_a( $queried_object, 'WP_Post' ) ) {
-			return get_the_author_meta( 'ID' );
+			return get_post_field( 'post_author', $queried_object->ID );
```
pfefferle, line 281: Why this change?

includes/class-blocks.php
```diff
@@ -287,12 +325,13 @@ public static function render_reply_block( $attrs ) {
...
 		if ( ! $show_embed ) {
+			$cleaned_url = str_replace( array( 'https://', 'http://' ), '', esc_url( $attrs['url'] ) );
 			$html .= sprintf(
-				'<p><a title="%2$s" aria-label="%2$s" href="%1$s" class="u-in-reply-to" target="_blank">%3$s</a></p>',
+				'<p><a title="%2$s" aria-label="%2$s" href="%1$s" class="u-in-reply-to" target="_blank" rel="noopener noreferrer">%3$s</a></p>',
 				esc_url( $attrs['url'] ),
 				esc_attr__( 'This post is a response to the referenced content.', 'activitypub' ),
 				// translators: %s is the URL of the post being replied to.
-				sprintf( __( '&#8620;%s', 'activitypub' ), \str_replace( array( 'https://', 'http://' ), '', esc_url( $attrs['url'] ) ) )
+				sprintf( __( '&#8620;%s', 'activitypub' ), esc_html( $cleaned_url ) )
```
pfefferle, line 334: what is the improvement here?

pfefferle:
Thanks for your contribution @Mukeshoad , I added some inline questions an feedback.

includes/wp-admin/import/class-mastodon.php
ghost: Thanks for catching that! You’re right — empty() already covers the isset() check. I’ll simplify it accordingly.

includes/class-blocks.php
ghost: Good point. Since these asset files are shipped with the plugin, existence checks aren’t really necessary. I’ll revert those changes to keep the code clean.

includes/class-blocks.php
Mukeshoad: Understood, I’ll simplify this back to the original approach and drop the file_exists() checks. Thanks for pointing that out.

includes/class-blocks.php
ghost: Understood, I’ll simplify this back to the original approach and drop the file_exists() checks. Thanks for pointing that out.

includes/class-blocks.php
ghost: That was an oversight on my end, I was aiming for consistency but I see now it changes behavior unnecessarily. I’ll revert this change to preserve the original logic.

includes/class-blocks.php
ghost: The intent was twofold:

Add rel="noopener noreferrer" for better security when opening links in a new tab.

Use esc_html() on the cleaned URL for stricter output escaping.

But I understand this may be unnecessary if we want to keep minimal diffs. I can revert or adjust based on your preference.

ghost:
> Thanks for your contribution @Mukeshoad , I added some inline questions an feedback.

Thanks a lot for the review and for the warm welcome, @pfefferle.

Just to give a bit of context, I’ve been working as a WordPress developer for about 13 years, but I’m relatively new to contributing here on GitHub. I’m still getting used to the workflow and best practices for open-source collaboration.

I really appreciate your detailed inline feedback. My main focus with this PR was to strengthen security (escaping, sanitization, nonce validation, etc.), and in some cases I may have been a little over-cautious with additional checks. Your comments are very helpful in understanding where the plugin already guarantees certain files/values, so I’ll adjust accordingly to keep the code lean and consistent.

I’ll work through your suggestions and update the PR. Thanks again for taking the time to review my changes, it’s a great learning experience to align my contributions with the project’s standards.

PR #2216, opened 2025-09-19
Add Perfecty Push integration for web push notifications

obenland:
I could have sworn there was a forum post asking for Web Push support. Well, I can't find it anymore.

## Proposed changes:

* Add new Perfecty Push integration class in `/integration/class-perfecty-push.php`
* Hook into ActivityPub `activitypub_handled_*` events to send web push notifications
* Support notifications for likes, reposts/boosts, comments, and new followers from the fediverse
* Add individual notification settings for users and blog actors (all types enabled by default)
* Include actor profile images in notifications with mp.jpg fallback for users without avatars
* Integrate settings into existing user profile and blog profile admin pages

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Install and activate the Perfecty Push plugin from WordPress.org
* Enable push notifications in browser when prompted by Perfecty Push
* Go to user profile page and verify "Push Notifications" settings appear
* Go to ActivityPub settings -> Blog Profile tab and verify "Push Notifications" settings appear
* Configure notification preferences (likes, reposts/boosts, comments, new followers)
* Have a fediverse account like, repost, comment on, or follow your ActivityPub account
* Verify web push notifications are received with correct titles, messages, and actor images
* Test that notification settings work independently for different users and blog actor

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

-   [x] Added - for new features
-   [ ] Changed - for changes in existing functionality
-   [ ] Deprecated - for soon-to-be removed features
-   [ ] Removed - for now removed features
-   [ ] Fixed - for any bug fixes
-   [ ] Security - in case of vulnerabilities

#### Message

Add Perfecty Push integration for web push notifications on ActivityPub events

</details>

PR #2223, opened 2025-09-22
Refactor notification system to use action hooks instead

pfefferle:
Deprecates the Notification class in favor of new `activitypub_handled_{type}` action hooks for all activity types. Updates all handler classes to trigger these hooks with consistent parameters, removes Notification usage, and updates integration and tests accordingly. This change improves extensibility and standardizes activity handling throughout the codebase.

For the context: https://github.com/Automattic/wordpress-activitypub/pull/2216#issuecomment-3313982333

## Proposed changes:
- Replace Notification usage with action hooks carrying consistent parameters (activity, user_id, state, context)
- Update handlers (`Accept`, `Reject`, `Follow`, `Move`, `Create`, `Delete`, `Like`, `Update`, `Undo`) to fire new hooks; adjust signatures to accept `$user_id`
- Update `Stream_Connector` to listen to `activitypub_handled_follow` and log accordingly; adjust tests to new method signatures

### Other information:

## Testing instructions:

* See unittests

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Standardized notification handling with new hooks for better extensibility and consistency.

</details>

### DIFF
--- a/includes/class-notification.php
+++ b/includes/class-notification.php
@@ -64,7 +64,22 @@ class Notification {
 	public function send() {
 		$type = \strtolower( $this->type );
 
+		/**
+		 * Action to send ActivityPub notifications.
+		 *
+		 * @deprecated unreleased Use "activitypub_handled_{$type}" instead.
+		 *
+		 * @param Notification $instance The notification object.
+		 */
 		\do_action_deprecated( 'activitypub_notification', array( $this ), 'unreleased', "activitypub_handled_{$type}" );
+
+		/**
+		 * Type-specific action to send ActivityPub notifications.
+		 *
+		 * @deprecated unreleased Use "activitypub_handled_{$type}" instead.
+		 *
+		 * @param Notification $instance The notification object.
+		 */
 		\do_action_deprecated( "activitypub_notification_{$type}", array( $this ), 'unreleased', "activitypub_handled_{$type}" );
 	}
 }
--- a/includes/class-notification.php
+++ b/includes/class-notification.php
@@ -9,6 +9,8 @@ namespace Activitypub;
 
 /**
  * Notification class.
+ *
+ * @deprecated unreleased Use action hooks like 'activitypub_handled_{type}' instead.
  */
 class Notification {
 	/**
@@ -48,6 +50,8 @@ class Notification {
 	 * @param int    $target   The WordPress User-Id.
 	 */
 	public function __construct( $type, $actor, $activity, $target ) {
+		\_deprecated_class( __CLASS__, 'unreleased', 'Use action hooks like "activitypub_handled_{type}" instead.' );
+
 		$this->type   = $type;
 		$this->actor  = $actor;
 		$this->object = $activity;
@@ -60,18 +64,7 @@ class Notification {
 	public function send() {
 		$type = \strtolower( $this->type );
 
-		/**
-		 * Action to send ActivityPub notifications.
-		 *
-		 * @param Notification $instance The notification object.
-		 */
-		do_action( 'activitypub_notification', $this );
-
-		/**
-		 * Type-specific action to send ActivityPub notifications.
-		 *
-		 * @param Notification $instance The notification object.
-		 */
-		do_action( "activitypub_notification_{$type}", $this );
+		\do_action_deprecated( 'activitypub_notification', array( $this ), 'unreleased', "activitypub_handled_{$type}" );
+		\do_action_deprecated( "activitypub_notification_{$type}", array( $this ), 'unreleased', "activitypub_handled_{$type}" );
 	}
 }
--- a/includes/handler/class-accept.php
+++ b/includes/handler/class-accept.php
@@ -54,10 +54,10 @@ class Accept {
 		/**
 		 * Fires after an ActivityPub Accept activity has been handled.
 		 *
-		 * @param array                $accept  The ActivityPub activity data.
-		 * @param int                  $user_id The local user ID.
-		 * @param bool                 $success Whether the follow request was successfully accepted.
-		 * @param \WP_Post | \WP_Error $result  The remote actor post or error.
+		 * @param array              $accept  The ActivityPub activity data.
+		 * @param int                $user_id The local user ID.
+		 * @param bool               $success Whether the follow request was successfully accepted.
+		 * @param \WP_Post|\WP_Error $result  The remote actor post or error.
 		 */
 		\do_action( 'activitypub_handled_accept', $accept, $user_id, $success, $result );
 	}
--- a/includes/handler/class-accept.php
+++ b/includes/handler/class-accept.php
@@ -48,18 +48,17 @@ class Accept {
 			return;
 		}
 
-		$reaction = Following::accept( $actor_post, $user_id );
-		$success  = ! \is_wp_error( $reaction );
+		$result = Following::accept( $actor_post, $user_id );
 
 		/**
 		 * Fires after an ActivityPub Accept activity has been handled.
 		 *
-		 * @param array      $accept   The ActivityPub activity data.
-		 * @param int|null   $user_id  The local user ID, or null if not applicable.
-		 * @param bool       $success  Whether the follow request was accepted successfully.
-		 * @param mixed      $reaction The result of accepting the follow request (contextual object).
+		 * @param array   $accept     The ActivityPub activity data.
+		 * @param int     $user_id    The local user ID.
+		 * @param mixed   $result     The result of accepting the follow request (contextual object).
+		 * @param WP_Post $actor_post The remote actor post.
 		 */
-		\do_action( 'activitypub_handled_accept', $accept, $user_id, $success, $reaction );
+		\do_action( 'activitypub_handled_accept', $accept, $user_id, $result, $actor_post );
 	}
 
 	/**
--- a/includes/handler/class-accept.php
+++ b/includes/handler/class-accept.php
@@ -55,10 +55,11 @@ class Accept {
 		 *
 		 * @param array      $accept   The ActivityPub activity data.
 		 * @param int|null   $user_id  The local user ID, or null if not applicable.
-		 * @param mixed      $reaction The result of accepting the follow request.
-		 * @param mixed|null $reaction The result of accepting the follow request (duplicate parameter).
+		 * @param bool       $success  Whether the follow request was accepted successfully.
+		 * @param mixed      $reaction The result of accepting the follow request (contextual object).
 		 */
-		\do_action( 'activitypub_handled_accept', $accept, $user_id, $reaction, $reaction );
+		$success = ! \is_wp_error( $reaction );
+		\do_action( 'activitypub_handled_accept', $accept, $user_id, $success, $reaction );
 	}
 
 	/**
--- a/includes/handler/class-accept.php
+++ b/includes/handler/class-accept.php
@@ -10,7 +10,6 @@ namespace Activitypub\Handler;
 use Activitypub\Collection\Following;
 use Activitypub\Collection\Outbox;
 use Activitypub\Collection\Remote_Actors;
-use Activitypub\Notification;
 
 use function Activitypub\object_to_uri;
 
@@ -22,19 +21,8 @@ class Accept {
 	 * Initialize the class, registering WordPress hooks.
 	 */
 	public static function init() {
-		\add_action(
-			'activitypub_inbox_accept',
-			array( self::class, 'handle_accept' ),
-			10,
-			2
-		);
-
-		\add_filter(
-			'activitypub_validate_object',
-			array( self::class, 'validate_object' ),
-			10,
-			3
-		);
+		\add_action( 'activitypub_inbox_accept', array( self::class, 'handle_accept' ), 10, 2 );
+		\add_filter( 'activitypub_validate_object', array( self::class, 'validate_object' ), 10, 3 );
 	}
 
 	/**
@@ -60,16 +48,17 @@ class Accept {
 			return;
 		}
 
-		Following::accept( $actor_post, $user_id );
+		$reaction = Following::accept( $actor_post, $user_id );
 
-		// Send notification.
-		$notification = new Notification(
-			'accept',
-			$actor_post->guid,
-			$accept,
-			$user_id
-		);
-		$notification->send();
+		/**
+		 * Fires after an ActivityPub Accept activity has been handled.
+		 *
+		 * @param array      $accept   The ActivityPub activity data.
+		 * @param int|null   $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $reaction The result of accepting the follow request.
+		 * @param mixed|null $reaction The result of accepting the follow request (duplicate parameter).
+		 */
+		\do_action( 'activitypub_handled_accept', $accept, $user_id, $reaction, $reaction );
 	}
 
 	/**
--- a/includes/handler/class-accept.php
+++ b/includes/handler/class-accept.php
@@ -48,17 +48,18 @@ class Accept {
 			return;
 		}
 
-		$result = Following::accept( $actor_post, $user_id );
+		$result  = Following::accept( $actor_post, $user_id );
+		$success = ! \is_wp_error( $result );
 
 		/**
 		 * Fires after an ActivityPub Accept activity has been handled.
 		 *
-		 * @param array    $accept     The ActivityPub activity data.
-		 * @param int      $user_id    The local user ID.
-		 * @param mixed    $result     The result of accepting the follow request (contextual object).
-		 * @param \WP_Post $actor_post The remote actor post.
+		 * @param array                $accept  The ActivityPub activity data.
+		 * @param int                  $user_id The local user ID.
+		 * @param bool                 $success Whether the follow request was successfully accepted.
+		 * @param \WP_Post | \WP_Error $result  The remote actor post or error.
 		 */
-		\do_action( 'activitypub_handled_accept', $accept, $user_id, $result, $actor_post );
+		\do_action( 'activitypub_handled_accept', $accept, $user_id, $success, $result );
 	}
 
 	/**
--- a/includes/handler/class-accept.php
+++ b/includes/handler/class-accept.php
@@ -53,10 +53,10 @@ class Accept {
 		/**
 		 * Fires after an ActivityPub Accept activity has been handled.
 		 *
-		 * @param array   $accept     The ActivityPub activity data.
-		 * @param int     $user_id    The local user ID.
-		 * @param mixed   $result     The result of accepting the follow request (contextual object).
-		 * @param WP_Post $actor_post The remote actor post.
+		 * @param array    $accept     The ActivityPub activity data.
+		 * @param int      $user_id    The local user ID.
+		 * @param mixed    $result     The result of accepting the follow request (contextual object).
+		 * @param \WP_Post $actor_post The remote actor post.
 		 */
 		\do_action( 'activitypub_handled_accept', $accept, $user_id, $result, $actor_post );
 	}
--- a/includes/handler/class-accept.php
+++ b/includes/handler/class-accept.php
@@ -56,7 +56,7 @@ class Accept {
 		 *
 		 * @param array              $accept  The ActivityPub activity data.
 		 * @param int                $user_id The local user ID.
-		 * @param bool               $success Whether the follow request was successfully accepted.
+		 * @param bool               $success True on success, false otherwise.
 		 * @param \WP_Post|\WP_Error $result  The remote actor post or error.
 		 */
 		\do_action( 'activitypub_handled_accept', $accept, $user_id, $success, $result );
--- a/includes/handler/class-accept.php
+++ b/includes/handler/class-accept.php
@@ -49,6 +49,7 @@ class Accept {
 		}
 
 		$reaction = Following::accept( $actor_post, $user_id );
+		$success  = ! \is_wp_error( $reaction );
 
 		/**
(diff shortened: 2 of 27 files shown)


includes/handler/class-accept.php
```diff
@@ -60,16 +48,18 @@ public static function handle_accept( $accept, $user_id ) {
...
-		);
-		$notification->send();
+		$reaction = Following::accept( $actor_post, $user_id );
+		$success  = ! \is_wp_error( $reaction );
+
+		/**
+		 * Fires after an ActivityPub Accept activity has been handled.
+		 *
+		 * @param array      $accept   The ActivityPub activity data.
+		 * @param int|null   $user_id  The local user ID, or null if not applicable.
```
obenland: Should just be `int` in this case, no? I'd expect the `is_wp_error()` check above to catch any invalid user_ids.

includes/handler/class-accept.php
```diff
@@ -60,16 +48,18 @@ public static function handle_accept( $accept, $user_id ) {
...
+		 * @param bool       $success  Whether the follow request was accepted successfully.
+		 * @param mixed      $reaction The result of accepting the follow request (contextual object).
+		 */
+		\do_action( 'activitypub_handled_accept', $accept, $user_id, $success, $reaction );
```
obenland: Instead of `$success`, how would you feel about making `$actor_post` the third parameter? That way that wouldn't have to be queried again.

Is `$reaction` a hold-over from saving likes and reposts? Could it be `$result`?

includes/handler/class-announce.php
```diff
@@ -114,13 +109,13 @@ public static function maybe_save_announce( $activity, $user_id ) {
...
 		/**
-		 * Fires after an Announce has been saved.
+		 * Fires after an ActivityPub Announce activity has been handled.
 		 *
-		 * @param array $activity The activity-object.
-		 * @param int   $user_id  The id of the local blog-user.
-		 * @param mixed $state    The state of the reaction.
-		 * @param mixed $reaction The reaction.
+		 * @param array      $activity The ActivityPub activity data.
+		 * @param int|null   $user_id  The local user ID, or null if not applicable.
```
obenland: When is `$user_id` null? It should be required by the Inbox.

includes/handler/class-delete.php
```diff
@@ -107,9 +120,11 @@ public static function maybe_delete_follower( $activity ) {
 
 		// Verify that Actor is deleted.
 		if ( ! is_wp_error( $follower ) && Tombstone::exists( $activity['actor'] ) ) {
-			Remote_Actors::delete( $follower->ID );
+			$state = Remote_Actors::delete( $follower->ID );
 			self::maybe_delete_interactions( $activity );
 		}
+
+		return $state ?? null;
```
obenland:
```suggestion
		return $state ?? false;
```

I assume `null` would be a failure result? Let's also add `@return` docs to this function

includes/handler/class-follow.php
```diff
@@ -44,46 +43,39 @@ public static function handle_follow( $activity, $user_id ) {
...
+		 * Fires after a Follow activity has been handled.
 		 *
-		 * @param string             $actor        The URL of the actor (follower) who initiated the follow.
-		 * @param array              $activity     The complete activity data of the follow request.
-		 * @param int                $user_id      The ID of the WordPress user being followed.
-		 * @param \WP_Post|\WP_Error $remote_actor The Actor object containing the new follower's data.
+		 * @param array      $activity     The ActivityPub activity data.
+		 * @param int|null   $user_id      The local user ID, or null if not applicable.
+		 * @param mixed      $remote_actor The WP_Post object of the remote actor/follower, or WP_Error if failed.
+		 * @param mixed|null $remote_actor The WP_Post object of the remote actor/follower (duplicate parameter).
```
obenland: When deprecating a hook, it's fine to change the parameters, too. I'd suggest keeping the original hook docs intact and adding a `@deprecated` annotation, then define the new hook after, with just one remote actor arg.

includes/handler/class-move.php
```diff
@@ -53,20 +54,15 @@ public static function handle_move( $activity ) {
 
 		$target_object = Remote_Actors::get_by_uri( $target_uri );
 		$origin_object = Remote_Actors::get_by_uri( $origin_uri );
+		$status        = null;
```
obenland: If we default to `false`, we could avoid the next conditional, be specific about the data type in hook docs, and still stay true to the success of the operation?

includes/class-notification.php
```diff
@@ -60,18 +64,7 @@ public function __construct( $type, $actor, $activity, $target ) {
...
-		 */
-		do_action( 'activitypub_notification', $this );
-
-		/**
-		 * Type-specific action to send ActivityPub notifications.
-		 *
-		 * @param Notification $instance The notification object.
-		 */
-		do_action( "activitypub_notification_{$type}", $this );
+		\do_action_deprecated( 'activitypub_notification', array( $this ), 'unreleased', "activitypub_handled_{$type}" );
```
obenland, line 74: Could we keep the docs and add `@deprecated` annotations?

includes/handler/class-accept.php
pfefferle: yea, I think AI tried to unify the docs over all hooks and because some have null...

includes/handler/class-accept.php
pfefferle: Good point, but I will change the order then. the third result should be the status and the forth a context.

includes/handler/class-follow.php
pfefferle: It is about the consistency with the other hooks, not with the deprecated one!

includes/handler/class-accept.php
```diff
@@ -60,16 +48,17 @@ public static function handle_accept( $accept, $user_id ) {
...
-		$notification->send();
+		$result = Following::accept( $actor_post, $user_id );
+
+		/**
+		 * Fires after an ActivityPub Accept activity has been handled.
+		 *
+		 * @param array   $accept     The ActivityPub activity data.
+		 * @param int     $user_id    The local user ID.
+		 * @param mixed   $result     The result of accepting the follow request (contextual object).
+		 * @param WP_Post $actor_post The remote actor post.
```
obenland:
```suggestion
		 * @param array    $accept     The ActivityPub activity data.
		 * @param int      $user_id    The local user ID.
		 * @param mixed    $result     The result of accepting the follow request (contextual object).
		 * @param \WP_Post $actor_post The remote actor post.
```

includes/handler/class-announce.php
```diff
@@ -106,21 +101,21 @@ public static function maybe_save_announce( $activity, $user_id ) {
...
+		 * Fires after an ActivityPub Announce activity has been handled.
 		 *
-		 * @param array $activity The activity-object.
-		 * @param int   $user_id  The id of the local blog-user.
-		 * @param mixed $state    The state of the reaction.
-		 * @param mixed $reaction The reaction.
+		 * @param array      $activity The ActivityPub activity data.
+		 * @param int        $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
+		 * @param mixed|null $result   The WP_Comment object of the created announce/repost comment, or null if creation failed.
```
obenland:
```suggestion
		 * @param array            $activity The ActivityPub activity data.
		 * @param int              $user_id  The local user ID.
		 * @param mixed            $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
		 * @param \WP_Comment|null $result   The WP_Comment object of the created announce/repost comment, or null if creation failed.
```

includes/handler/class-create.php
```diff
@@ -72,22 +61,22 @@ public static function handle_create( $activity, $user_id, $activity_object = nu
...
+		 * Fires after an ActivityPub Create activity has been handled.
 		 *
-		 * @param array                      $activity The activity-object.
-		 * @param int                        $user_id  The id of the local blog-user.
-		 * @param \WP_Comment|\WP_Error      $state    The comment object or WP_Error.
-		 * @param \WP_Comment|\WP_Error|null $reaction The reaction object or WP_Error.
+		 * @param array      $activity The ActivityPub activity data.
+		 * @param int.       $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
+		 * @param mixed|null $result   The WP_Comment object of the created comment, or null if creation failed.
```
obenland:
```suggestion
		 * @param array               $activity The ActivityPub activity data.
		 * @param int                 $user_id  The local user ID.
		 * @param int|false|\WP_Error $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
		 * @param \WP_Comment|null    $result   The WP_Comment object of the created comment, or null if creation failed.
```

includes/handler/class-delete.php
```diff
@@ -95,21 +98,35 @@ public static function handle_delete( $activity ) {
...
+
+		/**
+		 * Fires after an ActivityPub Delete activity has been handled.
+		 *
+		 * @param array      $activity The ActivityPub activity data.
+		 * @param int|null   $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
+		 * @param mixed|null $reaction The result of the delete operation (e.g., WP_Comment object or deletion status).
+		 */
+		\do_action( 'activitypub_handled_delete', $activity, $user_id, $state, $reaction );
```
obenland: Can we remove `$reaction`? It's always `null`.

includes/handler/class-delete.php
```diff
@@ -95,21 +98,35 @@ public static function handle_delete( $activity ) {
 				// Maybe handle Delete Activity for other Object Types.
 				break;
 		}
+
+		/**
+		 * Fires after an ActivityPub Delete activity has been handled.
+		 *
+		 * @param array      $activity The ActivityPub activity data.
+		 * @param int|null   $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
```
obenland:
```suggestion
		 * @param int        $user_id  The local user ID.
		 * @param bool       $state    True on successful deletion, false otherwise.
```

includes/handler/class-delete.php
```diff
@@ -95,21 +98,35 @@ public static function handle_delete( $activity ) {
...
+		 */
+		\do_action( 'activitypub_handled_delete', $activity, $user_id, $state, $reaction );
 	}
 
 	/**
 	 * Delete a Follower if Actor-URL is a Tombstone.
 	 *
 	 * @param array $activity The delete activity.
+	 *
+	 * @return mixed The state/result of the delete operation, or false if no action was taken.
```
obenland:
```suggestion
	 * @return bool True on success, false otherwise.
```
Should we return `true` when no action was taken to signal it wasn't a failure?

includes/handler/class-follow.php
```diff
@@ -44,46 +43,49 @@ public static function handle_follow( $activity, $user_id ) {
...
-		$notification->send();
+		/**
+		 * Fires after a Follow activity has been handled.
+		 *
+		 * @param array    $activity     The ActivityPub activity data.
+		 * @param int|null $user_id      The local user ID, or null if not applicable.
+		 * @param mixed    $remote_actor The WP_Post object of the remote actor/follower, or WP_Error if failed.
+		 * @param array    $activity     The ActivityPub activity data.
+		 */
+		\do_action( 'activitypub_handled_follow', $activity, $user_id, $remote_actor, $activity );
```
obenland: Can we remove the second `$activity`?

includes/handler/class-follow.php
```diff
@@ -44,46 +43,49 @@ public static function handle_follow( $activity, $user_id ) {
...
-			$activity,
-			$user_id
-		);
-		$notification->send();
+		/**
+		 * Fires after a Follow activity has been handled.
+		 *
+		 * @param array    $activity     The ActivityPub activity data.
+		 * @param int|null $user_id      The local user ID, or null if not applicable.
+		 * @param mixed    $remote_actor The WP_Post object of the remote actor/follower, or WP_Error if failed.
```
obenland:
```suggestion
		 * @param array              $activity     The ActivityPub activity data.
		 * @param int                $user_id      The local user ID.
		 * @param \WP_Post|\WP_Error $remote_actor The remote actor/follower, or WP_Error if failed.
```

includes/handler/class-follow.php
```diff
@@ -44,46 +43,49 @@ public static function handle_follow( $activity, $user_id ) {
...
 	 * Send Accept response.
 	 *
-	 * @param string             $actor           The Actor URL.
-	 * @param array              $activity_object The Activity object.
-	 * @param int                $user_id         The ID of the WordPress User.
-	 * @param \WP_Post|\WP_Error $remote_actor    The Actor object.
+	 * @param array      $activity_object The ActivityPub activity data.
+	 * @param int|null   $user_id         The local user ID, or null if not applicable.
+	 * @param mixed      $status          The WP_Post object of the remote actor/follower, or WP_Error if failed.
+	 * @param mixed|null $remote_actor    The WP_Post object of the remote actor/follower (duplicate parameter).
```
obenland: These seem to no longer line up with the args in the `activitypub_handled_follow` action.

includes/handler/class-inbox.php
```diff
@@ -61,12 +61,12 @@ public static function handle_inbox_requests( $data, $user_id, $type, $activity
...
+		 * Fires after an ActivityPub Inbox activity has been handled.
 		 *
-		 * @param array              $data     The data array.
-		 * @param int                $user_id  The ID of the local blog user.
-		 * @param \WP_Error|int      $id       The ID of the inbox item.
-		 * @param Activity|\WP_Error $activity The Activity object.
+		 * @param array      $data     The ActivityPub activity data.
+		 * @param int        $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $id       The ID of the inbox item that was created, or WP_Error if failed.
+		 * @param mixed|null $activity The Activity object or WP_Error from processing the inbox request.
```
```suggestion
		 * @param array              $data     The ActivityPub activity data.
		 * @param int                $user_id  The local user ID.
		 * @param \WP_Error|int      $id       The ID of the inbox item that was created, or WP_Error if failed.
		 * @param Activity|\WP_Error $activity The Activity object or WP_Error from processing the inbox request.
```

includes/handler/class-like.php
```diff
@@ -46,22 +46,22 @@ public static function handle_like( $like, $user_id ) {
...
+		 * Fires after an ActivityPub Like activity has been handled.
 		 *
-		 * @param array $like     The Activity array.
-		 * @param int   $user_id  The ID of the local blog user.
-		 * @param mixed $state    The state of the reaction.
-		 * @param mixed $reaction The reaction object.
+		 * @param array      $like     The ActivityPub activity data.
+		 * @param int        $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
+		 * @param mixed|null $result   The WP_Comment object of the created like comment, or null if creation failed.
```
obenland:
```suggestion
		 * @param array                            $like     The ActivityPub activity data.
		 * @param int                              $user_id  The local user ID.
		 * @param array|false|int|string|\WP_Error $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
		 * @param \WP_Comment|null                 $result   The WP_Comment object of the created like comment, or null if creation failed.
```

includes/handler/class-move.php
```diff
@@ -98,8 +86,18 @@ public static function handle_move( $activity ) {
...
+			$status = \wp_delete_post( $origin_object->ID );
 		}
+
+		/**
+		 * Fires after an ActivityPub Move activity has been handled.
+		 *
+		 * @param array $activity The ActivityPub activity data.
+		 * @param int   $user_id  The local user ID, or null if not applicable.
+		 * @param bool  $status   The status result of the move operation.
+		 * @param array $context  Array containing the origin and target objects: array( $origin_object, $target_object ).
```
obenland: If you want to pass them, it's probably fine to do as separate parameters.

includes/handler/class-reject.php
```diff
@@ -75,16 +63,17 @@ private static function reject_follow( $reject, $user_id ) {
...
-		$notification->send();
+		$status = Following::reject( $actor_post, $user_id );
+
+		/**
+		 * Fires after an ActivityPub Reject activity has been handled.
+		 *
+		 * @param array      $reject     The ActivityPub activity data.
+		 * @param int        $user_id    The local user ID, or null if not applicable.
+		 * @param mixed      $status     The status result of the rejection operation.
+		 * @param mixed|null $actor_post The WP_Post object of the actor being rejected, or WP_Error if failed.
```
obenland:
```suggestion
		 * @param array              $reject     The ActivityPub activity data.
		 * @param int                $user_id    The local user ID.
		 * @param \WP_Post|\WP_Error $status     Actor post on success, WP_Error on failure.
		 * @param \WP_Post           $actor_post The actor whose follow request was rejected.
```

includes/handler/class-undo.php
```diff
@@ -76,12 +71,13 @@ public static function handle_undo( $activity, $user_id ) {
...
-		 * Fires after an "Undo" activity has been handled.
+		 * Fires after an ActivityPub Undo activity has been handled.
 		 *
-		 * @param array    $activity The JSON "Undo" Activity.
-		 * @param int|null $user_id  The ID of the user who initiated the "Undo" activity otherwise null.
-		 * @param mixed    $state    The state of the "Undo" activity.
+		 * @param array      $activity The ActivityPub activity data.
+		 * @param int        $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
+		 * @param WP_Comment $comment  Always null for Undo activities.
```
obenland:
```suggestion
		 * @param array       $activity The ActivityPub activity data.
		 * @param int         $user_id  The local user ID.
		 * @param bool        $state    True on success, false on failure.
		 * @param \WP_Comment $comment  The comment that was deleted.
```

includes/handler/class-update.php
```diff
@@ -77,42 +75,54 @@ public static function handle_update( $activity ) {
...
+		 * Fires after an ActivityPub Update activity has been handled.
 		 *
-		 * @param array            $activity The complete Update activity data.
-		 * @param null             $user     Always null for Update activities.
-		 * @param int|array        $state    1 if comment was updated successfully, error data otherwise.
-		 * @param \WP_Comment|null $reaction The updated comment object if successful, null otherwise.
+		 * @param array      $activity The ActivityPub activity data.
+		 * @param int      $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
+		 * @param mixed|null $result   The WP_Comment object of the updated comment, or null if update failed.
```
```suggestion
		 * @param array            $activity The ActivityPub activity data.
		 * @param int              $user_id  The local user ID.
		 * @param mixed            $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
		 * @param \WP_Comment|null $result   The updated comment, or null if update failed.
```

includes/handler/class-update.php
```diff
@@ -77,42 +75,54 @@ public static function handle_update( $activity ) {
...
-		Remote_Actors::upsert( $actor );
+		$state = Remote_Actors::upsert( $actor );
+
+		/**
+		 * Fires after an ActivityPub Update activity has been handled.
+		 *
+		 * @param array      $activity The ActivityPub activity data.
+		 * @param int|null   $user_id  The local user ID, or null if not applicable.
+		 * @param mixed      $state    The state/result of the operation (e.g., comment ID, WP_Error, or status).
+		 * @param mixed|null $actor    Additional context or resulting object (e.g., comment, reaction), or null if not applicable.
```
```suggestion
		 * @param array         $activity The ActivityPub activity data.
		 * @param int.          $user_id  The local user ID.
		 * @param int|\WP_Error $state    Actor post ID on success, WP_Error on failure.
		 * @param array         $actor    Remote actor meta data.
```

includes/handler/class-update.php
```diff
@@ -77,42 +75,54 @@ public static function handle_update( $activity ) {
...
+		 */
+		\do_action( 'activitypub_handled_update', $activity, $user_id, $state, $actor );
```
obenland, line 126: With the introduction of this action, we might want to consider appending the type to differentiate between actor and interaction updates and not pass different data types in the same action.

obenland (commented):
One more thing I noticed was that we don't bail as much anymore and pass through WP_Errors to these hooks.

Hook naming (`*_handled_*`) and the way they executed up until now suggest that there's a reasonable expectation of success when they run. Like, the only reason these hooks would receive null or WP_Errors would be because there was some internal failure while handling a request.

I don't think we need to make these too generally applicable, that's what I'd consider `activitypub_inbox_{type}` hooks to be for. `activitypub_handled_` hooks should probably be more something to use if you want to do something in response to something the plugin did.

includes/handler/class-delete.php
pfefferle: I will spent some more time to think about the params. The idea is to have the same number and kind of params for every hook, to be as consistent as possible and to build hooks that work with the same number of params.

obenland: I think it's fine for that to be $activity and $user_id and for the rest to be type-specific (maybe a boolean $success instead of $state?). At the point where we're passing null or duplicate variables, it doesn't add much value, I think.

obenland (approved):

PR #2236, opened 2025-09-26
Refine handling of local content visibility for posts

pfefferle:
This PR aims to send an `Update` activity when a previously federated post is later set to local, reflecting changes to the audience (`to`, `cc` fields) and `in_reply_to`.

The benefit of this change (if it works as intended) over sending a `Delete` is that the post can be reactivated by simply changing its visibility.

Improves logic for 'activitypub_content_visibility' set to 'local', ensuring posts are only disabled if not already federated. Updates transformer logic to set correct recipients for local posts, prevents in-reply-to for local visibility, and adds/updates related unit tests to cover these scenarios.

For reference https://wordpress.org/support/topic/how-to-delete-federated-posts/

## Proposed changes:

* Send Updates when a post was published before and set to `local` afterwards
* Clear `in_reply_to` (not sure if that is needed)
* Only set Post-Author to `to` and `cc` (not if that is needed or if empty arrays would also do the job)

### Other information:

## Testing instructions:

* Publish a `public` post
* Change visibility to `local`
* Check Outbox if `Update` is properly created

* Write new post and set to `local` before publishing
* Check that no `Update` / `Create` Activities were created

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Improved handling of posts switched to 'local' visibility so they update correctly and can be re-enabled by changing visibility again.

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -337,13 +337,10 @@ function is_post_disabled( $post ) {
 
 	if (
 		! $disabled &&
-		\in_array( $visibility, array( ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL, ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE ), true )
+		\in_array( $visibility, array( ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL, ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE ), true ) &&
+		'federated' !== \get_post_meta( $post->ID, 'activitypub_status', true )
 	) {
-		if ( 'federated' === \get_post_meta( $post->ID, 'activitypub_status', true ) ) {
-			// If post was already federated, we need to send an update with the new visibility.
-			$disabled = false;
-
-		} else {
+			// If post wasn't federated, we don't need to send an update with the new visibility.
 			$disabled = true;
 		}
 	}
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -325,16 +325,27 @@ function is_post_disabled( $post ) {
 		return true;
 	}
 
-	$visibility = \get_post_meta( $post->ID, 'activitypub_content_visibility', true );
-
 	if (
-		ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL === $visibility ||
-		ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE === $visibility ||
 		! \post_type_supports( $post->post_type, 'activitypub' ) ||
 		'private' === $post->post_status ||
 		! empty( $post->post_password )
 	) {
 		$disabled = true;
+	}
+
+	$visibility = \get_post_meta( $post->ID, 'activitypub_content_visibility', true );
+
+	if (
+		! $disabled &&
+		\in_array( $visibility, array( ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL, ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE ), true )
+	) {
+		if ( 'federated' === \get_post_meta( $post->ID, 'activitypub_status', true ) ) {
+			// If post was already federated, we need to send an update with the new visibility.
+			$disabled = false;
+
+		} else {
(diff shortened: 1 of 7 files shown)


includes/transformer/class-base.php
```diff
@@ -206,6 +206,10 @@ protected function set_audience( $activity_object ) {
 			case ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE:
 				$activity_object->add_to( $mentions );
 				$activity_object->add_to( $replied_to );
+				break;
+			case ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL:
+				$activity_object->add_to( $actor->get_id() );
+				$activity_object->add_cc( $actor->get_id() );
```
pfefferle, line 212: not sure if we have to set the `to` and `cc` fields instead of keeping them empty.

includes/transformer/class-post.php
```diff
@@ -578,6 +578,10 @@ public function generate_reply_link( $block_content, $block ) {
 	 * @return string|array|null The in-reply-to URL of the post.
 	 */
 	protected function get_in_reply_to() {
+		if ( ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL === $this->get_content_visibility() ) {
+			return null;
+		}
```
pfefferle: not sure if we should clear the `in_reply_to` or simply keep it.

includes/transformer/class-base.php
obenland: Yeah me neither. My first instinct would have been to keep them empty. What prompted you to set them?

includes/functions.php
```diff
@@ -325,18 +325,29 @@ function is_post_disabled( $post ) {
...
+		\in_array( $visibility, array( ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL, ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE ), true )
+	) {
+		if ( 'federated' === \get_post_meta( $post->ID, 'activitypub_status', true ) ) {
+			// If post was already federated, we need to send an update with the new visibility.
+			$disabled = false;
+
+		} else {
+			$disabled = true;
+		}
+	}
```
obenland, line 345:
```suggestion
	if (
		! $disabled &&
		\in_array( $visibility, array( ACTIVITYPUB_CONTENT_VISIBILITY_LOCAL, ACTIVITYPUB_CONTENT_VISIBILITY_PRIVATE ), true ) &&
		'federated' !== \get_post_meta( $post->ID, 'activitypub_status', true )
	) {
			// If post wasn't federated, we don't need to send an update with the new visibility.
			$disabled = true;
		}
	}
```
Can be simplified since `$disabled` is already false in this conditional.

includes/transformer/class-base.php
pfefferle: I am not sure if it will be simply ignored with an empty to and cc, so I made it specific!?

pfefferle:
After some feedback from Mastodon contributors: This is actually not possible. We would have to send a `Delete` in this case.

The risk here: A `Delete` is irreversible and a re-publish is not possible then.

pfefferle:
@obenland the question is if we want to send a delete!?

obenland:
Maybe we could add a notice into the visibility selector in the editor, warning that it's a permanent change?

Other than that, it feels like a specific-enough action to me, to assume that it doesn't happen accidentally. I'd be fine with sending the delete even without a warning.

Jiwoon-Kim:
In the case of Misskey, there is an option for delete-and-edit. For WordPress federation, this could be implemented by actually deleting the original post and publishing a duplicate as a new post with a new ID.

Of course, any existing comments or interactions would be lost.

Jiwoon-Kim:
> For WordPress federation, this could be implemented by actually deleting the original post and publishing a duplicate as a new post with a new ID.

By the way, couldn’t this approach also be applied to actors? I have a hunch that threads.net might be using a similar method to implement federation opt-out.

PR #2240, opened 2025-09-29
Add QuoteRequest activity handler support

### LINKED ISSUE #2239: The plugin should be able to receive and approve QuoteRequest activity for FEP-044f implementation
To implement [FEP-044f](https://codeberg.org/fediverse/fep/src/branch/main/fep/044f/fep-044f.md)
, the WordPress ActivityPub plugin needs to support receiving and approving QuoteRequest activities.
This would enable proper interoperability with Mastodon and other software that follows the specification.

pfefferle:
Introduces a new handler for QuoteRequest activities as defined in FEP-044f, including registration in the handler initialization and updates to the activity class to support the new activity type and context. This enables processing and validation of QuoteRequest activities according to interaction policies.

Fixes #2239

Here is an example:

1. Quoted Post on Mastodon

Screenshot 2025-09-30 at 14 02 05

2. The `Accept` response

Screenshot 2025-09-30 at 14 02 39

3. The `QuoteAuthorization` object https://pfefferle.org/?p=4047&stamp=17034&activitypub

## Proposed changes:
- Adds a new QuoteRequest handler class with policy-based acceptance/rejection logic
- Registers the QuoteRequest activity type in the Activity class and handler initialization
- Creates a QuoteAuthorization extended object class for the activity context

### Other information:

## Testing instructions:

* Checkout this PR on one of your Blogs
* Publish a post
* Quote this post on Mastodon

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added support for QuoteRequest activities (FEP-044f), enabling proper handling, validation, and policy-based acceptance or rejection of quote requests.

</details>

### DIFF
--- a/includes/class-query.php
+++ b/includes/class-query.php
@@ -404,7 +404,7 @@ class Query {
 				'p'     => $post->ID,
 				'stamp' => $meta->meta_id,
 			),
-			\trailingslashit( \home_url() )
+			\home_url( '/' )
 		);
 
 		$activitypub_object = new Quote_Authorization();
--- a/includes/class-query.php
+++ b/includes/class-query.php
@@ -7,9 +7,13 @@
 
 namespace Activitypub;
 
+use Activitypub\Activity\Extended_Object\Quote_Authorization;
 use Activitypub\Collection\Actors;
 use Activitypub\Collection\Outbox;
 use Activitypub\Transformer\Factory;
+
+use function Activitypub\get_post_id;
+use function Activitypub\get_user_id;
 
 /**
  * Singleton class to handle and store the ActivityPub query.
@@ -137,6 +141,10 @@ class Query {
 	 */
 	private function prepare_activitypub_data() {
 		$queried_object = $this->get_queried_object();
+
+		if ( $queried_object instanceof \WP_Post && \get_query_var( 'stamp' ) ) {
+			return $this->maybe_get_stamp();
+		}
 
 		// Check for Outbox Activity.
 		if (
@@ -311,7 +319,7 @@ class Query {
 	 */
 	public function should_negotiate_content() {
 		$return           = false;
-		$always_negotiate = array( 'p', 'c', 'author', 'actor', 'preview', 'activitypub' );
+		$always_negotiate = array( 'p', 'c', 'author', 'actor', 'stamp', 'preview', 'activitypub' );
 		$url              = \wp_parse_url( $this->get_request_url(), PHP_URL_QUERY );
 		$query            = array();
 		\wp_parse_str( $url, $query );
@@ -367,5 +375,47 @@ class Query {
 	 */
 	public function set_old_host_request( $state = true ) {
 		$this->is_old_host_request = $state;
+	}
+
+	/**
+	 * Maybe get a QuoteAuthorization object from a stamp.
+	 *
+	 * @return bool True if the object was prepared, false otherwise.
+	 */
+	private function maybe_get_stamp() {
+		require_once ABSPATH . 'wp-admin/includes/post.php';
+
+		$stamp = \get_query_var( 'stamp' );
+		$meta  = \get_post_meta_by_id( (int) $stamp );
+
+		if ( ! $meta ) {
+			return false;
+		}
+
+		$post     = $this->get_queried_object();
+		$user_uri = get_user_id( $post->post_author );
+
+		if ( ! $user_uri ) {
+			return false;
+		}
+
+		$stamp_uri = \add_query_arg(
+			array(
+				'p'     => $post->ID,
+				'stamp' => $meta->meta_id,
+			),
+			\trailingslashit( \home_url() )
+		);
+
+		$activitypub_object = new Quote_Authorization();
+		$activitypub_object->set_id( $stamp_uri );
+		$activitypub_object->set_attributed_to( $user_uri );
+		$activitypub_object->set_interacting_object( $meta->meta_value );
+		$activitypub_object->set_interaction_target( get_post_id( $post->ID ) );
+
+		$this->activitypub_object    = $activitypub_object;
+		$this->activitypub_object_id = $activitypub_object->get_id();
+
+		return true;
 	}
 }
--- a/includes/class-query.php
+++ b/includes/class-query.php
@@ -12,9 +12,6 @@ use Activitypub\Collection\Actors;
 use Activitypub\Collection\Outbox;
 use Activitypub\Transformer\Factory;
 
-use function Activitypub\get_post_id;
-use function Activitypub\get_user_id;
-
 /**
  * Singleton class to handle and store the ActivityPub query.
  */
--- a/includes/handler/class-quote-request.php
+++ b/includes/handler/class-quote-request.php
@@ -38,8 +38,14 @@ class Quote_Request {
 	 * @param int   $user_id  The user ID.
 	 */
 	public static function handle_quote_request( $activity, $user_id ) {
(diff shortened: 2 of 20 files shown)


includes/handler/class-quote-request.php
```diff
@@ -0,0 +1,235 @@
...
+		if ( ! $post_id ) {
+			self::queue_reject( $activity, $user_id );
+			return;
+		}
+
+		$content_policy = \get_post_meta( $post_id, 'activitypub_interaction_policy_quote', true );
+
+		switch ( $content_policy ) {
+			case ACTIVITYPUB_INTERACTION_POLICY_ME:
+				self::queue_reject( $activity, $user_id, $post_id );
```
obenland: No `$state = false;` in this case? Is it considered a success?

includes/handler/class-quote-request.php
```diff
@@ -0,0 +1,235 @@
...
+
+		/**
+		 * Fires after an ActivityPub QuoteRequest activity has been handled.
+		 *
+		 * @param array $activity The ActivityPub activity data.
+		 * @param int   $user_id  The local user ID.
+		 * @param bool  $success  True on success, false otherwise.
+		 * @param null  $result   Reserved for future use. Currently always null for QuoteRequest activities.
+		 */
+		\do_action( 'activitypub_handled_quoterequest', $activity, $user_id, $state, null );
```
obenland:
```suggestion
		\do_action( 'activitypub_handled_quote_request', $activity, $user_id, $state, null );
```
I assume this was meant to carry over the camelCase version of the activity type? I'd suggest sticking with the snake case format of WordPress hooks for consistency. I guess there's a _possibility_ someone _could_ try to set up callbacks dynamically with activity types (not sure how), but I think it's fine for that edge case to convert the types to snake case.

> Reserved for future use.

There shouldn't be a need to do that. We can always add more arguments without breaking downstream callbacks.

includes/handler/class-quote-request.php
pfefferle: whoops

includes/handler/class-quote-request.php
```diff
@@ -0,0 +1,235 @@
...
+				'instrument' => 1,
+			)
+		);
+
+		$url = \add_query_arg(
+			array(
+				'p'     => $post_id,
+				'stamp' => $meta_id,
+			),
+			\trailingslashit( \home_url() )
```
obenland: I'm pretty sure `add_query_arg()` handles the correct slashing. Otherwise, it could also be `\home_url( '/' )`

includes/class-query.php
```diff
@@ -368,4 +376,46 @@ public function is_old_host_request() {
...
+		if ( ! $user_uri ) {
+			return false;
+		}
+
+		$stamp_uri = \add_query_arg(
+			array(
+				'p'     => $post->ID,
+				'stamp' => $meta->meta_id,
+			),
+			\trailingslashit( \home_url() )
```
obenland: Same as above

includes/handler/class-quote-request.php
```diff
@@ -0,0 +1,235 @@
...
+		 */
+		\do_action( 'activitypub_handled_quoterequest', $activity, $user_id, $state, null );
+	}
+
+	/**
+	 * ActivityPub inbox disallowed activity.
+	 *
+	 * @param array  $activity The activity array.
+	 * @param null   $user_id  The user ID.
+	 * @param string $type     The type of the activity.
```
obenland:
```suggestion
	 * @param array    $activity The activity array.
	 * @param int|null $user_id  The user ID.
	 * @param string   $type     The type of the activity.
```

includes/handler/class-quote-request.php
obenland:
```suggestion
				self::queue_reject( $activity, $user_id );
```
`queue_reject()` only accepts two parameters.

includes/handler/class-quote-request.php
```diff
@@ -0,0 +1,235 @@
...
+				break;
+		}
+
+		/**
+		 * Fires after an ActivityPub QuoteRequest activity has been handled.
+		 *
+		 * @param array $activity The ActivityPub activity data.
+		 * @param int   $user_id  The local user ID.
+		 * @param bool  $success  True on success, false otherwise.
+		 * @param null  $result   Reserved for future use. Currently always null for QuoteRequest activities.
```
obenland:
```suggestion
		 * @param null  $result   Reserved for future use. Currently, always null for QuoteRequest activities.
```

includes/handler/class-quote-request.php
```diff
@@ -0,0 +1,235 @@
...
+				'stamp' => $meta_id,
+			),
+			\trailingslashit( \home_url() )
+		);
+
+		$activity = new Activity();
+		$activity->set_type( 'Accept' );
+		$activity->set_actor( $actor->get_id() );
+		$activity->set_object( $activity_object );
+		$activity->set_result( $url );
```
obenland, line 156: Let's add `@method` docs for `set_result()` to `Generic_Object`.

includes/class-query.php
```diff
@@ -7,10 +7,14 @@
...
 namespace Activitypub;
 
+use Activitypub\Activity\Extended_Object\Quote_Authorization;
 use Activitypub\Collection\Actors;
 use Activitypub\Collection\Outbox;
 use Activitypub\Transformer\Factory;
 
+use function Activitypub\get_post_id;
+use function Activitypub\get_user_id;
+
```
obenland:
```suggestion
```
These should not be necessary, they're using the same namespace.

includes/class-query.php
```diff
@@ -368,4 +376,46 @@ public function is_old_host_request() {
...
+				'stamp' => $meta->meta_id,
+			),
+			\trailingslashit( \home_url() )
+		);
+
+		$activitypub_object = new Quote_Authorization();
+		$activitypub_object->set_id( $stamp_uri );
+		$activitypub_object->set_attributed_to( $user_uri );
+		$activitypub_object->set_interacting_object( $meta->meta_value );
+		$activitypub_object->set_interaction_target( get_post_id( $post->ID ) );
```
obenland, line 411: Let's add `@method` docs for `set_interacting_object()` and `set_interaction_target()` to `Generic_Object`.

includes/class-query.php
pfefferle: I thought I be specific, because they very much look like WordPress core funcs

includes/class-query.php
pfefferle: Can I add it to the class itself? this is very specific to the `Quote_Authorization`!?

obenland: Oh, absolutely! Yeah that makes sense

includes/handler/class-quote-request.php
```diff
@@ -0,0 +1,236 @@
...
+ * Handler for QuoteRequest activities.
+ *
+ * @see https://codeberg.org/fediverse/fep/src/branch/main/fep/044f/fep-044f.md
+ */
+class Quote_Request {
+	/**
+	 * Initialize the class, registering WordPress hooks.
+	 */
+	public static function init() {
+		\add_action( 'activitypub_inbox_quoterequest', array( self::class, 'handle_quote_request' ), 10, 2 );
```
obenland: I just realized we're running into the same hook name issue in the inbox. I think this is the first time we're handling a camelcase type? Maybe we could update the inbox to run `$type` through a snake_case function to create the hook name?

obenland (approved):

PR #2241, opened 2025-09-29
Add row action filters and Jetpack Reader link integration

### LINKED ISSUE #2156: Add Reader reference to Following list table in wp-admin
We should add action links to the Reader for individual follows and maybe a Button in the UI that points people to the reader to make it easier to find.

pfefferle:
Introduces new filters for customizing row actions in the Blocked Actors, Followers, and Following admin tables. Enhances Jetpack integration by conditionally registering hooks, adding a 'Reader' link to the Following list actions, and ensuring the ActivityPub Following UI is enabled when Jetpack is active. Also updates the Jetpack initialization condition in the integration loader.

Fixes #2156

## Proposed changes:
- Added filter hooks to all ActivityPub admin table classes for customizable row actions
- Enhanced Jetpack integration with a Reader link feature for the Following table
- Updated Jetpack initialization logic to support WordPress.com environments

### Other information:

## Testing instructions:

* Configure Jetpack or run on WP.com
* Check if Following is enabled
* Follow an Actor
* Check that Reader link is not shown when state is pending
* Check if Reader link is visible when state is no longer pending
* Click on reader link and check redirect

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

You can now follow people and see their updates right in the WordPress.com Reader when using Jetpack or WordPress.com.

</details>

### DIFF
--- a/includes/wp-admin/table/class-blocked-actors.php
+++ b/includes/wp-admin/table/class-blocked-actors.php
@@ -402,7 +402,7 @@ class Blocked_Actors extends \WP_List_Table {
 		 *
 		 * This filter is evaluated for each blocked actor item in the list table.
 		 *
-		 * @since 2.8.0
+		 * @since unreleased
 		 *
 		 * @param string[] $actions An array of row action links. Defaults are
 		 *                          'Unblock'.
--- a/includes/wp-admin/table/class-blocked-actors.php
+++ b/includes/wp-admin/table/class-blocked-actors.php
@@ -400,13 +400,12 @@ class Blocked_Actors extends \WP_List_Table {
 		/**
 		 * Filters the array of row action links on the Blocked Actors list table.
 		 *
-		 * The filter is evaluated only for non-hierarchical post types.
+		 * This filter is evaluated for each blocked actor item in the list table.
 		 *
 		 * @since 2.8.0
 		 *
 		 * @param string[] $actions An array of row action links. Defaults are
-		 *                          'Edit', 'Quick Edit', 'Restore', 'Trash',
-		 *                          'Delete Permanently', 'Preview', and 'View'.
+		 *                          'Unblock'.
 		 * @param array    $item    The current blocked actor item.
 		 */
 		$actions = apply_filters( 'activitypub_blocked_actors_row_actions', $actions, $item );
--- a/includes/wp-admin/table/class-blocked-actors.php
+++ b/includes/wp-admin/table/class-blocked-actors.php
@@ -397,6 +397,20 @@ class Blocked_Actors extends \WP_List_Table {
 			),
 		);
 
+		/**
+		 * Filters the array of row action links on the Blocked Actors list table.
+		 *
+		 * The filter is evaluated only for non-hierarchical post types.
+		 *
+		 * @since 2.8.0
+		 *
+		 * @param string[] $actions An array of row action links. Defaults are
+		 *                          'Edit', 'Quick Edit', 'Restore', 'Trash',
+		 *                          'Delete Permanently', 'Preview', and 'View'.
+		 * @param array    $item    The current blocked actor item.
+		 */
+		$actions = apply_filters( 'activitypub_blocked_actors_row_actions', $actions, $item );
+
 		return $this->row_actions( $actions );
 	}
 }
--- a/integration/class-jetpack.php
+++ b/integration/class-jetpack.php
@@ -99,7 +99,7 @@ class Jetpack {
 		$actions['reader'] = sprintf(
 			'<a href="%s" target="_blank">%s</a>',
 			esc_url( $url ),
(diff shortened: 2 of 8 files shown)


includes/wp-admin/table/class-blocked-actors.php
```diff
@@ -397,6 +397,19 @@ protected function handle_row_actions( $item, $column_name, $primary ) {
...
+		 *
+		 * This filter is evaluated for each blocked actor item in the list table.
+		 *
+		 * @since unreleased
+		 *
+		 * @param string[] $actions An array of row action links. Defaults are
+		 *                          'Unblock'.
+		 * @param array    $item    The current blocked actor item.
+		 */
+		$actions = apply_filters( 'activitypub_blocked_actors_row_actions', $actions, $item );
```
pfefferle, line 411: We could also call `parent::handle_row_actions` but then we would have to unhook all of the default `row_action` links.

integration/class-jetpack.php
```diff
@@ -68,4 +73,44 @@ public static function add_comment_types( $comment_types ) {
...
+	public static function add_reader_link( $actions, $item ) {
+		// Do not show the link for pending follow requests.
+		if ( 'pending' === $item['status'] ) {
+			return $actions;
+		}
+
+		$reader_id = \get_post_meta( $item['id'], '_activitypub_actor_feed', true );
+
+		if ( $reader_id ) {
+			$url = sprintf( 'https://wordpress.com/reader/feed/%d', (int) $reader_id );
```
obenland:
```suggestion
		$feed = \get_post_meta( $item['id'], '_activitypub_actor_feed', true );

		if ( isset( $feed['feed_id'] ) ) {
			$url = sprintf( 'https://wordpress.com/reader/feed/%d', (int) $feed['feed_id'] );
```

integration/class-jetpack.php
```diff
@@ -68,4 +73,44 @@ public static function add_comment_types( $comment_types ) {
...
+		} else {
+			$url = sprintf( 'https://wordpress.com/reader/feeds/lookup/%s', $item['identifier'] );
```
obenland:
```suggestion
			$url = sprintf( 'https://wordpress.com/reader/feeds/lookup/%s', rawurlencode( $item['identifier'] ) );
```

integration/class-jetpack.php
```diff
@@ -68,4 +73,44 @@ public static function add_comment_types( $comment_types ) {
...
+		}
+
+		$actions['reader'] = sprintf(
+			'<a href="%s" target="_blank">%s</a>',
+			esc_url( $url ),
+			esc_html__( 'View Feed', 'activitypub' )
+		);
+
+		return $actions;
```
obenland: `View Feed` should probably the first action. Maybe `View` could be enough? Essentially it's their "profile" in the Reader.

Image

integration/class-jetpack.php
```diff
@@ -20,11 +20,16 @@ class Jetpack {
...
-		\add_filter( 'jetpack_api_include_comment_types_count', array( self::class, 'add_comment_types' ) );
+		if ( ! \defined( 'IS_WPCOM' ) ) {
+			\add_filter( 'jetpack_sync_post_meta_whitelist', array( self::class, 'add_sync_meta' ) );
+			\add_filter( 'jetpack_sync_comment_meta_whitelist', array( self::class, 'add_sync_comment_meta' ) );
+			\add_filter( 'jetpack_sync_whitelisted_comment_types', array( self::class, 'add_comment_types' ) );
+			\add_filter( 'jetpack_json_api_comment_types', array( self::class, 'add_comment_types' ) );
+			\add_filter( 'jetpack_api_include_comment_types_count', array( self::class, 'add_comment_types' ) );
+		}
+
+		\add_filter( 'activitypub_following_row_actions', array( self::class, 'add_reader_link' ), 10, 2 );
```
obenland: Should we put these behind an `Jetpack::is_connection_ready()` check? They won't work without a connection.

integration/class-jetpack.php
pfefferle: works without, but happy to add encoding

integration/class-jetpack.php
pfefferle: Is this something that is relevant for all hooks?

/cc @jeherve

pfefferle: Is `Jetpack::is_connection_ready()` true for DotCom or do I have to use something like `IS_WPCOM || \Jetpack::is_connection_ready()`???

obenland: Oh, good thought. It looks like `is_connection_ready()` is `true` by default on Dotcom.

obenland (approved):

PR #2246, opened 2025-09-30
Improve user ID fallback and transformer logic

pfefferle:
Updated `get_user_id` to fall back to the blog user in blog mode or when the user is not found. Adjusted transformer factory to only create Post transformers if a valid user ID exists. Added and updated tests to cover new fallback and mode logic.

Atm. the post author always falls back to the Blog-User, if enabled or not. With this change, the transformer also checks if there is an Actor enabled before starting content-negotiation.

## Proposed changes:

* Implement proper fallback for `get_user_id`
* Check if there is a valid user before transforming object

### Other information:

## Testing instructions:

Before PR:

* Remove `activitypub` cap from all Blog-Users
* Set Actor mode to "Author Profiles Only"
* Load post JSON (`https://example.org/?p=123&activitypub`)
* The JSON loads with the Blog-User as `attributedTo`

After PR:

* HTML loads instead of JSON

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Posts now only fall back to the blog user when blog mode is enabled and no valid author exists, ensuring content negotiation only runs if an Actor is available.

</details>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1313,10 +1313,10 @@ function get_user_id( $id ) {
 		$user = Actors::get_by_id( Actors::BLOG_USER_ID );
 	} else {
 		$user = Actors::get_by_id( $id );
-	}
 
-	if ( \is_wp_error( $user ) ) {
-		$user = Actors::get_by_id( Actors::BLOG_USER_ID );
+		if ( \is_wp_error( $user ) ) {
+			$user = Actors::get_by_id( Actors::BLOG_USER_ID );
+		}
 	}
 
 	if ( \is_wp_error( $user ) ) {
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1300,12 +1300,24 @@ function get_content_warning( $post_id ) {
 /**
  * Get the ActivityPub ID of a User by the WordPress User ID.
  *
+ * Fall back to blog user if in blog mode or if user is not found.
+ *
  * @param int $id The WordPress User ID.
  *
- * @return string The ActivityPub ID (a URL) of the User.
+ * @return string|false The ActivityPub ID (a URL) of the User or false if not found.
  */
 function get_user_id( $id ) {
-	$user = Actors::get_by_id( $id );
+	$mode = \get_option( 'activitypub_actor_mode', 'default' );
+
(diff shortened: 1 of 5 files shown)


includes/functions.php
```diff
@@ -1300,12 +1300,24 @@ function get_content_warning( $post_id ) {
...
  *
- * @return string The ActivityPub ID (a URL) of the User.
+ * @return string|false The ActivityPub ID (a URL) of the User or false if not found.
  */
 function get_user_id( $id ) {
-	$user = Actors::get_by_id( $id );
+	$mode = \get_option( 'activitypub_actor_mode', 'default' );
+
+	if ( ACTIVITYPUB_BLOG_MODE === $mode ) {
+		$user = Actors::get_by_id( 0 );
```
```suggestion
		$user = Actors::get_by_id( Actors::BLOG_USER_ID );
```

includes/functions.php
```diff
@@ -1300,12 +1300,24 @@ function get_content_warning( $post_id ) {
...
+	} else {
+		$user = Actors::get_by_id( $id );
+	}
+
+	if ( \is_wp_error( $user ) ) {
+		$user = Actors::get_by_id( 0 );
```
```suggestion
		$user = Actors::get_by_id( Actors::BLOG_USER_ID );
```
This wp_error check only seems relevant to the `else` conditional of `Actors::get_by_id( $id );`, let's move it in there? If the result of `Actors::get_by_id( Actors::BLOG_USER_ID );` is a wp_error, there's no reason to try again.

includes/functions.php
```diff
@@ -1300,12 +1300,24 @@ function get_content_warning( $post_id ) {
...
+
+	if ( ACTIVITYPUB_BLOG_MODE === $mode ) {
+		$user = Actors::get_by_id( Actors::BLOG_USER_ID );
+	} else {
+		$user = Actors::get_by_id( $id );
+	}
+
+	if ( \is_wp_error( $user ) ) {
+		$user = Actors::get_by_id( Actors::BLOG_USER_ID );
+	}
```
obenland, line 1320: You don't want to move this check into the `else` case? It feels redundant for the `ACTIVITYPUB_BLOG_MODE === $mode` case.

obenland (approved):

PR #2247, opened 2025-09-30
Add external link icon to Jetpack reader feed link

pfefferle:
Updated the Jetpack integration to add the 'activitypub-external' class to the reader feed link, and added CSS to display a dashicon after external links. This visually indicates external links in the admin interface.

Screenshot 2025-09-30 at 13 30 44

## Proposed changes:

* 

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/integration/class-jetpack.php
+++ b/integration/class-jetpack.php
@@ -10,6 +10,7 @@ namespace Activitypub\Integration;
 use Activitypub\Collection\Followers;
 use Activitypub\Collection\Following;
 use Activitypub\Comment;
+use Automattic\Jetpack\Connection\Manager;
 
 /**
  * Jetpack integration class.
@@ -30,7 +31,7 @@ class Jetpack {
 
 		if (
 			( \defined( 'IS_WPCOM' ) && IS_WPCOM ) ||
-			( \class_exists( '\Jetpack' ) && \Jetpack::is_connection_ready() )
+			( \class_exists( 'Automattic\Jetpack\Connection\Manager' ) && ( new Manager() )->is_user_connected() )
 		) {
 			\add_filter( 'activitypub_following_row_actions', array( self::class, 'add_reader_link' ), 10, 2 );
 			\add_filter( 'pre_option_activitypub_following_ui', array( self::class, 'pre_option_activitypub_following_ui' ) );
@@ -104,7 +105,7 @@ class Jetpack {
 		return array_merge(
 			array(
 				'reader' => sprintf(
-					'<a href="%s" target="_blank">%s</a>',
+					'<a href="%s" target="_blank" class="activitypub-external">%s</a>',
 					esc_url( $url ),
 					esc_html__( 'View Feed', 'activitypub' )
 				),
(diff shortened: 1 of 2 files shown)


integration/class-jetpack.php
```diff
@@ -104,7 +104,7 @@ public static function add_reader_link( $actions, $item ) {
 		return array_merge(
 			array(
 				'reader' => sprintf(
-					'<a href="%s" target="_blank">%s</a>',
+					'<a href="%s" target="_blank" class="activitypub-external">%s</a>',
```
jeherve: You may be able to address the a11y issues by following the markup used in core:
https://github.com/WordPress/wordpress-develop/blob/6b6d2d1975cf72ace0f37ba440d380e16bdbb0d2/src/wp-admin/includes/dashboard.php#L1313

As an added bonus, if you use the `dashicons dashicons-external` classes you may not need to add any custom CSS, and you could leave Core handle everything.

pfefferle: It looks awful with the span :(

pfefferle: But if it is the WordPress way ;)

It is at least now consistent with the Site-Health.

Screenshot 2025-09-30 at 13 30 44

integration/class-jetpack.php
```diff
@@ -104,9 +104,11 @@ public static function add_reader_link( $actions, $item ) {
 		return array_merge(
 			array(
 				'reader' => sprintf(
-					'<a href="%s" target="_blank">%s</a>',
+					'<a href="%1$s" target="_blank">%2$s<span class="screen-reader-text"> %3$s</span><span aria-hidden="true" class="dashicons dashicons-external"></span></a>',
```
obenland, line 108:
```suggestion
					'<a href="%1$s" target="_blank">%2$s<span class="screen-reader-text"> %3$s</span><span aria-hidden="true" class="dashicons dashicons-external" style="font-size:1em;line-height:1.7;"></span></a>',
```
Let's make it a bit easier on the eyes:

Image

obenland (changes requested):

integration/class-jetpack.php
pfefferle: https://github.com/Automattic/wordpress-activitypub/pull/2247#discussion_r2391007255

obenland: Not sure what that's trying to say? My suggestion doesn't change the markup or affect a11y, as far as I'm aware.

If it's about not using custom CSS, in this case it serves a purpose.

integration/class-jetpack.php
jeherve: > It looks awful with the span :(

The icon alone looks terrible imo. :) I can't wait until it gets replaced with the "North East Arrow" character like in Gutenberg screens.

integration/class-jetpack.php
pfefferle: Consistency!?

Screenshot 2025-09-30 at 15 59 55

I added some CSS. I prefer that over inline CSS.

obenland (approved):

PR #2253, opened 2025-10-01
Release 7.5.0

pfefferle:
Release version 7.5.0


obenland (approved):
I added two more version replacements that the release script missed and created a fix for it in #2258. This is good to go.

PR #2278, opened 2025-10-02
Optimize and align validate_object methods across handlers

obenland:
Also fixes a potential fatal error for `Undo` activities from brid.gy. 

## Proposed changes:
* Optimize `validate_object` methods in Accept, Create, Reject, Undo, and Quote_Request handlers for better performance
* Replace expensive `array_diff()` + `array_keys()` operations with direct `isset()` checks
* Simplify object validation logic from `empty() || !is_array()` to `!isset() || !is_array()`
* Remove unnecessary `is_wp_error()` checks (request parameter is always `WP_REST_Request` in actual usage)
* Rename `$json_params` to `$activity` for improved code readability
* Align validation patterns across all handlers for consistency

### Other information:

All existing tests pass. Removed one obsolete test that was checking for an impossible scenario (WP_Error as request object).

## Testing instructions:

* Run `npm run env-test` to verify all PHP tests pass
* Verify that inbox validation still works correctly by sending various ActivityPub activities to the inbox


includes/handler/class-undo.php
```diff
@@ -85,36 +85,25 @@ public static function handle_undo( $activity, $user_id ) {
...
 			return false;
 		}
 
-		$required_object_attributes = array(
-			'id',
-			'type',
-			'actor',
-			'object',
-		);
+		if ( ! \is_array( $activity['object'] ) ) {
```
pfefferle, line 102: Can we first merge https://github.com/Automattic/wordpress-activitypub/pull/2284 and then have your changes afterwards? My PR adds support for URI-Objects at least for `Like`s and `Announce`s.

includes/handler/class-accept.php
```diff
@@ -72,36 +72,25 @@ public static function handle_accept( $accept, $user_id ) {
...
-			'type',
-			'actor',
-			'object',
-		);
+		if ( ! \is_array( $activity['object'] ) ) {
+			return false;
+		}
 
-		if ( ! empty( \array_diff( $required_object_attributes, \array_keys( $json_params['object'] ) ) ) ) {
+		if ( ! isset( $activity['object']['id'], $activity['object']['type'], $activity['object']['actor'], $activity['object']['object'] ) ) {
```
pfefferle, line 93: Have you changed it to `isset` because of any performance improvements?

I think the current version is way more readable and extensible then a huge list of issets on different levels on array keys!?

pfefferle:
> Remove unnecessary is_wp_error() checks (request parameter is always WP_REST_Request in actual usage)

The signature verification returns a `WP_Error` https://github.com/Automattic/wordpress-activitypub/blob/41efe963bbc658f18eae718a46a3eafb5069fd80/includes/rest/class-server.php#L74 does that have no impact on the `validate_object` verification?

pfefferle:
Ok, `$request` is always a `WP_REST_Request`. Is the signature verification called afterwards? Is there a way to run the signature verification earlier? We would save some time/resources then, because we wouldn't have to trigger the `validate_object` at all.

includes/handler/class-accept.php
pfefferle: ok, seems to be a performance thing!

obenland: Yup!

obenland:
We could move it to `rest_pre_dispatch` but at that point the request has not been matched with a handler yet, so would run for every REST API request. Currently every endpoint that needs signature verification specifies it for itself, which we would lose, too.

FWIW, I don't think it's a big deal, eventually a request has to pass all of these tests anyway.

pfefferle:
Then let's keep it as is!

pfefferle:
I think it is fine to merge (when the conflicts are resolved), I will update the validation callback through #2295 then.

pfefferle (dismissed):

pfefferle (approved):

PR #2290, opened 2025-10-03
Fix: issue #2286 – typo, improved quoting explanation, add missing i18n in advanced settings

### LINKED ISSUE #2286: I propose improving the string description for the ActivityPub plugin.
I propose improving the string description for the ActivityPub plugin.

Current:

> *"Quoting allows others to reshare your post while adding their own commentary."*

Problem:

* The term *"reshare"* is misleading, since reshare/boost corresponds to the **Announce** activity.
* A **Quote** is not a reshare, but an independent post that references the original one.

Improved version:

> *"Quoting allows others to create a new post that references your post, ~while adding their own commentary~."*

This makes the distinction between **Announce** (reshare/boost) and **Quote** (independent post with commentary) clearer and avoids confusion.

Jiwoon-Kim:
Fixes #2286

## Proposed changes:

* Fixed typo in example federated profile: `@bobz@example.com` → `@bob@example.com`
* Improved quoting description:
  **File**: src/editor-plugin/plugin.js:1 
  From: "Quoting allows others to reshare your post while adding their own commentary."
  To: "Quoting allows others to cite your post while adding their own commentary."
* ~Added missing i18n for advanced settings descriptions (Following User Interface, Inbox) in `includes/wp-admin/class-advanced-settings-fields.php: 244, 248, 251, 283, 287`~

### Other information:

## Testing instructions:

* ~Go to WordPress admin → Settings → ActivityPub → Advanced~
* ~Check that the "Following User Interface" and "Inbox" descriptions now have proper i18n applied~
* Verify the corrected strings in the editor plugin and readme
* https://wordpress.org/plugins/activitypub/
* https://translate.wordpress.org/projects/wp-plugins/activitypub/dev/ko/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=22278048&filters%5Btranslation_id%5D=135834849

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fixed typo in example, improve quoting description.

</details>


includes/wp-admin/class-advanced-settings-fields.php
```diff
@@ -241,14 +241,14 @@ public static function render_following_ui_field() {
 		<p>
 			<label>
 				<input type="checkbox" id="activitypub_following_ui" name="activitypub_following_ui" value="1" <?php checked( '1', $value ); ?> />
-				Display the "Following" interface in the admin menus and settings.
+				<?php \esc_html_e( 'Display the "Following" interface in the admin menus and settings.', 'activitypub' ); ?>
```
obenland: The strings in advanced settings are purposefully not translated, as this is more of a power-user tool and has a lot of long and complicated strings that we don't want to burden translators with. Please revert the changes in this file.

obenland (changes requested):
Thanks for the PR @Jiwoon-Kim.

Changes to editor plugin files need to be made in `/src/editor-plugin` and then build through `npm run build`. It looks like so far there's only a change in the build file.

includes/wp-admin/class-advanced-settings-fields.php
Jiwoon-Kim: Thanks for the feedback!
I was handling this as the Korean translation editor, so I thought it would be fine. These two strings were the only ones untranslated, and the rest were already translated, so I felt the translation was incomplete.
However, if you prefer consistency with the original approach, I can revert the changes in this file.

src/editor-plugin/plugin.js
```diff
@@ -160,10 +160,7 @@ const EditorPlugin = () => {
 
 			<SelectControl
 				label={ __( 'Who can quote this post?', 'activitypub' ) }
-				help={ __(
-					'Quoting allows others to reshare your post while adding their own commentary.',
-					'activitypub'
-				) }
+				help={ __( 'Quoting allows others to create a new post that references your post.', 'activitypub' ) }
```
obenland: The "allows others to create a new post" could be confusing—the setting doesn't influence whether others can create posts. Maybe something like `Quoting allows others to cite your post while adding their own commentary.`?

Jiwoon-Kim: It’s a bit ambiguous, isn’t it? This feature doesn’t actually create a comment or reply—it’s more like starting a new conversation in a separate post. Would this wording work?

“Quoting allows others to start a new post that references your post while adding their own commentary.”
Quoting allows others to create a new item that references your post or comment while adding their own commentary

obenland: What's ambiguous about `Quoting allows others to cite your post while adding their own commentary.`?

Jiwoon-Kim: Ah, I see. When translating into Korean, the word commentary overlaps with comment, and quote doesn’t feel very different from cite, so I suggested an alternative for clarity. But if the nuance is clearer in English, I’ll go with your suggestion. It’s 3 AM here, so I’ll make the change tomorrow.

Jiwoon-Kim: > Quoting lets others create a new post that references your post, while adding their own commentary.

> Quoting lets others cite your post in a new one, while adding their own commentary.

> Quoting lets others cite your post — effectively starting a new post that references it — while adding their own commentary.

obenland (approved):

PR #2294, opened 2025-10-06
Add option to hide followers and following lists

### LINKED ISSUE #45: Feature Request: hide followers
In Mastodon one can set "hide followers" which shows others only "myself" as follower, but hides all others. this would be greatly appreciated for WP activitypub too. otherwise people will receive a lot of spam because bots are able to scan all followers.

pfefferle:
This PR introduces a privacy feature that allows users to hide their followers and following lists from their ActivityPub profiles while maintaining the underlying follow relationships.

Fixes #45 

## Proposed changes:
- Adds `show_followers()` method to check if followers/following lists should be displayed
- Updates REST API controllers to conditionally include follower/following data based on privacy settings
- Implements UI settings for both blog-wide and individual user preferences

### Other information:

## Testing instructions:

* Go to settings and check/uncheck the Follower/Following settings
* See if the Follower/Following API endpoints show/hide the Followers/Followings accordingly

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added a privacy option to hide followers and following lists from profiles while keeping follow relationships intact.

</details>


includes/wp-admin/class-user-settings-fields.php
```diff
@@ -380,4 +388,20 @@ public static function blocked_keywords_callback() {
...
+
+	/**
+	 * Hide Social Graph field callback.
+	 */
+	public static function hide_followers_callback() {
+		$hide_followers = \get_user_option( 'activitypub_hide_social_graph', \get_current_user_id() );
+		?>
+		<label>
+			<input type="checkbox" name="activitypub_hide_social_graph" id="activitypub_hide_social_graph" value="1" <?php \checked( '1', $hide_followers ); ?> />
+			<?php \esc_html_e( 'Hide Followers and Following on Profile', 'activitypub' ); ?>
```
obenland, line 400: I don't know how important this is, but we don't really have a profile.

includes/wp-admin/class-user-settings-fields.php
```diff
@@ -380,4 +388,20 @@ public static function blocked_keywords_callback() {
...
+		</label>
+		<p class="description">
+			<?php esc_html_e( 'By default, these lists are public. Hiding them removes them from your profile, but people you follow will still see that you follow them.', 'activitypub' ); ?>
```
obenland:
```suggestion
			<?php esc_html_e( 'People you follow will still see that you follow them.', 'activitypub' ); ?>
```
I worry that the current phrasing could make folks feel like it's something that _should_ be hidden, when that's not the case. So rather than "warning" it's currently public, maybe we could just limit it to the expectation setting around followers still able to see the relationship?

obenland (dismissed):

includes/wp-admin/class-user-settings-fields.php
pfefferle: But you see the list when someone loads your "Profile" on Mastodon.

includes/wp-admin/class-blog-settings-fields.php
```diff
@@ -295,11 +303,26 @@ class="large-text"
...
+	 * Hide Social Graph field callback.
+	 */
+	public static function hide_followers_callback() {
+		?>
+		<label>
+			<input type="checkbox" name="activitypub_hide_social_graph" id="activitypub_hide_social_graph" value="1" <?php \checked( '1', \get_option( 'activitypub_hide_social_graph', '0' ) ); ?> />
+			<?php \esc_html_e( 'Hide Followers and Following on Profile', 'activitypub' ); ?>
+		</label>
+		<p class="description">
+			<?php esc_html_e( 'By default, these lists are public. Hiding them removes them from your profile, but people you follow will still see that you follow them.', 'activitypub' ); ?>
```
obenland:
```suggestion
			<?php esc_html_e( 'People you follow will still see that you follow them.', 'activitypub' ); ?>
```
Sorry, missed that one.

obenland (approved):

ghost:
thank you so much for your work on this issue, we are very glad that you have been able to implement this change!

includes/wp-admin/class-user-settings-fields.php
ghost: maybe you don't see a list on mastodon apps, but i noticed this because on https://browser.pub/ this info shows up if you load the followers page

PR #2297, opened 2025-10-06
Implement FEP-8fcf followers collection synchronization

pfefferle:
Adds support for FEP-8fcf to synchronize followers collections across servers. This includes generating and processing Collection-Synchronization headers, computing partial follower digests, exposing a new /followers-sync REST endpoint, and scheduling async reconciliation jobs when mismatches are detected. New methods and traits are introduced for digest computation, header parsing, and reconciliation, with hooks for monitoring sync events. This improves data consistency and interoperability with other ActivityPub servers supporting FEP-8fcf.

## Proposed changes:
- Adds Collection-Synchronization header generation and processing for FEP-8fcf protocol
- Implements partial follower digest computation using XOR'd SHA256 hashes
- Provides new `/followers/sync` REST endpoint for synchronization
- Includes async reconciliation scheduler for handling digest mismatches

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added support for FEP-8fcf follower synchronization, improving data consistency across servers with new sync headers, digest checks, and reconciliation tasks.

</details>


includes/collection/class-followers.php
```diff
@@ -481,4 +483,117 @@ public static function remove_blocked_actors( $value, $type, $user_id ) {
...
+	 * @return string|false The header value, or false if cannot generate.
+	 */
+	public static function generate_sync_header( $user_id, $authority ) {
+		$followers = self::get_by_authority( $user_id, $authority );
+		$followers = array_map(
+			function ( $post ) {
+				return $post->guid;
+			},
+			$followers
+		);
```
```suggestion
		$followers = \wp_list_pluck( $followers, 'guid' );
```

includes/handler/class-collection-sync.php
```diff
@@ -0,0 +1,182 @@
...
+ * @see https://codeberg.org/fediverse/fep/src/branch/main/fep/8fcf/fep-8fcf.md
+ */
+class Collection_Sync {
+	/**
+	 * Initialize the class, registering WordPress hooks.
+	 */
+	public static function init() {
+		\add_action( 'activitypub_inbox_create', array( self::class, 'handle_collection_synchronization' ), 10, 2 );
+
+		\add_filter( 'http_request_args', array( self::class, 'maybe_add_headers' ), -1, 2 );
```
obenland, line 31: Let's document the `-1` priority

includes/handler/class-collection-sync.php
```diff
@@ -0,0 +1,182 @@
...
+			$collection_type = 'followers';
+		}
+
+		if ( ! $collection_type ) {
+			// Unknown or unsupported collection type.
+			return;
+		}
+
+		// Get the actor URL for validation.
+		$actor_url = isset( $data['actor'] ) ? $data['actor'] : null;
```
obenland:
```suggestion
		$actor_url = $data['actor'] ?? false;
```

Not sure if that's much better. Could also have the next conditional check `( ! isset( $data['actor'] ) )` and then use `$data['actor']` in subsequent calls.

includes/rest/class-followers-controller.php
```diff
@@ -75,6 +75,52 @@ public function register_routes() {
...
+					),
+				),
+				array(
+					'methods'             => \WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_partial_followers' ),
+					'permission_callback' => array( 'Activitypub\Rest\Server', 'verify_signature' ),
+					'args'                => array(
+						'authority' => array(
+							'description' => 'The URI authority to filter followers by.',
+							'type'        => 'string',
```
obenland, line 99: Is this a URI? Or could it be just a domain?

includes/rest/class-followers-controller.php
```diff
@@ -139,6 +185,61 @@ function ( $item ) use ( $context ) {
...
+		}
+
+		// Get partial followers filtered by authority.
+		$followers = Followers::get_by_authority( $user_id, $authority );
+		$followers = array_map(
+			function ( $post ) {
+				return $post->guid;
+			},
+			$followers
+		);
```
obenland:
```suggestion
		$followers = \wp_list_pluck( $followers, 'guid' );
```

includes/scheduler/class-collection-sync.php
```diff
@@ -0,0 +1,102 @@
...
+	 *
+	 * @param string $type      The collection type (e.g., 'followers').
+	 * @param int    $user_id   The local user ID.
+	 * @param string $actor_url The remote actor URL.
+	 * @param array  $params    The Collection-Synchronization header parameters.
+	 */
+	public static function schedule_reconciliation( $type, $user_id, $actor_url, $params ) {
+		// Schedule async processing to avoid blocking the inbox.
+		\wp_schedule_single_event(
+			time() + 60, // Process in 1 minute.
```
obenland:
```suggestion
			time() + MINUTE_IN_SECONDS,
```

includes/scheduler/class-collection-sync.php
```diff
@@ -0,0 +1,102 @@
...
+	 * @param string $actor_url The remote actor URL.
+	 * @param array  $params    The Collection-Synchronization header parameters.
+	 */
+	public static function reconcile_followers( $user_id, $actor_url, $params ) {
+		if ( empty( $params['url'] ) ) {
+			return;
+		}
+
+		// Fetch the authoritative partial followers collection.
+		$data = Http::get_remote_object( $params['url'], 300 ); // Cache for 5 minutes.
```
obenland:
```suggestion
		$data = Http::get_remote_object( $params['url'], 5 * MINUTE_IN_SECONDS );
```

includes/scheduler/class-collection-sync.php
```diff
@@ -0,0 +1,102 @@
...
+		if ( \is_wp_error( $data ) || ! isset( $data['orderedItems'] ) || ! \is_array( $data['orderedItems'] ) ) {
+			return;
+		}
+
+		$remote_followers = $data['orderedItems'];
+
+		// Get our authority.
+		$home_authority = get_url_authority( \home_url() );
+
+		$accepted = Following::get_by_authority( $user_id, $home_authority );
```
obenland, line 71: For my understanding, these are accounts from the local site that this local actor follows?

obenland (commented):
Fedify turned this off by default recently. Do you happen know what led to this decision and should it have an effect on us adopting this?

obenland:
Is there a good way to test this?

pfefferle:
> [Followers collection synchronization](https://fedify.dev/manual/send#followers-collection-synchronization) ([FEP-8fcf](https://w3id.org/fep/8fcf)) is now opt-in rather than automatic. This feature must now be explicitly enabled through the [syncCollection](https://jsr.io/@fedify/fedify@1.5.0/doc/~/SendActivityOptionsForCollection.syncCollection) option, giving developers more control over when to include followers collection digests. This change improves network efficiency by reducing unnecessary synchronization traffic.

https://github.com/fedify-dev/fedify/discussions/224

I am fine adding (yet another) setting to opt in ☺️ 

The final state already works with filters, so I might spend some time thinking about an opt-in/opt-out mechanism. It’s probably fine if the sync doesn’t run on every `Create`, we could even “store” the last sync instead.

Overall, I think this is a good feature, especially for self-hosted sites. It might generate some traffic, but since it runs asynchronously, that’s acceptable. WordPress blogs often miss pings when they’re under heavy load, so in the long run, this could actually reduce traffic by preventing the plugin from sending Activities to too many false followers.

pfefferle:
> Is there a good way to test this?

This is as tricky as the signature PR.

I think the easiest way is to somehow produce a broken state and trigger the sync.

I wanted to run this on my own Blog next week, but wanted to wait for a final state of the PR.

includes/rest/class-followers-controller.php
pfefferle: It is the domain part but with the protocol.

The authority of `https://example.org/test?abc` is `https://example.org` so maybe we could call it `host` instead!?

includes/scheduler/class-collection-sync.php
pfefferle: Get all `Remote_Actors` a local User follows that match the authority from the incoming `Create` request.

includes/handler/class-collection-sync.php
pfefferle: I am not sure it has to be `-1`.

I will run some more tests next week. I debugged so much to get this working with Unittests and forgot to re-review this one.

pfefferle: It has to be! I added a documentation.

pfefferle:
@obenland now it runs only once a week (outgoing and incoming).

includes/rest/class-followers-controller.php
obenland:
```suggestion
							'type'        => 'string',
							'format'      => 'uri',
							'pattern'     => '^https?://[^/]+$',
```
With this we should be able to skip the `// Validate authority format.` part below. Just the pattern would be enough for that actually. The uri format just sends it through `sanitize_url()`.

includes/handler/class-collection-sync.php
```diff
@@ -0,0 +1,223 @@
...
+
+		if ( ! isset( $body['type'] ) || 'Create' !== $body['type'] ) {
+			return $args;
+		}
+
+		// Only send header if we haven't sent one to this authority in the last day.
+		$inbox_authority = get_url_authority( $url );
+		$user_id         = $args['user_id'] ?? 0;
+
+		if ( ! $user_id || ! $inbox_authority ) {
```
obenland: Would this bail on headers for the blog user?

includes/functions.php
```diff
@@ -1790,3 +1790,33 @@ function extract_name_from_uri( $uri ) {
...
+	$authority = $parsed['scheme'] . '://' . $parsed['host'];
+
+	if ( ! empty( $parsed['port'] ) ) {
+		$default_ports = array(
+			'http'  => 80,
+			'https' => 443,
+		);
+
+		if ( ! isset( $default_ports[ $parsed['scheme'] ] ) || $default_ports[ $parsed['scheme'] ] !== $parsed['port'] ) {
+			$authority .= ':' . $parsed['port'];
```
obenland: It doesn't look like the `get_partial_followers` endpoint accounts for ports, in case the intention is for them to match.

includes/class-signature.php
```diff
@@ -465,4 +465,97 @@ public static function generate_digest( $body ) {
...
+	 * The digest is created by XORing together the individual SHA256 digests
+	 * of each follower's ID.
+	 *
+	 * @see https://codeberg.org/fediverse/fep/src/branch/main/fep/8fcf/fep-8fcf.md
+	 *
+	 * @param array $collection The user ID whose followers to compute.
+	 *
+	 * @return string|false The hex-encoded digest, or false if no followers.
+	 */
+	public static function compute_collection_digest( $collection ) {
```
obenland: Optional: Maybe `get_collection_digest()`?

docs/collection-synchronization.md
```diff
@@ -0,0 +1,223 @@
...
+   - If the remote actor is NOT in the remote followers list, reject the local follow (remote server no longer recognizes it)
+4. For **pending** following relationships:
+   - If the remote actor IS in the remote followers list, accept the pending follow (remote server already accepted it)
+   - If the remote actor is NOT in the remote followers list, reject the pending follow (remote server doesn't recognize it)
+
+The reconciliation is handled asynchronously via WordPress's cron system to avoid blocking inbox processing.
+
+This is implemented in `includes/scheduler/class-collection-sync.php`.
+
+## Components
```
obenland: A lot of the information in this doc is very specific and feels like it has a high likelihood of becoming outdated easily (file and function names, software versions, etc)

Maybe instead of a separate doc, we could pick the most pertinent parts and drop them in the file documentation of the `Collection_Sync` handler class or something?

includes/handler/class-collection-sync.php
pfefferle: Good catch!!!

includes/handler/class-collection-sync.php
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	/**
+	 * Get the frequency for Collection-Synchronization headers.
+	 *
+	 * @return int Frequency in seconds.
+	 */
+	private static function get_frequency() {
+		/**
+			 * Filter the frequency of Collection-Synchronization headers sent to a given authority.
```
obenland:
```suggestion
		 * Filter the frequency of Collection-Synchronization headers sent to a given authority.
```

includes/rest/class-followers-controller.php
```diff
@@ -139,6 +187,56 @@ function ( $item ) use ( $context ) {
...
+		$authority = $request->get_param( 'authority' );
+
+		// Validate authority format.
+		if ( ! preg_match( '#^https?://[^/]+$#', $authority ) ) {
+			return new \WP_Error(
+				'invalid_authority',
+				\__( 'Invalid authority format.', 'activitypub' ),
+				array( 'status' => 400 )
+			);
+		}
```
obenland:
```suggestion
```
With the pattern in the args definition, this should no longer be necessary.

includes/scheduler/class-collection-sync.php
```diff
@@ -0,0 +1,102 @@
...
+				Following::accept( $following, $user_id );
+				unset( $remote_followers[ $key ] );
+			}
+		}
+
+		/**
+		 * Action triggered after reconciliation is complete.
+		 *
+		 * @param int    $user_id         The local user ID that triggered the reconciliation.
+		 * @param string $actor_url       The remote actor URL.
```
obenland:
```suggestion
		 * @param int    $user_id   The local user ID that triggered the reconciliation.
		 * @param string $actor_url The remote actor URL.
```

includes/collection/class-followers.php
```diff
@@ -567,4 +569,108 @@ public static function remove_blocked_actors( $value, $type, $user_id ) {
...
+		return Signature::get_collection_digest( $follower_ids );
+	}
+
+	/**
+	 * Get partial followers collection for a specific instance.
+	 *
+	 * Returns only followers whose ID shares the specified URI authority.
+	 * Used for FEP-8fcf synchronization.
+	 *
+	 * @param int    $user_id  The user ID whose followers to get.
```
```suggestion
	 * @param int    $user_id   The user ID whose followers to get.
```

includes/handler/class-collection-sync.php
```diff
@@ -0,0 +1,223 @@
...
+
+	/**
+	 * Add Collection-Synchronization header to `Create` activities (FEP-8fcf).
+	 *
+	 * This method adds the Collection-Synchronization header to outgoing `Create` activities.
+	 *
+	 * @see https://codeberg.org/fediverse/fep/src/branch/main/fep/8fcf/fep-8fcf.md
+	 *
+	 * @param array  $args    The HTTP request arguments.
+	 * @param string $url     The request URL.
```
obenland:
```suggestion
	 * @param array  $args The HTTP request arguments.
	 * @param string $url  The request URL.
```

obenland (commented):
Sorry, just a couple of minor things I noticed, then it should be good to go

obenland (approved):

PR #2300, opened 2025-10-07
Pass `totalItems` to `prepare_collection_response`

pfefferle:
Refactors the REST collection response preparation to accept an explicit `$total_items` parameter instead of relying on the `totalItems` value from the response array. This improves accuracy in pagination calculations by allowing controllers to pass the correct total item count directly.

## Proposed changes:
- Updated method signature for `prepare_collection_response` to accept `$total_items` parameter
- Modified all controller calls to pass explicit total item counts
- Enhanced outbox totalItems calculation to count published posts more accurately

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Improved REST collection pagination by using explicit total item counts for more accurate results.

</details>

### DIFF
--- a/includes/rest/class-outbox-controller.php
+++ b/includes/rest/class-outbox-controller.php
@@ -76,6 +76,8 @@ class Outbox_Controller extends \WP_REST_Controller {
 				'schema' => array( $this, 'get_item_schema' ),
 			)
 		);
+
+		\add_filter( 'activitypub_rest_outbox_array', array( $this, 'overload_total_items' ), 10, 3 );
 	}
 
 	/**
@@ -268,5 +270,23 @@ class Outbox_Controller extends \WP_REST_Controller {
 		$this->schema = $schema;
 
 		return $this->add_additional_fields_schema( $this->schema );
+	}
+
+	/**
+	 * Overload total items.
+	 *
+	 * The `totalItems` property is used by Mastodon to show the overall
+	 * number of federated posts.
+	 *
+	 * @param array $response The response array.
+	 *
+	 * @return array The modified response array.
+	 */
+	public function overload_total_items( $response ) {
+		if ( isset( $response['totalItems'] ) ) {
+			$response['totalItems'] = 100;
+		}
+
+		return $response;
 	}
 }
--- a/includes/rest/class-outbox-controller.php
+++ b/includes/rest/class-outbox-controller.php
@@ -77,7 +77,7 @@ class Outbox_Controller extends \WP_REST_Controller {
 			)
 		);
 
-		\add_filter( 'activitypub_rest_outbox_array', array( $this, 'overload_total_items' ), 10, 3 );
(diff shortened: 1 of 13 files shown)


includes/rest/class-outbox-controller.php
```diff
@@ -198,7 +210,7 @@ public function get_items( $request ) {
 			$response['orderedItems'][] = $this->prepare_item_for_response( $outbox_item, $request );
 		}
 
-		$response = $this->prepare_collection_response( $response, $request );
+		$response = $this->prepare_collection_response( $response, $request, $outbox_query->found_posts );
```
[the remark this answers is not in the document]
pfefferle: this is on purpose

includes/rest/class-outbox-controller.php
```diff
@@ -175,10 +175,22 @@ public function get_items( $request ) {
...
 		);
 
+		$post_types = \get_option( 'activitypub_support_post_types', array( 'post' ) );
+
+		if ( $user_id > Actors::BLOG_USER_ID ) {
+			$count_posts            = \count_user_posts( $user_id, $post_types, true );
+			$response['totalItems'] = \intval( $count_posts );
+		} else {
+			foreach ( $post_types as $post_type ) {
+				$count_posts             = \wp_count_posts( $post_type );
```
obenland: I assume both `count_user_posts()` and `wp_count_posts()` include unfederated posts, not sure how you feel about that?

includes/rest/class-outbox-controller.php
obenland: Instead of passing `$outbox_query->found_posts` to `prepare_collection_response()`, couldn't we just keep it as the default? `'totalItems'` should be the correct amount before the response gets passed to `prepare_collection_response()`.

includes/rest/class-outbox-controller.php
pfefferle: That was what we used before. I fear that custom queries would make the request unnecessarily slow!?

includes/rest/class-outbox-controller.php
```diff
@@ -273,4 +271,49 @@ public function get_item_schema() {
...
+				'author'      => $request->get_param( 'user_id' ),
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+				'meta_query'  => array(
+					array(
+						'key'     => 'activitypub_status',
+						'compare' => 'EXISTS',
+					),
+				),
+				'fields'      => 'ids',
+				'nopaging'    => true,
```
obenland: Since we don't do anything with these IDs, I wonder if it would be a faster query if we just looked for 1 post, similar to what we're doing with comments below?

obenland (approved):

PR #2302, opened 2025-10-08
Integrate Federated Reply block with Reader post share

obenland:
https://github.com/user-attachments/assets/df766cd8-afa5-45f2-aecc-951c6affa886

## Proposed changes:

* Adds logic to replace WP.com Reader's 'Share Post' feature with the Federated Reply block when sharing posts. 
* Adapts post share parameters and redirects the request.

### Other information:

## Testing instructions:

* Apply this PR to a WP.com-connected test site.
* Follow a Federated account in the WP.com reader and pull up their feed.
* On a post, click "Repost" and select your test site.
* Make sure it correctly redirects to the Editor with the Reply block added.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Integrate Federated Reply block with WP.com Reader's post share functionality, allowing users to reply to ActivityPub posts directly from the Reader.

</details>

### DIFF
--- a/.github/changelog/2302-from-description
+++ b/.github/changelog/2302-from-description
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Integrate Federated Reply block with WP.com Reader's post share functionality, allowing users to reply to ActivityPub posts directly from the Reader.
--- a/integration/class-jetpack.php
+++ b/integration/class-jetpack.php
@@ -24,11 +24,6 @@ class Jetpack {
 	 * Initialize the class, registering WordPress hooks.
 	 */
 	public static function init() {
-		// Replace the Reader's "Share Post" with the Federated Reply block.
-		if ( isset( $_GET['is_post_share'], $_GET['url'] ) && $_GET['is_post_share'] ) { // phpcs:ignore WordPress.Security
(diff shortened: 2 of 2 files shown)


pfefferle (dismissed):
Nice and simple 😍

obenland:
@pfefferle Sorry, I had a night to think about it, I think I prefer it in a callback where it doesn't run on every page load.

pfefferle (approved):

Jiwoon-Kim:
Reblog on Tumblr and Share on Facebook are not actually **`Announce`** activities in ActivityPub — they’re conceptually closer to **`Quote`**.
The pull request *“Integrate Federated Reply block with Reader post share (#2302)”* treats the embedding of a Federated Reply block as a repost.

However, I personally believe it would be more appropriate to implement a **(tentatively named) Federated Quote block** first, and then integrate it with the **WordPress.com Reader’s “Repost”** feature.

Since **`add/share-post-override`** has already been merged into trunk, it might be better to handle this in a follow-up patch — for example, **`change/match-quote-to-repost`** — to align repost behavior with the semantics of quote-type activities.

PR #2303, opened 2025-10-09
Add email notifications for new quote events

pfefferle:
Implements email notifications when a user's post is quoted, including user and blog-level settings, email template, and tests. Updates admin and settings UI to allow enabling/disabling quote notifications. Adds comprehensive PHPUnit tests for the new notification logic.

Screenshot 2025-10-09 at 12 32 44

## Proposed changes:
- Adds email notification functionality for quote events with user and blog-level settings
- Creates a new email template specifically for quote notifications
- Implements comprehensive PHPUnit test coverage for the notification logic

### Other information:

## Testing instructions:

* Checkout on a test site
* Quote a blog-post with a mastodon user
* Check Mail

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added email notifications for quoted posts, with user and blog-level settings and a new customizable email template.

</details>

### DIFF
--- a/includes/class-mailer.php
+++ b/includes/class-mailer.php
@@ -22,8 +22,8 @@ class Mailer {
 
 		\add_action( 'activitypub_inbox_follow', array( self::class, 'new_follower' ), 10, 2 );
 		\add_action( 'activitypub_inbox_create', array( self::class, 'direct_message' ), 10, 2 );
-		/** After @see \Activitypub\Handler\Create::handle_create() */
-		\add_action( 'activitypub_inbox_create', array( self::class, 'mention' ), 20, 2 );
+		
+		\add_action( 'activitypub_inbox_create', array( self::class, 'mention' ), 20, 2 ); // After @see \Activitypub\Handler\Create::handle_create()
 
 		\add_action( 'activitypub_handled_quote_request', array( self::class, 'quote' ), 10, 3 );
 	}
--- a/includes/class-mailer.php
+++ b/includes/class-mailer.php
@@ -420,6 +420,10 @@ class Mailer {
 			$quoted_url = object_to_uri( $activity['object']['quoteUrl'] );
 		}
 
+		if ( ! $quoted_url ) {
+			return;
+		}
+
 		$template_args = array(
 			'activity'   => $activity,
 			'actor'      => $actor,
--- a/includes/class-mailer.php
+++ b/includes/class-mailer.php
@@ -424,11 +424,19 @@ class Mailer {
 			return;
 		}
 
+		// Try to get the quoted post title.
+		$quoted_title = null;
+		$post_id      = \url_to_postid( $quoted_url );
+		if ( $post_id ) {
+			$quoted_title = \get_the_title( $post_id );
+		}
+
 		$template_args = array(
-			'activity'   => $activity,
-			'actor'      => $actor,
-			'user_id'    => $user_id,
-			'quoted_url' => $quoted_url,
+			'activity'     => $activity,
+			'actor'        => $actor,
+			'user_id'      => $user_id,
+			'quoted_url'   => $quoted_url,
+			'quoted_title' => $quoted_title,
 		);
 
 		/* translators: 1: Blog name, 2: Actor name */
--- a/includes/class-mailer.php
+++ b/includes/class-mailer.php
@@ -22,7 +22,10 @@ class Mailer {
 
 		\add_action( 'activitypub_inbox_follow', array( self::class, 'new_follower' ), 10, 2 );
 		\add_action( 'activitypub_inbox_create', array( self::class, 'direct_message' ), 10, 2 );
-		\add_action( 'activitypub_inbox_create', array( self::class, 'mention' ), 20, 2 );  /** After @see \Activitypub\Handler\Create::handle_create() */
+		/** After @see \Activitypub\Handler\Create::handle_create() */
+		\add_action( 'activitypub_inbox_create', array( self::class, 'mention' ), 20, 2 );
+
+		\add_action( 'activitypub_handled_quote_request', array( self::class, 'quote' ), 10, 3 );
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


includes/class-mailer.php
```diff
@@ -360,6 +361,112 @@ public static function mention( $activity, $user_id ) {
...
+		$actor = get_remote_metadata_by_actor( $activity['actor'] );
+		if ( ! $actor || \is_wp_error( $actor ) ) {
+			return;
+		}
+
+		$actor = self::normalize_actor( $actor );
+
+		// Get the quoted post/object.
+		// For QuoteRequest activities, object is the quoted URL string.
+		// For regular quote posts, object.quoteUrl contains the quoted URL.
```
obenland: Let's update this to adhering to [WordPress docs standards for multiline comments](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-2-multi-line-comments).

includes/class-mailer.php
```diff
@@ -360,6 +361,112 @@ public static function mention( $activity, $user_id ) {
...
+		$subject = \sprintf( \esc_html__( '[%1$s] Quote from: %2$s', 'activitypub' ), \esc_html( \get_option( 'blogname' ) ), \esc_html( $actor['name'] ) );
+
+		\ob_start();
+		\load_template( ACTIVITYPUB_PLUGIN_DIR . 'templates/emails/new-quote.php', false, $template_args );
+		$html_message = \ob_get_clean();
+
+		$alt_function = function ( $mailer ) use ( $actor, $activity, $quoted_url ) {
+			$content = '';
+			// For QuoteRequest, instrument contains the quote post URL.
+			// For regular quotes, object contains the quote post.
```
obenland: Let's update this to adhering to [WordPress docs standards for multiline comments](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-2-multi-line-comments).

includes/class-mailer.php
```diff
@@ -360,6 +361,112 @@ public static function mention( $activity, $user_id ) {
...
+			$quote_object = $activity['instrument'] ?? $activity['object'];
+			if ( is_array( $quote_object ) && ! empty( $quote_object['content'] ) ) {
```
obenland, line 452:
```suggestion
			if ( ! empty( $quote_object['content'] ) ) {
```
We should be able to simplify this?

includes/class-mailer.php
```diff
@@ -360,6 +361,112 @@ public static function mention( $activity, $user_id ) {
...
+				$message .= $content . "\r\n\r\n";
+			}
+
+			if ( $quoted_url ) {
+				/* translators: Quoted post URL */
+				$message .= \sprintf( \esc_html__( 'Your post: %s', 'activitypub' ), \esc_url( $quoted_url ) ) . "\r\n";
+			}
+
+			// Get the quote post URL.
+			$quote_url = is_array( $quote_object ) && ! empty( $quote_object['id'] ) ? $quote_object['id'] : ( $activity['instrument'] ?? '' );
```
obenland:
```suggestion
			$quote_url = empty( $quote_object['id'] ) ? $activity['instrument'] ?? '' : $quote_object['id'];
```
Can it be simplified to this? if it can, is `$quote_object['id']` actually not set in that case and it could be 
```			
			$quote_url = $quote_object['id'] ?? $activity['instrument'] ?? '';
```

includes/class-mailer.php
```diff
@@ -360,6 +361,112 @@ public static function mention( $activity, $user_id ) {
...
+		// Get the quoted post/object.
+		// For QuoteRequest activities, object is the quoted URL string.
+		// For regular quote posts, object.quoteUrl contains the quoted URL.
+		$quoted_url = null;
+		if ( is_string( $activity['object'] ) ) {
+			// QuoteRequest format.
+			$quoted_url = $activity['object'];
+		} elseif ( ! empty( $activity['object']['quoteUrl'] ) ) {
+			// Regular quote post format.
+			$quoted_url = $activity['object']['quoteUrl'];
```
obenland:
```suggestion
			$quoted_url = object_to_uri( $activity['object']['quoteUrl'] );
```
Not sure if this would be helpful? Could it be a link object?

templates/emails/new-quote.php
```diff
@@ -0,0 +1,68 @@
...
+// Load header.
+require __DIR__ . '/parts/header.php';
+?>
+
+<h1><?php esc_html_e( 'Your post was quoted!', 'activitypub' ); ?></h1>
+
+<p>
+	<?php
+	/* translators: %s: The name of the person who quoted the post. */
+	$message = __( 'Looks like someone&#8217;s sharing one of your posts! Your post was just quoted by %s on the Fediverse. Here&#8217;s what they said:', 'activitypub' );
```
obenland:
```suggestion
	$message = __( 'Looks like someone&#8217;s shared one of your posts! Your post was just quoted by %s on the Fediverse. Here&#8217;s what they said:', 'activitypub' );
```

obenland: The "Here's what they said:" part only works with an embed. We might need two strings depending on whether there's an object to display or not.

includes/class-mailer.php
pfefferle: the spec is a bit fuzzy here.

includes/class-mailer.php
```diff
@@ -360,6 +361,117 @@ public static function mention( $activity, $user_id ) {
...
+		 * For regular quote posts, object.quoteUrl contains the quoted URL.
+		*/
+		$quoted_url = null;
+		if ( is_string( $activity['object'] ) ) {
+			// QuoteRequest format.
+			$quoted_url = $activity['object'];
+		} elseif ( ! empty( $activity['object']['quoteUrl'] ) ) {
+			// Regular quote post format.
+			$quoted_url = object_to_uri( $activity['object']['quoteUrl'] );
+		}
```
obenland, line 421: if `$quoted_url` is `null`, should we bail here?

templates/emails/new-quote.php
```diff
@@ -0,0 +1,68 @@
...
+// Load header.
+require __DIR__ . '/parts/header.php';
+?>
+
+<h1><?php esc_html_e( 'Your post was quoted!', 'activitypub' ); ?></h1>
+
+<p>
+	<?php
+	/* translators: %s: The name of the person who quoted the post. */
+	$message = __( 'Looks like someone&#8217;s shared one of your posts! Your post was just quoted by %s on the Fediverse. Here&#8217;s what they said:', 'activitypub' );
```
obenland: The "Here's what they said:" part only works with an embed. We might need two strings depending on whether there's an object to display or not?

For like and reblog notifications on Dotcom we include the post title and link to the post as well.
Something like that?

```php
	$comment_author = '<a href="' . esc_url( $args['actor']['url'] ) . '">' . esc_html( $args['actor']['webfinger'] ) . '</a>';
	$allowed_tags   = array('a' => array('href' => array() ) );

	if ( ! empty( $args['post_title'], $args['quoted_url'] ) ) :
		/* translators: 1: author link, 2: post URL, 3: post title */
		$message = __( '%1$s just quoted <a href="%2$s">&#8220;%3$s&#8221;</a> on the Fediverse. Here&#8217;s what they said:', 'activitypub' );
		printf(
			wp_kses( $message, $allowed_tags ),
			wp_kses( $comment_author, $allowed_tags ),
			esc_url( $args['quoted_url'] ),
			esc_html( $args['post_title'] )
		);
	elseif ( ! empty( $args['quoted_url'] ) ) :
		/* translators: 1: author link, 2: post URL */
		$message = __( '%1$s just quoted <a href="%2$s">one of your posts</a> on the Fediverse. Here&#8217;s what they said:', 'activitypub' );
		printf(
			wp_kses( $message, $allowed_tags ),
			wp_kses( $comment_author, $allowed_tags ),
			esc_url( $args['quoted_url'] )
		);
	else :
		/* translators: %s: author link */
		$message = __( '%s just quoted one of your posts on the Fediverse. Here&#8217;s what they said:', 'activitypub' );
		printf(
			wp_kses( $message, $allowed_tags ),
			wp_kses( $comment_author, $allowed_tags )
		);
	endif;
```

obenland: Maybe "just shared" instead of "just quoted" to avoid overlap with `Your post was quoted!` one line above?

obenland: I think it would also be fine to drop "Looks like someone&#8217;s shared one of your posts!"

obenland:
A couple more thoughts on this:

- [ ] A quote post represents a reblog at the same time, triggering email notifications for the new "comment". We should either not send the reblog notification or… figure something else out that doesn't result in two emails for the same action. (Including Dotcom)
- [x] The quote-url-only case, where there is no object to embed is still a mystery to me. In that case we don't seem to link to the quoting post either, just display the URL of the post that was quoted? Do we not have the URL of the quote post? (if we do, we should fetch it and embed it 🙂)

obenland:
`Mailer::quote()` and the quote template should be ready now. I think it's reasonable to expect that we'll always have a post to embed, similar to DM and Mention emails.

The two remaining items of uncertainty for me:

- [ ] We have tests for create activities triggering Quote email notifications and account for Notes with `quoteurl` properties, but we're not currently wired in to trigger quote email notifications on Create. What should be done there?
- [x] A quote post represents a reblog at the same time, triggering email notifications for the new "comment". We should either not send the reblog notification or… figure something else out that doesn't result in two emails for the same action. (Including Dotcom)

pfefferle:
@obenland Mastodon does not send an `Announce` in the case of a Quote, so there is no duplication.

Screenshot 2025-10-15 at 10 58 11

pfefferle:
We could add Quotes as Comments though. This way we could re-use (update) the comment Mails.

obenland:
> Mastodon does not send an Announce in the case of a Quote, so there is no duplication.

You don't get reblog comment notifications in your testing?

image

pfefferle:
No, not even live!

pfefferle:
@obenland have you clicked repost and quote?

pfefferle:
(because you can do both actions on the same item)

obenland:
Not that I know of. It gives me two options and I clicked quote.

obenland:
> We could add Quotes as Comments though. This way we could re-use (update) the comment Mails.

Yeah not sure I have an opinion on that, tbh, I'd just go with your preference on this.

pfefferle:
@obenland can you try to reproduce if you receive two mails or not? I tried it several times and can not reproduce it.

obenland:
Interesting, I can't reproduce it anymore, no. Checking that off the list.

obenland:
> We have tests for create activities triggering Quote email notifications and account for Notes with quoteurl properties, but we're not currently wired in to trigger quote email notifications on Create.

What should we do with that?

PR #2305, opened 2025-10-09
Add scheduled inbox purge and retention setting

pfefferle:
Introduces a scheduled task to purge old inbox items, similar to the existing outbox purge. Adds an advanced setting for configuring the inbox retention period, including UI and option registration. Ensures site performance by allowing administrators to control how long inbox items are kept.

## Proposed changes:
- Adds a new advanced setting for configuring inbox retention period (default 360 days)
- Implements a daily scheduled task to automatically purge old inbox items
- Provides UI for administrators to control the inbox retention setting

### Other information:

## Testing instructions:

* See unittests

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added a scheduled task and setting to automatically purge old inbox items, helping maintain site performance and storage control.

</details>

### DIFF
--- a/includes/class-scheduler.php
+++ b/includes/class-scheduler.php
@@ -123,11 +123,11 @@ class Scheduler {
 	 * @return void
 	 */
 	public static function deregister_schedules() {
-		wp_unschedule_hook( 'activitypub_update_remote_actors' );
-		wp_unschedule_hook( 'activitypub_cleanup_remote_actors' );
-		wp_unschedule_hook( 'activitypub_reprocess_outbox' );
-		wp_unschedule_hook( 'activitypub_outbox_purge' );
-		wp_unschedule_hook( 'activitypub_inbox_purge' );
+		\wp_unschedule_hook( 'activitypub_update_remote_actors' );
+		\wp_unschedule_hook( 'activitypub_cleanup_remote_actors' );
+		\wp_unschedule_hook( 'activitypub_reprocess_outbox' );
+		\wp_unschedule_hook( 'activitypub_outbox_purge' );
+		\wp_unschedule_hook( 'activitypub_inbox_purge' );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


obenland:
Instead of introducing separate options and cron jobs, could this be combined with the Outbox purging?

I noticed the retention period being double that of the Outbox. Is 6 months not enough? I'd expect there to be vastly more posts in the Inbox than the Outbox, especially with Following enabled.

pfefferle:
The inbox is what will be used by Apps in the future, so I thought it should be at least a different setting! We can reduce it though.

pfefferle:
I think we should keep them separated to be more flexible.

obenland (commented):
Let's not show a UI for the option and keep it CLI or `wp-admin/options.php` only.

pfefferle:
@obenland like discussed in slack, I removed the settings for Inbox/Outbox purge.

includes/class-scheduler.php
```diff
@@ -321,6 +329,40 @@ public static function purge_outbox() {
...
+				'post_type'   => Inbox::POST_TYPE,
+				'post_status' => 'any',
+				'fields'      => 'ids',
+				'numberposts' => -1,
+				'date_query'  => array(
+					array(
+						'before' => $date->format( 'Y-m-d' ),
+					),
+				),
+			)
```
obenland, line 353:
```suggestion
		$days     = (int) \get_option( 'activitypub_inbox_purge_days', 180 );
		$post_ids = \get_posts(
			array(
				'post_type'   => Inbox::POST_TYPE,
				'post_status' => 'any',
				'fields'      => 'ids',
				'numberposts' => -1,
				'date_query'  => array(
					array(
						'before' => gmdate( 'Y-m-d', time() - ( $days * DAY_IN_SECONDS ) ),
					),
				),
			)
```
Would this work? Might be a bit simpler.

obenland (commented):
I think this all looks good, had just one question:

includes/class-scheduler.php
pfefferle: let's see what the tests say!

obenland: A different level of cowboy coding 😄

obenland (approved):

PR #2311, opened 2025-10-13
Add support for ActivityPub Reader-Item post type

pfefferle:
Introduces a new 'ap_post' post type and related taxonomies for handling ActivityPub objects, including registration, CRUD operations, and taxonomy management. Updates create and update handlers to process non-interaction objects using the new Objects collection. Enhances debug functionality to support the new post type and taxonomies.

## Proposed changes:
- Added new Objects collection class with methods for creating, updating, and retrieving ActivityPub objects
- Modified Create and Update handlers to differentiate between interactions (replies) and objects, routing each to appropriate collections
- Enhanced debug functionality to support the new post type and taxonomies

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added a new ap_object post type and taxonomies for storing and managing incoming ActivityPub objects, with updated handlers

</details>


includes/class-sanitize.php
```diff
@@ -182,4 +182,23 @@ public static function webfinger( $value ) {
...
+	public static function content( $content ) {
+		$content = \make_clickable( $content );
+		$content = \wpautop( $content );
+		$content = \wp_kses_post( $content );
+
+		if ( ! site_supports_blocks() ) {
+			return $content;
+		}
+
+		return Blocks::html_to_blocks( $content );
```
pfefferle: No idea if that makes sense, because we do not really allow to edit the remote posts!?

Jiwoon-Kim: Perhaps it may be necessary to allow editing in order to permanently cache a DeepL-translated version in the `contentMap`?

pfefferle: You should not change the content at all, even for translations! I agree that we might want to also store the content in the contentMap, that might match the local lang though.

includes/handler/class-create.php
```diff
@@ -34,14 +35,38 @@ public static function init() {
...
 
+		$success = ( false !== $result && ! \is_wp_error( $result ) );
+
+		/**
+		 * Fires after an ActivityPub Create activity has been handled.
+		 *
+		 * @param array                            $activity The ActivityPub activity data.
+		 * @param int                              $user_id  The local user ID.
+		 * @param bool                             $success  True on success, false otherwise.
+		 * @param array|string|int|\WP_Error|false $result   The WP_Comment object of the created comment, or null if creation failed.
```
obenland: I'm not sure how we ended up with `array|string|int|\WP_Error|false` but we should review that. With `self::create_object()` (`self::create_post()`?), this can now contain a `WP_Post`.

includes/handler/class-create.php
obenland: Up until now, this `return` prevented the `activitypub_handled_create` action from firing—with these changes that is no longer the case. Same for the `return` in the `is_self_ping()` conditional below ⬇️ .
Not sure how to best solve this

includes/handler/class-create.php
```diff
@@ -61,23 +86,41 @@ public static function handle_create( $activity, $user_id, $activity_object = nu
...
+		if ( ! \is_wp_error( $check_dupe ) ) {
+			/**
+			 * Fires when a Create activity is received for an existing object.
+			 *
+			 * @param array                          $activity        The activity-object.
+			 * @param int                            $user_id         The id of the local blog-user.
+			 * @param \Activitypub\Activity\Activity $activity_object The activity object.
+			 */
+			\do_action( 'activitypub_inbox_update', $activity, $user_id, $activity_object );
+			return;
```
obenland: Same as [above](https://github.com/Automattic/wordpress-activitypub/pull/2311/files#r2426698245).

includes/handler/class-update.php
```diff
@@ -43,7 +45,7 @@ public static function handle_update( $activity, $user_id ) {
 			case 'Organization':
 			case 'Service':
 			case 'Application':
-				self::update_actor( $activity, $user_id );
+				self::handle_actor_update( $activity, $user_id );
```
obenland: Nitpick: Terms like `handle` and `manage` are quite ambiguous. Could we keep `update_actor()` and call it `update_object()` below?

includes/class-post-types.php
```diff
@@ -345,6 +347,78 @@ public static function register_outbox_post_type() {
...
+				'map_meta_cap'        => true,
+				'public'              => false,
+				'show_in_rest'        => true,
+				'rewrite'             => false,
+				'query_var'           => false,
+				'supports'            => array( 'title', 'editor', 'author', 'custom-fields', 'excerpt', 'comments' ),
+				'delete_with_user'    => true,
+				'can_export'          => true,
+				'exclude_from_search' => true,
+				'taxonomies'          => array( 'ap_tag' ),
```
obenland: Should `ap_object_type` be part of this list?

includes/class-post-types.php
```diff
@@ -345,6 +347,78 @@ public static function register_outbox_post_type() {
...
+					'edit_item'     => \__( 'Edit Tag', 'activitypub' ),
+					'update_item'   => \__( 'Update Tag', 'activitypub' ),
+					'add_new_item'  => \__( 'Add New Tag', 'activitypub' ),
+					'new_item_name' => \__( 'New Tag Name', 'activitypub' ),
+					'menu_name'     => \__( 'Tags', 'activitypub' ),
+				),
+				'public'       => false,
+				'query_var'    => true,
+				'show_in_rest' => true,
+				'rewrite'      => array( 'slug' => 'ap_tag' ),
```
obenland: This should be the default?

includes/class-post-types.php
```diff
@@ -345,6 +347,78 @@ public static function register_outbox_post_type() {
...
+				'exclude_from_search' => true,
+				'taxonomies'          => array( 'ap_tag' ),
+			)
+		);
+
+		\register_taxonomy(
+			'ap_tag',
+			array( Posts::POST_TYPE ),
+			array(
+				'labels'       => array(
```
obenland: For non-hierarchical taxonomies, these labels might be the default, according to `get_taxonomy_labels()`.

includes/class-post-types.php
```diff
@@ -345,6 +347,78 @@ public static function register_outbox_post_type() {
...
+					'edit_item'     => \__( 'Edit Object Type', 'activitypub' ),
+					'update_item'   => \__( 'Update Object Type', 'activitypub' ),
+					'add_new_item'  => \__( 'Add New Object Type', 'activitypub' ),
+					'new_item_name' => \__( 'New Object Type Name', 'activitypub' ),
+					'menu_name'     => \__( 'Object Types', 'activitypub' ),
+				),
+				'public'       => false,
+				'query_var'    => true,
+				'show_in_rest' => true,
+				'rewrite'      => array( 'slug' => 'ap_object_type' ),
```
obenland: Same as above, I think this is the default.

includes/handler/class-create.php
pfefferle: yes, because it is no `create` but an `update`. this object already exists, so it will run through the update process and trigger `activitypub_handled_update`

pfefferle, line 89: this is a different case!

includes/handler/class-create.php
pfefferle: Same comment as above ;)

includes/handler/class-create.php
obenland: What I'm trying to say is that it will now run `activitypub_inbox_update` and then `activitypub_handled_create`. It no longer returns early in `handle_create()`.

pfefferle: I see!

pfefferle:
@obenland how should we proceed with this PR? Keep it open and merge every other follow up PR into this branch or see that we can merge this one first?

obenland:
Either way, I'm happy for this one to go in soon. Can give it another look today

includes/class-post-types.php
```diff
@@ -345,6 +347,65 @@ public static function register_outbox_post_type() {
 		);
 	}
 
+	/**
+	 * Register the Object post type.
+	 */
+	public static function register_object_post_type() {
```
obenland: Should this be renamed, too? And the comment? 
`register_post_post_type()` is probably not great. Maybe `register_ap_post_post_type()`? We could also leave it

includes/handler/class-update.php
```diff
@@ -58,7 +60,7 @@ public static function handle_update( $activity, $user_id ) {
 			case 'Video':
 			case 'Event':
 			case 'Document':
-				self::update_interaction( $activity, $user_id );
+				self::handle_object_update( $activity, $user_id );
```
```suggestion
				self::update_object( $activity, $user_id );
```

includes/handler/class-create.php
```diff
@@ -34,14 +35,42 @@ public static function init() {
...
+			$result = self::create_interaction( $activity, $user_id, $activity_object );
+		} else { // Handle non-interaction objects.
+			$result = self::create_post( $activity, $user_id, $activity_object );
+		}
+
+		if ( false === $result ) {
 			return;
 		}
 
+		$success = ( false !== $result && ! \is_wp_error( $result ) );
```
obenland:
```suggestion
		$success = ! \is_wp_error( $result );
```
`$result` is never false at this point.

includes/handler/class-create.php
```diff
@@ -54,30 +83,48 @@ public static function handle_create( $activity, $user_id, $activity_object = nu
...
+	}
+
+	/**
+	 * Handle non-interaction posts like posts.
+	 *
+	 * @param array                          $activity        The activity-object.
+	 * @param int                            $user_id         The id of the local blog-user.
+	 * @param \Activitypub\Activity\Activity $activity_object Optional. The activity object. Default null.
+	 *
+	 * @return \WP_Post|\WP_Error The post on success or WP_Error on failure.
```
obenland:
```suggestion
	 * @return \WP_Post|\WP_Error|false The post, WP_Error on failure, false if not processed.
```

obenland (approved):

Issue #2146, opened 2025-09-05
An analysis of the idea to implement an image object block in the WordPress Block Editor.


Jiwoon-Kim:
An analysis of the idea to implement an image object block in the WordPress Block Editor.

Based on the current progress, I believe it is appropriate to open an issue for this. The following content likely pertains to information that should be included in the `json-ld` of a media attachment post.

1.  Included in the body (`content`). 2. Included in the attachment. 3. Included in comments.

-----

### 1\. Included in the body (`content`)

Inclusion of an ActivityPub object within the body itself is a **non-standard interaction**. However, it appears to be a feasible feature to implement within WordPress.

Currently, an image included in the body is not an ActivityPub object; it's an HTML array (typically within a gallery block). However, it could be made interactive via the Jetpack carousel.

In federation, the `wp-block-gallery` is replaced with `ap-block-gallery`.
This would allow all images to become interactive image objects if the follower is a WordPress instance.

**Feature Idea**

  * **Support Attachment Post Links in the Gallery Block**
    → Provide a compatibility option to the image object block in the block editor. When inserting images into a gallery, it would allow the retrieval of an image from a `?attachment_id=123` permalink instead of a simple file URL.
    → There's no need to load the entire object; you only need to retrieve the file path URL, the URL of the `post type: attachment`, and the author ID. This could be considered a non-standard interaction and might be a Jetpack-specific mechanism.

  * **Image-Level Federation**
    → This allows for direct interaction with a media library's `Image` object, rather than a simple hotlink to an image within the post body.

  * **Image Carousel and Comments**
    → On my instance, images can be viewed individually in the image carousel (=Gallery UI).
    → A user on a remote instance can directly send actions like comments or likes to a specific `Image Object`.
    → In short, it enables interaction at the image level, not just with the entire post.

**Technical Considerations**

  * The ActivityPub plugin must add logic to convert `?attachment_id=123` into an `Image Object`.
  * Interactions like comments or likes must be mapped to direct actions with the `Image` object in the `ActivityStreams 2.0` specification.

-----

### 2\. Included in the attachment

This model involves directly including a **remote instance's `Image` object in the `attachment` array of your post (`Note` or `Article`)**, rather than simply embedding an `<img>` tag. This would create a concept like a WordPress **"Shared Media Library."**

According to GPT, the core question is whether the `attachment` must be an "image from the original instance." In the specification, **that's not a requirement.**

[...]

Issue #1914, opened 2025-07-10
Generalize how remote objects and link types may be stored


Menrath:
Follow up of https://github.com/Automattic/wordpress-activitypub/issues/1297

Currently most remote objects are only stored as comments. This is fine, as comments are Notes only and there is no other remote comment than replies to original posts.

The Friends plugin (@akirk) stores incoming objects too, probably in it's own way. The Event Bridge does too, again a different implementation.

I suggest  to start with introduce a unified way how to mark `posts` of all post types as "remote with ActivityPub source", e.g. via a unified meta-key that references an actor (`ap_actor`).

Maybe think of allowing relationships between a comment and a `post`, replies are mostly Notes, but do they have to be?

Menrath:
I mention this yesterday in a discussion with @akirk. Maybe a custom post status, e.g. `activitypub_remote` along with a post_meta `_ap_actor`, could be proposed and advertised.

obenland:
@Menrath [Is this fixed with the new `ap_post` post type](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/handler/class-create.php#L44)?

Menrath:
@obenland I will review it next week!

Menrath:
@obenland 

**TL/DR:** I guess to answer that question I have to really work with the code hands on and migrate features of the Event Bridge to it, not just read and review!

I guess it is not solved completely, but the recent changes around the inbox/attachments/`ap_post` etc are huge improvements and are laying a good foundation for a lot of cool applications to come!

I looked into the code, but I think it's easier to give you two example scenarios, than to point out each line of code where a filter/change etc. could be useful at this moment.

1. Event-Bridge like aggregation of events. Following other ActivityPub profiles in order to list their events too. I guess I could hook into the follow table to insert an boolean trigger to activate the behaviour for each followed Actor (not just follow, but also use that Actor as an event source and list their events). When receiving remote `ap_actor`s I need to create an post with the corresponding event post type. 

It is complex. I guess I could currently hook and disable the creation of an `ap_post` and still create an event post. The attachment class and the `ap_post` class are quite tensely linked, I guess I would have a hard time using it on another post_type than `ap_post`. In order to give better feedback, I need to really dive into migrating the existing features and make use of your modernisations and really go to the point where there are missing parts, it's hard to tell from an observer point of view. I can start coding so next week. 

[...]

PR #2314, opened 2025-10-13
Add Attachments processor class for handling ActivityPub media

obenland:
## Proposed changes:
  Implements a dual storage strategy where Mastodon imports continue using WordPress attachment posts while
  ActivityPub inbox items use direct file storage in `uploads/activitypub/ap_posts/{post_id}/`.

  **New methods added:**
  - `import_files()` - Public method for direct file import (used by ap_posts)
  - `save_file()` - Downloads and saves files to `uploads/activitypub/ap_posts/{post_id}/`, returns array with
  `url`, `mime_type`, and `alt` keys
  - `import_inline_files()` - Processes inline images for file storage
  - `append_files_to_content()` - Appends file-based media to content
  - `generate_files_markup()` - Creates block markup from file data
  - `get_files_gallery_block()` - Gallery block for file-based attachments
  - `delete_directory()` - Public helper to delete post's file directory

  **Methods removed:**
  - `maybe_hide_from_media_library()` - No longer needed, Mastodon import attachments should be visible in the
  user's media library
  - `has_updated_attachments()` - Simplified to inline empty check

  **Key improvements:**
  - Uses WP_Filesystem API exclusively for all file operations
  - Only handles remote URLs (no local file logic needed for inbox items)
  - Public `$ap_posts_dir` property for consistent path references
  - Removed unused function parameters
  - Well-documented return types with hash notation for array keys
  - Simple folder-per-post structure for easy cleanup

  ### Other information:

  ## Testing instructions:

  * All 1212 tests passing
  * Tests updated to verify file creation instead of attachment posts
  * Test cleanup properly removes activitypub directories
  * Mastodon import tests still pass (no regression)
  * Tests use public `$ap_posts_dir` property for path consistency

  **Technical details:**
  The implementation uses a simple folder-per-post structure where each `ap_post` gets its own directory at
  `uploads/activitypub/ap_posts/{post_id}/`. When a post is deleted, the entire directory is recursively removed
  using `WP_Filesystem::delete()`.

  Files are tracked by the directory structure itself rather than post meta, keeping the implementation simple 
  and avoiding the need for complex deduplication logic.

  ### Changelog entry

  <details>

  #### Significance

  - [x] Minor

  #### Type

  - [x] Changed - for changes in existing functionality

  #### Message

  Use direct file storage for ActivityPub inbox attachments instead of creating WordPress attachment posts. 
  Mastodon imports continue using attachment posts as before.

  </details>


includes/collection/class-posts.php
```diff
@@ -164,4 +184,38 @@ private static function add_taxonomies( $post_id, $activity_object ) {
...
+		$new_urls = array_map( 'Activitypub\object_to_uri', $attachments );
+
+		// Extract URLs from existing attachments.
+		$stored_urls = array();
+		foreach ( $attached_media as $attachment ) {
+			$stored_urls[] = \get_post_meta( $attachment->ID, '_activitypub_source_url', true );
+		}
+
+		// Compare the arrays.
+		return array_filter( $stored_urls ) !== $new_urls;
```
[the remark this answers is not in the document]
obenland: Fixed in [4ff81f7](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/4ff81f7b3f8410a7b535f0184b8d209b3b644619).

pfefferle:
I ran some tests and have some feedback not directly to the code, but the handling:

1. We should not add external Attachments to the Media-Library. This is very spammy and floods the internal media handling with external media.
2. When I delete a post, the Media-Attachment will not be deleted with the post.
3. It seems that the current attachment handling has problems with `Update` activities. When I delete a post without deleting the Attachment, the new one was generated without the Attachment.
4. The current version does not side-load inline images. I think we should support inline images and ignore attachments that reference the same one.

Screenshot 2025-10-15 at 10 29 49

pfefferle (changes requested):
^

Jiwoon-Kim:
My consistent opinion is that media file direct links and media objects should be clearly distinguished. The attachment page is the ideal place to include the JSON-LD representation of a media object.
I’ve suggested ideas for improving the media library several times before, and I plan to write a detailed post about it in the Discussions soon.

The content of a remote image object is essentially a hotlink.
It’s also possible to use it without uploading it to on-site storage, depending on the implementation.

Jiwoon-Kim:
Yes, I’m aware that Mastodon locally caches all attachment images, but WordPress operates under a very different architecture.

My current idea is to load all attachment JSON-LD metadata first, then apply lazy loading for inline images or gallery blocks.

This would provide better performance and flexibility without requiring immediate local uploads.

That said, we still need to consider how platforms like Mastodon, Pixelfed, and Threads handle this, since the appropriate approach may depend on the object type — for example, Article objects might need different attachment-fetching behavior compared to Note objects.

Jiwoon-Kim:
might need a page like ?remote_attachment_id, but it may only need to be applied to Articles and not to Notes.

Jiwoon-Kim:
In the case of directly attached media files, a ?remote_attachment_id page is not generated.
Such a page is created only for media objects.

In both cases, the default behavior is to maintain hotlinking (remote linking).

If a license extension is applied, the file is optionally cached locally and the file URL is replaced with a local one — but the original URI is preserved — only when explicit reuse permission is specified.

It is normal for the remote media attachment page to remain even after the remote article is deleted, because it implies that the media can still be reused through the media library.

I’ll post it soon in the discussion tab under the title “Media Upload Endpoint and Remote Media Library.”

Jiwoon-Kim:
https://github.com/Automattic/wordpress-activitypub/discussions/2329#discussioncomment-14709021

> If there is no media library, implementing the media object endpoint will be difficult, so the existence of a media object can be assumed to mean that a media library and media upload endpoint exist.
>If a media object exists, it can be reused. However, in the case of a direct media file link, it cannot be proven in ActivityPub that it is the same entity, so it is one-time. Therefore, in the case of WordPress, there is no reason to store such media locally.

https://make.wordpress.org/core/2023/07/07/media-library/

Jiwoon-Kim:
https://github.com/Automattic/wordpress-activitypub/discussions/2345

obenland:
@pfefferle Side-loaded images are now hidden in the Media Library, they get deleted when their parent post gets deleted, and inline images are being parsed and replaced as well. It's probably not perfect, but should be a good step forward.

This should be ready for another round of reviews.

Jiwoon-Kim:
For example, if the hosting policy offers unlimited databases but only 1 GB of storage space, would that still be sufficient for normal use?

pfefferle:
~The image is still added twice, if it is inline and an attachment. I think we should de-duplicate them and not add the attachment if the image is already inline.~

Pardon me! This works like a charm and I simply messed up my test!

includes/class-attachments.php
```diff
@@ -0,0 +1,417 @@
...
+		$gallery .= '<figure class="wp-block-gallery has-nested-images columns-default is-cropped">';
+
+		foreach ( $attachment_ids as $id ) {
+			$image_src = \wp_get_attachment_image_src( $id, 'large' );
+			if ( ! $image_src ) {
+				continue;
+			}
+
+			$caption  = \get_post_field( 'post_content', $id );
+			$gallery .= "\n<!-- wp:image {\"id\":{$id},\"sizeSlug\":\"large\",\"linkDestination\":\"none\"} -->\n";
```
pfefferle, line 686: we do not parse the content for blocks, why should we add the images as such?

includes/class-attachments.php
```diff
@@ -0,0 +1,417 @@
...
+	/**
+	 * Process attachments from an ActivityPub object and attach them to a post.
+	 *
+	 * @param array $attachments Array of ActivityPub attachment objects.
+	 * @param int   $post_id     The post ID to attach files to.
+	 * @param int   $author_id   Optional. User ID to set as attachment author. Default 0.
+	 *
+	 * @return array Array of attachment IDs.
+	 */
+	public static function process( $attachments, $post_id, $author_id = 0 ) {
```
pfefferle: `process` is a bit vague and does not really make clear what exactly the function is doing. Maybe `import_and_attach`?

includes/class-attachments.php
```diff
@@ -0,0 +1,417 @@
...
+		if ( ! empty( $attachment_data['name'] ) ) {
+			$mime_type = $attachment_data['mediaType'] ?? '';
+			if ( 'image' === strtok( $mime_type, '/' ) ) {
+				$post_data['meta_input']['_wp_attachment_image_alt'] = $attachment_data['name'];
+			}
+		}
+
+		// Flag to filter out from Media Library.
+		if ( Posts::POST_TYPE === get_post_type( $post_id ) ) {
+			$post_data['meta_input']['_activitypub_import'] = '1';
```
pfefferle: instead of true/false, what about storing the source (importer, inbox, ...)?

includes/class-attachments.php
```diff
@@ -0,0 +1,417 @@
...
+
+				if ( \is_wp_error( $attachment_id ) ) {
+					continue;
+				}
+			}
+
+			$new_url = \wp_get_attachment_url( $attachment_id );
+			if ( $new_url ) {
+				$url_mappings[ $image_url ] = $new_url;
+				$content                    = str_replace( $image_url, $new_url, $content );
```
pfefferle:
```suggestion
				$content                    = \str_replace( $image_url, $new_url, $content );
```

includes/class-attachments.php
```diff
@@ -0,0 +1,417 @@
...
+	 * Normalize an ActivityPub attachment object to a standard format.
+	 *
+	 * @param mixed $attachment The attachment data (array or object).
+	 *
+	 * @return array|false Normalized attachment data or false on failure.
+	 */
+	private static function normalize_attachment( $attachment ) {
+		// Convert object to array if needed.
+		if ( is_object( $attachment ) ) {
+			$attachment = get_object_vars( $attachment );
```
pfefferle:
```suggestion
			$attachment = \get_object_vars( $attachment );
```

includes/class-attachments.php
```diff
@@ -0,0 +1,417 @@
...
+	/**
+	 * Normalize an ActivityPub attachment object to a standard format.
+	 *
+	 * @param mixed $attachment The attachment data (array or object).
+	 *
+	 * @return array|false Normalized attachment data or false on failure.
+	 */
+	private static function normalize_attachment( $attachment ) {
+		// Convert object to array if needed.
+		if ( is_object( $attachment ) ) {
```
pfefferle:
```suggestion
		if ( \is_object( $attachment ) ) {
```

includes/class-attachments.php
```diff
@@ -0,0 +1,417 @@
...
+		// Add alt text for images.
+		if ( ! empty( $attachment_data['name'] ) ) {
+			$mime_type = $attachment_data['mediaType'] ?? '';
+			if ( 'image' === strtok( $mime_type, '/' ) ) {
+				$post_data['meta_input']['_wp_attachment_image_alt'] = $attachment_data['name'];
+			}
+		}
+
+		// Flag to filter out from Media Library.
+		if ( Posts::POST_TYPE === get_post_type( $post_id ) ) {
```
pfefferle:
```suggestion
		if ( Posts::POST_TYPE === \get_post_type( $post_id ) ) {
```

includes/class-attachments.php
obenland: @pfefferle I think it would be helpful for me to understand which functions we prefix and which ones we don't. It's gotten a bit inconsistent recently. Based on what I've seen I assumed it's WordPress-defined functions but not PHP core functions. Is that not the case?

pfefferle: I always try to prefix both (when possible) (happens that I miss(ed) some).

pfefferle: https://github.com/search?q=repo%3AAutomattic%2Fwordpress-activitypub%20%5Cstr_replace&type=code

pfefferle: https://github.com/search?q=repo%3AAutomattic%2Fwordpress-activitypub+%5Ctrim&type=code

includes/class-attachments.php
obenland: Good question! This was taken from the Mastodon importer, that has its posts converted into regular WP posts.

~~Speaking of Mastodon imports, for them we do want their images to show up in the Media Library, so I'll have to update that.~~ It already does that. Phew.

pfefferle: Good point! yes, I would like to see the imported images.

includes/class-attachments.php
obenland: I went with `import()`, I hope that's good. `attach` is kind of implied from there.

includes/class-attachments.php
Jiwoon-Kim: I think having a dedicated media library for remote objects would be a good approach, because it’s actually very common for WordPress installations to include virtual folder support through plugins like FileBird (and in some cases, even implement user-specific folders).

Of course, we’ll have to see how it turns out in practice, but there’s also a possibility that the core media library structure itself might change in the future — such discussions are already taking place.

Anyway, if you insert a hotlinked image into an Image block and then include it within a Gallery block, the gallery can indeed display the hotlinked image properly.

Jiwoon-Kim: https://github.com/Automattic/wordpress-activitypub/issues/2146

tests/phpunit/tests/includes/class-test-attachments.php
```diff
@@ -0,0 +1,617 @@
...
+	}
+
+	/**
+	 * Test inline image processing without attachments.
+	 *
+	 * @covers ::import_inline_images
+	 */
+	public function test_process_inline_images_only() {
+		// Create a post with inline images.
+		$post_content = '<p>Check out this image: <img src="https://example.com/image1.jpg" alt="Test image"> and this one <img src="https://example.com/image2.png"  alt=""/></p>';
```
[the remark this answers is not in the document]
obenland: Fixed in [cdc0c43](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/cdc0c43661fb029517056112da890ec9cffaf8e1).

includes/class-attachments.php
```diff
@@ -0,0 +1,417 @@
...
+		// Don't filter if we're querying attachments for an ap_post.
+		if ( $query->get( 'post_parent' ) ) {
+			$parent_post = \get_post( $query->get( 'post_parent' ) );
+			if ( $parent_post && Posts::POST_TYPE === $parent_post->post_type ) {
+				return;
+			}
+		}
+
+		// Check if the query is already explicitly looking for _activitypub_import.
+		$meta_query = $query->get( 'meta_query' ) ?: array(); // phpcs:ignore Universal.Operators.DisallowShortTernary
```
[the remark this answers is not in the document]
obenland: I don't think this is going to work, unless we'd pass `null` as a default value. I guess we could pass an empty array as the default?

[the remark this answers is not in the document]
obenland: Updated in [3da9b0a](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/3da9b0ab54012d600141ea47e13724477387eca7).

Menrath:
I haven't reviewed that, but just by reading the title and desc, does this also target https://github.com/Automattic/wordpress-activitypub/issues/1914 ?

Jiwoon-Kim:
@Menrath Personally, I think it’s better to handle post-type objects like `Article` and `Note` separately from media-type objects like `Image`, `Video`, and `Audio`.
Since the `uploadMedia` endpoint is already defined and object retrieval is supported,
I believe we could implement a **local user `uploadMedia` endpoint** on the frontend through extensions of the **Media Library** and **Attachment Page** features.

* [https://www.w3.org/TR/activitypub/#uploading-media](https://www.w3.org/TR/activitypub/#uploading-media)
* [https://www.w3.org/wiki/SocialCG/ActivityPub/MediaUpload#uploadMedia](https://www.w3.org/wiki/SocialCG/ActivityPub/MediaUpload#uploadMedia)
* [https://www.w3.org/TR/activitypub/#retrieving-objects](https://www.w3.org/TR/activitypub/#retrieving-objects)

However, I’m not entirely sure whether it’s truly necessary to parse the `attachment` property from a remote actor’s `Note` or `Article` and store it in the local media library.
For example, even if a WordPress hosting plan provides an unlimited database, it might only offer **1 GB of storage**.
If the system automatically caches all external media from remote actors, the local site’s storage could quickly fill up with spam, effectively preventing any further media uploads.

That’s why I suggested that **direct file links** should be treated as **hotlinks**,
and that we should only record them in the database (as something like `remote_attachment_page`)
when the actor has an `uploadMedia` endpoint and the object is a genuine media object.

In any case, implementing **remote media local caching** now could still be useful later,
since it would provide reusable infrastructure when implementing **selective local caching** in the future.
So while I don’t oppose the current PR, I can’t say I fully agree with its direction.

---

I wanted to suggest that WordPress could independently implement a **`sharedMedia` endpoint** to enable the reuse of remote media objects.

obenland:
> does this also target https://github.com/Automattic/wordpress-activitypub/issues/1914 ?

@Menrath Probably not this one specifically, but mayve a combination of #2311 and #2363

obenland:
@pfefferle This should be ready for another look when you get a chance

Jiwoon-Kim:
I discovered something quite shocking: **you should never use an `Image` object directly inside the `attachment` property of a WordPress post.**

pfefferle:
Why do you use the word "shocking" in this regards! I can't see why this is shocking!

pfefferle:
@Jiwoon-Kim can you please stop using clickbaity titles/wordings like that!

Jiwoon-Kim:
I’m sorry about that. I had just finished reviewing all the related object types, properties, and endpoints, and the conclusion I reached was honestly quite shocking to me personally. I wrote the post before I had fully calmed down from that excitement.

pfefferle:
If you think there is something wrong with the implementation, please file a bug! Explain the issue, be precise and link to the relevant code/documentation!

pfefferle:
@obenland What about hooking the Attachment importer into `do_action( "save_post_{$post->post_type}", $post_id, $post, $update );` instead of a direct calling in the `create` and `update` functions?

Not every user has enough storage or maybe uses a different form of caching, so it might be nice to be able to unhook the caching mechanism!?

Jiwoon-Kim:
I will make my final statement regarding this PR.
An image file **must always be wrapped within an Image object**.
However, if the Image object does not contain an `id`, there is no way for WordPress to assign an attachment ID to it.
Therefore, **if the Image object itself cannot be cached, the image file must not be cached either**.

I believe the **media upload mechanism should be perfectly symmetrical** with the remote media object/file caching process.
To remain consistent with WordPress’s native media upload workflow, this approach should be enforced.

> In particular, servers MUST append an `id` property to the object, and SHOULD include the uploaded and/or processed file paths in the object's `url` property.
> [https://www.w3.org/wiki/SocialCG/ActivityPub/MediaUpload#uploadMedia](https://www.w3.org/wiki/SocialCG/ActivityPub/MediaUpload#uploadMedia)

pfefferle:
@Jiwoon-Kim as I already tried to explain to you multiple times: this is not about "**uploading**" data! we are working on S2S communication, not on C2S!

this is a simple local caching of remote images that are part of a post (`Note` or `Article`), the exact same way, any other platform (including Mastodon) is doing it!

Jiwoon-Kim:
@pfefferle 
The PR in question is as follows:

> Remote URL support: Downloads and imports media from the ActivityPub network using `media_handle_sideload()`
> Verify images are saved to the WordPress media library with correct metadata

What I mean is that this is **not an upload**, but rather a **download**.
And since the images are saved to the **WordPress Media Library** along with their metadata, that means an **attachment page** is being generated.

Did I misunderstand something here?

My point is: since servers MUST append an id when uploading media, the same requirement should apply when downloading media. Without an id, WordPress cannot assign or map an attachment ID for caching.

pfefferle:
> Did I misunderstand something here?

The Attachment is only used as a container to be able to better handle the image. It is easier to "connect" the image to a post and to update/delete/... it when it is stored as Attachment.

pfefferle (changes requested):
@obenland the imported attachment will be federated (added to the outbox) if you have the "Media" Post-Type enabled in the settings.

We have to add an exception for that in the Transformer and Scheduler!

See https://github.com/Automattic/wordpress-activitypub/pull/2372

Jiwoon-Kim:
Exactly. That’s why I’m saying that **remote image objects should be cached (both the file and the object itself) into the local media library only when there is an intention to reuse that object**.

And in that case, the object **must include an `id`**, because the WordPress `attachment` post type **represents the media object itself** — not just the file.

pfefferle:
@Jiwoon-Kim that is a valid point, but the ID does not even help here, because we can't federate items with the id of another server!

@obenland this really speaks against storing remote data as local Attachments!

Jiwoon-Kim:
Hmm, I agree that at this moment we can’t directly federate items using a remote server’s `id`.
However, we should still consider specific cases like **cloning or importing emojis from other servers**, where reusing a remote object’s metadata (including its `id`) becomes meaningful and even necessary.

For example:

Cloning emoji from other servers

- https://fedi.tips/using-custom-emoji-on-your-mastodon-server/
- https://docs.akkoma.dev/stable/configuration/custom_emoji/#stealing-emoji
- https://misskey-hub.net/en/docs/for-admin/features/managing-emojis/#copying-emoji-from-another-instance


In such scenarios, maintaining the original `id` (or mapping it to a local equivalent) could improve **interoperability and reference consistency** across federated servers.

pfefferle:
@Jiwoon-Kim let's keep the focus on image caching in this PR please!

Jiwoon-Kim:
I think in the case of WordPress, it would make more sense to first implement an uploadMedia endpoint for the C2S media library upload.
That way, we can have a more concrete discussion about S2S image caching later on.
Ultimately, it seems the core issue comes down to uploading and downloading.

pfefferle:
@Jiwoon-Kim we're stuck in a loop! we already have federated media based on the classic inbox/outbox pattern:

* https://github.com/Automattic/wordpress-activitypub/blob/c9950fb3238be1e614bb7d7eba039ed241170293/includes/scheduler/class-post.php#L25
* https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/transformer/class-attachment.php

if we can/should improve that, let's discuss it in a different thread.

the `uploadMedia` endpoint is a C2S API that needs OAuth authorization, so it can't be used by S2S without any modification! we are not planning to do any WordPress specialties (yet), because WordPress is still the underdog in the fediverse. we are behind most platforms when it comes to the reader experience, so we have to first catch up before we think about innovating the WordPress to WordPress communication.

the plan of this PR is, to have a **simple** image caching, nothing more.

Jiwoon-Kim:
Would it be possible to **disable the generation of additional image sizes** (such as 150, 300, 1024, and scaled versions) when remote image files are cached?

---

**P.S.**
I now understand that the `uploadMedia` endpoint belongs to the **C2S layer** and cannot be used within the **S2S** context.
However, I still believe that implementing an `uploadMedia`-like mechanism is important, because **the `attachment` property of a federated post should ideally include the WordPress `attachment` post type’s ID**.

Of course, I understand that there are currently **no plans for WordPress-specific features**.
That said, I’m also considering the possibility of a **media library pull model via GET requests**, as an alternative to the traditional **outbox push model using POST requests** — essentially separating the **C2S `uploadMedia` endpoint** from a potential **S2S `mediaLib` endpoint** (for WordPress or other CMS specialties).

pfefferle:
> Would it be possible to disable the generation of additional image sizes (such as 150, 300, 1024, and scaled versions) when remote image files are cached?

This is already the case.

Jiwoon-Kim:
https://github.com/Automattic/wordpress-activitypub/discussions/2377#discussioncomment-14796297

includes/class-attachments.php
```diff
@@ -0,0 +1,672 @@
...
+		}
+
+		// Get mime type.
+		$mime_type = $attachment_data['mediaType'] ?? '';
+		if ( empty( $mime_type ) && \function_exists( 'mime_content_type' ) ) {
+			$mime_type = mime_content_type( $filepath );
+		}
+
+		return array(
+			'url'       => $base_url . '/' . $filename,
```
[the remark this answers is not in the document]
obenland: Fixed in [4bfb8e7](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/4bfb8e7a366fe0ab48d85dfa71d69cf179585478).

.github/changelog/2314-from-description
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: added
```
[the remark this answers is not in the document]
obenland: Added is fine.

tests/phpunit/tests/includes/class-test-attachments.php
```diff
@@ -0,0 +1,617 @@
...
+
+	/**
+	 * Set up each test.
+	 */
+	public function set_up() {
+		parent::set_up();
+
+		// Mock HTTP requests only for remote attachment tests.
+		\add_filter( 'pre_http_request', array( $this, 'mock_download_url' ), 10, 3 );
+		\add_filter( 'wp_delete_file', '__return_empty_string' ); // Prevent actual file deletion during tests.
```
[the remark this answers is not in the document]
obenland: It's an empty check, which an empty string will pass.

includes/class-attachments.php
```diff
@@ -0,0 +1,667 @@
...
+		 * @param array  $files  Array of file data arrays.
+		 */
+		$custom_markup = \apply_filters( 'activitypub_files_media_markup', '', $files );
+
+		if ( ! empty( $custom_markup ) ) {
+			return $custom_markup;
+		}
+
+		// Default to block markup.
+		$type = strtok( $files[0]['mime_type'], '/' );
```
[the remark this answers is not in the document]
obenland: Fixed in [e8444d9](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/e8444d9ceae594b8abf65791747ddc6429b626c0).

.github/changelog/2314-from-description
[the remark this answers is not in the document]
obenland: This is fine.

includes/collection/class-posts.php
```diff
@@ -52,6 +53,11 @@ public static function add( $activity, $user_id ) {
 
 		self::add_taxonomies( $post_id, $activity_object );
 
+		// Process attachments if present.
+		if ( ! empty( $activity_object['attachment'] ) ) {
+			Attachments::import_files( $activity_object['attachment'], $post_id );
```
pfefferle: What about hooking the Attachment importer into do_action( "save_post_{$post->post_type}", $post_id, $post, $update ); instead of a direct calling in the create and update functions?

https://github.com/Automattic/wordpress-activitypub/pull/2314#issuecomment-3450091373

includes/class-attachments.php
```diff
@@ -0,0 +1,665 @@
...
+	/**
+	 * Import attachments from an ActivityPub object and attach them to a post.
+	 *
+	 * @param array $attachments Array of ActivityPub attachment objects.
+	 * @param int   $post_id     The post ID to attach files to.
+	 * @param int   $author_id   Optional. User ID to set as attachment author. Default 0.
+	 *
+	 * @return array Array of attachment IDs.
+	 */
+	public static function import( $attachments, $post_id, $author_id = 0 ) {
```
pfefferle, line 66:
```suggestion
	public static function import_for_post( $attachments, $post_id, $author_id = 0 ) {
```

to be a bit more specific and to follow the `delete_attachments_with_post` naming!?

includes/class-attachments.php
```diff
@@ -0,0 +1,665 @@
...
+		);
+	}
+
+	/**
+	 * Append media to post content.
+	 *
+	 * @param int   $post_id        The post ID.
+	 * @param int[] $attachment_ids Array of attachment IDs.
+	 */
+	private static function append_media_to_content( $post_id, $attachment_ids ) {
```
pfefferle:
```suggestion
	private static function append_media_to_post_content( $post_id, $attachment_ids ) {
```

includes/class-attachments.php
```diff
@@ -0,0 +1,665 @@
...
+	 * @param int     $post_id The post ID.
+	 * @param array[] $files {
+	 *     Array of file data arrays.
+	 *
+	 *     @type string $url       Full URL to the file.
+	 *     @type string $mime_type MIME type of the file.
+	 *     @type string $alt       Alt text for the file.
+	 * }
+	 */
+	private static function append_files_to_content( $post_id, $files ) {
```
pfefferle:
```suggestion
	private static function append_files_to_post_content( $post_id, $files ) {
```

includes/class-attachments.php
obenland: If `delete_attachments_with_post` is the inspiration, I think I'd rather change that to `delete_attachments`. "for post" or "with post" feel redundant

includes/class-attachments.php
obenland: Fixed in [0876ad0](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/0876ad0406f2df89a49e90c168419aaefa3da7c6).

includes/class-attachments.php
obenland: Fixed in [0876ad0](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/0876ad0406f2df89a49e90c168419aaefa3da7c6).

includes/class-attachments.php
obenland: Fixed in [0876ad0](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/0876ad0406f2df89a49e90c168419aaefa3da7c6). I ended up just re-using `delete_directory()`.

pfefferle: That was not the only reason, I thought it might be nice to have all post specific functions prefixed because we want to (at least) also have an attachment importer for comments.

includes/collection/class-posts.php
obenland: Could we do this in a follow up?

It's not quite clear to me how we'd get the object's attachments array from a `save_post` callback, so there might be more to it than just shuffling some function calls around.

pfefferle: It could be a follow up.

The idea is to simply have the save_post/update_post as trigger, to then check the attachments instead of having it called in the collection.

Jiwoon-Kim:
If the **`icon`** property of an [[Actor Object](https://www.w3.org/TR/activitypub/#actor-objects)](https://www.w3.org/TR/activitypub/#actor-objects) supports **ImageObject** and arrays, then it could work similarly to Facebook profiles — where each user can leave comments with their profile image attached.

Now that I think about it, it’s quite strange that the **avatar (`icon`) source** for a WordPress Actor is **Gravatar**. That’s an external source — and as far as I know, external sources aren’t supposed to be used in federation for **trust and reliability reasons**.

If that’s acceptable, then technically there’s no reason why **hotlinked images** included in a post couldn’t also be federated, right?

Also, I just got curious — when deleting a user in WordPress, there’s an option to **reassign all content to another user**.
I haven’t tested this, but what happens to the **outbox items or Activity objects** in that case?

I just wanted to leave a comment here since the recent PRs have been focused on **remote actor avatar handling** and **ActivityPub media processing**, and I’ve also been dealing with some **account-related database issues** on WordPress.com lately.

---

By the way, I just saw this PR: [[mastodon/mastodon#36322](https://github.com/mastodon/mastodon/pull/36322)](https://github.com/mastodon/mastodon/pull/36322).
It seems Mastodon now allows converting a **Note** into an **Article** via an **Update activity**, doesn’t it?

includes/class-attachments.php
obenland: Updated in [6820a9a](https://github.com/Automattic/wordpress-activitypub/pull/2314/commits/6820a9ab9b4afb9d37f8a82e86b37b47a1ccf523). It now has a public `import_post_files()` method in preparation for a future `import_comment_files()`. Under the hood they'll use the same private operations methods.

pfefferle:
@Jiwoon-Kim this is not directly related to this PR, can you please start this as discussion instead, to keep the focus on the code changes.

obenland:
@pfefferle This should be ready for another review

includes/class-attachments.php
```diff
@@ -0,0 +1,729 @@
...
+		$activitypub_dir = self::get_storage_paths( $post_id, 'post' )['basedir'];
+
+		if ( $wp_filesystem->is_dir( $activitypub_dir ) ) {
+			$wp_filesystem->delete( $activitypub_dir, true );
+		}
+	}
+
+	/**
+	 * Import attachments from an ActivityPub object and attach them to a post.
+	 *
```
pfefferle, line 59: I would at least mention that this is used by the importer!?

includes/collection/class-posts.php
```diff
@@ -52,6 +53,11 @@ public static function add( $activity, $user_id ) {
 
 		self::add_taxonomies( $post_id, $activity_object );
 
+		// Process attachments if present.
+		if ( ! empty( $activity_object['attachment'] ) ) {
+			Attachments::import_post_files( $activity_object['attachment'], $post_id );
```
pfefferle, line 58: do we need to provide an option to unhook local storage?

pfefferle (approved):

Jiwoon-Kim:
Conditional Media Handling Architecture: Differentiating Upload Paths and Managing Non-Attached ActivityStreams Objects
https://github.com/Automattic/wordpress-activitypub/discussions/2382#discussioncomment-14814249

includes/class-attachments.php
obenland: Could you tell me more about that? We don't usually document callers, what makes this different?

pfefferle: I always stumble upon that function, because it still has an unprefixed/unsuffixed name `import` but it is very specific to the post import used by the mastodon import. I think it would be nice to give a hint about these two differences and when to use which function.

includes/collection/class-posts.php
obenland: Hm, that's a good question. Maybe it should be one that allows disabling the creation of `ap_post`s entirely?
I'd probably consider sideloading attachments part of the core functionality of the "save incoming posts" feature.

pfefferle: I think it makes no sense to disable the creation of ap_posts entirely, because this is literally not using the follow functionality. But I think it is a valid usecase for users with limited webspace, to not store the images locally but to stick with hotlinking.

(that is why I proposed hooking into save/update post hooks btw.)

PR #2327, opened 2025-10-17
Refactor actor lookup order for HTTP URIs

pfefferle:
Moved the check for locally stored remote actors to occur after blog actor detection in class-actors.php. Updated class-query.php to handle blog actor ID for home/front page URLs before falling back to author ID lookup. This improves accuracy in actor resolution for blog and remote actors.

## Proposed changes:
- Remove `Remote_Actors::get_by_uri` lookup because it is not functional and slows down lookups.
- In query resolution, map home/front page URLs to BLOG_USER_ID before falling back to author ID resolution.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Improved actor resolution by prioritizing blog actor detection before remote actor checks and refining home page URL handling.

</details>

### DIFF
--- a/includes/collection/class-actors.php
+++ b/includes/collection/class-actors.php
@@ -259,13 +259,6 @@ class Actors {
 					return self::BLOG_USER_ID;
 				}
 
-				// Check locally stored remote Actor.
-				$post = Remote_Actors::get_by_uri( $uri );
-
-				if ( ! \is_wp_error( $post ) ) {
-					return $post->ID;
-				}
-
 				return new \WP_Error(
 					'activitypub_no_user_found',
 					\__( 'Actor not found', 'activitypub' ),
--- a/includes/collection/class-actors.php
+++ b/includes/collection/class-actors.php
@@ -222,13 +222,6 @@ class Actors {
 			// Check for http(s) URIs.
 			case 'http':
 			case 'https':
-				// Check locally stored remote Actor.
-				$post = Remote_Actors::get_by_uri( $uri );
-
-				if ( ! \is_wp_error( $post ) ) {
-					return $post->ID;
-				}
-
 				// Check for http(s)://blog.example.com/@username.
 				$resource_path = \wp_parse_url( $uri, PHP_URL_PATH );
 
@@ -264,6 +257,13 @@ class Actors {
 					normalize_url( home_url() ) === $normalized_uri
 				) {
 					return self::BLOG_USER_ID;
+				}
+
+				// Check locally stored remote Actor.
+				$post = Remote_Actors::get_by_uri( $uri );
+
+				if ( ! \is_wp_error( $post ) ) {
+					return $post->ID;
 				}
 
 				return new \WP_Error(
--- a/tests/phpunit/tests/includes/class-test-query.php
+++ b/tests/phpunit/tests/includes/class-test-query.php
@@ -525,7 +525,7 @@ class Test_Query extends \WP_UnitTestCase {
 	 * @covers ::get_activitypub_object
 	 */
 	public function test_home_page_actor_mode() {
-		\add_option( 'activitypub_actor_mode', ACTIVITYPUB_ACTOR_MODE );
+		\update_option( 'activitypub_actor_mode', ACTIVITYPUB_ACTOR_MODE );
 
 		$actor_queries = array();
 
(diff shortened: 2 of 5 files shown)


tests/phpunit/tests/includes/class-test-query.php
```diff
@@ -518,4 +518,77 @@ public function test_maybe_get_stamp_invalid_author() {
...
+		$object = Query::get_instance()->get_activitypub_object();
+
+		remove_filter( 'query', $query_filter );
+
+		$message = 'Should not query Remote_Actors table for home page.';
+		if ( ! empty( $actor_queries ) ) {
+			$message .= ' Found queries: ' . wp_json_encode( $actor_queries );
+		}
+
+		$this->assertNull( $object, 'Home page should return an object' );
```
obenland: One of these feels off. Should it be null or an object? If an object, do we expect a specific object that we can test for?

tests/phpunit/tests/includes/class-test-query.php
```diff
@@ -518,4 +518,77 @@ public function test_maybe_get_stamp_invalid_author() {
...
+		$object = Query::get_instance()->get_activitypub_object();
+
+		\remove_filter( 'query', $query_filter );
+
+		$message = 'Should not query Remote_Actors table for home page.';
+		if ( ! empty( $actor_queries ) ) {
+			$message .= ' Found queries: ' . wp_json_encode( $actor_queries );
+		}
+
+		$this->assertNotNull( $object, 'Home page should return an object' );
```
obenland, line 589: Same as above.

includes/collection/class-actors.php
```diff
@@ -222,13 +222,6 @@ public static function get_id_by_resource( $uri ) {
 			// Check for http(s) URIs.
 			case 'http':
 			case 'https':
-				// Check locally stored remote Actor.
-				$post = Remote_Actors::get_by_uri( $uri );
-
-				if ( ! \is_wp_error( $post ) ) {
-					return $post->ID;
-				}
-
```
obenland, line 231: The PR description says "Moved the check for locally stored remote actors to occur after blog actor detection" but that doesn't seem to be the case?

Is there a reason `maybe_get_virtual_object()` calls `Actors::get_by_various()` at all? could we not just return either the Blog or Application object based on whatever query parameters we need?

tests/phpunit/tests/includes/class-test-query.php
pfefferle: It should be null, because the Blog-User is disabled in that case.

includes/collection/class-actors.php
pfefferle: > The PR description says "Moved the check for locally stored remote actors to occur after blog actor detection" but that doesn't seem to be the case?

Yes, my Copilot budget was expired, so it didn't catch up :) (I update the code).

The fallback to `Remote_Actors::get_by_uri( $uri )` is not functional, because it returns a Post-ID and we only check for User-IDs, so I removed it.

pfefferle: > Is there a reason maybe_get_virtual_object() calls Actors::get_by_various() at all? could we not just return either the Blog or Application object based on whatever query parameters we need?

The Blog-User and the Application users have several Identifiers, so I would have to clone the complete code of `Actors::get_by_various()`.

For example:

Blog:

* `http://example.org/`
* `http://example.org/@example.org`
* `http://example.org/?author=0`

Application:

* `http://example.org/wp-json/activitypub/1.0/application`
* `http://example.org/@application`
* `http://example.org/?author=-1`

And because `is_home` is not very reliable (see Slack discussion), I decided to [remove it](https://github.com/Automattic/wordpress-activitypub/pull/2327/commits/6b31c175df7d75876396feed1d009f187978f303) and optimize the lookup instead. Happy for any advice!

tests/phpunit/tests/includes/class-test-query.php
pfefferle: I updated the description.

tests/phpunit/tests/includes/class-test-query.php
pfefferle: It does not really matter what object it is, I want to simply proof that there is no DB lookup in this case.

obenland (approved):

includes/collection/class-actors.php
obenland: No I think that makes sense. Thanks!

PR #2328, opened 2025-10-17
Add fallback to trigger create on update failure

pfefferle:
Update handling now triggers the 'activitypub_inbox_create' action if an update fails due to a missing post or comment. This ensures that create logic is invoked for objects not found during update requests. Includes a new test to verify the fallback behavior.

## Proposed changes:

* The ActivityPub update handler was modified to add a fallback mechanism.
* If an "Update" activity fails (e.g., the actor cannot be found or updated), the handler now triggers the activitypub_inbox_create action as a fallback.
* This ensures that "Create" logic is executed when "Update" cannot be processed, improving robustness and compatibility.
* New tests were added to verify that the fallback (activitypub_inbox_create) is triggered correctly when an update fails.
* The tests simulate update failures and check that the fallback action is called.
* Indentation and formatting were improved in the test file to comply with coding standards.

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added fallback to trigger create handling when updates fail for missing posts or comments, ensuring objects are properly created.

</details>

### DIFF
--- a/.github/changelog/2328-from-description
+++ b/.github/changelog/2328-from-description
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added fallback to trigger create handling when updates fail for missing posts or comments, ensuring objects are properly created.
--- a/includes/collection/class-posts.php
+++ b/includes/collection/class-posts.php
@@ -84,8 +84,8 @@ class Posts {
 
 		if ( ! $post_id ) {
 			return new \WP_Error(
-				'activitypub_object_not_found',
-				\__( 'Object not found', 'activitypub' ),
+				'activitypub_post_not_found',
+				\__( 'Post not found', 'activitypub' ),
 				array( 'status' => 404 )
 			);
 		}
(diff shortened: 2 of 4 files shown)


obenland (approved):

Jiwoon-Kim:
In this context, when you say “create,” you’re referring to backfilling an object from a remote server, right?
Not the actual Create activity itself, but the process of fetching and reconstructing the object from its remote source.

pfefferle:
When the code is not able to `Update` an object, it tries to handle it as a `Create`... in `SQL` you would call it an `UPSERT`.

Jiwoon-Kim:
Thank you for the explanation.
When I first created an interaction between my WordPress, Mastodon, and Misskey accounts, I once edited a post that had been written **before** federation started. So in that case, it must have received an **Update** activity **without** a preceding **Create**.

However, the object still appeared as an updated entity.
That’s why I assumed that—since the **Update activity** itself already contains the full **object content**—if the system checks the cached remote object list and finds that no object exists for that URI, it automatically creates (backfills) an object cache while saving the state as “update.”

PR #2330, opened 2025-10-17
Add support for quote reactions and improve quote detection

pfefferle:
Introduces a new 'quote' comment type, updates interaction handling to extract and process quote links from content, and enhances reply detection to recognize quote-inline patterns. Includes comprehensive tests for quote extraction and reply identification.

Screenshot 2025-11-05 at 15 24 55

supersedes #2303 

## Proposed changes:
* Adds a new 'quote' comment type with appropriate metadata and translations
* Implements quote link extraction from HTML content with regex pattern matching
* Enhances reply detection to recognize quote-inline patterns in addition to standard inReplyTo properties

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Added support for quote comments, improving detection and handling of quoted replies and links in post interactions.

</details>


Jiwoon-Kim:
What do you think of this idea?

When a QuoteRequest activity is received, it’s stored in the `wp_comment` table with `comment_type=QuoteRequest` and `comment_status=moderated`.
When the comment is approved, an `Accept` activity is triggered.
This method is also used to display the quote count.

In the case of Misskey quotes, when the `quoteUri` property exists, it indicates that the post has been quoted — however, it isn’t reflected in the quote count.

[https://docs.akkoma.dev/stable/development/ap_extensions/#quote-posts](https://docs.akkoma.dev/stable/development/ap_extensions/#quote-posts)

> Old Misskey only understood and modern Misskey still prefers the _misskey_quote property for this. Similar some other older software used quoteUrl or quoteURL.
> All current implementations with quote support understand quoteUri.

~To quote a Misskey note in WordPress, the `quoteUri` property must be supported.~
Since FEP-044f has been implemented, notifications can be received when detected, but quoting is not possible.

pfefferle:
https://mastodon.social/@pfefferle/115405195331557281

pfefferle:
Close it for now and see how the Reader might help.

includes/class-comment.php
```diff
@@ -636,7 +636,7 @@ public static function register_comment_types() {
 			array(
 				'label'          => __( 'Reposts', 'activitypub' ),
 				'singular'       => __( 'Repost', 'activitypub' ),
-				'description'    => __( 'A repost on the indieweb is a post that is purely a 100% re-publication of another (typically someone else\'s) post.', 'activitypub' ),
+				'description'    => __( 'A repost (or Announce) is when a post appears in the timeline because someone else shared it, while still showing the original author as the source.', 'activitypub' ),
```
obenland: Are these translations used anywhere?

includes/functions.php
```diff
@@ -618,7 +618,20 @@ function is_activity_public( $data ) {
...
-	return ! empty( $data['object']['inReplyTo'] );
+	if ( ! empty( $data['object']['inReplyTo'] ) ) {
+		return true;
+	}
+
+	if ( empty( $data['object']['content'] ) ) {
+		return false;
+	}
+
+	// very simple check for quote content.
```
```suggestion
	// Very simple check for quote content.
```

includes/collection/class-interactions.php
```diff
@@ -34,10 +34,19 @@ class Interactions {
 	public static function add_comment( $activity ) {
 		$comment_data = self::activity_to_comment( $activity );
 
-		if ( ! $comment_data || ! isset( $activity['object']['inReplyTo'] ) ) {
+		if ( ! $comment_data ) {
 			return false;
 		}
 
+		if ( empty( $activity['object']['inReplyTo'] ) ) {
+			$activity = self::extract_quote_link( $activity );
```
obenland: Let's return the URL of the quoted post here and pass it directly to `\url_to_postid()`, since that's what we're after, the comment_post_ID.

obenland:
Maybe we could use incoming `QuoteRequest`s to create these quote comments.

pfefferle:
@obenland forgot to add the new attributes, that's why they were not added to the inbox. I changed the code to work with the quote attribute (and fallbacks).

includes/functions.php
```diff
@@ -621,6 +621,22 @@ function is_activity_reply( $data ) {
...
+/**
+ * Check if passed Activity is a quote.
+ *
+ * Checks for quote properties: quote, quoteUrl, quoteUri, or _misskey_quote.
+ *
+ * @param array $data The Activity object as array.
+ *
+ * @return boolean True if a quote, false if not.
+ */
+function is_activity_quote( $data ) {
```
obenland: You'll love this, but I think this should be `is_quote_activity()`. It's not checking whether it's a quote of an activity, but an activity of the type "quote", if that makes sense.

pfefferle: It was meant to read like "**is** the **quote** an **activity**".

obenland (dismissed):

obenland (approved):

PR #2339, opened 2025-10-20
Ensure `support_post_types` option returns an array

### LINKED ISSUE #2332: Foreach argument warning when no supported post-types selected
### Quick summary

When the `activitypub_support_post_types` option is set to have no post types selected, an empty string is returned by `\get_option( 'activitypub_support_post_types', array( 'post' ) )` triggering a php warning: 'Warning: foreach() argument must be of type array|object, string given in /home/sjapags/www/wp-content/plugins/activitypub/includes/class-post-types.php on line 41'.

### Steps to reproduce

1. Go to /wp-admin/options-general.php?page=activitypub&tab=settings
2. Delesect all post types for the 'Supported post types' option
3. Make sure the relevant wp-config constants are defined to output PHP warnings to screen / log file.
4. Check a screen / the log file to observe the PHP warning.

### Site owner impact

### Severity

Minor

No revenue impact

Make sure you have at least one post-type selected.

pfefferle:
Adds a filter to guarantee the 'activitypub_support_post_types' option always returns an array, preventing potential type errors when the option is empty or not set.

Fixes #2332

## Proposed changes:

* Adds a new filter callback support_post_types_ensure_array to validate the option value

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Added a safeguard to ensure the plugin works correctly even when no post types are selected.

</details>


obenland:
For the settings field [we cast the option to an array](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/wp-admin/class-settings-fields.php#L287), could we do the same in `Post_Types` to fix that warning?

pfefferle:
You mean, simply cast the return instead of returning an array when empty? Sure!

obenland:
Apologies, I should have been clearer.

I meant, could we cast the option call to an array [where the warning gets thrown in `Post_Types`](https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/class-post-types.php#L42), like we do in settings fields?

```diff
- foreach ( \get_option( 'activitypub_support_post_types', array( 'post' ) ) as $post_type ) {
+ foreach ( (array) \get_option( 'activitypub_support_post_types', array( 'post' ) ) as $post_type ) {
```

pfefferle:
The problem is, that this is not the only place! And having to add that to every call seemed too fragile to me: https://github.com/search?q=repo%3AAutomattic%2Fwordpress-activitypub%20get_option(%20%27activitypub_support_post_types%27&type=code

We already missed to add it to every call!

pfefferle:
But happy to add it to every call instead...

obenland:
That's helpful, thanks! Want to remove the unnecessary type cast from settings field and the rest cache integration?

obenland (approved):

PR #2341, opened 2025-10-20
Disable Post `Create` and `Update` by default

pfefferle:
Introduces a new 'activitypub_create_posts' option to control whether posts can be created or updated via ActivityPub. Updates Posts::add and Posts::update to require a user ID and store it in post meta. Adjusts handlers and tests to accommodate the new parameter and option.

## Proposed changes:
- Adds a new 'activitypub_create_posts' option that defaults to false, requiring explicit opt-in for post creation/updates
- Updates Posts::add and Posts::update methods to require and store user ID in post meta
- Modifies handlers to check the new option before allowing post operations

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


includes/collection/class-posts.php
```diff
@@ -114,6 +117,11 @@ public static function update( $activity ) {
 			return $post_id;
 		}
 
+		$post_meta = \get_post_meta( $post_id, '_activitypub_user_id', false );
+		if ( \is_array( $post_meta ) && ! \in_array( (string) $user_id, $post_meta, true ) ) {
+			\add_post_meta( $post_id, '_activitypub_user_id', $user_id );
```
obenland, line 122: Is this to determine who to serve that content to? Maybe we should just add all user_ids that follow that actor?

When someone adds a new follow, we should probably add that user_id to that list as well, so they have access to existing posts?

pfefferle: I think we should process every inbox request separately, because we have to get the visibility right. Only because it ends up in the Posts steam of one user does not mean that is meant for every other follower.

pfefferle: I have an idea for the shared inbox that might help with that.

pfefferle: this would require a major update though, because it is a breaking change 😱

pfefferle: Let's not do it as part of this PR, because it brings a new layer of complexity we have to solve in general!

obenland (approved):

PR #2342, opened 2025-10-20
Validate IP before resolving comment author domain

pfefferle:
Added a check to ensure the comment author's IP address is present and valid before attempting to resolve the hostname with gethostbyaddr. This prevents potential errors when the IP is missing or invalid.

Fixes https://wordpress.org/support/topic/php-error-warning-17/

## Proposed changes:
- Added IP validation using `filter_var()` with `FILTER_VALIDATE_IP` before calling `gethostbyaddr()`
- Initialized `$comment_author_domain` variable to an empty string as a fallback
- Wrapped the hostname resolution in a conditional check to ensure it only runs with valid IPs

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Added a safety check to prevent errors when resolving comment author hostnames without a valid IP address.

</details>


includes/class-mailer.php
```diff
@@ -87,7 +87,11 @@ public static function comment_notification_text( $message, $comment_id ) {
 		}
 
 		$post                  = \get_post( $comment->comment_post_ID );
-		$comment_author_domain = \gethostbyaddr( $comment->comment_author_IP );
+		$comment_author_domain = '';
+		// Only attempt to resolve hostname if we have a valid IP address.
```
obenland, line 92:
```suggestion

		// Only attempt to resolve hostname if we have a valid IP address.
```
I'm a big fan of setting defaults and conditionally changing them, and doing that in one block to make it easy to see they belong together!

Unless there's an inline comment, where making it easy to catch takes precedence for me? Let's give comments room to breath

includes/class-mailer.php
```diff
@@ -87,7 +87,11 @@ public static function comment_notification_text( $message, $comment_id ) {
...
+		if ( ! empty( $comment->comment_author_IP ) && \filter_var( $comment->comment_author_IP, FILTER_VALIDATE_IP ) ) {
```
obenland:
```suggestion
		if ( filter_var( $comment->comment_author_IP, FILTER_VALIDATE_IP ) ) {
```
Presumably the `filter_var()` check on its own is enough? Not sure if there's a speed concern where we'd want to avoid calling it when there's an empty value.

obenland (dismissed):
I think this is good to go as-is.

There are two optional, OCD-satisfying suggestions if you feel inclined, but I consider them optional.

obenland (approved):
💖

PR #2349, opened 2025-10-21
Shared-Inbox: Add support for public activity delivery to all users

pfefferle:
Improve `get_local_recipients` to also support Public posts and Follower Collections.

Introduces `Actors::get_all_ids()` to retrieve all local ActivityPub user IDs and updates `Inbox_Controller` to deliver public activities to all such users. Adds `Following::get_user_ids()` for retrieving local followers of a remote actor. Includes new tests for these methods and public activity recipient logic.

## Proposed changes:
- Introduces `Actors::get_all_ids()` to retrieve IDs of all local ActivityPub-capable users
- Adds `Following::get_user_ids()` to get local followers of a remote actor
- Updates `Inbox_Controller::get_local_recipients()` to deliver public activities to all users and handle follower collections

### Other information:

## Testing instructions:

* See unittests

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Improved delivery of public and follower activities by expanding local recipient handling to include all ActivityPub-capable users and follower collections.

</details>


includes/collection/class-actors.php
```diff
@@ -370,38 +370,48 @@ public static function get_collection() {
...
+	 */
+	public static function get_all() {
+		$user_ids = self::get_all_ids();
+
+		$actors = array_map(
+			function ( $user_id ) {
+				return self::get_by_id( $user_id );
+			},
+			$user_ids
+		);
```
```suggestion
		$actors = array_map( array( self::class, 'get_by_id' ), $user_ids );
```

includes/rest/class-inbox-controller.php
```diff
@@ -284,13 +288,28 @@ public function get_item_schema() {
 	 * @return array An array of user IDs who are the recipients of the activity.
 	 */
 	private function get_local_recipients( $activity ) {
+		// Public activity, deliver to all local ActivityPub users.
+		if ( is_activity_public( $activity ) ) {
+			return Actors::get_all_ids();
```
obenland, line 293: I think you tried to explain this to me before, if that's the case please bear with me.

If the activity is public it might not have followers defined as its audience? Is that why it goes to everyone?
Would we run into a situation where User A subscribed to the actor for this activity and now User B gets those posts in their timeline?

includes/rest/class-inbox-controller.php
```diff
@@ -284,13 +288,28 @@ public function get_item_schema() {
...
+
 		$recipients = extract_recipients_from_activity( $activity );
 		$user_ids   = array();
 
 		foreach ( $recipients as $recipient ) {
 
 			if ( ! is_same_domain( $recipient ) ) {
-				continue;
+				$collection = Http::get_remote_object( $recipient );
+				// If it is a remote actor we can skip it.
```
obenland, line 304:
```suggestion

				// If it is a remote actor we can skip it.
```

includes/collection/class-following.php
```diff
@@ -419,6 +419,27 @@ public static function check_status( $user_id, $post_id ) {
...
 	}
 
+	/**
+	 * Get local user IDs following a given remote actor.
+	 *
+	 * @param string $actor_url The actor URL.
+	 *
+	 * @return int[] List of local user IDs following the actor.
+	 */
+	public static function get_user_ids( $actor_url ) {
```
obenland: The name is a bit confusing to me. These are the IDs of local followers for this following, not the user IDs of following?

How would you feel about `get_follower_ids()` to make that a bit clearer?

includes/rest/class-inbox-controller.php
pfefferle: 🤔

good point... maybe we should also use the `get_user_ids` (or how we end up calling the function).

includes/collection/class-following.php
pfefferle: I thought about a proper name a lot and AI was not very helpful, so I am happy with anything!!! 😂

obenland (approved):

includes/rest/class-inbox-controller.php
obenland: @pfefferle Looks like this was never resolved. Could we revisit this? I find it quite confusing to get posts from accounts I'm not following.

PR #2354, opened 2025-10-21
Fix public key retrieval for GoToSocial profiles

obenland:
Fixes https://github.com/Automattic/wordpress-activitypub/discussions/2319

## Summary

Public key retrieval was failing for GoToSocial profiles because GoToSocial uses path-based key URLs (e.g., `/users/username/main-key`) instead of fragment-based URLs (e.g., `/users/username#main-key`) like Mastodon.

The existing code only stripped URL fragments using `strip_fragment_from_url()`, which had no effect on GoToSocial's path-based URLs. This caused the system to attempt fetching the key URL directly, which returns only a partial actor object without the required fields, failing validation in `Remote_Actors::upsert()`.

```json
{
  "@context": [
    "https://w3id.org/security/v1",
    "https://www.w3.org/ns/activitystreams"
  ],
  "id": "https://gotosocial-production-f7fa.up.railway.app/users/obietester",
  "preferredUsername": "obietester",
  "publicKey": {
    "id": "https://gotosocial-production-f7fa.up.railway.app/users/obietester/main-key",
    "owner": "https://gotosocial-production-f7fa.up.railway.app/users/obietester",
    "publicKeyPem": "-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAzcBN0ce7t2EnJTHoxTGW\nI/5ptpoT4L3DztbB5Xd2xDklRhUZDrB1l2p/z9Dct1ZU8KtQNpGe9KAGeUpgFksx\nBCQt2tX0LlpBHTks5ifQlczVo+YYM2hzSJAh9ElWX/Y6rcZRzCs8B6zMAc+NWvgm\nFJa+uvo9HqnLvTDSJwodIDLoZ3i9hjYKnb65xhgRjSBiLr9ljv5lfCSFDEjNJUHL\nW93FWB9RKrC3WA9+tk2c62Ak9MV5/8tSlWLCwlPwfduGWuDhweQ3RLjF+7ALrI9v\nEKuTzFhxXAaFBOE6Zi2gF5ma9zGikCRayllORbYmKC15V808+VIZZ+G0tFrUB6Zl\nZwIDAQAB\n-----END PUBLIC KEY-----\n"
  },
  "type": "Person"
}
```

## Proposed changes:
* Fix public key retrieval for GoToSocial profiles by handling path-based key URLs
* Updated `Remote_Actors::get_public_key()` to strip both fragment-based (`#main-key`) and path-based (`/main-key`) key suffixes
* Added test coverage for GoToSocial-style URL pattern

### Other information:

## Testing instructions:

* Run the existing test suite: `npm run env-test -- --filter=Test_Remote_Actors`
* All 12 tests should pass, including the new test case for GoToSocial-style URLs
* Optionally test with a real GoToSocial instance at https://gotosocial-production-f7fa.up.railway.app/users/obietester/main-key

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fix public key retrieval for GoToSocial profiles with path-based key URLs.

</details>


pfefferle:
If you use `Http::get_remote_object` instead of the function it will work out of the box without the workaround, because there is no inbox verification!

pfefferle:
But that will bypass the local cache!

🤔

includes/collection/class-remote-actors.php
```diff
@@ -535,7 +535,11 @@ public static function normalize_identifier( $actor ) {
 	 * @return resource|\WP_Error The public key resource or WP_Error.
 	 */
 	public static function get_public_key( $key_id ) {
-		$actor = get_remote_metadata_by_actor( \strip_fragment_from_url( $key_id ) );
+		$actor_url = \strip_fragment_from_url( $key_id ); // Strip fragment from URL (Mastodon-style #main-key).
+		$actor_url = \preg_replace( '#/main-key$#', '', $actor_url ); // Also handle GoToSocial-style /main-key path suffix.
+
+		$actor = get_remote_metadata_by_actor( $actor_url );
```
pfefferle:
```suggestion
		$actor_url = \strip_fragment_from_url( $key_id ); // Strip fragment from URL (Mastodon-style #main-key).
		
		// Check local storage.
		$actor = self::get_by_uri( $actor_url );
		
		if ( ! \is_wp_error( $actor ) ) {
			$actor = \json_decode( $actor->post_content, true );
		} else {
			// Get remote object instead.
			$actor = Http::get_remote_object( $actor );
		}
```

pfefferle: Stripping a URL fragment shouldn’t be an issue, since fragments aren’t sent to servers, the result should remain the same.

However, removing paths might work for GoToSocial but could break other implementations, so I’d prefer a more conservative approach.

Alternatively, we could drop the Inbox requirement instead.

pfefferle (changes requested):

includes/collection/class-remote-actors.php
obenland: I don't mind going that route, don't feel super strongly about it. ~~It does result in duplicate requests for GoToSocial actors and actors that are unavailable though.~~

> removing paths might work for GoToSocial but could break other implementations

This should be a pretty narrow use-case, only for public key requests, and the replace is specific to `/main-key`. If other softwares use that format and don't fall back to the full actor object with it removed, it would probably be enough of an edge case to fix individually, IMO.

> we could drop the Inbox requirement instead.

Not sure if that would work as it would update the full actor object with the partial one for GoToSocial profiles.

includes/collection/class-remote-actors.php
```diff
@@ -536,6 +536,11 @@ public static function normalize_identifier( $actor ) {
 	 */
 	public static function get_public_key( $key_id ) {
 		$actor = get_remote_metadata_by_actor( \strip_fragment_from_url( $key_id ) );
```
pfefferle: We should also replace this with the local lookup, because now we try to resolve it from the remote resource, try to store it, and then fail, to then try to resolve it again.

includes/collection/class-remote-actors.php
pfefferle: the `main-key` path can also be the username, so `https://example.com/authors/main-key` would fail in that case.

includes/collection/class-remote-actors.php
obenland: Fixed in [42b187c](https://github.com/Automattic/wordpress-activitypub/pull/2354/commits/42b187c2331cdf46a85486e37fb7bd09eae02b92)

pfefferle (approved):

PR #2357, opened 2025-10-22
Refactor inbox to support deduplicated inbox-recipient storage

pfefferle:
This PR refactors the inbox system to support shared inbox functionality where activities are stored once with multiple recipients as metadata entries, rather than creating duplicate posts per recipient. Key changes include updating meta registration to allow multiple `_activitypub_user_id` values, introducing new shared inbox hooks, and adding recipient management methods.

## Proposed changes:

- Introduces `activitypub_inbox_shared` hooks for efficient multi-recipient processing
- Updates inbox collection to store recipients as separate meta entries
- Adds migration to clean up existing inbox items for fresh shared inbox structure

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Refactored the inbox system to use a shared inbox, storing activities once with multiple recipients for improved efficiency and reduced duplication.

</details>


Jiwoon-Kim:
Since this idea is more of a tentative hypothesis, I preferred to discuss it in the comments rather than formally submitting it as an issue.

Summary of the idea:

- Treat activities as objects and store them in the wp_post table.
- Implement Inbox and Outbox using taxonomies.
- Question: Should remote activities and local activities be separated into different CPTs (custom post types)?
- Reason: Remote activities create remote objects, whereas local activities are generated from local objects.

pfefferle:
Inbox/Outbox are already CPTs: https://github.com/Automattic/wordpress-activitypub/blob/trunk/includes/class-post-types.php

Jiwoon-Kim:
Yes, since the Inbox and Outbox are implemented as CPTs, I suspected that for local-to-local activities there could be a conflict between the URIs of the activity in the sender's Outbox and the receiver's Inbox.

pfefferle:
Shouldn't be an issue!

Jiwoon-Kim:
Yes, I’m glad to hear it’s not an issue. Intuitively, I imagined that a single activity would exist as a duplicate object in both the Inbox and the Outbox. I don’t fully understand why that would be incorrect or why it’s not a problem, but as long as it works properly in practice, that’s what matters.

---

Exactly, that makes sense. Intuitively, it does feel odd to imagine the same activity “existing twice” — once in the sender’s Outbox and once in the receiver’s Inbox — as if it were a duplicated object.

The reason it’s not actually a problem is that in ActivityPub, collections like Outbox and Inbox are just containers or references to activities. The activity itself is identified by its URI, and each actor’s collection merely points to it. So even if the same activity appears in multiple collections, it’s not considered a duplication in a harmful sense — it’s just the protocol keeping track of where the activity is relevant.

As long as the flow works correctly (sending, receiving, processing activities), the system behaves as expected, and you don’t need to worry about “conflict” between Inbox and Outbox entries.

pfefferle:
I will update the other handlers if that is how we want to proceed.

Jiwoon-Kim:
I’ve been thinking about it, and I believe it’s better to implement the outbox as a custom post type (CPT) and the shared inbox as a taxonomy.

Because the outbox already represents local activity types, if we implement remote activities as a custom post type (CPT) and set the inbox as a shared taxonomy between the two CPTs, it will allow us to receive both local and remote activities seamlessly.

includes/collection/class-inbox.php
```diff
@@ -27,32 +27,72 @@ class Inbox {
...
+	 */
+	const CONTEXT_SHARED_INBOX = 'shared_inbox';
+
 	/**
 	 * Add an activity to the inbox.
 	 *
-	 * @param Activity|\WP_Error $activity The Activity object.
-	 * @param int                $user_id  The id of the local blog-user.
+	 * @param Activity|\WP_Error $activity   The Activity object.
+	 * @param int|array          $recipients The id(s) of the local blog-user(s).
```
obenland, line 48: Let's only document the new type and use `_deprecated_argument()` if `$recipients` is an int and convert it.

```php
if ( is_int( $recipients ) {
	\_deprecated_argument();

	$recipients = array( $recipients );
}
```

includes/collection/class-inbox.php
```diff
@@ -27,32 +27,72 @@ class Inbox {
...
+		// Sanitize recipients.
+		$recipients = \array_map( 'absint', $recipients );
+		$recipients = \array_unique( $recipients );
+		// Filter out null/false but keep 0 (blog user).
+		$recipients = \array_filter(
+			$recipients,
+			function ( $value ) {
+				return null !== $value && false !== $value;
+			}
+		);
```
obenland: Is it possible to still have `null` and `false` values after `absint()`? We might be able to just remove this.

includes/collection/class-inbox.php
```diff
@@ -27,32 +27,72 @@ class Inbox {
...
+		\sort( $recipients );
```
obenland: Does this matter?

includes/collection/class-inbox.php
```diff
@@ -27,32 +27,72 @@ class Inbox {
...
+
+		if ( empty( $recipients ) ) {
+			return new \WP_Error(
+				'activitypub_inbox_no_recipients',
+				\__( 'No valid recipients provided', 'activitypub' ),
```
obenland: Is this a user-facing string? If not, we could avoid translation work by not sending it through `__()`.

includes/collection/class-inbox.php
```diff
@@ -222,4 +269,116 @@ public static function undo( $id ) {
...
+	 * Check if a user is a recipient of an inbox activity.
+	 *
+	 * @param int $post_id The inbox post ID.
+	 * @param int $user_id The user ID to check.
+	 *
+	 * @return bool True if user is a recipient, false otherwise.
+	 */
+	public static function has_recipient( $post_id, $user_id ) {
+		$recipients = self::get_recipients( $post_id );
+		return \in_array( (int) $user_id, $recipients, true );
```
obenland, line 286:
```suggestion

		return \in_array( (int) $user_id, $recipients, true );
```

includes/collection/class-inbox.php
```diff
@@ -222,4 +269,116 @@ public static function undo( $id ) {
...
+		// Check if already a recipient.
+		if ( self::has_recipient( $post_id, $user_id ) ) {
+			return true;
+		}
+
+		// Add new recipient as separate meta entry.
+		$result = \add_post_meta( $post_id, '_activitypub_user_id', $user_id, false );
+
+		if ( $result ) {
+			\clean_post_cache( $post_id );
```
obenland: `add_post_meta()` deletes its own meta cache, is there more that needs updating?

includes/collection/class-inbox.php
```diff
@@ -222,4 +269,116 @@ public static function undo( $id ) {
...
+		// Allow 0 for blog user, but reject negative values.
+		if ( $user_id < 0 ) {
+			return false;
+		}
+
+		// Delete the specific meta entry with this value.
+		$result = \delete_post_meta( $post_id, '_activitypub_user_id', $user_id );
+
+		if ( $result ) {
+			\clean_post_cache( $post_id );
```
obenland: Same as above

includes/collection/class-inbox.php
```diff
@@ -222,4 +269,116 @@ public static function undo( $id ) {
...
+	 * Remove a recipient from an inbox activity.
+	 *
+	 * @param int $post_id The inbox post ID.
+	 * @param int $user_id The user ID to remove.
+	 *
+	 * @return bool True on success, false on failure.
+	 */
+	public static function remove_recipient( $post_id, $user_id ) {
+		$user_id = (int) $user_id;
+		// Allow 0 for blog user, but reject negative values.
```
obenland, line 324:
```suggestion

		// Allow 0 for blog user, but reject negative values.
```

includes/collection/class-inbox.php
```diff
@@ -222,4 +269,116 @@ public static function undo( $id ) {
...
+
+		if ( \is_wp_error( $post ) ) {
+			return $post;
+		}
+
+		// Check if user is a recipient.
+		if ( ! self::has_recipient( $post->ID, $user_id ) ) {
+			return new \WP_Error(
+				'activitypub_inbox_not_recipient',
+				\__( 'User is not a recipient of this activity', 'activitypub' ),
```
obenland: Is this a user-facing string? If not, we could avoid translation work by not sending it through `__()`.

includes/handler/class-inbox.php
```diff
@@ -21,20 +21,48 @@ class Inbox {
...
+		self::handle_inbox_requests( $data, $user_ids, $type, $activity, $context );
+	}
+
+	/**
+	 * Handles "Inbox" requests.
+	 *
+	 * Supports both single user_id (int) and multiple user_ids (array).
+	 *
+	 * @param array              $data         The data array.
+	 * @param int|array          $user_ids     The id of the local blog-user, or array of user IDs.
```
obenland, line 59: Let's only document the new type and use `_deprecated_argument()` if `$user_ids` is an int and convert it.

includes/rest/class-inbox-controller.php
```diff
@@ -168,28 +174,39 @@ public function create_item( $request ) {
...
-				\do_action( 'activitypub_inbox', $data, $user_id, $type, $activity );
-
-				/**
-				 * ActivityPub inbox action for specific activity types.
-				 *
-				 * @param array              $data     The data array.
-				 * @param int                $user_id  The user ID.
-				 * @param Activity|\WP_Error $activity The Activity object.
-				 */
-				\do_action( 'activitypub_inbox_' . $type, $data, $user_id, $activity );
```
obenland, line 191: Would adding the context not be enough for these actions to stay here? What's the benefit of moving them to run on `activitypub_inbox_shared`?

includes/class-migration.php
```diff
@@ -1051,4 +1055,28 @@ public static function sync_jetpack_following_meta() {
 			\do_action( 'added_post_meta', ...$meta );
 		}
 	}
+
+	/**
+	 * Cleanup inbox items for shared inbox migration.
```
obenland:
```suggestion
	 * Clean up inbox items for shared inbox migration.
```
This is similar to setup and set up, where cleanup is the noun and clean up the verb.

includes/class-migration.php
```diff
@@ -1051,4 +1055,28 @@ public static function sync_jetpack_following_meta() {
...
+	 *
+	 * Deletes all existing inbox items to prepare for the new shared inbox structure
+	 * where activities are stored once with multiple recipients as metadata.
+	 */
+	private static function cleanup_inbox() {
```
obenland:
```suggestion
	private static function clean_up_inbox() {
```

obenland (commented):
Mostly small stuff and some questions for me to better understand things

includes/collection/class-inbox.php
pfefferle: Int is still fine.

includes/handler/class-inbox.php
pfefferle: it needs to still support both, because the `activitypub_inbox` still only sends ints. no need to deprecate/change that.

includes/rest/class-inbox-controller.php
pfefferle: To make the code more readable and to be able to unhook them... I would love to deprecate them over time, so I thought it might be nice to hide them a bit.

includes/collection/class-inbox.php
obenland: The code docs talk about backwards compatibility for ints, which indicates to me that they're no longer expected as parameters? The only place the function is called from, other than tests, is `Activitypub\Handler\Inbox::handle_inbox_requests()`, which also expects an array of user_ids.

If you want to keep it compatible with ints for potential future use, let's just remove the back compat language.

includes/collection/class-inbox.php
```diff
@@ -27,32 +27,65 @@ class Inbox {
...
 		}
 
-		$item = self::get_by_guid( $activity->get_id() );
+		// Backward compatibility: convert single user_id to array.
+		if ( ! \is_array( $recipients ) ) {
+			$recipients = array( (int) $recipients );
+		}
+
+		// Sanitize recipients.
+		$recipients = \array_map( 'absint', $recipients );
```
```suggestion
		// Sanitize recipients.
		$recipients = \array_map( 'absint', (array) $recipients );
```

includes/collection/class-inbox.php
```diff
@@ -27,32 +27,65 @@ class Inbox {
...
+				'No valid recipients provided',
+				array( 'status' => 400 )
+			);
+		}
+
+		// Check if activity already exists (by GUID).
+		$existing = self::get_by_guid( $activity->get_id() );
+
+		// If activity exists, add new recipients to it.
+		if ( ! \is_wp_error( $existing ) && $existing instanceof \WP_Post ) {
```
```suggestion
		if ( $existing instanceof \WP_Post ) {
```

includes/collection/class-inbox.php
```diff
@@ -88,6 +121,13 @@ public static function add( $activity, $user_id ) {
 			\kses_init_filters();
 		}
 
+		// Add recipients as separate meta entries after post is created.
+		if ( ! \is_wp_error( $id ) && $id > 0 ) {
```
```suggestion
		if ( ! \is_wp_error( $id ) ) {
```
I'd expect post Ids to always be more than 0.

includes/handler/class-inbox.php
```diff
@@ -21,20 +21,48 @@ class Inbox {
...
+	 * @param string             $type         The type of the activity.
+	 * @param Activity|\WP_Error $activity     The Activity object.
+	 * @param string             $context      The context of the request (Inbox_Collection::CONTEXT_INBOX or Inbox_Collection::CONTEXT_SHARED_INBOX).
+	 */
+	public static function handle_inbox_requests( $data, $user_ids, $type, $activity, $context = Inbox_Collection::CONTEXT_INBOX ) {
+		$inbox_id = null;
```
obenland: Could this be `$result`? It feels a bit less confusing than setting `$inbox_id` to WP_Errors and checking `$inbox_id` to see whether it should be added to the Inbox or not. Although it doesn't feel super great either.

includes/handler/class-inbox.php
```diff
@@ -59,23 +86,39 @@ public static function handle_inbox_requests( $data, $user_id, $type, $activity
...
 
-		if ( $success ) {
-			$id = Inbox_Collection::add( $activity, $user_id );
+		if ( ! \is_wp_error( $inbox_id ) ) {
+			// Pass user_id as-is (can be int or array) to Inbox_Collection::add().
+			$inbox_id = Inbox_Collection::add( $activity, $user_ids );
 		}
 
+		// Normalize user_id to array for action hooks.
+		$user_ids = is_array( $user_ids ) ? $user_ids : array( $user_ids );
```
```suggestion
		$user_ids = (array) $user_ids;
```
This might be good to do at the top of the function so `Inbox_Collection::add()` can take advantage of it, too.

includes/rest/class-inbox-controller.php
obenland: I feel like it makes it much harder to read and understand. I'm happy for us to deprecate them, but I'm not sure they. need to be moved into a callback to do that?

includes/rest/class-inbox-controller.php
```diff
@@ -156,9 +159,12 @@ public function create_item( $request ) {
 		} else {
 			$recipients = $this->get_local_recipients( $data );
 
+			// Filter out blocked recipients.
+			$allowed_recipients = array();
 			foreach ( $recipients as $user_id ) {
-				// Check user-specific blocks for this recipient.
-				if ( Moderation::activity_is_blocked_for_user( $activity, $user_id ) ) {
+				if ( ! Moderation::activity_is_blocked_for_user( $activity, $user_id ) ) {
```
obenland: Regardless of what we decide about the fait of `trigger_inbox_actions()`, let's keep checking for something truthy here.

includes/rest/class-inbox-controller.php
pfefferle: They should be out of the way... Implementers should (in the best case) not see them... The ideal solution (in my mind) would be to move them into a `deprecated.php` file... but I a have no strong feelings about that...

includes/collection/class-inbox.php
pfefferle: AI thought I wanted to deprecate the int! Sorry :(

obenland (approved):

Issue #2054, opened 2025-08-05
Issue: Following Behavior Inconsistency


Jiwoon-Kim:
When attempting to follow a threads.net profile using the blog profile (group actor) (e.g., @newsfeed@travel-in-busan.com), the follow request remains in a pending state and does not complete.

In contrast, when using a user profile (e.g., @thaumiel999@travel-in-busan.com) to follow the same threads.net profile, the follow request is processed successfully.

Additionally, when the user profile @thaumiel999@travel-in-busan.com attempts to follow the blog profile (group) @newsfeed@travel-in-busan.com, the request also remains pending. However, there is currently no way to approve this follow request from the blog profile in the Settings > ActivityPub > Followers admin interface.

obenland:
I'm not sure that's a plugin issue and suspect that it's something where Threads might limit follows to Person actors.

Jiwoon-Kim:
@obenland Is the current ActivityPub plugin also limiting follows to Person actors?
If so, then the behavior I observed and your explanation would make sense together as part of the same logic.

obenland:
> Is the current ActivityPub plugin also limiting follows to Person actors?

No, there is no restriction on actor type.

Jiwoon-Kim:
I was curious whether following is somehow limited within the same instance because of this part:

> Additionally, when the user profile @thaumiel999@travel-in-busan.com attempts to follow the blog profile (group) @newsfeed@travel-in-busan.com, the request also remains pending. However, there is currently no way to approve this follow request from the blog profile in the Settings > ActivityPub > Followers admin interface.

obenland:
That might be a bug? Not sure

obenland:
Looks like [that's by design](https://github.com/Automattic/wordpress-activitypub/blob/500aef9cbe1ebe29a781313fece154affbabf37e/includes/class-dispatcher.php#L279-L285). Not sure if we want to change that for Follow activities.

Jiwoon-Kim:
Yes, I tested this by creating two accounts on different sites and attempting to follow, and the requests are also set to pending. It seems like this is either intentional behavior or a bug.

<img width="3161" height="1789" alt="Image" src="https://github.com/user-attachments/assets/c9df198f-4a25-4d1f-8ae0-b405a47e6bf9" />

Jiwoon-Kim:
It’s possible that following behavior differs depending on whether it happens within the same instance or across instances.

[...]

PR #2359, opened 2025-10-22
Require Shared-Inbox

pfefferle:
The Shared-Inbox is essential for the work on the Reader and helps deduplicating Activities.

Eliminates the shared inbox option, its settings field, health check display, and associated logic. The shared inbox endpoint is now always provided, simplifying configuration and code maintenance.

## Proposed changes:
- Removed the `activitypub_shared_inbox` setting registration and its associated UI field
- Simplified `get_endpoints()` methods to always return the shared inbox endpoint
- Removed constant-based feature flag handling for `ACTIVITYPUB_SHARED_INBOX_FEATURE`

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Simplified configuration by always enabling the shared inbox and removing its separate setting, UI field, and related logic.

</details>

### DIFF
--- a/.github/changelog/2359-from-description
+++ b/.github/changelog/2359-from-description
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Simplified configuration by always enabling the shared inbox and removing its separate setting, UI field, and related logic.
--- a/includes/class-options.php
+++ b/includes/class-options.php
@@ -18,7 +18,6 @@ class Options {
 	public static function init() {
 		\add_filter( 'pre_option_activitypub_actor_mode', array( self::class, 'pre_option_activitypub_actor_mode' ) );
 		\add_filter( 'pre_option_activitypub_authorized_fetch', array( self::class, 'pre_option_activitypub_authorized_fetch' ) );
-		\add_filter( 'pre_option_activitypub_shared_inbox', array( self::class, 'pre_option_activitypub_shared_inbox' ) );
 		\add_filter( 'pre_option_activitypub_vary_header', array( self::class, 'pre_option_activitypub_vary_header' ) );
 
 		\add_filter( 'pre_option_activitypub_allow_likes', array( self::class, 'maybe_disable_interactions' ) );
@@ -76,25 +75,6 @@ class Options {
 		}
 
 		if ( ACTIVITYPUB_AUTHORIZED_FETCH ) {
-			return '1';
-		}
-
-		return '0';
-	}
-
-	/**
-	 * Pre-get option filter for the Shared Inbox.
-	 *
-	 * @param string $pre The pre-get option value.
-	 *
-	 * @return string If the constant is defined, return the value, otherwise return the pre-get option value.
-	 */
-	public static function pre_option_activitypub_shared_inbox( $pre ) {
-		if ( ! \defined( 'ACTIVITYPUB_SHARED_INBOX_FEATURE' ) ) {
-			return $pre;
-		}
-
-		if ( ACTIVITYPUB_SHARED_INBOX_FEATURE ) {
 			return '1';
 		}
 
--- a/includes/model/class-blog.php
+++ b/includes/model/class-blog.php
@@ -391,15 +391,9 @@ class Blog extends Actor {
 	 * @return string[]|null The endpoints.
 	 */
 	public function get_endpoints() {
-		$endpoints = null;
-
-		if ( \get_option( 'activitypub_shared_inbox' ) ) {
-			$endpoints = array(
-				'sharedInbox' => get_rest_url_by_path( 'inbox' ),
-			);
-		}
-
-		return $endpoints;
+		return array(
+			'sharedInbox' => get_rest_url_by_path( 'inbox' ),
+		);
 	}
 
 	/**
--- a/includes/model/class-user.php
+++ b/includes/model/class-user.php
@@ -317,15 +317,9 @@ class User extends Actor {
 	 * @return string[]|null The endpoints.
 	 */
 	public function get_endpoints() {
-		$endpoints = null;
-
-		if ( \get_option( 'activitypub_shared_inbox' ) ) {
-			$endpoints = array(
-				'sharedInbox' => get_rest_url_by_path( 'inbox' ),
-			);
-		}
-
-		return $endpoints;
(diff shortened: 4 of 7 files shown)


Jiwoon-Kim:
For local actors:
The **Inbox** can receive both **local** and **remote activities**.
However, the **Outbox** can only contain **local activities by default** — even when those activities refer to remote objects.

For remote actors:
Only **remote activities** can be created.

It’s not a definitive proposal, but perhaps it would be better to **store remote and local activities separately in the posts table**,
and implement the **Inbox** as **post meta**?

I think of the **Outbox** as forming the **profile feed**, and the **Inbox** as forming the **following feed**.

pfefferle:
Do you have an example of a local inbox activity?

Jiwoon-Kim:
In the case of a **local actor**, activities are created **based on objects**.
In contrast, a **remote actor** creates **objects based on received activities**.

Jiwoon-Kim:
> Do you have an example of a local inbox activity?

In cases where a follow activity occurs between two local users, the Inbox can indeed include a local activity.

obenland (approved):

Jiwoon-Kim:
Perhaps it might be better to first implement the following feed based on **local follow activities** before handling **remote follow (federated) activities**.
Here’s how I imagine the local follow flow:

```mermaid
sequenceDiagram
    participant Alice (local actor)
    participant Bob (local actor)
    participant Outbox (Alice)
    participant Inbox (Bob)
    participant Following Feed

    Alice->>Outbox: Create "Follow" Activity (target: Bob)
    Outbox-->>Bob: Deliver "Follow" Activity (local)
    Bob->>Inbox: Store local "Follow" Activity
    Bob-->>Bob: Accept Activity (optional)
    Inbox-->>Following Feed: Update local following feed
    Note over Alice,Bob: This process demonstrates a local Follow Activity<br>before implementing remote follow (federated) handling.
```

This example demonstrates how **Alice** follows **Bob**, assuming both actors exist on the same **local server**.

---

## Example: Local "Follow" Activity (ActivityPub JSON)

This example shows the full lifecycle of a **local follow** interaction, represented according to the **ActivityPub specification**.

---

### 1. Activity stored in **Alice’s Outbox** (Create)

Alice initiates a **Follow** activity targeting Bob.
This activity is stored in her **Outbox** collection.

```json
{
  "id": "https://example.com/activities/998877", 
  "type": "Follow",
  "actor": "https://example.com/@alice", 
  "object": "https://example.com/@bob", 
  "to": [
    "https://example.com/@bob/inbox" 
  ],
  "published": "2025-10-22T12:00:00Z"
}
```

---

### 2. Activity delivered and stored in **Bob’s Inbox**

Alice’s Outbox then **delivers** this `Follow` activity to Bob’s Inbox (`https://example.com/@bob/inbox`).
The same activity object is stored in Bob’s Inbox.

```json
{
  "id": "https://example.com/activities/998877",
  "type": "Follow",
  "actor": "https://example.com/@alice",
  "object": "https://example.com/@bob",
  "to": [
    "https://example.com/@bob/inbox"
  ],
  "published": "2025-10-22T12:00:00Z"
}
```

---

### 3. Bob **Accepts** Alice’s Follow request

When Bob accepts the follow request, his Outbox generates an **Accept** activity.
This activity references Alice’s original `Follow` activity and is sent to **Alice’s Inbox**, allowing Alice to confirm that her follow was accepted.

```json
{
  "id": "https://example.com/activities/998878",
  "type": "Accept",
  "actor": "https://example.com/@bob", 
  "object": {
    "type": "Follow",
    "id": "https://example.com/activities/998877",
    "actor": "https://example.com/@alice",
    "object": "https://example.com/@bob"
  },
  "to": [
    "https://example.com/@alice/inbox"
  ],
  "published": "2025-10-22T12:01:00Z"
}
```

---

This JSON flow clearly shows that:

* **Local actors** can both **send** and **receive** activities within the same domain.
* **Follow** and **Accept** form a complete interaction loop that can be used to implement a **local following feed**, even before introducing federated (remote) actors.

pfefferle:
@Jiwoon-Kim Are you aware that this already works for remote Actors? It is already possible to follow users and you can already receive their messages.

Why should we handle local follow requests differently from remote follows? They should simply work out of the box in the same way, it already works for remote users!?

Jiwoon-Kim:
Are you aware that if you send a follow to another local user’s handle, it doesn’t move past the pending state?

pfefferle:
@Jiwoon-Kim this is a bug we have to fix then.

obenland:
@pfefferle https://github.com/Automattic/wordpress-activitypub/issues/2054#issuecomment-3201234388

Jiwoon-Kim:
I am currently reviewing whether the following approach is valid based on the assumptions I have made.

**Handling Local Follow/Accept Activities**

In WordPress, a single taxonomy can be assigned to multiple post types, such as `post_type=activity` and `post_type=remote-activity`.

Example admin URL:

```
https://yourdomain.com/wp-admin/term.php
    ?taxonomy=inbox
    &tag_ID=123
    &post_type=local-activity
```

Example `register_taxonomy` code:

```php
register_taxonomy(
    'my_taxonomy',          // Taxonomy name
    ['post', 'custom_type'], // Array of post types to attach
    [
        'label' => 'My Taxonomy',
        'hierarchical' => true
    ]
);
```

This allows us to categorize both local and remote activities using the same taxonomy structure, which could simplify handling local follow and accept activities in a unified manner.

pfefferle:
@obenland this check was added, to not produce duplicate reactions (comments, likes, ...) so we would have to revise that "fix".

Jiwoon-Kim:
I don’t know the exact reason why local follows are not being processed correctly. However, I wanted to examine whether there is a need to distinguish between local and remote activities (including follows) when implementing a shared Inbox, Inbox, and Outbox.

Summary of the idea:

Treat activities as objects and store them in the wp_post table.

Implement Inbox and Outbox using taxonomies.

Question: Should remote activities and local activities be separated into different CPTs (custom post types)?

Reason: Remote activities create remote objects, whereas local activities are generated from local objects.

PR #2368, opened 2025-10-24
Remove ap_follower from Stream exclusion

obenland:
It's no longer in use.

### DIFF
--- a/integration/load.php
+++ b/integration/load.php
@@ -176,8 +176,6 @@ add_filter(
 	'wp_stream_posts_exclude_post_types',
 	function ( $post_types ) {
 		$post_types[] = 'ap_actor';
-		// @todo remove in one of the next versions
-		$post_types[] = 'ap_follower';
 		$post_types[] = 'ap_extrafield';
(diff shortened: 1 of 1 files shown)


integration/load.php
```diff
@@ -176,8 +176,6 @@ function register_stream_connector( $classes ) {
 	'wp_stream_posts_exclude_post_types',
 	function ( $post_types ) {
 		$post_types[] = 'ap_actor';
-		// @todo remove in one of the next versions
-		$post_types[] = 'ap_follower';
 		$post_types[] = 'ap_extrafield';
 		$post_types[] = 'ap_extrafield_blog';
```
obenland, line 180: Not sure if we want to add `ap_post` here?

pfefferle: Yes, let's add `ap_post`!

obenland: Fixed in [b36c3d0](https://github.com/Automattic/wordpress-activitypub/pull/2368/commits/b36c3d02fbb0b0d443fd298134dc92c1146a3e4d).

pfefferle (approved):

PR #2376, opened 2025-10-27
Add scheduled processing for inbox activities

pfefferle:
Refactors inbox activity handling to reduce duplicate processing through scheduled batching. Activities are now cached with associated user IDs for 10 minutes, then processed in a single batch after a 15-second delay, preventing redundant processing of the same activity.

## Proposed changes:
- Introduces scheduled batch processing via `schedule_inbox_request` and `process_inbox_request` methods
- Adds `deduplicate` method to merge duplicate inbox items created during race conditions
- Adds `add_recipients` helper method for bulk recipient assignment

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Improved inbox performance by batching and deduplicating activities, reducing redundant processing and improving handling during high activity periods.

</details>


pfefferle:
@obenland what do you think about using a schedule to collect all requests that come in through the actor inboxes, to then store them in one single request?

Any early ideas? This was just an early stage idea and I am happy to find better names and/or caching options.

includes/handler/class-inbox.php
```diff
@@ -37,14 +40,28 @@ public static function init() {
...
+		// Store activity in inbox collection once.
+		$inbox_item = Inbox_Collection::get_by_guid( $activity_id );
+		if ( \is_wp_error( $inbox_item ) ) {
+			Inbox_Collection::add( $activity, (array) $user_ids );
+		} else {
+			Inbox_Collection::add_recipients( $inbox_item->ID, (array) $user_ids );
+		}
+
+		\wp_clear_scheduled_hook( 'activitypub_process_inbox_request', array( $activity_id ) );
+		\wp_schedule_single_event( time() + 15, 'activitypub_process_inbox_request', array( $activity_id ) );
```
pfefferle: I am not sure how many time can/should be between requests!?

includes/collection/class-inbox.php
```diff
@@ -330,6 +330,20 @@ public static function remove_recipient( $post_id, $user_id ) {
 		return \delete_post_meta( $post_id, '_activitypub_user_id', $user_id );
 	}
 
+	/**
+	 * Add multiple recipients to an existing inbox activity.
+	 *
+	 * @param int   $post_id   The inbox post ID.
+	 * @param int[] $user_ids  The user ID or array of user IDs to add.
+	 *
+	 * @return void
```
```suggestion
	 * @param int   $post_id  The inbox post ID.
	 * @param int[] $user_ids The user ID or array of user IDs to add.
```

includes/collection/class-inbox.php
```diff
@@ -358,4 +372,45 @@ public static function get_by_guid_and_recipient( $guid, $user_id ) {
...
+
+		// If only one post, return it.
+		if ( count( $duplicates ) === 1 ) {
+			return $duplicates[0];
+		}
+
+		// Keep the first post, merge recipients from others and delete.
+		$primary = $duplicates[0];
+
+		foreach ( array_slice( $duplicates, 1 ) as $duplicate ) {
```
```suggestion
		}
		
		// Keep the first post, all others are duplicates.
		$primary = array_shift( $duplicates );
		
		foreach ( $duplicates as $duplicate ) {
```
If `$duplicates` is empty it won't iterate over the array

includes/handler/class-inbox.php
```diff
@@ -37,14 +40,28 @@ public static function init() {
...
-	public static function maybe_handle_inbox_request( $data, $user_ids, $type, $activity, $context = Inbox_Collection::CONTEXT_INBOX ) {
+	public static function schedule_inbox_request( $data, $user_ids, $type, $activity, $context = Inbox_Collection::CONTEXT_INBOX ) {
 		// Ignore shared inbox requests to prevent duplicate processing.
 		if ( Inbox_Collection::CONTEXT_SHARED_INBOX === $context ) {
 			return;
 		}
 
-		// Process inbox requests.
-		self::handle_inbox_requests( $data, $user_ids, $type, $activity, $context );
+		$activity_id = isset( $data['id'] ) ? $data['id'] : null;
```
```suggestion
		$activity_id = $data['id'] ?? null;
```
Would this be a good use case for `object_to_uri()`?

includes/handler/class-inbox.php
```diff
@@ -37,14 +40,28 @@ public static function init() {
...
+		if ( ! $activity_id ) {
+			self::handle_inbox_requests( $data, $user_ids, $type, $activity, $context );
+			return;
+		}
+
+		// Store activity in inbox collection once.
+		$inbox_item = Inbox_Collection::get_by_guid( $activity_id );
+		if ( \is_wp_error( $inbox_item ) ) {
+			Inbox_Collection::add( $activity, (array) $user_ids );
```
obenland: Calling `Inbox_Collection::add()` directly doesn't allow Inbox storage to be skipped.

What if we made `Inbox_Collection::add()` do the `get_by_guid()` check internally to prevent duplicates and return the primary post? That way we could re-use `handle_inbox_requests()` for every user individually and then schedule a deduplication event if the context is `CONTEXT_INBOX`

pfefferle: That is already the case, good point!

pfefferle:
@obenland I deprecated the inbox handler. You were right to question it, it is a central part of the inbox and we should not have a second layer of actions/hooks.

includes/rest/class-actors-inbox-controller.php
```diff
@@ -103,6 +104,8 @@ public function register_routes() {
 				'schema' => array( $this, 'get_item_schema' ),
 			)
 		);
+
+		\add_action( 'activitypub_inbox_create_item', array( self::class, 'process_create_item' ) );
```
obenland: I'd expect this not to work, as `register_routes` should only run on `rest_api_init`. Does it?
We could consider moving this into `Dispatcher` or `Handler`.

pfefferle: I mainly kept it here because it is a very simple and small schedule and to have it all together... I feared that it might get confusing to have it in the dispatcher... but maybe this can be solved with good phpdoc!?

pfefferle: > I'd expect this not to work, as register_routes should only run on rest_api_init.

Good point!

obenland (approved):

PR #2378, opened 2025-10-27
Move default visibility logic to UI and classic editor

pfefferle:
This PR removes the server-side `default_post_meta_data` filter and shifts the logic for determining default ActivityPub content visibility to the UI layer. The change ensures consistent behavior between block and classic editors by implementing the default visibility logic directly in the editor plugin and classic editor meta box, allowing for more flexible UI-driven defaults based on post age and federation status.

Fixes https://wordpress.org/support/topic/cant-set-posts-to-public-on-fediverse/

@obenland the Classic Editor settings came by accident. I planned to only add the visibility, but claude copied the whole block editor sidebar and I have to admit that I quite like it ;)

Screenshot 2025-10-28 at 08 13 11

## Proposed changes:
- Removes server-side filter and related PHP tests
- Adds visibility logic to block editor plugin with comprehensive JS tests
- Expands classic editor meta box with full ActivityPub settings

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [X] Minor
-   [ ] Major

#### Type

#### Message

Moved default visibility handling from the server to the editor UI, ensuring consistent and flexible ActivityPub visibility settings across both block and classic editors.

</details>


integration/class-classic-editor.php
```diff
@@ -7,6 +7,8 @@
 
 namespace Activitypub\Integration;
 
+use PHPUnit\TestFixture\A;
```
obenland: I'm surprised this survived the precommit hook

integration/class-classic-editor.php
```diff
@@ -18,73 +20,188 @@ class Classic_Editor {
...
+				\update_post_meta( $post_id, 'activitypub_content_visibility', $visibility );
+			}
+		}
+
+		// Save quote interaction policy.
+		if ( isset( $_POST['activitypub_interaction_policy_quote'] ) ) {
+			$quote_policy = \sanitize_text_field( \wp_unslash( $_POST['activitypub_interaction_policy_quote'] ) );
+			if ( \in_array( $quote_policy, array( 'anyone', 'followers', 'me' ), true ) ) {
+				\update_post_meta( $post_id, 'activitypub_interaction_policy_quote', $quote_policy );
+			}
```
obenland: This could probably be simplified to a meta key check in `$_POST` and then running `update_post_meta()` with it. Sanitization and handling of empty values should be covered by the register_meta calls and filters.

src/editor-plugin/__tests__/plugin.test.js
```diff
@@ -0,0 +1,151 @@
+/**
+ * @jest-environment jsdom
+ */
+
+describe( 'EditorPlugin getDefaultVisibility', () => {
+	// Helper function to simulate the getDefaultVisibility logic
+	const getDefaultVisibility = ( meta, postDate ) => {
```
obenland: Can we put this in a central location with an export so it can be reused?

obenland (commented):
Can you help me understand the benefit of "more flexible UI-driven defaults" in three places over `default_post_meta_data()`?

src/editor-plugin/__tests__/plugin.test.js
pfefferle: sure!

pfefferle:
Have you read the linked .org issue? With the default value, set in the options hook, and the deletion of the default values in a different one, there is no way to change the default for posts older than a month.

The only two options are to not delete the default or to move the code to the form handlings instead. I prefer this one (☺️), because the other solution produces a lot of default meta spam.

But maybe you see a third solution!?

obenland (dismissed):

obenland (approved):

PR #2389, opened 2025-10-29
Fix Prettify

pfefferle:
Modified the checkout step in the format workflow to use the pull request's head ref, repository, and authentication token. This ensures the workflow checks out the correct code from forks and external repositories.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -13,7 +13,9 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v5
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Setup Node.js
         uses: actions/setup-node@v6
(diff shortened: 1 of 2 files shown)


obenland:
Any idea _why_ this broke?

pfefferle:
No idea

obenland:
I assume the bug was introduced in https://github.com/Automattic/wordpress-activitypub/pull/2277. It's possible the workflow hasn't actually needed to commit anything since then and it's only been hitting the "No formatting changes needed" path

.github/workflows/format.yml
```diff
@@ -13,7 +13,8 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v5
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
obenland: Is the repository key needed?

pfefferle: let's check

pfefferle: nope! seems to be not needed

obenland (approved):

PR #2399, opened 2025-10-30
Social Web UI: Add feed reader

pfefferle:
| Screenshot 2025-10-31 at 12 28 11    | Screenshot 2025-10-31 at 12 28 04 |
| -------- | ------- |

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


pfefferle:
@obenland this should be ready for a first feedback (if you find some time)!

I will change from the wp/v2 endpoint after I make progress on https://github.com/Automattic/wordpress-activitypub/pull/2472 next week.

src/social-web/store/actions.ts
```diff
@@ -62,4 +69,20 @@ export const actions = {
 			yield actions.setLoading( 'interactions', false );
 		}
 	},
+
+	*fetchFeed() {
```
obenland: Is this needed with us using entityRecords? If there's anything we need to preload, [we can do that on the PHP side](https://github.com/Automattic/wordpress-activitypub/pull/2422/files#diff-ba0c22268e740c51f9e2ac46372bdf1f2fabf7cbdbdb864376b9beb048ec416cR35-R55).

src/social-web/routes/feed/stage.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { useFeed } from '../../hooks/use-feed';
+import {
+	titleField,
+	dateField,
+	statusField,
+	excerptField,
+	metadataField,
+	createContentField,
+} from '../../components/fields';
+import { getFeedActions } from './FeedActions';
```
obenland:
```suggestion
import { getFeedActions } from './feed-actions';
```
Among all the confusing mix of cases, filenames are in kebap-case, I think

src/social-web/routes/feed/stage.tsx
```diff
@@ -0,0 +1,145 @@
...
+		direction: 'desc',
+	},
+	search: '',
+	filters: [],
+	fields: [ 'metadata', 'title.rendered', 'excerpt.rendered' ],
+	layout: {},
+};
+
+const defaultLayouts = {
+	table: {
```
obenland: Maybe we don't need to provide the table view for now? It looks pretty broken and I'm not sure about how useful it is?

In the Site Editor, the table view at `/wp-admin/site-editor.php?p=%2Fpage&layout=table` takes over the entire width of the content area, which feels more appropriate. Still, not sure if that is a useful view to read a feed.

src/social-web/components/layout/index.tsx
```diff
@@ -7,65 +7,61 @@
...
-const FollowingInspector = lazy(
-	() => import( /* webpackChunkName: "social-web/following" */ '../../routes/following/inspector' )
-);
-const InteractionInspector = lazy(
-	() => import( /* webpackChunkName: "social-web/interactions" */ '../../routes/interactions/inspector' )
-);
+// Import route stages
+import FeedStage from '../../routes/feed/stage';
+import FollowingStage from '../../routes/following/stage';
+import InteractionsStage from '../../routes/interactions/stage';
```
obenland: What's the benefit of removing lazy loading?

src/social-web/routes/feed/stage.tsx
```diff
@@ -0,0 +1,145 @@
...
+	const { feed, isResolving, totalItems, totalPages } = useFeed( {
+		perPage: view.perPage || 20,
+		page: view.page || 1,
+		orderBy: view.sort?.field || 'date',
+		order: view.sort?.direction || 'desc',
+		search: view.search || '',
+	} );
+
+	const fields: Field< FeedPost >[] = useMemo(
+		() => [ metadataField, titleField, excerptField, createContentField(), dateField, statusField ],
```
obenland: `createContentField()` should be `contentField` and not a function.

I'd remove the status field, it doesn't feel helpful.

obenland (commented):
The basic functionality seems largely in place and we can build on that. The stuff I left feedback on felt worth fixing before we start iterating on it.

src/social-web/components/layout/index.tsx
pfefferle: None... This might simply be the result of a merge-conflict 😢

obenland (approved):

obenland:
hm, this is actually not working for me. `ap_posts` returns a list of posts but they don't get displayed

obenland:
Is it the missing pre-loading?

obenland (changes requested):
There are no posts displayed, currently

pfefferle:
@obenland have you properly added the correct post-metas?

obenland:
You'll have to break that down for me 🙂 It doesn't just work out of the box with ap_posts?

pfefferle:
@obenland no! not all blog users have subscribed to the same content especially when it is about private posts or direct messages!

See: https://github.com/Automattic/wordpress-activitypub/blob/1570996c771a8d89930d4a692e4534d293cccf19/includes/collection/class-inbox.php#L132 and https://github.com/Automattic/wordpress-activitypub/blob/1570996c771a8d89930d4a692e4534d293cccf19/includes/collection/class-inbox.php#L306

pfefferle:
Here is the used filter: https://github.com/Automattic/wordpress-activitypub/pull/2399/files#diff-ceb59dc4107c120554c144aa5f601712077313b8b65bc0a0a844b0125272461fR703

includes/class-post-types.php
```diff
@@ -651,6 +654,75 @@ public static function filter_ap_actor_query_by_follower( $args, $request ) {
...
+	}
+
+	/**
+	 * Filter ap_post REST query to only show posts for the current user.
+	 *
+	 * @param array $args Query arguments.
+	 *
+	 * @return array Modified query arguments.
+	 */
+	public static function filter_ap_post_by_user( $args ) {
```
pfefferle, line 703: @obenland here is the filter that is used, to only show `ap_post`s that are visible to the logged-in user.

obenland:
The ap_post api endpoint returns a list of posts for me, they just don't get displayed in the app. None of the dataview component is, as a matter of fact.

If there was an empty result, I'd expect the empty message to show up, which it doesn't.

Screenshot 2025-11-17 at 9 17 19 AM

pfefferle:
Ah, it seems to be an issue with "Sort by Title". Can you change to "Sort by Date", to be sure that we have the same issue!?

pfefferle:
I removed "Sort by Title"!

obenland:
The last change didn't seem to help, I'm afraid. I'm sorry I don't have more to offer currently, I'm a bit hamstrung by not having any posts locally to test with at the moment.

Does it work for you?

> Can you change to "Sort by Date", to be sure that we have the same issue!?

DataViews don't display for me, so there's no sorting that I can change, I'm afraid.

pfefferle:
@obenland it does

Screenshot 2025-11-17 at 16 37 40

can you add a `_activitypub_user_id` Post-Meta that has you WordPress User ID to one (or all) of your `ap_post`s?

obenland (approved):
It works locally for me, so let's get it in and iterate

PR #2412, opened 2025-10-31
Refactor Mastodon import to use arrays, add tests

### LINKED ISSUE #2405: Mastodon Imports fail with error
### Quick summary

```
Fatal error: Uncaught Error: Cannot use object of type stdClass as array in /var/www/html/wp-content/plugins/wordpress-activitypub/includes/functions.php:557

Stack trace: 
#0 /var/www/html/wp-content/plugins/wordpress-activitypub/includes/functions.php(538): Activitypub\extract_recipients_from_activity_property('bto', Array) 
#1 /var/www/html/wp-content/plugins/wordpress-activitypub/includes/functions.php(608): Activitypub\extract_recipients_from_activity(Array) 
#2 /var/www/html/wp-content/plugins/wordpress-activitypub/includes/wp-admin/import/class-mastodon.php(287): Activitypub\is_activity_public(Array) 
#3 /var/www/html/wp-content/plugins/wordpress-activitypub/includes/wp-admin/import/class-mastodon.php(247): Activitypub\WP_Admin\Import\Mastodon::import_posts() 
#4 /var/www/html/wp-content/plugins/wordpress-activitypub/includes/wp-admin/import/class-mastodon.php(82): Activitypub\WP_Admin\Import\Mastodon::import() 
#5 /var/www/html/wp-admin/admin.php(365): Activitypub\WP_Admin\Import\Mastodon::dispatch() 
#6 {main} thrown in /var/www/html/wp-content/plugins/wordpress-activitypub/includes/functions.php on line 557
```

See https://wordpress.org/support/topic/import-mastodon-beta/page/2/#post-18701387

### Steps to reproduce

Try to import a mastodon zip file.

### Site owner impact

### Severity

Minor

_No response_

_No response_

pfefferle:
Refactored Mastodon import logic to consistently use associative arrays instead of stdClass objects, improving compatibility and reliability. Updated Blocks and Mastodon classes, adjusted related docblocks, and added comprehensive PHPUnit tests for the Mastodon import process, including edge cases for public/private posts, hashtags, summaries, duplicates, and filter hooks.

Fixes #2405 

## Proposed changes:
- Modified `json_decode()` to use associative flag, ensuring all decoded JSON is in array format
- Updated all references to use array syntax `['key']` instead of object syntax `->key`
- Added comprehensive test coverage for the import functionality

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Refactored Mastodon import handling to use consistent array-based data, improving reliability and compatibility across all import scenarios.

</details>

### DIFF
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -279,8 +279,8 @@ class Blocks {
 	/**
 	 * Converts content to blocks before saving to the database.
 	 *
-	 * @param array  $data The post data to be inserted.
-	 * @param object $post The Mastodon Create activity.
+	 * @param array $data The post data to be inserted.
+	 * @param array $post The Mastodon Create activity.
 	 *
 	 * @return array
 	 */
@@ -297,8 +297,9 @@ class Blocks {
 		$data['post_content'] = \rtrim( \implode( PHP_EOL, $blocks ), PHP_EOL );
 
 		// Add reply block if it's a reply.
-		if ( null !== $post->object->inReplyTo ) {
-			$reply_block          = \sprintf( '<!-- wp:activitypub/reply {"url":"%1$s","embedPost":true} /-->' . PHP_EOL, \esc_url( $post->object->inReplyTo ) );
+		$in_reply_to = $post['object']['inReplyTo'] ?? null;
+		if ( null !== $in_reply_to ) {
+			$reply_block          = \sprintf( '<!-- wp:activitypub/reply {"url":"%1$s","embedPost":true} /-->' . PHP_EOL, \esc_url( $in_reply_to ) );
 			$data['post_content'] = $reply_block . $data['post_content'];
 		}
 
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -297,9 +297,8 @@ class Blocks {
 		$data['post_content'] = \rtrim( \implode( PHP_EOL, $blocks ), PHP_EOL );
 
 		// Add reply block if it's a reply.
-		$in_reply_to = $post['object']['inReplyTo'] ?? null;
-		if ( null !== $in_reply_to ) {
-			$reply_block          = \sprintf( '<!-- wp:activitypub/reply {"url":"%1$s","embedPost":true} /-->' . PHP_EOL, \esc_url( $in_reply_to ) );
+		if ( ! empty( $post['object']['inReplyTo'] ) ) {
+			$reply_block          = \sprintf( '<!-- wp:activitypub/reply {"url":"%1$s","embedPost":true} /-->' . PHP_EOL, \esc_url( $post['object']['inReplyTo'] ) );
 			$data['post_content'] = $reply_block . $data['post_content'];
 		}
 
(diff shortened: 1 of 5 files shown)


includes/class-blocks.php
```diff
@@ -297,8 +297,9 @@ function ( $paragraph ) {
 		$data['post_content'] = \rtrim( \implode( PHP_EOL, $blocks ), PHP_EOL );
 
 		// Add reply block if it's a reply.
-		if ( null !== $post->object->inReplyTo ) {
-			$reply_block          = \sprintf( '<!-- wp:activitypub/reply {"url":"%1$s","embedPost":true} /-->' . PHP_EOL, \esc_url( $post->object->inReplyTo ) );
+		$in_reply_to = $post['object']['inReplyTo'] ?? null;
+		if ( null !== $in_reply_to ) {
+			$reply_block          = \sprintf( '<!-- wp:activitypub/reply {"url":"%1$s","embedPost":true} /-->' . PHP_EOL, \esc_url( $in_reply_to ) );
```
obenland:
```suggestion
		if ( ! empty( $post['object']['inReplyTo'] ) ) {
			$reply_block          = \sprintf( '<!-- wp:activitypub/reply {"url":"%1$s","embedPost":true} /-->' . PHP_EOL, \esc_url( $post['object']['inReplyTo'] ) );
```

tests/phpunit/tests/includes/wp-admin/import/class-test-mastodon.php
```diff
@@ -0,0 +1,799 @@
...
+							'type'      => 'Note',
+							'content'   => '<p>Another public post</p>',
+							'published' => '2024-01-15T11:00:00Z',
+							'to'        => array( 'https://www.w3.org/ns/activitystreams#Public' ),
+							'tag'       => array(),
+						),
+					),
+				),
+			)
+		);
```
obenland, line 95: I've not thought this through yet, but I wonder if we should create a fixture with a "full" export, including attachments, that we could re-use. Not sure how much sense that makes?

pfefferle: I like the idea, but can we have that a different PR?

obenland (approved):

Jiwoon-Kim:
I’ve been thinking about why post migration is so difficult in ActivityPub, and I believe the biggest reason is the immutability of IDs.
While other properties can be changed, the id must remain constant.

The second issue is visibility — particularly when posts are set to followers-only.
You can migrate your followings, but followers cannot be migrated, which means the visibility scope can change after migration.
The complication deepens with the inReplyTo object, since its visibility should inherit from the parent post — and this dependency can make post-level migration nearly impossible to handle correctly.

The third reason is that what we usually call “migration” is actually actor-level migration, not object-level migration.
A post object may appear to belong to its actor, but in essence, both the actor and the post are independent entities.

And most importantly — we’re not migrating the full set of activities, inbox, and outbox data, right?
That’s why followers can’t be transferred when an account is migrated.

PR #2419, opened 2025-11-05
Notes: do not prepend post titles to reply posts

jeherve:
## Proposed changes:

We currently force the addition of the post title to notes, so that title is included before the post content if it exists.

This does not make sense when the note includes a Federated Reply block. In those situations, you would expect the beginning of the note to be a @mention of the person you are replying to, not the post title.

> [!NOTE]
> This PR is also an opportunity to add an extra parameter to the existing `activitypub_object_content_template`, so it's easier for folks to customize their content template based off the type of the post.

### Other information:

## Testing instructions:

* Write and publish 2 posts:
    * One that includes a Federated reply block in reply to a random post on the Fediverse.
    * One that does not.
* Both posts should have a post title, but should have very content in the post so they're automatically classified as "Notes".
* View the posts on the frontend, and add `/activitypub` to the end of the post URL
    * The post title should only be prepended to the post that does not include the reply block.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -900,9 +900,7 @@ class Post extends Base {
 			$template = '';
 
 			// If the post is a note and not a reply, force the inclusion of the post title.
-			$is_reply = null !== $this->get_in_reply_to();
-
-			if ( 'Note' === $type && ! $is_reply ) {
+			if ( 'Note' === $type && empty( $this->get_in_reply_to() ) ) {
 				$template .= "[ap_title type=\"html\"]\n\n";
 			}
 
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -899,7 +899,10 @@ class Post extends Base {
 		if ( 'wordpress-post-format' === $post_format_setting ) {
 			$template = '';
 
-			if ( 'Note' === $type ) {
+			// If the post is a note and not a reply, force the inclusion of the post title.
+			$is_reply = null !== $this->get_in_reply_to();
+
+			if ( 'Note' === $type && ! $is_reply ) {
 				$template .= "[ap_title type=\"html\"]\n\n";
 			}
 
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -899,8 +899,15 @@ class Post extends Base {
 		if ( 'wordpress-post-format' === $post_format_setting ) {
 			$template = '';
 
-			// If the post is a note and not a reply, force the inclusion of the post title.
-			if ( 'Note' === $type && empty( $this->get_in_reply_to() ) ) {
+			/*
+			 *If the post is a note, not a reply, and does not have mentions
+			 * force the inclusion of the post title.
+			 */
+			if (
+				'Note' === $type
+				&& empty( $this->get_in_reply_to() )
+				&& empty( $this->get_mentions() )
+			) {
 				$template .= "[ap_title type=\"html\"]\n\n";
 			}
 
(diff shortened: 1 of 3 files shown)


includes/transformer/class-post.php
```diff
@@ -894,11 +894,15 @@ protected function get_post_content_template() {
...
 		$post_format_setting = \get_option( 'activitypub_object_type', ACTIVITYPUB_DEFAULT_OBJECT_TYPE );
+		$type                = $this->get_type();
 
 		if ( 'wordpress-post-format' === $post_format_setting ) {
 			$template = '';
 
-			if ( 'Note' === $this->get_type() ) {
+			// If the post is a note and not a reply, force the inclusion of the post title.
+			$is_reply = null !== $this->get_in_reply_to();
+
```
pfefferle: @jeherve Should we also check for `$this->get_mentions()` (@-replies) or do you think replies are fine for now?

includes/transformer/class-post.php
```diff
@@ -894,11 +894,15 @@ protected function get_post_content_template() {
...
+			if ( 'Note' === $type && ! $is_reply ) {
```
```suggestion
			if ( 'Note' === $type && empty( $this->get_in_reply_to() ) ) {
```
Would this work?

It would be great to have a unit test for this

includes/transformer/class-post.php
jeherve: I worry this would catch longer posts, that would only happen to mention Fediverse users like we mention someone by their name and link to their blog or Twitter account today. 

Here is an example:
https://herve.bzh/where-did-all-the-wordpress-editors-go/

includes/transformer/class-post.php
```diff
@@ -915,8 +919,9 @@ protected function get_post_content_template() {
 		 *
 		 * @param string   $template  The template string containing shortcodes.
 		 * @param \WP_Post $item The WordPress post object being transformed.
+		 * @param string   $type ActivityStreams 2.0 Object-Type for the post.
```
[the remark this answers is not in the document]
jeherve: Done in cac66546b3668f96465512bda51048883a2f65b4

includes/transformer/class-post.php
jeherve: After our discussion, I added this in since it's unlikely note posts would be longer posts.

d39a915facdcbd85b12068436149c137e3eb7aaf

includes/transformer/class-post.php
jeherve: Works for me! c289589abbde6696e917b4ec8d92256c94a18e80

jeherve:
> It would be great to have a unit test for this

Sure thing. I worked on that in bfdfbc8d302cefb6d513cb2017ce331c2daa6862

includes/transformer/class-post.php
```diff
@@ -894,11 +894,20 @@ protected function get_post_content_template() {
...
-			if ( 'Note' === $this->get_type() ) {
+			/*
+			 * If the post is a note, not a reply, and does not have mentions
+			 * force the inclusion of the post title.
+			 */
+			if (
+				'Note' === $type
+				&& empty( $this->get_in_reply_to() )
+				&& empty( $this->get_mentions() )
+			) {
```
pfefferle, line 910: Shouldn't it be either or?

```suggestion
				'Note' === $type &&
				( empty( $this->get_in_reply_to() ) || 
				empty( $this->get_mentions() ) )
			) {
```

pfefferle: Oh no, it checks for empty! You are right!!! Forget my comment 🫣

pfefferle (approved):
Thanks @jeherve ☺️

PR #2436, opened 2025-11-10
Allow short ternary (Elvis) operator in codebase

pfefferle:
Removes all inline phpcs ignore comments for the Universal.Operators.DisallowShortTernary rule and updates phpcs.xml to exclude this rule, permitting use of the short ternary (Elvis) operator throughout the codebase.

## Proposed changes:
- Updated PHPCS configuration to allow short ternary operators
- Removed 11 `phpcs:ignore` comments across 8 files where the Elvis operator is used

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


phpcs.xml
```diff
@@ -31,4 +31,9 @@
 			</property>
 		</properties>
 	</rule>
+
+	<!-- Allow short ternary operators (Elvis operator) -->
```
obenland:
```suggestion
```
This is pretty self-evident. And if we can keep Elvis out of the codebase, that's an added benefit. 🙂

obenland (dismissed):

pfefferle:
@obenland btw. is the short-ternary operator allowed on WP.com?

pfefferle:
It is not allowed on WP.com, so we have to keep it as is.

obenland:
> It is not allowed on WP.com, so we have to keep it as is.

We could add an exception for the activitypub plugin folder in the global phpcs.xml on Dotcom. There's precedent.

obenland (approved):

PR #2437, opened 2025-11-10
Ignore announces from blog actor in Announce handler

pfefferle:
Announces originating from the blog actor are now ignored in the Announce handler to prevent self-announcing behavior. Added corresponding unit tests to verify that announces from the blog actor are not processed, while those from external and same-domain non-blog actors are handled as expected.

## Proposed changes:

* Added logic to ignore announces from the blog actor while allowing announces from regular user actors and external actors
* Added comprehensive test coverage for the new filtering behavior
* Updated existing tests to use external actor URLs instead of local user URLs

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Prevented self-announcing by ignoring announces from the blog actor, while still processing announces from user and external actors.

</details>

### DIFF
--- a/includes/handler/class-announce.php
+++ b/includes/handler/class-announce.php
@@ -42,10 +42,10 @@ class Announce {
 			return;
 		}
 
-		$actor_id = object_to_uri( $announcement['actor'] );
 		// Ignore announces from the blog actor.
-		$actor_resource_id = Actors::get_id_by_resource( $actor_id );
-		if ( is_same_domain( $actor_id ) && ! is_wp_error( $actor_resource_id ) && Actors::BLOG_USER_ID === $actor_resource_id ) {
+		$actor_id = object_to_uri( $announcement['actor'] );
+		$user_id  = Actors::get_id_by_resource( $actor_id );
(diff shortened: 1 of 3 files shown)


includes/handler/class-announce.php
```diff
@@ -40,6 +42,13 @@ public static function handle_announce( $announcement, $user_ids, $activity = nu
 			return;
 		}
 
+		// Ignore announces from the blog actor.
+		$actor_id = object_to_uri( $announcement['actor'] );
+		$user_id  = Actors::get_id_by_resource( $actor_id );
+		if ( is_same_domain( $actor_id ) && ! \is_wp_error( $user_id ) && Actors::BLOG_USER_ID === $user_id ) {
```
obenland:
```suggestion
		if ( Actors::BLOG_USER_ID === Actors::get_id_by_resource( $announcement['actor'] ) ) {
```
Is all of this other stuff necessary? Tests still pass without it

pfefferle: It is not necessary, but I thought it might speed up checks if we do not have the lookup when the domain is no local one.

obenland (approved):

PR #2451, opened 2025-11-12
Fix: PHPdoc types in Base_Objects

Menrath:
Likely there are more errors.

Almost everything could be an array! Though I agree not encouraging the use in the setters for certain things, when parsing remote stuff it should be considered!

### Changelog entry

<details>

#### Significance

- [x] Patch
- [ ] Minor
- [ ] Major

#### Type

#### Message

Corrected PHPDoc types in Base_Object class to accurately reflect that certain properties can return arrays.

</details>


includes/activity/class-base-object.php
```diff
@@ -55,7 +55,7 @@
  * @method string[]|null        get_summary_map()        Gets the summary map property of the object.
  * @method array[]|null         get_tag()                Gets the tag property of the object.
  * @method string[]|string|null get_to()                 Gets the primary recipients of the object.
- * @method string               get_type()               Gets the type of the object.
+ * @method string[]|string      get_type()               Gets the type of the object.
```
obenland: Is there an example you could share for a type that's an array?

Menrath: I can't recall. But I heard someone having received such data in federation, where he would just pick the first item then. All I know is that the spec allows it. Though it's really common not to use it.

includes/activity/class-base-object.php
```diff
@@ -40,7 +40,7 @@
  * @method string[]|string|null get_in_reply_to()        Gets the objects this object is in reply to.
  * @method array|null           get_interaction_policy() Gets the interaction policy property of the object.
  * @method array|null           get_likes()              Gets the collection of likes for this object.
- * @method string|null          get_location()           Gets the physical or logical locations associated with the object.
+ * @method array|string|null    get_location()           Gets the physical or logical locations associated with the object.
```
Menrath, line 43: @obenland I am not sure: couldn't the `location` also be a `Place` object?

obenland: Yes, happy to change this

includes/activity/class-base-object.php
obenland: Hm, interesting. This might need more work than just the docs update to make it work.

I'm on the fence about leaving it until we run into a bug with this, vs documenting and not handling it.

Menrath: Totally true. But it could be a nice reminder :)

During one of the recent conferences we discussed that sometimes multi-typing might be handy, but everyone opted not to do so, because the fear is too strong too break federation with others that might not be able to handle it, or have appropriate fallbacks!

includes/activity/class-base-object.php
Menrath: I thought again. I'd vote for leaving it like that. Thinks might get too bloated then :/

includes/activity/class-base-object.php
obenland: > But it could be a nice reminder :)

My concern is more about expectations and confidence. If we document `type` as possibly being an array we should make sure all corresponding code accounts for it to avoid ambiguity in the future and avoid creating areas  of code that people don't want to touch because it could break in unexpected ways.

obenland:
```suggestion
 * @method string               get_type()               Gets the type of the object.
```
I'm no longer on the fence 🙂 Let's keep it as-is for now and account for any activities that use it when we come across them

obenland:
@Menrath Could you add a changelog message as well, following [the guidance here](https://github.com/Automattic/wordpress-activitypub/pull/2451#issuecomment-3523279835)? I'm unable to add it myself on external PRs. Should be all that's left.

pfefferle:
I think we an also add the "Skip Changelog" label, because this change is not really user-facing.

pfefferle (approved):

PR #2454, opened 2025-11-13
Add contextual skills for development workflows

obenland:
## Proposed changes:
This PR implements a comprehensive skill system to provide context-aware documentation for development workflows:

- activitypub-dev-cycle: Environment setup, testing, linting, and builds
- activitypub-php-conventions: WordPress coding standards and patterns
- activitypub-pr-workflow: PR creation and review processes
- activitypub-federation: ActivityPub protocol specification
- activitypub-testing: PHPUnit and Playwright test patterns
- activitypub-release: Version management and releases
- activitypub-integrations: Third-party plugin integration

Skills use cross-references to reduce duplication while maintaining context-appropriate information. Federation skill focuses on protocol understanding rather than implementation details.

### Other information:

## Testing instructions:

* Review the new `.claude/skills/` directory structure
* Verify each skill file has appropriate content and cross-references
* Check that `.gitignore` properly tracks skills while ignoring `*.local.*` files
* Review `CLAUDE.md` for documentation on the skills system


obenland:
@pfefferle This will need some more work and polish, there's also still a lot of overlap. I'm curious if this would be helpful to make Claude work better for us. Curious for your feedback/thoughts.

See
https://www.claude.com/blog/skills
https://www.anthropic.com/engineering/equipping-agents-for-the-real-world-with-agent-skills
https://docs.claude.com/en/docs/agents-and-tools/agent-skills/overview

pfefferle:
I’m not entirely convinced yet. I like the idea of improving how we use AI, but I’d prefer a more generic approach. We’re both using Claude right now, but that could change, and I don’t think we should have to rebuild this whole structure every time we switch models.

On top of that, the examples, class structure, and coding standards files feel like they duplicate what already exists in the codebase. The examples and structure are already reflected in the code, and our coding standards are defined in `.editorconfig` and `phpcs.xml`. Keeping separate files for all of this means they’ll need to be updated every time we make structural changes, which adds unnecessary maintenance overhead.

.claude/skills/activitypub-integrations/SKILL.md
```diff
@@ -0,0 +1,155 @@
+---
```
pfefferle, line 1: I think these files are also helpful for devs... what about placing them as README.md in the integrations folder and check how we can load that with claude?

pfefferle:
I think a lot of these files provide generally helpful information. Let's see how we can place them in the folders of the features they describe and simply point claude to these? (see my comment to the integrations file)

Maybe we can also re-use our docs folder or move some of these there?

tests/README.md
```diff
@@ -0,0 +1,240 @@
+# Testing Reference
```
pfefferle, line 1: let's have this in the tests folder.

.claude/skills/activitypub-php-conventions/examples.md
```diff
@@ -0,0 +1,799 @@
...
+- [Transformer Example](#transformer-example)
+- [Handler Example](#handler-example)
+- [REST Endpoint Example](#rest-endpoint-example)
+- [Integration Example](#integration-example)
+
+## Complete Class Example
+
+### Full-featured ActivityPub Class
+
+```php
```
pfefferle: let's check if we can link classes instead of having the duplicated code!

docs/code-linting.md
```diff
@@ -0,0 +1,346 @@
+# Linting and Code Quality Reference
```
pfefferle, line 1: this could also be copied into the docs folder.

.claude/skills/activitypub-dev-cycle/environment-setup.md
```diff
@@ -0,0 +1,397 @@
+# Environment Setup Reference
```
pfefferle: this is an even better version of https://github.com/Automattic/wordpress-activitypub/blob/trunk/docs/development-environment.md

so let's merge these and update the linking!?

obenland:
I gave this another iteration and removed about 30% of the volume in favor of references to existing docs, etc.

@pfefferle Let me know what you think?

pfefferle (approved):

PR #2499, opened 2025-11-19
Fix attachment filename generation for URLs with query parameters

obenland:
## Proposed changes:
* Strip query parameters from URLs before generating attachment filenames
* Prevents "Filename too long" filesystem errors
* Fixes issues with Instagram CDN URLs and other services that use long query strings

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

1. Process an ActivityPub activity with an attachment URL that has query parameters (e.g., Instagram CDN URL)
2. Verify the file is saved successfully without errors
3. Check that the filename is clean without query parameters

**Before:** 
```
PHP Warning: copy(.../582455906_17930138757114605_3007218277985087381_n.jpgstpdst-jpg_e35_tt6_nc_cat101ccb7-5_nc_sid18de74efgeyJlZmdfdGFnIjoiRkVFRC5iZXN0X2ltYWdlX3VybGdlbi5DMyJ9_nc_ohcne9zSzVebdsQ7kNvwGMeb6Q_nc_ocAdmzsGgwxVX3rDrTgZriddqj37P8s3eVEtAa8tNuy5DnRzJUT7nFkflSQgNVztbdQCIpM6rhfFycR1UGy1LYI2oP_nc_adz-m_nc_cid0_nc_zt23_nc_htscontent.cdninstagram.com_nc_gideIO13gOfATKzYZ-ERRIiugoh00_Afg0LbbH0emgIK49rQawuJMuNFn3fJZCkBf5zfuflpG90Qoe6922C08B): Failed to open stream: Filename too long
```

**After:** 
File saved successfully as `582455906_17930138757114605_3007218277985087381_n.jpg`

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Added - for new features
-   [ ] Changed - for changes in existing functionality
-   [ ] Deprecated - for soon-to-be removed features
-   [ ] Removed - for now removed features
-   [x] Fixed - for any bug fixes
-   [ ] Security - in case of vulnerabilities

#### Message

Fix "Filename too long" errors when downloading attachments from URLs with query parameters (e.g., Instagram CDN URLs).

</details>

PR #2500, opened 2025-11-19
Add WP-CLI command to reprocess inbox items

obenland:
## Proposed changes:
* Add `wp activitypub reprocess_inbox` command for local development
* Delete existing comments/posts before reprocessing to ensure clean state
* Enable `activitypub_create_posts` option in local development for testing

### Other information:

## Testing instructions:

1. Set up local development environment with WP-CLI
2. Process an ActivityPub activity to create an inbox item
3. Run `wp activitypub reprocess_inbox <post_id>` where `<post_id>` is an ap_inbox post ID
4. Verify that:
   - Existing comments/posts are deleted
   - The activity is reprocessed
   - New comments/posts are created with the original timestamps

**Example usage:**
```bash
# Reprocess inbox item with ID 123
wp activitypub reprocess_inbox 123
```

**Expected output:**
```
Reprocessing inbox item 123...
Deleted comment 456 (source_id: object ID)
Deleted 1 comment(s) and 0 post(s).
Success: Inbox item 123 has been reprocessed.
```

### Changelog entry

### DIFF
--- a/local/class-cli.php
+++ b/local/class-cli.php
@@ -227,11 +227,11 @@ class Cli extends \WP_CLI_Command {
 		if ( isset( $activity_data['object'] ) && is_array( $activity_data['object'] ) && isset( $activity_data['object']['id'] ) ) {
 			$object_id = $activity_data['object']['id'];
 
-			$comment = Comment::object_id_to_comment( $object_id );
-			if ( $comment ) {
-				\wp_delete_comment( $comment->comment_ID, true );
+			$object_comment = Comment::object_id_to_comment( $object_id );
+			if ( $object_comment ) {
+				\wp_delete_comment( $object_comment->comment_ID, true );
 				++$deleted_comments;
-				\WP_CLI::log( sprintf( 'Deleted comment %d (source_id: object ID)', $comment->comment_ID ) );
+				\WP_CLI::log( sprintf( 'Deleted comment %d (source_id: object ID)', $object_comment->comment_ID ) );
 			}
 
 			// Delete ap_post with GUID matching the object ID.
--- a/local/class-cli.php
+++ b/local/class-cli.php
@@ -210,42 +210,48 @@ class Cli extends \WP_CLI_Command {
 			\WP_CLI::error( 'Failed to decode activity data.' );
 		}
 
-		// Delete existing comments created from this activity.
+		// Only delete existing artifacts for activity types that create new ones.
+		$should_delete = isset( $activity_data['type'] ) && in_array( $activity_data['type'], array( 'Create', 'Like', 'Announce' ), true );
+
 		$deleted_comments = 0;
 		$deleted_posts    = 0;
 
-		// Delete comments with source_id matching the activity ID.
-		$comment = Comment::object_id_to_comment( $activity_id );
-		if ( $comment ) {
-			\wp_delete_comment( $comment->comment_ID, true );
-			++$deleted_comments;
-			\WP_CLI::log( sprintf( 'Deleted comment %d (source_id: activity ID)', $comment->comment_ID ) );
-		}
-
-		// Delete comments and posts with source_id/GUID matching the activity object ID.
-		if ( isset( $activity_data['object'] ) && is_array( $activity_data['object'] ) && isset( $activity_data['object']['id'] ) ) {
-			$object_id = $activity_data['object']['id'];
-
-			$object_comment = Comment::object_id_to_comment( $object_id );
-			if ( $object_comment ) {
-				\wp_delete_comment( $object_comment->comment_ID, true );
+		if ( $should_delete ) {
+			// Delete comments with source_id matching the activity ID.
+			$comment = Comment::object_id_to_comment( $activity_id );
+			if ( $comment ) {
+				\wp_delete_comment( $comment->comment_ID, true );
 				++$deleted_comments;
-				\WP_CLI::log( sprintf( 'Deleted comment %d (source_id: object ID)', $object_comment->comment_ID ) );
+				\WP_CLI::log( sprintf( 'Deleted comment %d (source_id: activity ID)', $comment->comment_ID ) );
 			}
 
-			// Delete ap_post with GUID matching the object ID.
-			$ap_post = Posts::get_by_guid( $object_id );
-			if ( ! \is_wp_error( $ap_post ) ) {
-				\wp_delete_post( $ap_post->ID, true );
-				++$deleted_posts;
-				\WP_CLI::log( sprintf( 'Deleted ap_post %d (GUID: object ID)', $ap_post->ID ) );
-			}
-		}
+			// Delete comments and posts with source_id/GUID matching the activity object ID.
+			if ( isset( $activity_data['object']['id'] ) && is_array( $activity_data['object'] ) ) {
+				$object_id = $activity_data['object']['id'];
 
-		if ( $deleted_comments > 0 || $deleted_posts > 0 ) {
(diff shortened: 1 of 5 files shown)


local/class-cli.php
```diff
@@ -168,4 +171,87 @@ private function generate_comments( $args, $assoc_args ) {
...
+				\WP_CLI::log( sprintf( 'Deleted comment %d (source_id: object ID)', $comment->comment_ID ) );
+			}
+
+			// Delete ap_post with GUID matching the object ID.
+			$ap_post = Posts::get_by_guid( $object_id );
+			if ( ! \is_wp_error( $ap_post ) ) {
+				\wp_delete_post( $ap_post->ID, true );
+				++$deleted_posts;
+				\WP_CLI::log( sprintf( 'Deleted ap_post %d (GUID: object ID)', $ap_post->ID ) );
+			}
```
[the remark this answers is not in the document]
obenland: At that point it should create a fatal error so we can fix it.

local/class-cli.php
```diff
@@ -168,4 +171,87 @@ private function generate_comments( $args, $assoc_args ) {
...
+		}
+
+		if ( $deleted_comments > 0 || $deleted_posts > 0 ) {
+			\WP_CLI::log( sprintf( 'Deleted %d comment(s) and %d post(s).', $deleted_comments, $deleted_posts ) );
+		} else {
+			\WP_CLI::log( 'No existing comments or posts found to delete.' );
+		}
+
+		// Call the process_create_item method to reprocess.
+		Actors_Inbox_Controller::process_create_item( $activity_id );
```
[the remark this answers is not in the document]
obenland: This is on purpose. We don't want to skip on a duplicated item.

local/class-cli.php
```diff
@@ -168,4 +171,87 @@ private function generate_comments( $args, $assoc_args ) {
...
+
+		// Delete comments and posts with source_id/GUID matching the activity object ID.
+		if ( isset( $activity_data['object'] ) && is_array( $activity_data['object'] ) && isset( $activity_data['object']['id'] ) ) {
+			$object_id = $activity_data['object']['id'];
+
+			$comment = Comment::object_id_to_comment( $object_id );
+			if ( $comment ) {
+				\wp_delete_comment( $comment->comment_ID, true );
+				++$deleted_comments;
+				\WP_CLI::log( sprintf( 'Deleted comment %d (source_id: object ID)', $comment->comment_ID ) );
```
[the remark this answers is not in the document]
obenland: Fixed in [f390ee1](https://github.com/Automattic/wordpress-activitypub/pull/2500/commits/f390ee1315d636f19fa9267a6a489d80764178ea).

pfefferle (changes requested):
I would assume that a method called `reprocess` would reprocess any kind of Inbox Item, not only `Create`s (aside from that, the current implementation would handle any kind of Item as a `Create`). The Inbox stores every incoming Activity (except `Delete`s), so a reprocess has to take care of properly reprocessing `Announce`s, `Like`s, `Undo`s and all other types, the plugin supports.

To have a proper reprocess we should simply call the same `do_action` that is triggered by the inbox, to run through the same detection and handler process, a normal Activity would run through.

local/class-cli.php
```diff
@@ -168,4 +171,87 @@ private function generate_comments( $args, $assoc_args ) {
...
+		if ( ! $activity_data ) {
+			\WP_CLI::error( 'Failed to decode activity data.' );
+		}
+
+		// Delete existing comments created from this activity.
+		$deleted_comments = 0;
+		$deleted_posts    = 0;
+
+		// Delete comments with source_id matching the activity ID.
+		$comment = Comment::object_id_to_comment( $activity_id );
```
pfefferle: There is no check if the Activity is a `Create` or `Update`!

obenland: Fixed in [531d041](https://github.com/Automattic/wordpress-activitypub/pull/2500/commits/531d0418fa7785a728004637f9ed9ebcbc1720ab)

local/class-cli.php
```diff
@@ -168,4 +170,110 @@ private function generate_comments( $args, $assoc_args ) {
...
+			\WP_CLI::error( 'Failed to decode activity data.' );
+		}
+
+		// Only delete existing artifacts for activity types that create new ones.
+		$should_delete = isset( $activity_data['type'] ) && in_array( $activity_data['type'], array( 'Create', 'Like', 'Announce' ), true );
+
+		$deleted_comments = 0;
+		$deleted_posts    = 0;
+
+		if ( $should_delete ) {
```
pfefferle: why deleting existing comments/posts? the plugin should be able to handle these cases automatically.

local/class-cli.php
```diff
@@ -168,4 +170,110 @@ private function generate_comments( $args, $assoc_args ) {
...
+			}
+		} else {
+			\WP_CLI::log( sprintf( 'Skipping deletion for %s activity type.', $activity_data['type'] ) );
+		}
+
+		// Bypass signature verification for internal reprocessing.
+		\add_filter( 'activitypub_defer_signature_verification', '__return_true' );
+
+		// Create internal REST request to the shared inbox endpoint.
+		$request = new \WP_REST_Request( 'POST', '/' . \ACTIVITYPUB_REST_NAMESPACE . '/inbox' );
```
pfefferle: I am not sure if this is always possible. Some Activities could have been sent to the Actors inbox with an empty audience. In these cases, the Activity wouldn't be reprocessed from the shared-inbox.

obenland: Fixed in [80f4b6a](https://github.com/Automattic/wordpress-activitypub/pull/2500/commits/80f4b6a753fbc7ff75de8650a07e790a22c279e8).

local/class-cli.php
obenland: It doesn't seem to. When reprocessing create activities, ap_posts don't get updated. Not sure if this is a bug, but for now deleting does what I need this command to do.

obenland:
@pfefferle Thanks for your feedback, I think this is in a better shape than it was last night.

This is about as much time I'd want to invest in this PR. If there are any additional points of feedback, I think we're quickly approaching a negative return for time invested, and I'd want to find a different approach.

pfefferle:
> ...and I'd want to find a different approach.

What is the purpose of the PR?

obenland:
To allow reprocessing of inbox items in local development to test changes in ap_post parsing. Test if changes like in #2499 are working.

pfefferle:
> This is about as much time I'd want to invest in this PR. If there are any additional points of feedback, I think we're quickly approaching a negative return for time invested, and I'd want to find a different approach.

I’m not so sure about that then. I think the time investment is still worthwhile! If the reprocessing isn’t working correctly, then an `Update` Activity from a remote server won’t work properly either.

pfefferle:
@obenland can you provide me some examples to test why the reprocess only work with the deletion?

obenland:
The way I tested it was to change an ap_post item, remove some content for example, then reprocess its inbox item and see if the content was restored.

pfefferle:
@obenland found the issue! will work on a fix!

pfefferle:
@obenland should be fixed now!

obenland:
@pfefferle Yes, looks good. Has some unit tests failing now.

You'll have to accept the PR when you're happy with it ;)

pfefferle:
Now that I’ve seen the effects of the restructuring, I’m not sure “happy with it” is the right way to put it. :(

pfefferle (approved):

PR #2501, opened 2025-11-19
Simplify post creation logic in Create handler

pfefferle:
We should always store posts with the release of the reader.

Removed redundant option check for 'activitypub_create_posts' when handling non-interaction objects. Now, post creation is attempted by default if the activity is not a reply or quote.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


pfefferle:
@obenland Maybe this also solves the issue you have with the empty timeline when testing the reader on your Atomic site.

obenland (approved):

PR #2506, opened 2025-11-20
Add featured image support to Social Web reader

pfefferle:
## Summary
Adds featured image display capability to the ActivityPub Social Web reader feed view.

## Changes
- **Backend**: Store first image from ActivityPub attachments as `_activitypub_featured_image_url` post meta
- **REST API**: Register `featured_image` field on `ap_post` post type
- **Frontend**: New `featuredImageField` component for DataViews
- **Types**: Add `featured_image` to `FeedPost` interface and `useFeed` hook defaults
- **Styling**: Flex-based layout with image positioned on right side via CSS

## Implementation Details
- Featured images are cached locally during attachment import process
- Stored in post meta for efficient retrieval (no filesystem scanning)
- Meta is cleaned up on post updates
- Users can enable/disable the image field via DataViews properties panel
- Image appears as first field in field order, styled to display on the right

## Testing
1. Follow an ActivityPub actor that posts images
2. View the Social Web feed page
3. Enable the "Image" field in DataViews properties
4. Verify images display on the right side of feed items

## Checklist
- [x] Code follows WordPress coding standards
- [x] Changes are properly documented
- [x] TypeScript types are updated
- [x] Build succeeds without errors


obenland:
Can you help me understand the benefit of designating a featured image?

pfefferle:
that was a feature that was requested at the meetup! to have at least one image show up in the stream view.

/cc @jeherve

obenland:
Have at least one picture show up when excerpt is selected?

pfefferle:
yes

pfefferle:
Updated the featured image implementation:

- Featured image is now displayed within the excerpt field instead of as a separate column
- Image appears below the excerpt text with proper spacing
- Image sizing: 300px max width/height, maintains aspect ratio
- Only shows in excerpt view, not in content view
- Responsive design with auto width/height

pfefferle:
## Copilot Feedback Review

I've reviewed all 7 Copilot comments on this PR. Here's the status:

### ✅ Fixed

**HTML entities in alt text (excerpt/index.tsx:36)**
- Valid accessibility issue
- Fixed by adding `decodeEntities()` to the image alt attribute
- Ensures titles with entities like `&amp;`, `&quot;` are properly displayed to screen readers

### ❌ Not Applicable

The remaining 6 comments reference the standalone `featured-image` field component, which has been removed from this implementation. The featured image is now integrated directly into the excerpt field instead of being a separate column. The referenced code is no longer used or exported.

**Specific comments not applicable:**
1. Browser support for `:has()` pseudo-class - references unused component
2. Duplication between featured-image and excerpt components - excerpt now handles both
3. Unused `register()` call return value - standard REST API pattern, not an issue
4. Featured image width constraints - component no longer in use
5. REST API response structure - standard WordPress pattern
6. Additional duplication comment - resolved by integration approach

The new implementation displays the featured image within the excerpt field, maintaining aspect ratio with a 300px max dimension constraint. This provides a cleaner UI with the image positioned below the excerpt text.

obenland:
@jeherve @pfefferle Can you help me understand the need for this?

> to have at least one image show up in the stream view.

Why wouldn't I use the content view in that case?

The more I use the app myself, the more I wonder if we should remove the excerpt view all-together. It looks pretty bad for Note content.

pfefferle:
I am quite the opposite. some full blog post breaks the readability a lot!

obenland:
Sure, but I wouldn't expect those to show up as Notes? And if they do that's probably something we should address with the platform they use.

Maybe we should add that short form/long form switch sooner rather than later.

pfefferle:
Are we talking about the same thing? This is about having the excerpt in the main view not?

What has this todo with the object-type filter?

pfefferle:
> Maybe we should add that short form/long form switch sooner rather than later.

what is that?

includes/collection/class-posts.php
```diff
@@ -325,7 +326,22 @@ private static function maybe_import_attachments( $activity_object, $post_id ) {
...
+					if ( isset( $file['mime_type'] ) && 0 === strpos( $file['mime_type'], 'image/' ) ) {
+						\update_post_meta(
+							$post_id,
+							'_activitypub_featured_image_url',
+							$file['url']
+						);
+						break; // Only save the first image.
+					}
+				}
+			}
```
obenland, line 344: This should happen as part of `Attachments::import_post_files()`

includes/class-post-types.php
```diff
@@ -406,6 +406,17 @@ public static function register_post_post_type() {
...
+		\register_post_meta(
+			Posts::POST_TYPE,
+			'_activitypub_featured_image_url',
+			array(
+				'type'              => 'string',
+				'single'            => true,
+				'description'       => 'URL of the featured image from ActivityPub attachments.',
+				'sanitize_callback' => 'sanitize_url',
+			)
+		);
```
obenland, line 419: Instead of using custom storage we should add `'thumbnail'` support to the custom post type registration and re-use Core APIs.

src/social-web/components/fields/content/index.tsx
```diff
@@ -47,13 +48,37 @@ export const contentField: Field< FeedPost > = {
...
 						dangerouslySetInnerHTML={ { __html: content || '<p>\u00A0</p>' } }
 					/>
+					{ hasFeaturedImage && (
+						<img
+							src={ item.featured_image }
+							alt={ decodeEntities( item.title?.rendered || __( 'Post image', 'activitypub' ) ) }
+							className="activitypub-feed-content__image"
+							loading="lazy"
+						/>
+					) }
```
obenland, line 58: We shouldn't add a featured image to Notes, it'll likely duplicate an image that gets displayed as part of the content.

src/social-web/components/fields/content/index.tsx
```diff
@@ -47,13 +48,37 @@ export const contentField: Field< FeedPost > = {
...
+			return <div className="activitypub-feed-excerpt">{ '\u00A0' }</div>;
+		}
+
+		return (
+			<div className="activitypub-feed-excerpt">
+				{ plainText && <div className="activitypub-feed-excerpt__text">{ plainText }</div> }
+				{ hasFeaturedImage && (
+					<img
+						src={ item.featured_image }
+						alt={ decodeEntities( item.title?.rendered || __( 'Post image', 'activitypub' ) ) }
```
obenland:
```suggestion
						alt={ decodeEntities( item.title?.rendered || __( 'Post Thumbnail', 'activitypub' ) ) }
```

src/social-web/components/fields/featured-image/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+	label: __( 'Image', 'activitypub' ),
+	enableHiding: true,
+	enableSorting: false,
+	getValue: ( { item }: { item: FeedPost } ) => item.featured_image || '',
+	render: ( { item }: { item: FeedPost } ) => {
+		if ( ! item.featured_image ) {
+			return null;
+		}
+
+		const altText = item.title?.rendered || __( 'Post image', 'activitypub' );
```
obenland:
```suggestion
		const altText = item.title?.rendered || __( 'Post Thumbnail', 'activitypub' );
```

obenland (commented):
I still don't think this is needed. If you feel strongly about having it we should not process them and display them for Note types, though.

src/social-web/components/fields/content/index.tsx
pfefferle: Yes, that was an unfinalized port from the excerpt/content system to the new "magic" content system.

includes/class-post-types.php
pfefferle: but doesn't that need an attachment?

PR #2510, opened 2025-11-20
Add type filter for feed based on ap_object_type taxonomy

pfefferle:
## Summary
Adds a button-based filter to the ActivityPub feed that allows users to filter posts by object type (Article, Note, etc.) using the `ap_object_type` taxonomy.

## Changes
- Add `ap_object_type` field to `FeedPost` TypeScript interface
- Add `apObjectType` parameter to `useFeed` hook with REST API integration
- Create `TypeFilter` component using WordPress ButtonGroup
- Integrate TypeFilter into feed stage with localStorage persistence
- Filter buttons appear in DataViews header, aligned to the right

## Features
- **Button-based UI**: Clean ButtonGroup interface with "All", "Article", "Note" buttons
- **localStorage persistence**: Selected filter is remembered across browser sessions
- **REST API integration**: Properly filters posts via the `ap_object_type` taxonomy parameter
- **Reactive updates**: Feed updates immediately when filter selection changes

## Test plan
- [ ] View the feed in the ActivityPub Social Web reader
- [ ] Verify filter buttons appear in the header on the right side
- [ ] Click "Article" button and verify only Article posts are shown
- [ ] Click "Note" button and verify only Note posts are shown
- [ ] Click "All" button and verify all posts are shown
- [ ] Refresh the page and verify the selected filter persists
- [ ] Clear localStorage and verify it defaults to "All"

## Screenshots

Screenshot 2025-11-20 at 18 36 35


obenland:
Nice! Is that something we could add to the DataView settings?

pfefferle:
I tried... will give it another 30 mins...

pfefferle:
@obenland I am not sure if I like that more!? https://github.com/Automattic/wordpress-activitypub/pull/2510/commits/025afc94734e0ee833efbcde0a6b8a7d54664bc8

pfefferle:
Might be nicer if we have some more filters in the future (single actor feeds could maybe be handled as filter too)

obenland:
Hm, it looks like I don't have any articles…

obenland:
Are articles being associated correctly in the inbox? Did you create your test-articles manually?

I can't get the selector to show up, which has me wondering if that's because I only have note posts

pfefferle:
I have added some Articles, by hand and through the inbox.

obenland:
Hm, I do have one article post…

obenland:
oh! I was looking in the wrong place

pfefferle:
@obenland I just re-tested with my blogs and it works as expected.

src/social-web/routes/feed/stage.tsx
```diff
@@ -53,6 +54,11 @@ export default function FeedStage( { onSelectItem }: FeedStageProps ) {
 		[]
 	);
 
+	// Fetch ap_object_type taxonomy terms for filter
+	const { records: apObjectTypes } = useEntityRecords( 'taxonomy', 'ap_object_type', {
+		per_page: 100,
```
obenland: We could probably drop that? I assume we'd only ever expect 2 types to be returned? Definitely less than 10?

We could also consider implementing the `elements` array as a static array to avoid this request and make the labels translatable.

src/social-web/routes/feed/stage.tsx
```diff
@@ -125,18 +131,44 @@ export default function FeedStage( { onSelectItem }: FeedStageProps ) {
...
+	}, [ view.filters ] );
+
 	const { feed, isResolving, totalItems, totalPages } = useFeed( {
 		perPage: view.perPage || 20,
 		page: view.page || 1,
 		orderBy: view.sort?.field || 'date',
 		order: view.sort?.direction || 'desc',
 		search: view.search || '',
 		userId: activeActorId,
+		apObjectType: apObjectTypeFilter,
```
obenland:
```suggestion
		filters: view.filters,
```

Should we pass all `view.filters` here and let `useFeed()` dissect them?

src/social-web/routes/feed/stage.tsx
```diff
@@ -125,18 +131,44 @@ export default function FeedStage( { onSelectItem }: FeedStageProps ) {
...
+			label: __( 'Type', 'activitypub' ),
+			enableHiding: false,
+			enableSorting: false,
+			elements:
+				apObjectTypes?.map( ( term: any ) => ( {
+					value: term.id,
+					label: term.name,
+				} ) ) || [],
+			filterBy: {
+				operators: [ 'isAny' ],
```
obenland:
```suggestion
				operators: [ 'is' ],
```

We'd probably only want one to be selected at any time?

src/social-web/routes/feed/stage.tsx
```diff
@@ -125,18 +131,44 @@ export default function FeedStage( { onSelectItem }: FeedStageProps ) {
...
+			},
```
obenland:
```suggestion
				render: () => null,
			},
```
Claude felt like it was important to define this explicitly. Not sure if it actually is.

src/social-web/routes/feed/stage.tsx
```diff
@@ -125,18 +131,44 @@ export default function FeedStage( { onSelectItem }: FeedStageProps ) {
...
+		} ),
+		[ apObjectTypes ]
+	);
```
obenland: Let's move `apObjectTypeField` into its own field component, alongside the others.

src/social-web/routes/feed/stage.tsx
pfefferle: I am not sure about that, but let's start with one

pfefferle: It does have to be isAny... a taxonomy seem to not work otherwise:

> 'is' operator:
> - Used for single-value comparisons
>  - Matches when the field value exactly equals the filter value
>  - Best for boolean or single-select filters
>
>  'isAny' operator:
>  - Used for multi-value/enum comparisons
>  - Matches when the field value is any one of the selected filter values
>  - Required when you have elements defined (multiple options to choose from)
>  - Allows users to select multiple types (e.g., filter for "Article OR Note")
>
>  In your case, since you defined elements with multiple object types (Article, Note, etc.) and users can potentially select multiple types to filter by, WordPress DataViews expects the 'isAny' operator to
  properly:
>
>  1. Display the filter UI with checkboxes for multiple selections
>  2. Apply the filter logic correctly (show posts that match ANY of the selected types)
>  3. Store and persist the filter state properly
>
>  With 'is', DataViews likely didn't recognize it as a valid enum-style filter and either didn't display the filter UI at all or didn't persist the selection correctly. The 'isAny' operator tells DataViews: "this
>   field has predefined options (elements) and we want to filter by any matching value."

includes/class-post-types.php
```diff
@@ -706,6 +706,17 @@ function ( $params ) {
 					'sanitize_callback' => 'absint',
 				);
 
+				$params['ap_object_type'] = array(
+					'description'       => __( 'Filter posts by ActivityPub object type.', 'activitypub' ),
```
obenland:
```suggestion
					'description'       => 'Filter posts by ActivityPub object type.',
```

includes/class-post-types.php
```diff
@@ -706,6 +706,17 @@ function ( $params ) {
...
+					'type'              => 'array',
+					'items'             => array(
+						'type' => 'integer',
+					),
+					'sanitize_callback' => function ( $value ) {
+						return array_map( 'absint', (array) $value );
+					},
```
obenland: Is the built-in sanitization not taking care of this? We could set `'minimum' => 0` if we want to avoid negative numbers. Alternatively we could validate the passed value by checking if an ap_object_type term with that ID exists

obenland (commented):
In bbfe23d0c002569e74d36c2c710de7b347a06c1a I moved all of the entity stuff into the field definition, too, and refactored it to enable single type selections, I hope that's okay. I meant to provide all that stuff as feedback but it got to be a lot and pretty intricate.

The main missing thing is making these term names more accessible. `Article` and `Note` are technical terms (we literally save what activities send in their type), and not necessarily the best way to describe what we're trying to convey. I'm also not sure they're translated or translatable?

obenland:
In b2d41d2c, updated to limit the filter to notes and articles and make the labels translatable. I'm not sure "Short form" and "Long form" are the best labels, happy for us to change those. It doesn't seem to be easily possible to not add the filter when there's only one term.

pfefferle:
I am not a fan of limiting the types! This is a similar pattern, Mastodon used (or still uses) an what made `Article`s a second class object on the Fediverse!

I will think about a way to better support all kinds of Objects and to translate at least most of them.

pfefferle:
@obenland Added some small changes to have best of both worlds. I also changed the translatable text a bit... I think "Long form" and "Short form" are no common (end-user friendly) terms...

obenland (approved):

PR #2515, opened 2025-11-20
Remove manual data cleanup from PHPUnit tests

obenland:
## Summary

This PR removes all manual `wp_delete_post()`, `wp_delete_comment()`, and `wp_delete_user()` calls from test methods and teardown functions across 45 test files.

I think performance gains will be marginal but it should help with avoiding adding more of this through AI following existing practices.

## Why?

The WordPress test framework's `WP_UnitTestCase_Base::tear_down_after_class()` automatically calls `_delete_all_data()` which handles cleanup of:
- All posts
- All comments  
- All users
- All terms
- All other test data

Manual cleanup is redundant.

## Changes

- **45 files modified**
- **532 lines removed** containing manual delete calls
- **10 lines added** (formatting adjustments)
- All cleanup now handled automatically by WordPress test framework

## Testing

- ✅ All 1372 tests pass
- ✅ No deprecation warnings
- ✅ Code linter passes
- ✅ Test execution time: ~28 seconds

## Benefits

1. **Cleaner code**: Removes hundreds of lines of redundant cleanup
2. **Improved reliability**: Consistent cleanup behavior across all tests
3. **Easier maintenance**: No need to remember manual cleanup in new tests
4. **Future-proof**: Framework cleanup handles all data types automatically

### DIFF
--- a/tests/phpunit/tests/includes/activity/class-test-activity.php
+++ b/tests/phpunit/tests/includes/activity/class-test-activity.php
@@ -225,8 +225,5 @@ class Test_Activity extends \WP_UnitTestCase {
 		$activity->set_object( $activity_object );
 
 		$this->assertContains( 'https://example.com/author/456', $activity->get_to() );
-
-		// Delete user.
-		\wp_delete_user( $user_id );
 	}
 }
--- a/tests/phpunit/tests/includes/class-test-activitypub.php
+++ b/tests/phpunit/tests/includes/class-test-activitypub.php
@@ -63,8 +63,6 @@ class Test_Activitypub extends \WP_UnitTestCase {
 			)
 		);
 		$this->assertEquals( $expected, \get_post_meta( $post_id, $meta_key, true ) );
-
-		\wp_delete_post( $post_id, true );
 	}
 
 	/**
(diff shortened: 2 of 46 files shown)


pfefferle:
It looks like you're taking the opposite approach with test data compared to filters: https://github.com/Automattic/wordpress-activitypub/pull/2522

Shouldn’t we be consistent, either explicitly adding and removing everything, or removing everything by default?

obenland:
There's a difference in how data and filters need to be handled, as hooks are dependent on the WordPress request lifecycle, whereas data is not.

Core- and plugin-registered hooks need to be preserved between test classes. #2522 aims to reduce the risk of accidentally removing hooks that shouldn't be removed by replacing `remove_all_actions()` and `remove_all_filters()` calls with callback-specific ones. It's essentially an extension of #2414.

Data on the other hand can be (and in many cases needs to be) deleted between classes. [`WP_UnitTestCase_Base::tear_down_after_class()`](https://github.com/WordPress/wordpress-develop/blob/cd301d0c833e6d69ab8b6502c1820b7190085d73/tests/phpunit/includes/abstract-testcase.php#L91) handles that by default for all data that was added during the run of individual test classes.

So I think these two PRs do create consistency, removing everything by default for data, and explicitly adding and removing everything for hooks.

pfefferle (dismissed):

pfefferle (dismissed):

pfefferle (approved):

PR #2523, opened 2025-11-21
Fix infinite scroll bugs: clickable items and page reload

obenland:
## Summary
Fixes two critical bugs in the infinite scroll implementation for the social-web feed while eliminating an infinite update loop.

## Changes
- **Fix unclickable items during loading**: Removed `isLoadingMore` from DataViews `isLoading` prop so items on page 1 remain clickable while page 2+ is loading
- **Fix page reload behavior**: Don't persist `paged` URL parameter when infinite scroll is enabled, ensuring reloads always start from page 1 (standard infinite scroll UX)
- **Fix selection cleanup**: Check `allLoadedRecords` instead of `feed` to prevent premature selection clearing during page transitions
- **Remove infinite loop**: Removed actor switching effect that was causing infinite update loop due to `normalizedView` creating new object references on every render

## Known Issue
The actor switching state cleanup (Bug 1) has been removed to prevent the infinite loop. This will need to be addressed next with a different approach that doesn't trigger the loop.

## Test Plan
- [x] Verify no "Maximum update depth exceeded" error in console
- [x] Verify items on page 1 remain clickable while scrolling to load page 2
- [x] Verify page reload on page 3 returns to page 1
- [ ] Test user switching behavior (known to have issues)


obenland:
@pfefferle I think I'd just close this and create a new one based on tanstack router

PR #2529, opened 2025-11-24
Optimize Social Web reader for mobile devices

pfefferle:
### Mobile Navigation (WordPress Site Editor Pattern)
* Implement progressive one-region-at-a-time navigation on mobile devices (screens < 782px)
* Navigation flow: Sidebar → Stage → Inspector
* Hash-based routing for mobile state management (`#/`, `#/feed`, `#/feed/123`)
* Site-hub logo/icon functions as back button to navigate between levels
* Object type filters and hashtags navigate to feed stage on mobile after selection
* Mobile header uses dark background throughout (matches sidebar theme)

### Mobile Optimizations
* Make inspector full-screen on mobile devices
* Optimize feed/stream view for smaller screens with better spacing and typography
* Hide sidebar and stage automatically when inspector is open on mobile
* Improve text readability with responsive font sizes and hyphenation
* Make all images responsive within content areas
* Move mobile headers outside themed Panel surfaces for consistent styling

### Technical Implementation
* Hash change event listeners for bidirectional navigation
* CSS data attributes (`data-mobile-view`) for mobile state management
* Generic CSS selectors for contextual mobile header styling
* Lazy-loaded route components for performance
* WordPress 782px breakpoint for mobile detection

| Site-Hub | Stage | Inspector |
|---|---|---|
| Screenshot 2025-11-27 at 12 29 01 | Screenshot 2025-11-27 at 12 29 04 | Screenshot 2025-11-27 at 12 29 08 |

### Other information:

### Mobile Navigation Flow
* Open the Social Web reader on a mobile device or resize browser to < 782px width
* Initial load should show sidebar with navigation menu, object type filters, and popular tags
* Click on an object type filter or tag:
  - Should navigate to feed stage view
  - Site-hub header should appear at top with back button
  - Feed should show filtered content
* Click site-hub logo/icon:
  - Should navigate back to sidebar view
* Click on a post in the feed:
  - Should navigate to inspector view (full screen)
  - Site-hub header should show back button
* Click site-hub logo/icon from inspector:
  - Should navigate back to feed stage
* Click site-hub logo/icon from stage:
  - Should navigate back to sidebar

### General Mobile Testing
* Verify inspector takes up full screen on mobile
* Verify sidebar and feed are properly hidden when not active
* Verify content is readable with appropriate spacing
* Test on different mobile screen sizes (iPhone, Android, tablets)
* Verify all touch targets are appropriately sized
* Check that browser back/forward buttons work correctly with hash navigation
* Verify dark theme is consistent throughout all mobile views

### Changelog entry

No changelog entry required for this PR.


obenland:
Mobile views should work like they do for the Site Editor, where it only shows the site hub at the top.

.github/changelog/2529-from-description
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Optimize Social Web reader for mobile devices with WordPress Site Editor-style navigation: progressive one-region-at-a-time view (Sidebar → Stage → Inspector) with hash-based routing and back button navigation.
```
obenland, line 4:
```suggestion
```

We don't need a changelog here.

src/social-web/components/site-hub/index.tsx
```diff
@@ -71,9 +104,7 @@ function SiteHub() {
 							size="compact"
 							className="site-hub__command-button"
 							icon={ search }
-							onClick={ () => openCommandCenter() }
```
obenland, line 74: No more command palette? This might not be intentional?

src/social-web/components/site-hub/index.tsx
```diff
@@ -7,24 +7,28 @@
...
 import SiteIcon from '../site-icon';
 import './style.scss';
 
-function SiteHub() {
+interface SiteHubProps {
+	onNavigateBack?: () => void;
+	selectedItemId?: string | number | null;
+}
+
+function SiteHub( { onNavigateBack, selectedItemId }: SiteHubProps = {} ) {
```
obenland: The SiteHub changes feel quite drastic. Would a `SiteHubMobile` component like in site-edit be helpful here? As a separate component we'd also be able to render it conditionally in JS vs through CSS.

It might also be helpful to not separate out the Sidebar component, but defined the sidebar in Layout with nested components.

src/social-web/components/layout/style.scss
```diff
@@ -51,21 +56,63 @@
...
+		> .site-hub {
+			display: none; // No header in sidebar view
+		}
+		.sidebar-region {
+			display: flex;
+		}
+		.stage-region {
+			display: none;
+		}
+	}
```
obenland: Let's use `useViewportMatch` from `@wordpress/compose` to conditionally render the right components in JS, rather than hiding them through CSS. It reduces memory, is easier to debug, and avoids rendering components that are invisible.

src/social-web/components/layout/index.tsx
```diff
@@ -185,13 +210,27 @@ export function Layout() {
 
 	const showInspector = !! selectedItemId;
 
+	// Determine mobile view state for CSS classes
+	// Root: showing sidebar only (no section navigated to yet)
+	// Stage: showing stage (section selected, no item)
+	// Inspector: showing inspector (item selected)
+	const isAtRoot = checkIsAtRoot();
+	const mobileView = showInspector ? 'inspector' : ! isAtRoot ? 'stage' : 'sidebar';
```
obenland: This is also something where `useViewportMatch` would be helpful.

src/social-web/components/site-hub/index.tsx
```diff
@@ -34,16 +38,45 @@ function SiteHub() {
...
+	useEffect( () => {
+		const handleHashChange = () => {
+			setCurrentHash( window.location.hash );
+		};
+
+		window.addEventListener( 'hashchange', handleHashChange );
+		return () => {
+			window.removeEventListener( 'hashchange', handleHashChange );
+		};
+	}, [] );
```
obenland: I think we found the limits of our simple router setup.

Asking to switch to Tanstack before moving forward with these styles feels like a lot, but this introduces quite a bit of tech debt otherwise. What do you think?

obenland (commented):
Functionally a big step forward, nice work!

It also uncovers quite a bit of deferred fundamentals we should probably address before moving forward with this. Mainly a more robust routing setup and possibly giving Layout more say over how Sidebar gets rendered. I think [`edit-site`'s Layout component](https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-site/src/components/layout/index.js) would be a good example to follow.

obenland:
Thanks for making these updates!

I'm still a bit conflicted. Prop drilling keeps making returns, like with `onNavigateBack` in `SiteHub` or `onNavigate` in `Sitebar`, `ObjectTypes`, and `PopularTags`, and we're still manually navigating to make it all work. Maybe this approach, and by extension orienting ourselves at edit-site, is not the best way forward?

I took a step back and took another look at boot's approach, which resulted in #2606. It has a little bit of a different UX in how it handles the sidebar, but might make for a more future-proof direction? I'm curious what you think

pfefferle:
> I'm still a bit conflicted. Prop drilling keeps making returns, like with onNavigateBack in SiteHub or onNavigate in Sitebar, ObjectTypes, and PopularTags, and we're still manually navigating to make it all work. Maybe this approach, and by extension orienting ourselves at edit-site, is not the best way forward?

I still do not fully know what I am doing, so I think even that approach can be optimized.

> I took a step back and took another look at boot's approach, which resulted in https://github.com/Automattic/wordpress-activitypub/pull/2606. It has a little bit of a different UX in how it handles the sidebar, but might make for a more future-proof direction? I'm curious what you think

I like it! I am also happy to go that path. I simply took your credo "It should fit into the look and feel of WordPress" as "requirement" and this PR is the outcome of "do it like the site-editor is doing it". And that is the only point against the other approach: It does not look like anything else in the current WP-Admin!

obenland:
> that is the only point against the other approach: It does not look like anything else in the current WP-Admin!

Yeah fair, the promise there is that it'll make upgrading to using Boot easier and visually more stable. 🤞

pfefferle:
closed in favor of #2606.

PR #2535, opened 2025-11-25
Reader: Alignment experiments

pfefferle:
## Summary
Ensures DataViews list items maintain a consistent width only when the inspector is closed, preventing content from expanding to fill all available space.

Screenshot 2025-11-25 at 12 26 30

## Changes
- Added `max-width: 50%` to `.dataviews-view-list__item-wrapper` only when inspector is NOT present
- Uses `:not(:has(.inspector-region))` selector to detect when inspector is closed
- When inspector is open, items use their natural width to properly fill available space

## Visual Result  
- When inspector is closed: Items are constrained to 50% width, matching the width they would have when inspector is open
- When inspector is open: Items use natural width, filling the available space next to the inspector
- Content maintains consistent layout whether inspector state changes

## Related
Branched from `add/reader`


obenland:
Let's expand on this?

This could be a good opportunity to remove the "Feed" title and move the subtitle below "Social Web" in the sidebar. With that, we can limit the width of the entire dataview component and center it.

We should probably give it a fixed width, so the 50% don't get too narrow on intermediate size screens like tables.

pfefferle:
> we can limit the width of the entire dataview component and center it. 

I am not sure about centering it... this might be very noisy because it "jumps" around when the inspector opens. but sure let's play around with it. Feel free to give the PR a more generic title to use it for experimenting!

pfefferle:
I like removing the headline. The centered design is not my favorite. It feels bumpy when you click on the items.

pfefferle:
Maybe in combination with having the inspector always open.

obenland:
Can you help me understand the bumpiness? In my testing the change is instant when the inspector opens. It feels similar to any other website with centered content changing its width.

pfefferle:
> In my testing the change is instant when the inspector opens.

That is what I meant with "bumpiness". The inspector changes the alignment completely and I have to refocus to continue scrolling. I tried to find an english word for "unruhig".

> It feels similar to any other website with centered content changing its width.

Do you have an example for a site using a centered design and opens something to the right? (Maybe I'm just not used to it.)

obenland:
Closing in favor of #2607.

PR #2541, opened 2025-11-25
Prevent loading indicators for empty filter sections

obenland:
## Summary

- Skip rendering the tags and object type filter components entirely while loading, instead of showing "Loading" text.
- This avoids a poor UX where users see a loading indicator flash briefly before it disappears when there are no terms to display.

## Test Plan

- [ ] Open Social Web feed on a site with no tags - verify no loading indicator appears and then vanishes.
- [ ] Open Social Web feed on a site with only one object type - verify no loading indicator appears and then vanishes.
- [ ] Verify filters render and work correctly when tags/object types are available.

---

Changelog: Skip


pfefferle (approved):
Nice!

PR #2560, opened 2025-11-28
Add relay mode for ActivityPub federation

pfefferle:
> if all you have is a hammer, everything looks like a nail

so how would a WordPress plugin dev implement a relay? 😉

## Proposed changes:

This PR adds relay mode functionality to the ActivityPub plugin. When enabled, the blog actor automatically forwards all public activities it receives to all its followers, functioning as an ActivityPub relay server.

* Add relay mode setting in advanced settings page with warning about resource usage
* Implement automatic forwarding of public activities using action hooks
* Change blog actor type to 'Service' when relay mode is active
* Automatically manage related options (blog identifier, actor mode) when toggling relay mode

### Other information:

## Testing instructions:

1. Go to 'Settings > ActivityPub > Advanced'
2. Enable the "Relay Mode" checkbox
3. Save settings
4. Verify that:
   - The blog actor identifier changes to 'relay'
   - The blog actor type changes to 'Service' when viewing the actor endpoint
   - Actor mode is set to blog-only
5. Send a public activity to the blog actor from another ActivityPub server
6. Verify that the activity is forwarded (announced) to all followers of the blog actor
7. Disable relay mode
8. Verify that previous settings are restored

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add relay mode to forward public activities to all followers.

</details>


pfefferle:
@obenland the idea was, to use the plugin, to run a WordPress (or Full-Article) relay, that can be added as default for at least WordPress.com blogs. The question is how well that scales on high load.

obenland:
Should we do that on the Dotcom side first? This could probably be a blog plugin or something?

It feels like such a fringe feature that I'd hesitate to put it into the plugin proper.

pfefferle:
@obenland I removed the settings so that it is only possible to activate the "relay mode" if you know what you are doing.

pfefferle:
> Should we do that on the Dotcom side first? This could probably be a blog plugin or something?

It is so much harder to make the needed changes from the outside. I agree, that we should not make the setting user facing, but I would keep it as a part of the plugin.

obenland:
The `Blog::get_type()` changes are the only ones that are not added through a hook callback, so making that filterable should be the only thing standing in the way of doing it on Dotcom. Something like https://gist.github.com/obenland/32197e7322ee1bad75f6979107378834

Up to you, if you feel strongly about having it available outside of it

pfefferle:
If I've learned anything during my time working on plugins, it's that having a bunch of plugins to create a single functionality always results in poor UX.

obenland (approved):
> Up to you, if you feel strongly about having it available outside of it

Looks like you do 🙂

PR #2575, opened 2025-12-03
Store and retrieve webfinger acct for remote actors

pfefferle:
## Proposed changes:
* Store webfinger acct in `_activitypub_acct` post meta when creating/updating remote actors
* Retrieve and populate webfinger in actor object when loading remote actors from database
* Generate webfinger from actor ID/URL when not already available

### Other information:

- [x] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Create or update a remote actor
* Verify that `_activitypub_acct` post meta is stored with the webfinger address
* Load a remote actor from the database
* Verify that the actor object has the webfinger field populated
* Test with actors that have webfinger and those that need it generated
* Run PHPUnit tests: `./vendor/bin/phpunit tests/phpunit/tests/includes/collection/class-test-remote-actors.php`

### Changelog entry

- [x] Automatically create a changelog entry from the details below.

<details>

<summary>Changelog Entry Details</summary>

#### Significance

- [ ] Patch
- [x] Minor
- [ ] Major

#### Type

- [x] Added - for new features
- [ ] Changed - for changes in existing functionality
- [ ] Deprecated - for soon-to-be removed features
- [ ] Removed - for now removed features
- [ ] Fixed - for any bug fixes
- [ ] Security - in case of vulnerabilities

#### Message

Store and retrieve webfinger acct for remote actors to improve identification and reduce lookups

</details>

PR #2576, opened 2025-12-03
Refactor webfinger resolution to avoid code duplication

obenland:
## Summary

This PR refactors the webfinger resolution logic in response to review feedback on #2575:

- Refactors `get_acct()` to parse actor JSON directly (similar to `get_avatar_url()`), avoiding infinite loop with `get_actor()`
- Simplifies `get_actor()` to use `get_acct()` for webfinger resolution, removing duplicated logic
- Updates admin table classes to use `$actor->get_webfinger()` directly since it's now populated by `get_actor()`

This eliminates code duplication while ensuring the webfinger is stored and retrieved correctly.

## Test plan

- [x] PHPUnit tests pass for `Remote_Actors`
- [x] PHPUnit tests pass for `Webfinger`
- [x] Linting passes

### DIFF
--- a/includes/collection/class-remote-actors.php
+++ b/includes/collection/class-remote-actors.php
@@ -452,17 +452,7 @@ class Remote_Actors {
 		}
 
 		if ( ! $actor->get_webfinger() ) {
-			$webfinger = \get_post_meta( $post->ID, '_activitypub_acct', true );
-
-			if ( ! $webfinger ) {
-				$acct = Webfinger::uri_to_acct( $actor->get_id() );
-				if ( \is_wp_error( $acct ) ) {
-					$acct = Webfinger::guess( $actor );
-				}
-				$webfinger = Sanitize::webfinger( $acct );
-			}
-
-			$actor->set_webfinger( $webfinger );
+			$actor->set_webfinger( self::get_acct( $post->ID ) );
 		}
 
 		return $actor;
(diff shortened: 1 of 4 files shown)


obenland:
@pfefferle I created a separate PR with these suggestions to make it easier to discard if they're no good.

`Remote_Actor::get_avatar_url()` already parse the json directly so I thought we could do something similar in `get_acct()` to avoid the infinite loop.

`get_acct()` was only used by the admin table classes and I updated them to get the webfinger directly from the actor now, making `get_acct()` essentially an internal method for `get_actor()`.

includes/collection/class-remote-actors.php
```diff
@@ -645,8 +635,12 @@ public static function get_acct( $id ) {
 		$acct = Webfinger::uri_to_acct( $post->guid );
 
 		if ( \is_wp_error( $acct ) ) {
-			$actor = self::get_actor( $post );
-			$acct  = Webfinger::guess( $actor );
+			$actor = Actor::init_from_json( $post->post_content );
```
pfefferle, line 638: that is also what claude suggested, but I am not sure what the better way is. have the "heavy code" in get_actor or in get_acct!?

obenland: Having it in `get_acct` makes `get_actor` easier to read. With acct being persisted now, we'll also only have to go through it once (beyond grabbing the meta).

PR #2585, opened 2025-12-04
Route public replies to all actors instead of just followers

### LINKED ISSUE #2537: Revisit public activity delivery logic to prevent non-followers from receiving posts
## Context

In PR #2349, a concern was raised about public activity delivery logic that was never fully resolved before merge.

## The Problem

The current implementation may deliver public posts to users who are not following the actor. This creates a confusing user experience where users receive posts from accounts they haven't subscribed to.

From the original discussion:
> User A subscribed to the actor for this activity and now User B gets those posts in their timeline?

## Proposed Solution

As suggested by @pfefferle in the original review, the implementation should "also use the `get_user_ids`" approach to ensure more targeted delivery that respects follower relationships.

## Impact

Users are currently seeing posts from accounts they're not following, which creates noise and confusion in their timelines.

## References

- Original discussion: https://github.com/Automattic/wordpress-activitypub/pull/2349#discussion_r2448393980

pfefferle:
## What does this PR do?

Routes public ActivityPub replies to followers of the reply author, while maintaining the ability to add additional explicitly-addressed recipients.

## What problem does it solve?

Fixes #2537
Related to #2539

When receiving public replies via ActivityPub, they were only being delivered to followers of the person replying. This implementation ensures followers receive the reply while still allowing the function to process additional recipients from the activity's addressing fields.

## How did you implement your changes?

Modified `get_local_recipients()` in `includes/rest/class-inbox-controller.php`:
- Initialize `$user_ids` array at the beginning of the method
- For public activities, populate `$user_ids` with followers of the actor
- Remove early return, allowing the function to continue processing
- The subsequent recipient extraction logic can add any additional explicitly-addressed recipients to the follower list
- This maintains backward compatibility while fixing the conversation continuity issue

## Testing

- Tested with incoming public reply activities to ensure they reach followers
- Verified additional recipients in the addressing fields are still processed
- Confirmed private/direct activities are unaffected by the change

## Other information:

## Testing instructions:

* Set up WordPress instance with ActivityPub plugin
* Follow a remote user from Actor A
* Have Actor B (not followed by Actor A) reply to a public post from the followed user
* Verify Actor A can see the reply in their inbox

### Changelog entry

<details>

#### Significance

- [x] Patch
- [ ] Minor
- [ ] Major

#### Type

#### Message

Route public replies to all actors to ensure conversation continuity.

</details>

### DIFF
--- a/includes/rest/class-inbox-controller.php
+++ b/includes/rest/class-inbox-controller.php
@@ -17,7 +17,6 @@ use Activitypub\Moderation;
 use function Activitypub\camel_to_snake_case;
 use function Activitypub\extract_recipients_from_activity;
 use function Activitypub\is_activity_public;
-use function Activitypub\is_activity_reply;
 use function Activitypub\is_collection;
 use function Activitypub\is_same_domain;
 use function Activitypub\user_can_activitypub;
@@ -369,16 +368,13 @@ class Inbox_Controller extends \WP_REST_Controller {
 	 * @return array An array of user IDs who are the recipients of the activity.
 	 */
 	private function get_local_recipients( $activity ) {
+		$user_ids = array();
+
 		if ( is_activity_public( $activity ) ) {
-			if ( is_activity_reply( $activity ) ) {
-				return Actors::get_all_ids();
-			} else {
-				return Following::get_follower_ids( $activity['actor'] );
-			}
+			$user_ids = Following::get_follower_ids( $activity['actor'] );
 		}
 
 		$recipients = extract_recipients_from_activity( $activity );
-		$user_ids   = array();
 
 		foreach ( $recipients as $recipient ) {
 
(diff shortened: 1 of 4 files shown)


includes/rest/class-inbox-controller.php
```diff
@@ -368,8 +369,12 @@ public function get_item_schema() {
...
 	 */
 	private function get_local_recipients( $activity ) {
-		if ( is_activity_public( $activity ) ) { // Public activity, deliver to followers of the actor.
-			return Following::get_follower_ids( $activity['actor'] );
+		if ( is_activity_public( $activity ) ) {
+			if ( is_activity_reply( $activity ) ) {
+				return Actors::get_all_ids();
+			} else {
+				return Following::get_follower_ids( $activity['actor'] );
+			}
```
obenland:
```suggestion
		$user_ids = array();
		if ( is_activity_public( $activity ) ) {
			$user_ids = Following::get_follower_ids( $activity['actor'] );
		}
```
And then removing the `$user_ids   = array();` in line 383. I'm curious if that works?

Then we'd default to followers and add any specific account mentioned in replies.

obenland: oh, done in 50cd03d.

obenland (approved):
Nice

PR #2594, opened 2025-12-05
Add HTTP signatures to WebFinger requests

### LINKED ISSUE #2593: Unable to tag gotosocial accounts?
### Quick summary

Tagging a gotosocial account from Wordpress AP plugin fails.

### Steps to reproduce

Screenshot below shows a post that attempted to tag a gotosocial Fediverse account, but that the handle did not parse as a tag, only as plain text.

I asked the account holder about it and they said:

_So there's no obvious errors in the gotosocial logs, which is annoying, but I can see that the gotosocial server is returning "401 Unauthorized" status to requests from nham.co.uk to /users/ftmf on our side. It looks like gotosocial, unlike Mastodon, requires those requests to be signed. So it's likely this is something that needs to be addressed in the WordPress ActivityPub plugin, I would think._

Image

### Site owner impact

### Severity

Moderate

_No response_

_No response_

obenland:
Fixes #2593

## Proposed changes:
* Updates `Webfinger::get_data()` to use `Http::get()` for signed HTTP requests.
* Uses `http_request_args` filter to override the Accept header to `application/jrd+json`.

Some ActivityPub implementations like GoToSocial require HTTP signatures for all requests, including WebFinger lookups. Previously, WebFinger requests were made without signatures, causing 401 Unauthorized errors when trying to resolve mentions for users on these platforms.

### Other information:

## Testing instructions:

* Verify WebFinger lookups work against standard Mastodon instances.
* Verify WebFinger lookups work against GoToSocial instances (requires a GoToSocial account to test mentions).
* Run existing WebFinger unit tests with `npm run env-test -- --filter=Webfinger`.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Mentions now work with GoToSocial and other platforms that require signed requests.

</details>

### DIFF
--- a/.github/changelog/2594-from-description
+++ b/.github/changelog/2594-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Mentions now work with GoToSocial and other platforms that require signed requests.
--- a/includes/class-webfinger.php
+++ b/includes/class-webfinger.php
@@ -215,14 +215,16 @@ class Webfinger {
 			\rawurlencode( $identifier )
 		);
 
-		$response = \wp_safe_remote_get(
-			$webfinger_url,
-			array(
-				'headers' => array( 'Accept' => 'application/jrd+json' ),
-			)
-		);
+		$set_accept_header = function ( $args ) {
+			$args['headers']['Accept'] = 'application/jrd+json';
+			return $args;
+		};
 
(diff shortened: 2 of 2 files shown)


obenland:
I don't think this is needed after all.

PR #2604, opened 2025-12-08
Cache expensive operations in Post transformer

pfefferle:
## Proposed changes:

This PR adds caching for expensive operations in the Post transformer to improve performance by avoiding redundant computation during object serialization.

**Cached properties:**
* content: Processed ActivityPub content with shortcodes and filters
* summary: Post summary/excerpt
* tags: Post tags including hashtags
* attachment: Post attachments
* mentions: Extracted mentions from content
* in_reply_to: Reply-to URLs from reply blocks

All cached properties use `false` as a sentinel value to indicate "not yet computed", providing a consistent pattern across all properties and correctly handling nullable return values.

**Content length check removed:**
The content length check in `get_type()` has been removed due to inconsistent usage. Previously, the type determination would sometimes use raw content and sometimes use processed content depending on call order, leading to unpredictable behavior. The type is now determined solely based on post format, title support, and post type.

### Other information:

## Testing instructions:

1. Checkout the branch
2. Run the test suite: `npx wp-env run tests-cli -- /var/www/html/wp-content/plugins/activitypub/vendor/bin/phpunit -c /var/www/html/wp-content/plugins/activitypub/phpunit.xml.dist`
3. Verify all 1443 tests pass
4. Test creating ActivityPub posts to ensure content, tags, and attachments are correctly generated
5. Verify performance improvement when accessing Post transformer properties multiple times
6. Confirm that short posts with titles are now correctly identified as Articles (not Notes based on length)

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Cache expensive operations in Post transformer to improve performance.

</details>

### DIFF
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -364,7 +364,9 @@ class Post extends Base {
 		$max_media = (int) \apply_filters( 'activitypub_max_image_attachments', $max_media );
 
 		if ( 0 === $max_media ) {
-			return array();
+			$this->attachment = array();
+
+			return $this->attachment;
 		}
 
 		$media = array(
--- a/includes/transformer/class-post.php
+++ b/includes/transformer/class-post.php
@@ -42,6 +42,48 @@ class Post extends Base {
 	private $actor_object = null;
 
 	/**
+	 * The content.
+	 *
+	 * @var string|false False indicates not yet computed.
+	 */
+	private $content = false;
+
+	/**
+	 * The summary.
+	 *
+	 * @var string|null|false False indicates not yet computed.
+	 */
+	private $summary = false;
+
+	/**
+	 * The tags.
+	 *
+	 * @var array|false False indicates not yet computed.
+	 */
+	private $tags = false;
+
+	/**
+	 * The attachment.
+	 *
+	 * @var array|false False indicates not yet computed.
+	 */
+	private $attachment = false;
+
+	/**
+	 * The mentions.
+	 *
+	 * @var array|false False indicates not yet computed.
+	 */
+	private $mentions = false;
+
+	/**
+	 * The in_reply_to.
+	 *
+	 * @var string|array|null|false False indicates not yet computed.
+	 */
+	private $in_reply_to = false;
+
+	/**
 	 * Transforms the WP_Post object to an ActivityPub Object
 	 *
 	 * @return \Activitypub\Activity\Base_Object The ActivityPub Object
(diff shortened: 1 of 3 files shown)


includes/transformer/class-post.php
```diff
@@ -381,14 +431,10 @@ protected function get_type() {
...
-		$content   = \wp_strip_all_tags( $this->item->post_content );
-
 		// Check if the post has a title.
 		if (
-			! $has_title ||
-			! $this->item->post_title ||
-			\strlen( $content ) <= ACTIVITYPUB_NOTE_LENGTH
+			! \post_type_supports( $this->item->post_type, 'title' ) ||
+			! $this->item->post_title
 		) {
```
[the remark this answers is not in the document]
pfefferle: this is on purpose, because it used different content versions.

includes/transformer/class-post.php
obenland: Should this be assigned to `$this->attachment`?

includes/transformer/class-post.php
```suggestion
// Check if the post has no title.
if ( ! \post_type_supports( $this->item->post_type, 'title' ) || ! $this->item->post_title ) {
```

includes/transformer/class-post.php
```diff
@@ -464,12 +516,20 @@ protected function get_summary() {
 			return null;
 		}
 
+		if ( false !== $this->summary ) {
+			return $this->summary;
+		}
```
obenland, line 520: There's a difference in behavior with `get_content()`, where drafts are calculated on every call, whereas here this conditional prevents that.

Not sure what's more correct but I assume we'd want to be consistent?

includes/transformer/class-post.php
```diff
@@ -596,14 +667,17 @@ protected function get_in_reply_to() {
...
 
 		if ( 1 === count( $reply_urls ) ) {
-			return \current( $reply_urls );
+			$this->in_reply_to = \current( $reply_urls );
+			return $this->in_reply_to;
 		}
 
-		return \array_values( \array_unique( $reply_urls ) );
+		$this->in_reply_to = \array_values( \array_unique( $reply_urls ) );
+		return $this->in_reply_to;
```
obenland, line 684:
```suggestion

		return $this->in_reply_to;
```

includes/transformer/class-post.php
```diff
@@ -596,14 +667,17 @@ protected function get_in_reply_to() {
 		}
 
 		if ( empty( $reply_urls ) ) {
-			return null;
+			$this->in_reply_to = null;
+			return $this->in_reply_to;
```
obenland, line 673:
```suggestion

			return $this->in_reply_to;
```

includes/transformer/class-post.php
```diff
@@ -596,14 +667,17 @@ protected function get_in_reply_to() {
...
 		}
 
 		if ( 1 === count( $reply_urls ) ) {
-			return \current( $reply_urls );
+			$this->in_reply_to = \current( $reply_urls );
+			return $this->in_reply_to;
```
obenland, line 679:
```suggestion

			return $this->in_reply_to;
```

obenland (approved):

PR #2609, opened 2025-12-09
Add image optimization for imported attachments

pfefferle:
## Proposed changes:

* Add automatic image optimization for imported ActivityPub attachments to reduce file sizes
* Resize large images to max 1200px dimension while maintaining aspect ratio
* Convert images to WebP format when supported by the server (falls back to JPEG when not)
* Skip GIFs (may be animated) and SVGs (vector graphics) to preserve their special properties

### Other information:

Existing tests were updated to handle the new WebP conversion behavior.

## Testing instructions:

* Set up a local WordPress environment with wp-env
* Have another ActivityPub-enabled instance (like Mastodon) post an image larger than 1200px
* Follow that account and receive the post
* Check the imported image in `wp-content/uploads/activitypub/` - it should be:
  * Resized to max 1200px on the longest edge
  * Converted to WebP format (or JPEG if WebP not supported)
  * Significantly smaller file size than the original

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add image optimization for imported attachments (resize to 1200px max, convert to WebP).

</details>

### DIFF
--- a/includes/class-attachments.php
+++ b/includes/class-attachments.php
@@ -643,13 +643,21 @@ class Attachments {
 		// Check if WebP is supported.
 		$can_webp = $editor->supports_mime_type( 'image/webp' );
 
-		// Determine output format.
+		// Determine output format and save.
 		if ( $can_webp ) {
+			// Convert to WebP.
 			$new_path = \preg_replace( '/\.[^.]+$/', '.webp', $file_path );
 			$result   = $editor->save( $new_path, 'image/webp' );
 		} elseif ( \in_array( $mime_type, array( 'image/png', 'image/webp' ), true ) ) {
 			// Keep original format for potentially transparent images when WebP not available.
-			$result = $needs_resize ? $editor->save( $file_path ) : true;
+			if ( ! $needs_resize ) {
+				// No changes needed.
+				return array(
+					'path'    => $file_path,
+					'changed' => false,
+				);
+			}
+			$result = $editor->save( $file_path );
 		} else {
 			// Convert to JPEG when WebP not available.
 			$new_path = \preg_replace( '/\.[^.]+$/', '.jpg', $file_path );
@@ -663,18 +671,17 @@ class Attachments {
 			);
 		}
 
-		// If format changed, delete the original.
-		if ( is_array( $result ) && isset( $result['path'] ) && $result['path'] !== $file_path ) {
+		// Handle result - $result is always an array from $editor->save().
+		$result_path = $result['path'] ?? $file_path;
+
+		// If path changed (format conversion), delete the original file.
+		if ( $result_path !== $file_path ) {
 			\wp_delete_file( $file_path );
-			return array(
-				'path'    => $result['path'],
-				'changed' => true,
-			);
 		}
 
 		return array(
-			'path'    => $file_path,
-			'changed' => $needs_resize,
+			'path'    => $result_path,
+			'changed' => true,
 		);
 	}
 
--- a/includes/class-attachments.php
+++ b/includes/class-attachments.php
@@ -452,7 +452,7 @@ class Attachments {
 		}
 
 		// Optimize images before sideloading (resize and convert to WebP).
-		$optimized = self::optimize_image( $tmp_file );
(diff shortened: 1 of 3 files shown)


pfefferle:
@obenland in addition to #2608, I think it makes sense to also reduce the size an image format for remote images. I added the optimization to both, the imports and the remote posts. I am happy to have it for remote posts only though, because you might want to have the original image when importing your Mastodon posts.

includes/class-attachments.php
```diff
@@ -551,6 +581,96 @@ private static function save_file( $attachment_data, $object_id, $object_type )
...
+	 *
+	 * @return array{path: string, changed: bool}|\WP_Error The optimized file path and whether it changed, or WP_Error.
+	 */
+	private static function optimize_image( $file_path, $max_dimension ) {
+		// Check if it's an image.
+		$mime_type = \wp_check_filetype( $file_path )['type'] ?? '';
+		if ( ! $mime_type || ! \str_starts_with( $mime_type, 'image/' ) ) {
+			return array(
+				'path'    => $file_path,
+				'changed' => false,
```
obenland: Let's remove the "changed" information and only return the file path. It shouldn't matter whether it changed or not.

includes/class-attachments.php
```diff
@@ -444,15 +451,31 @@ private static function save_attachment( $attachment_data, $post_id, $author_id
...
+		$original_name = \basename( \wp_parse_url( $attachment_data['url'], PHP_URL_PATH ) );
+
+		// Update filename extension if format changed.
+		if ( $optimized['changed'] ) {
+			$new_ext       = \pathinfo( $tmp_file, PATHINFO_EXTENSION );
+			$original_name = \preg_replace( '/\.[^.]+$/', '.' . $new_ext, $original_name );
+		}
+
 		$file_array = array(
-			'name'     => \basename( \wp_parse_url( $attachment_data['url'], PHP_URL_PATH ) ),
```
obenland, line 449: Can we use `$tmp_file` here instead of `$attachment_data['url']`? That would allow us to remove the added 7 lines above.

includes/class-attachments.php
```diff
@@ -444,15 +451,31 @@ private static function save_attachment( $attachment_data, $post_id, $author_id
...
+			'name'     => $original_name,
 			'tmp_name' => $tmp_file,
 		);
 
 		// Prepare attachment post data.
+		// Clear mime type if format changed so WordPress auto-detects it.
+		$mime_type = $optimized['changed'] ? '' : ( $attachment_data['mediaType'] ?? '' );
 		$post_data = array(
-			'post_mime_type' => $attachment_data['mediaType'] ?? '',
+			'post_mime_type' => $mime_type,
```
obenland: Let's just remove this and let WP handle it? I've not checked how it works under the hood, this is just based on the comment in line 475.

includes/class-attachments.php
pfefferle: this was an initial feedback, because it will always be handled as changed otherwise!?

obenland: What are the consequences of assuming they changed when they didn't?

includes/class-attachments.php
pfefferle: We still need to derive the filename from the original URL because `wp_tempnam()` creates files with random suffixes (e.g., `test.jpg12345`) that don't have proper extensions. I've simplified the code by:

1. Moving the filename derivation earlier
2. Renaming the temp file to have a proper extension before optimization
3. Then updating only the extension after optimization

This ensures `wp_check_filetype()` can properly detect the mime type while keeping the original filename for WordPress.

obenland (approved):
Let's get this in for now.

If you don't mind, I'd like to take a stab at iterating on this, it feels like there's some room for simplification across the class.

Unique filenames is one of them. I was about to suggest using `wp_unique_filename()` instead of `get_unique_path()` and then realized that `save_file()` also has it own way of creating a unique name. Stuff like that.

PR #2611, opened 2025-12-09
Respect WordPress "show avatars" setting for remote actors

pfefferle:
See https://sueden.social/@freuwesen/115667842788940001

## Proposed changes:
* Respect WordPress "show avatars" setting in the Followers block by conditionally rendering avatar images server-side with PHP.
* Also adds the setting check to the block editor preview for consistent behavior.
* Maintains existing behavior for ActivityPub comments (using the Avatars class).

### Other information:

## Testing instructions:

* Go to **Settings → Discussion**
* Uncheck **Show Avatars**
* Add a **Followers block** to a page/post
* Verify that follower avatars are not displayed in either the editor or frontend
* Re-enable **Show Avatars**
* Verify that follower avatars are displayed again

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Respect WordPress "show avatars" setting in the Followers block.

</details>


obenland:
By extension, what would "show avatars" being false mean for the Reactions block?

includes/collection/class-remote-actors.php
```diff
@@ -655,9 +655,16 @@ public static function get_acct( $id ) {
...
-	 * @return string The avatar URL or empty string if not found.
+	 * @return string The avatar URL or default avatar if not found or avatars disabled.
 	 */
 	public static function get_avatar_url( $id ) {
+		$default_avatar_url = ACTIVITYPUB_PLUGIN_URL . 'assets/img/mp.jpg';
+
+		// Respect WordPress "show avatars" setting.
+		if ( ! \get_option( 'show_avatars' ) ) {
+			return $default_avatar_url;
+		}
```
obenland: I'm not sure we should fall back to the default avatar. WordPress' handling of the option is a matter of not displaying avatars of users interacting with the site in the front end, which is reflected in [the original question](https://sueden.social/@freuwesen/115667842788940001).

Instead, we should check all instances where we display avatars, and only do that conditionally to this option. Probably our blocks, for the most part.

obenland (changes requested):

includes/collection/class-remote-actors.php
pfefferle: Have you read the linked mastodon post with the comments?

obenland: The one I linked to? Yes. Is there anything in specific that I missed?

pfefferle: The thing is, that he want to have the information of the users but not show images. see his posts and the likes/shares there.

https://bunte-kuechenabenteuer.de/flammkuchen-mit-birne-und-ziegenkaese-3-portionen/ 

(the 👍 and ♻️ are a like and a share with no image).

pfefferle: But because it was also requested: We can show a simple bar with the total numbers instead then.

Something like:

`5 Comments | 12 Likes | 3 Shares | 7 Quotes`

obenland: > We can show a simple bar with the total numbers instead then.

yeah that sounds good. we should also look at the followers block etc

pfefferle: ok, I will rework this pr then, to focus on comments and followers/followings. I will keep the facepiles aside and we fix that with a different view.

obenland (approved):

obenland:
Maybe worth an exploration before merging this: Is there a way to use core-data to get this setting through the API instead of having to pass it manually.

`getEntityRecord( 'root', 'site' )?.show_avatars` or something?

pfefferle:
## Exploration: Using core-data for `show_avatars` setting

I explored whether we can use `@wordpress/core-data` to get the `show_avatars` setting via `getEntityRecord( 'root', 'site' )?.show_avatars`.

### Findings

**WordPress core does NOT expose `show_avatars` via the REST API.**

The `site` entity in `@wordpress/core-data` gets its fields from `/wp/v2/settings`, which only includes settings registered with `show_in_rest => true`. Looking at WordPress core's [`register_initial_settings()`](https://developer.wordpress.org/reference/functions/register_initial_settings/), `show_avatars` is not among the registered settings.

The settings that ARE exposed include: title, description, timezone, date/time formats, posts_per_page, show_on_front, default_comment_status, etc. - but not avatar-related settings.

### Options

1. **Register it ourselves** - We could add:
   ```php
   add_action( 'rest_api_init', function() {
       register_setting( 'discussion', 'show_avatars', array(
           'show_in_rest' => true,
           'type'         => 'boolean',
       ) );
   } );
   ```
   Then use: `select( 'core' ).getEntityRecord( 'root', 'site' )?.show_avatars`

2. **Keep current approach** - Use `wp_localize_script` with `window._activityPubOptions`

### Recommendation

Option 1 (registering the setting) would work and allow using core-data. The setting would then be available at both the REST endpoint and via `getEntityRecord( 'root', 'site' )`.

However, note that accessing `/wp/v2/settings` requires `manage_options` capability, so this would only work for administrators in the editor. For non-admin users, we'd still need the `wp_localize_script` fallback.

Would you prefer we implement the core-data approach with a fallback, or keep the current implementation?

pfefferle:
That is how claude says: "Computer says no!" ;)

obenland:
I'm reading

> 1. Register it ourselves

PR #2616, opened 2025-12-10
Fix actor switcher to respect actor mode and capabilities

obenland:
## Summary
- Move settings registration from `class-settings.php` to `class-options.php` so settings are available outside wp-admin via `init` hook.
- Add schema with enum values for `activitypub_actor_mode` setting to expose it via REST API.
- Update actor switcher to check user capabilities via `canUser` on `ap_extrafield` and `ap_extrafield_blog` post types.
- Add capability mappings to extra field post types (`activitypub` for user fields, `manage_options` for blog fields).
- Auto-correct active actor when mode changes make current actor unavailable.
- Link to appropriate settings page based on available actor (blog profile vs user profile).

## Test plan
- [ ] Set actor mode to "blog only" - verify user cannot switch to user actor.
- [ ] Set actor mode to "user only" - verify user cannot switch to blog actor.
- [ ] Set actor mode to "both" - verify user can switch between actors.
- [ ] As a user without `activitypub` capability, verify they can only use blog actor (if enabled).
- [ ] As a user without `manage_options` capability, verify they can only use user actor (if enabled).
- [ ] Verify clicking actor switcher links to correct settings page based on mode.


obenland:
@pfefferle This required a bit more of a fundamental change than initially anticipated.

In order for the `actor_mode` option to be available in the `wp/v2/settings` endpoint, it needs to be a registered setting with `show_in_rest` enabled. To make that possible, I move the settings registration from `admin_init`, where it's only available in wp-admin and not the rest api, to `init`. 

The consequence of doing that was that sanitization callbacks were now enabled for tests, which made tests fail (see test fixes). So I wonder if this is something we'd want to do for trunk, more generally, and possibly add an upgrade routine to send all registered options through `sanitize_option()` again. I wonder if this was the missing link in our string value troubles with options in the past.

pfefferle:
Generally yes but I am not sure I like the migration to options. Is that the right place?

I think we have to also document that a bit better, at least in the PHPDoc!

obenland:
It felt like the best place to me, just because of the integration with the Options API through [sanitization](https://github.com/WordPress/wordpress-develop/blob/6.9/src/wp-includes/option.php#L3073) and [defaults](https://github.com/WordPress/wordpress-develop/blob/6.9/src/wp-includes/option.php#L3076).

Can you tell me more about the kind of documentation you're looking for?

pfefferle:
> It felt like the best place to me, just because of the integration with the Options API through [sanitization](https://github.com/WordPress/wordpress-develop/blob/6.9/src/wp-includes/option.php#L3073) and [defaults](https://github.com/WordPress/wordpress-develop/blob/6.9/src/wp-includes/option.php#L3076).

Makes sense!

> Can you tell me more about the kind of documentation you're looking for?

Moving them to `init` is a quite expensive change, because we do register all the settings on every page load then! I think we should document why we have done that (at least in the phpdoc).

What about a less "expensive" approach with registering the settings on `admin_init` and `rest_api_init` instead?

```PHP
\add_action( 'admin_init', array( self::class, 'register_settings' ), 11 );
\add_action( 'rest_api_init', array( self::class, 'register_settings' ), 11 );
```

includes/class-options.php
```diff
@@ -147,32 +146,29 @@ public static function register_settings() {
...
 		);
 
 		\register_setting(
 			'activitypub',
 			'activitypub_allow_reposts',
 			array(
-				'type'              => 'integer',
-				'description'       => 'Allow reposts.',
-				'default'           => '1',
-				'sanitize_callback' => 'absint',
```
pfefferle: why have you removed them? does WordPress implicitly check because of the `type`?

obenland: Fixed in [57e4db0](https://github.com/Automattic/wordpress-activitypub/pull/2616/commits/57e4db0667775ed9615c45b285fac019ffcfe8e3).

I had to remove them because when they were hooked to `init` they were applied in tests and it caused strict tests with stringified values to fail.

pfefferle (approved):

PR #2617, opened 2025-12-10
Add summary display style to Reactions block

### LINKED ISSUE #1815: Feature Request: Like/Report-Block like classic social media
### What

On my blog https://kaffeeringe.de/ I try to achieve the classic look of a social media account. I css-customized the follower-block as a header to show the number of posts and the number of followers.

### Why

I would like to show the like- and boost-count in the list-view like it looks in Mastodon, X, Facebook.

### How

🔁 3 Reposts   ⭐ 5 Likes   🗨️ 1 Comment

This could be a block or a set of shortcodes with these counts, so I can build it myself.

pfefferle:
Fixes #1815

## Proposed changes:

Add a new "Compact" display style to the Reactions block that shows reaction counts inline without avatar facepiles. Both styles use the same underlying Reactions component - the Compact style simply hides the avatars in the facepile row and modal dropdown.

### Features:
- **Facepile style** (default): Shows avatar facepiles with reaction counts
- **Compact style**: Shows only reaction count buttons inline, without avatars
- Automatically defaults to Compact style when WordPress "Show Avatars" setting is disabled (Settings → Discussion)
- When avatars are disabled, the Facepile style option is hidden from the block styles panel

| Compact | Facepile |
|---|---|
| Compact style screenshot | Facepile style screenshot |

### Other information:

## Testing instructions:

1. Go to the block editor and insert a "Fediverse Reactions" block
2. In the block toolbar, click the styles icon to switch between "Facepile" and "Compact" styles
3. When "Compact" style is selected, verify:
   - No avatar facepiles are shown
   - Reaction buttons are displayed inline
   - Clicking a button opens a dropdown without avatars
4. Test that the block renders correctly on the frontend with both styles
5. Go to Settings → Discussion and disable "Show Avatars"
6. Insert a new Reactions block and verify it defaults to "Compact" style
7. Verify the "Facepile" style option is hidden when avatars are disabled

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add compact display style to Reactions block that hides avatars.

</details>


src/reactions/edit.js
```diff
@@ -36,20 +39,102 @@ const DUMMY_REACTIONS = {
...
+		if ( hasInitialized.current ) {
+			return;
+		}
+		hasInitialized.current = true;
+
+		// Only apply default if no style has been explicitly chosen yet.
+		const hasStyleClass = className?.includes( 'is-style-' );
+		if ( ! showAvatars && ! hasStyleClass ) {
+			setAttributes( {
+				className: ( className ? className + ' ' : '' ) + 'is-style-summary',
```
obenland: We add `clsx` as a dependency in `add/router`, maybe we could move that up into this PR and use it here.

src/reactions/edit.js
```diff
@@ -36,20 +39,102 @@ const DUMMY_REACTIONS = {
...
+				displayStyle: 'summary',
+			} );
+		}
+	}, [ className, showAvatars, setAttributes ] );
+
+	// Sync displayStyle attribute with className when style changes.
+	const classNameStyle = className?.includes( 'is-style-summary' ) ? 'summary' : 'facepile';
+	if ( classNameStyle !== displayStyle ) {
+		setAttributes( { displayStyle: classNameStyle } );
+	}
```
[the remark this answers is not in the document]
obenland: This looks like good feedback

src/reactions/edit.js
```diff
@@ -65,13 +150,45 @@ export default function Edit( { __unstableLayoutClassNames } ) {
...
+			{ isSummaryStyle && (
+				<InspectorControls>
+					<PanelBody title={ __( 'Summary Settings', 'activitypub' ) }>
+						<ToggleControl
+							label={ __( 'Show Comments', 'activitypub' ) }
+							help={ __( 'Include the comment count in the summary.', 'activitypub' ) }
+							checked={ showComments }
+							onChange={ ( value ) => setAttributes( { showComments: value } ) }
+							__nextHasNoMarginBottom
+						/>
```
obenland: I'd expect the dummy block in the editor to update when this control gets toggled, and remove the comments.

src/reactions/edit.js
```diff
@@ -65,13 +150,45 @@ export default function Edit( { __unstableLayoutClassNames } ) {
...
+						<ToggleControl
+							label={ __( 'Show empty Reactions', 'activitypub' ) }
+							help={ __( 'Display Reaction types even when they have no count.', 'activitypub' ) }
+							checked={ showEmpty }
+							onChange={ ( value ) => setAttributes( { showEmpty: value } ) }
+							__nextHasNoMarginBottom
+						/>
```
obenland: Same as above, let's remove reactions that don't have a count, in the dummy that would be quotes.

Should we make this available to facepile style as well?

src/reactions/edit.js
```diff
@@ -65,13 +150,45 @@ export default function Edit( { __unstableLayoutClassNames } ) {
...
+						<TextControl
+							label={ __( 'Separator', 'activitypub' ) }
+							help={ __( 'Character(s) to separate each item.', 'activitypub' ) }
+							value={ separator }
+							onChange={ ( value ) => setAttributes( { separator: value } ) }
+							__next40pxDefaultSize
+							__nextHasNoMarginBottom
+						/>
```
obenland: I'm not in love with this one. Is it necessary to have that?

It feels like it could be simplified with a 20px margin between reaction types, no separator.

src/reactions/render.php
```diff
@@ -38,133 +55,167 @@
...
-);
+		$counts[ $_type ] = array(
+			'count' => $count,
+			'label' => $type_object['label'],
+		);
+	}
+
+	// Don't render if there are no items to show.
+	if ( empty( $counts ) ) {
+		ob_end_clean();
```
obenland:
```suggestion
		ob_end_clean();
		echo '<!-- Reactions block: No reactions found. -->';
```

src/reactions/render.php
```diff
@@ -206,14 +257,56 @@ class="reaction-label wp-element-button"
...
+		<template data-wp-each="context.modal.items">
+			<li class="reaction-item">
+				<a data-wp-bind--href="context.item.url" target="_blank" rel="noopener noreferrer">
+					<img
+						alt=""
+						data-wp-bind--alt="context.item.name"
+						data-wp-bind--src="context.item.avatar"
+						data-wp-on--error="callbacks.setDefaultAvatar"
+						src=""
+					/>
```
obenland: We should not show this when show_avatars is false

src/reactions/render.php
```diff
@@ -206,14 +257,56 @@ class="reaction-label wp-element-button"
 		</div>
```
obenland: If feel like we shouldn't show avatars in the facepile if show_avatars is false.

src/reactions/edit.js
```diff
@@ -65,13 +150,45 @@ export default function Edit( { __unstableLayoutClassNames } ) {
...
+			) }
 			<InnerBlocks
 				template={ TEMPLATE }
 				allowedBlocks={ [ 'core/heading' ] }
 				templateLock={ 'all' }
 				renderAppender={ false }
 			/>
-			<Reactions postId={ getCurrentPostId() } fallbackReactions={ DUMMY_REACTIONS } />
+			{ isSummaryStyle ? (
+				<Summary separator={ separator } />
```
obenland: Let's pass it the post ID, like in `Reactions` and display the actual reactions if available.

src/reactions/render.php
```diff
@@ -38,133 +55,167 @@
...
-wp_interactivity_config(
-	'activitypub/reactions',
-	array(
-		'defaultAvatarUrl' => ACTIVITYPUB_PLUGIN_URL . 'assets/img/mp.jpg',
-		'namespace'        => ACTIVITYPUB_REST_NAMESPACE,
-	)
-);
+		$counts[ $_type ] = array(
+			'count' => $count,
+			'label' => $type_object['label'],
```
obenland: This seems to be missing singular labels.

```php
		$label = sprintf(
			_n(
				$type_object['count_single'],
				$type_object['count_plural'],
				$count,
				'activitypub'
			),
			number_format_i18n( $count )
		);
```

src/reactions/render.php
pfefferle: how would that look like 🤔

the whole idea of this PR was to not use facepiles when avatars are disabled!?

obenland: If we don't just want to show the labels, which I'd understand, we could not make the default style available.

Thinking further on this, should we still offer the modals with the user lists in the summary style?

obenland: maybe the difference between the two styles could just be CSS and whether the avatars are displayed or not.

pfefferle: it was a request to have that style even with avatars enabled.

pfefferle: btw. I changed the code to auto add the summary version to the post by default when avatars are disabled.

pfefferle: This is only an issue if you add it on purpose.

src/reactions/edit.js
obenland: It looks like this is still missing.

src/reactions/render.php
```diff
@@ -38,140 +54,179 @@
...
+	// Collect counts for summary view.
+	$counts = array();
+
+	// Add comment count if enabled.
+	if ( $attributes['showComments'] ) {
+		$comment_count = get_comments_number( $_post_id );
+		if ( $attributes['showEmpty'] || $comment_count > 0 ) {
+			$counts['comments'] = array(
+				'count' => $comment_count,
+				'label' => _n( 'Comment', 'Comments', $comment_count, 'activitypub' ),
```
obenland: The way label is being rendered below, we need to include the comment count into the label string.

Image

src/reactions/render.php
```diff
@@ -38,140 +54,179 @@
...
-	array(
-		'defaultAvatarUrl' => ACTIVITYPUB_PLUGIN_URL . 'assets/img/mp.jpg',
-		'namespace'        => ACTIVITYPUB_REST_NAMESPACE,
-	)
-);
+		// phpcs:disable WordPress.WP.I18n
+		$label = sprintf(
+			_n(
+				$type_object['count_single'],
+				$type_object['count_plural'],
```
obenland: Hm, that's unfortunate. Looks like the translation strings only accept numeric values, so we can't pass a string here. As a result, the number will not be bold here, as it is in the editor. I also just noticed there's a difference in capitalization.

Image

Image

pfefferle:
@obenland with all the feedback above and working on the fixes, I now think it really makes sense to re-use the Reaction Block for both!

> maybe the difference between the two styles could just be CSS and whether the avatars are displayed or not.

It is not only CSS, but the current version is a somehow Avatarless version of the Reactions Block. The `Summary` style simply disables the facepiles (avatars), but keeps the dropdown and shows the "buttons" inline.

| Summary | Facepiles |
|---|---|
| Screenshot 2025-12-12 at 09 19 59 | Screenshot 2025-12-12 at 09 20 03 |

src/reactions/render.php
```diff
@@ -38,6 +51,21 @@
...
 
+/*
+ * Determine display style - compact style hides avatars.
+ * For auto-hooked blocks without explicit style, use avatar setting to determine style.
+ */
+$has_style_class = isset( $attributes['className'] ) && strpos( $attributes['className'], 'is-style-' ) !== false;
+if ( ! $has_style_class ) {
+	$default_style              = get_option( 'show_avatars', true ) ? 'facepile' : 'compact';
+	$attributes['className']    = trim( ( $attributes['className'] ?? '' ) . ' is-style-' . $default_style );
+	$attributes['displayStyle'] = $default_style;
```
obenland:
```suggestion
	$attributes['className']    = trim( ( $attributes['className'] ?? '' ) . ' is-style-' . $default_display_style );
	$attributes['displayStyle'] = $default_display_style;
```

src/reactions/render.php
```diff
@@ -38,6 +51,21 @@
...
+}
+
+$is_compact   = 'compact' === $attributes['displayStyle'];
+$show_avatars = ! $is_compact;
```
obenland:
```suggestion
$show_avatars = 'facepile' === $attributes['displayStyle'];
```

includes/class-blocks.php
```diff
@@ -84,7 +84,18 @@ public static function register_blocks() {
...
+		$reactions_args = array();
+		if ( ! \get_option( 'show_avatars', true ) ) {
+			$reactions_args['styles'] = array(
+				array(
+					'name'      => 'compact',
+					'label'     => __( 'Compact', 'activitypub' ),
+					'isDefault' => true,
+				),
+			);
+		}
```
obenland, line 91: This doesn't seem to work for me. When I insert a new Reactions block with avatars disabled, it still default to facepile for me. Maybe something we need to do on the JS side

obenland (commented):
@pfefferle This update is excellent.

The one question remaining for me is wether we should force compact display when show_avatar is false, regardless of choice of style, particularly for blocks that are already inserted.

pfefferle:
> The one question remaining for me is wether we should force compact display when show_avatar is false, regardless of choice of style, particularly for blocks that are already inserted.

I think either way, we should have a consistent handling over all blocks.

includes/class-blocks.php
pfefferle: This is mainly to deactivate the "Facepile Style" in the sidebar. I have not tested your case tbh. 🫣

includes/class-blocks.php
```diff
@@ -84,7 +84,18 @@ public static function register_blocks() {
...
+		// Register reactions block, conditionally removing facepile style if avatars are disabled.
+		$reactions_args = array();
+		if ( ! \get_option( 'show_avatars', true ) ) {
+			$reactions_args['styles'] = array(
+				array(
+					'name'      => 'compact',
+					'label'     => __( 'Compact', 'activitypub' ),
+					'isDefault' => true,
+				),
+			);
```
obenland:
```suggestion
			$reactions_args['styles'] = array();
```

Removing all styles when we don't show avatars might be the best way forward?

With only Compact, the block still displays a Styles section but you can't change it, which feels awkward. With empty styles it defaults to compact without showing the styles section.

Image

obenland (commented):
I think that's all I have. Not sure if I covered all scenarios in testing, but it should be pretty easy to fix/update if there's more we find later.

obenland (approved):

kaffeeringe:
Cool! Now it would be nice to have a comments count block in the same style. ;-)

PR #2625, opened 2025-12-11
Fix Followers block error when social graph is hidden

obenland:
## Summary

Fixes the Fediverse Followers block throwing an error when the "Hide Social Graph" privacy option is enabled. This regression was introduced in #2294.

The issue occurs because the ActivityPub followers endpoint does not return `orderedItems` when the social graph is hidden, causing `Cannot read properties of undefined (reading 'length')` in the block editor.

This fix replaces the custom ActivityPub endpoint with WordPress's built-in REST API for fetching followers in the block editor, and properly respects the social graph visibility setting throughout.

## Changes

- Add REST query parameter support for `ap_actor` post type to filter by followed user.
- Expose follower meta fields (`_activitypub_acct`, `_activitypub_avatar_url`) to REST API.
- Refactor `edit.js` to use `useEntityRecords` instead of `apiFetch`.
- Filter users with hidden social graph from the block's user selector dropdown.
- Show warning notice in editor for existing blocks whose selected user has hidden their social graph.
- Add privacy check in `render.php` to skip rendering when social graph is hidden.
- REST API returns no results when querying followers for users with hidden social graph.
- Add `show_in_rest` to user meta registration for JavaScript access.
- Update unit tests to verify social graph visibility is respected.

## Test Plan

- [ ] Enable "Hide Social Graph" in user profile settings (under ActivityPub section).
- [ ] Insert Fediverse Followers block in a post.
- [ ] Verify the user with hidden social graph is not available in the user selector dropdown.
- [ ] If editing an existing block with a user who now has hidden social graph, verify warning notice appears.
- [ ] Verify the block does not display followers on the frontend when social graph is hidden.
- [ ] Disable "Hide Social Graph" setting.
- [ ] Verify block works correctly and user appears in dropdown again.
- [ ] Test with blog user (ID 0) - same behavior should apply.

<details open>

Significance:
- [ ] Patch: backwards-compatible bug fix.
- [x] Minor: added functionality in a backwards-compatible manner.
- [ ] Major: broke backwards compatibility; could be a minor in 0.x versions.

Type:
- [ ] Added: new feature.
- [x] Changed: updated existing functionality.
- [ ] Deprecated: marked for future removal.
- [ ] Removed: deleted code or feature.
- [ ] Fixed: bug fix.
- [ ] Security: fixed vulnerability.

Changelog entry: Respect "Hide Social Graph" setting in Followers block editor and frontend.

</details>


includes/class-post-types.php
```diff
@@ -572,6 +597,45 @@ public static function register_ap_actor_rest_field() {
...
+	public static function filter_ap_actor_rest_query( $args, $request ) {
+		$following = $request->get_param( 'activitypub_following' );
+
+		if ( $following ) {
+			$args['meta_query'] = array( // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+				array(
+					'key'   => Followers::FOLLOWER_META_KEY,
+					'value' => $following,
+				),
+			);
```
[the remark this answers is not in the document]
obenland: Fixed in [0460fe7](https://github.com/Automattic/wordpress-activitypub/pull/2625/commits/0460fe7170788fc3cef874c095170e6317c4868b).

includes/class-post-types.php
```diff
@@ -572,6 +597,45 @@ public static function register_ap_actor_rest_field() {
...
+	 */
+	public static function filter_ap_actor_rest_query( $args, $request ) {
+		$following = $request->get_param( 'activitypub_following' );
+
+		if ( $following ) {
+			$args['meta_query'] = array( // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+				array(
+					'key'   => Followers::FOLLOWER_META_KEY,
+					'value' => $following,
+				),
```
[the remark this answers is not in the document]
obenland: Fixed in [0460fe7](https://github.com/Automattic/wordpress-activitypub/pull/2625/commits/0460fe7170788fc3cef874c095170e6317c4868b).

includes/class-post-types.php
```diff
@@ -572,6 +597,45 @@ public static function register_ap_actor_rest_field() {
...
+	 * Register custom query parameters for the ap_actor REST endpoint.
+	 *
+	 * @param array $params The existing collection parameters.
+	 *
+	 * @return array Modified collection parameters.
+	 */
+	public static function register_ap_actor_rest_query_params( $params ) {
+		$params['activitypub_following'] = array(
+			'description' => \__( 'Filter actors by who they follow (WordPress user ID).', 'activitypub' ),
+			'type'        => 'integer',
```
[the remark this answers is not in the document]
obenland: Fixed in [0460fe7](https://github.com/Automattic/wordpress-activitypub/pull/2625/commits/0460fe7170788fc3cef874c095170e6317c4868b).

pfefferle:
Several things:

* Shouldn't we disable the block completely when the feature is disabled? I think "Hide Followers and Following on Profile" includes the Blocks, not only the API-Endpoint.
* If you change the User in the Block Settings to the Blog-User, it keeps the Followers of the Former (real) User and adds some Dummy Users (Bob Developer, Eve Hacker, ...) and it does not store the change, because after a reload it falls back to the User.

More feedback inline.

includes/class-post-types.php
```diff
@@ -572,6 +598,48 @@ public static function register_ap_actor_rest_field() {
...
+	 * @param array $params The existing collection parameters.
+	 *
+	 * @return array Modified collection parameters.
+	 */
+	public static function register_ap_actor_rest_query_params( $params ) {
+		$params['activitypub_following'] = array(
+			'description'       => \__( 'Filter actors by who they follow (WordPress user ID).', 'activitypub' ),
+			'type'              => 'integer',
+			'sanitize_callback' => 'absint',
+			'validate_callback' => 'is_user_member_of_blog',
```
pfefferle: I am not sure the `is_user_member_of_blog` works here, because the blog user has ID `0`.

includes/class-post-types.php
```diff
@@ -572,6 +598,48 @@ public static function register_ap_actor_rest_field() {
...
+	 * @param array            $args    Query arguments.
+	 * @param \WP_REST_Request $request The REST request.
+	 *
+	 * @return array Modified query arguments.
+	 */
+	public static function filter_ap_actor_rest_query( $args, $request ) {
+		$following = $request->get_param( 'activitypub_following' );
+
+		// Only filter if user ID provided and social graph is visible or current user can edit that user.
+		if ( $following && ( Actors::show_social_graph( $following ) || \current_user_can( 'edit_user', $following ) ) ) {
```
pfefferle: This check `\current_user_can( 'edit_user', $following )` might also not work for the Blog-User (id `0`).

obenland:
> Shouldn't we disable the block completely when the feature is disabled? I think "Hide Followers and Following on Profile" includes the Blocks, not only the API-Endpoint.

I wondered the same and came to the conclusion that it's a block that needs to be inserted manually.

To complicate matters, I noticed this while testing #2617 and the Followers blocks I had in the test posts showed up broken. So they were already inserted and then I checked the box to hide followers. In that scenario, we'd probably want to not show them?

Nice catch with the blog user, obviously the one thing I didn't test. I tried to create e2e around this but was unsuccessful.

includes/class-post-types.php
obenland: Fixed in [5660fc9](https://github.com/Automattic/wordpress-activitypub/pull/2625/commits/5660fc9689fae6c673bc071692963ad31fffdf78).

includes/class-post-types.php
obenland: Fixed in [5660fc9](https://github.com/Automattic/wordpress-activitypub/pull/2625/commits/5660fc9689fae6c673bc071692963ad31fffdf78).

obenland:
@pfefferle I updated the PR to 
* show a notice in the block if the selected user has sharing disabled.
* For newly inserted blocks it filters out all users (incl blog) that have it disabled.
* In the frontend it shows an HTML comment instead of the block when sharing is disabled.

pfefferle:
The idea behind hiding followers is that no one should be able to query the list at all. Re-enabling the AP-Actors Post-Type API endpoint makes this possible again through a different path. At that point, follower “hiding” relies on the assumption that no one will query the `wp/v2` endpoint.

pfefferle:
I think we should stick to the ActivityPub API endpoint and simply have the check if the user has the endpoint enabled or not!?

pfefferle (dismissed):

pfefferle (approved):

PR #2634, opened 2025-12-15
Improve Reader empty state based on following status

pfefferle:
Fixes #

## Proposed changes:
* Add contextual empty state messages in the Reader based on whether the user is following anyone
* When not following anyone: show engaging message with link to the following page (URL differs for blog vs user actors)
* When following but no posts: show message about checking back for updates
* Add `followed_by` REST API parameter for `ap_actor` post type to query actors a user follows
* Add `useFollowing` hook to fetch following count via REST API
* Add unit tests for the new hook

Screenshot 2025-12-15 at 17 54 26

### Other information:

## Testing instructions:

* Go to the ActivityPub Reader page (`/wp-admin/index.php?page=activitypub-social-web`)
* With no followers, you should see "Your feed is waiting to come alive. Start following people on the Fediverse" with a link
* The link should go to:
  - `options-general.php?page=activitypub&tab=following` for the blog actor (ID 0)
  - `users.php?page=activitypub-following-list` for user actors
* Follow someone, then the message should change to "Nothing new from the people you follow. Check back soon for fresh updates."

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Improve Reader empty state to show contextual messages based on following status.

</details>


src/app/routes/feed/stage.tsx
```diff
@@ -186,6 +193,32 @@ export default function FeedStage(): ReactNode {
...
+				<>
+					{ __( 'Your feed is waiting to come alive.', 'activitypub' ) }{ ' ' }
+					<a href={ followingUrl }>{ __( 'Start following people on the Fediverse', 'activitypub' ) }</a>
+				</>
+			);
+		}
+
+		// Default: following people but no posts yet.
+		return __( 'Nothing new from the people you follow. Check back soon for fresh updates.', 'activitypub' );
+	};
```
obenland: Let's extract renderEmptyStateContent() into its own EmptyState component. All of the things it depends on can be pulled through hooks.

src/app/routes/feed/stage.tsx
```diff
@@ -186,6 +193,32 @@ export default function FeedStage(): ReactNode {
...
 
+	// Generate empty state content based on following status.
+	const renderEmptyStateContent = (): ReactNode => {
+		// If search or filters are active, show simple "no results" message.
+		if ( normalizedView.search || ( normalizedView.filters && normalizedView.filters.length > 0 ) ) {
+			return __( 'No posts found.', 'activitypub' );
+		}
+
+		// If not following anyone, show link to following page.
+		if ( followingResolved && ! followingCount ) {
```
obenland: Hm, this adds quite a bit of complexity where it might not be strictly needed and we might get a better user experience without it.

Even if someone already follows a (single) user, there's still a good change they don't post frequently enough to fill a timeline, especially when starting out. Could we generalize the prompt to follow _more_ people instead?

src/app/routes/feed/stage.tsx
```diff
@@ -186,6 +193,32 @@ export default function FeedStage(): ReactNode {
...
+			const followingUrl =
+				activeActorId === 0
+					? addQueryArgs( 'options-general.php', { page: 'activitypub', tab: 'following' } )
+					: addQueryArgs( 'users.php', { page: 'activitypub-following-list' } );
+
+			return (
+				<>
+					{ __( 'Your feed is waiting to come alive.', 'activitypub' ) }{ ' ' }
+					<a href={ followingUrl }>{ __( 'Start following people on the Fediverse', 'activitypub' ) }</a>
```
obenland:
```suggestion
					{ createInterpolateElement(
						__(
							'Your feed is waiting to come alive. <a>Follow more people on the Fediverse</a> to see their posts here.',
							'activitypub'
						),
						{
							a: <a href={ followingUrl } />,
						}
					) }
```
Using `createInterpolateElement` would make this easier to translate for RTL languages.

src/app/routes/feed/stage.tsx
```diff
@@ -301,16 +334,7 @@ export default function FeedStage(): ReactNode {
...
-				<p>
-					{ normalizedView.search || ( normalizedView.filters && normalizedView.filters.length > 0 )
-						? __( 'No posts found.', 'activitypub' )
-						: __(
-								'No posts found in your feed. Posts from ActivityPub actors you follow will appear here.',
-								'activitypub'
-						  ) }
-				</p>
-			}
+			empty={ <p>{ renderEmptyStateContent() }</p> }
```
obenland:
```suggestion
			empty={ <EmptyState /> }
```
We should be able to pass a zero-prop component here.

obenland (changes requested):

src/app/routes/feed/stage.tsx
pfefferle: You mean one single text, that does not care why the list is empty and always show the follow (more) people link!?

obenland: Unless there are filters active, yes. Maybe something like [my suggestion below](https://github.com/Automattic/wordpress-activitypub/pull/2634#discussion_r2620487259)?

includes/class-post-types.php
```diff
@@ -627,32 +627,6 @@ public static function register_ap_actor_rest_field() {
...
-				$args['meta_query'] = array(); // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
-			}
-
-			$args['meta_query'][] = array(
-				'key'   => Followers::FOLLOWER_META_KEY,
-				'value' => $request['follower_of'],
-			);
-		}
-
-		return $args;
```
obenland, line 655: This seems unrelated to the PR? The callback is needed for the Followers list that's currently disabled.

obenland (commented):
Nice, much cleaner!

includes/class-post-types.php
pfefferle: PR incoming

pfefferle: This should no longer be needed!

obenland: Which PR was that?

My understanding is that the callback is needed for the Followers list that's currently disabled.

pfefferle: Ah, I see! Sorry!

pfefferle:
@obenland done!

obenland (approved):

PR #2635, opened 2025-12-15
Fix infinite re-render loop in Reader feed stage

pfefferle:
## Proposed changes:

Fixes an infinite re-render loop that caused "Maximum update depth exceeded" warnings in the Reader feed.

### Root Cause

The `@wordpress/views` hook's `updateView` function includes `onChangeQueryParams` in its dependency array. When `onChangeQueryParams` was defined inline in the component, it created a new function reference on every render. This caused `updateView` to change reference on every render, which triggered useEffects that depended on it, creating an infinite loop.

Additionally, inline empty arrays like `filters: view.filters || []` and `records || []` created new array references on every render, causing the `useFeed` hook to recalculate and triggering dependent useEffects repeatedly.

### Fix

1. **Memoize `onChangeQueryParams`**: Wrapped the callback in `useCallback` with an empty dependency array so it maintains a stable reference.

2. **Remove `updateView` from useEffect deps**: The actor-switch useEffect already has a guard condition (`prevActiveActorId.current !== activeActorId`) that prevents repeated calls, so `updateView` doesn't need to be in the dependency array.

3. **Use stable default arrays**: Created module-level constants (`DEFAULT_FIELDS`, `DEFAULT_FILTERS`, `EMPTY_FEED`) instead of inline empty arrays to prevent new references on each render.

### Other information:

## Testing instructions:

* Go to the ActivityPub Reader page (`/wp-admin/index.php?page=activitypub-social-web`)
* Open browser console
* Verify no "Maximum update depth exceeded" warning appears
* Navigate between tabs, switch actors, use filters - confirm no infinite loop warnings


pfefferle:
@obenland maybe you have some test content on your local instance you can test it with. I wiped my db to work on #2634 🤨

obenland (approved):
Nice work!

PR #2650, opened 2025-12-17
Fix TypeError when useView returns undefined

pfefferle:
## Summary

Fixes a `TypeError: can't access property "page", y is undefined` error that occurs when the `useView` hook from `@wordpress/views` returns `undefined` during initialization.

## Changes

- Add nullish coalescing guards (`view ?? DEFAULT_VIEW`) to all components and hooks that use `useView`
- Affected files:
  - `src/app/components/empty-state/index.tsx`
  - `src/app/routes/feed/stage.tsx`
  - `src/app/routes/followers/stage.tsx`
  - `src/app/hooks/use-feed-filters.ts`
  - `src/app/hooks/use-tag-filter.ts`
  - `src/app/hooks/use-object-type-filter.ts`

## Test plan

- [ ] Open the Reader page in WP Admin
- [ ] Verify no TypeError occurs on initial load
- [ ] Verify feed filtering works correctly
- [ ] Verify followers page loads without errors


obenland:
Is it not just a matter of providing a default view in EmtpyState?

PR #2656, opened 2025-12-17
Add missing config files to .distignore

pfefferle:
## Proposed changes:
* Add `webpack.config.js` to `.distignore` to exclude it from distribution packages
* Update `.gitignore` to ignore full `.vscode` directory instead of just `.vscode/settings.json`

### Other information:

N/A - configuration file update only.

## Testing instructions:

* Verify webpack.config.js is excluded when building a distribution package

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -53,3 +53,4 @@ src
 tsconfig.json
 tests
 vendor
+webpack.config.js
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,7 @@ _site
 .phpunit.cache
 .phpunit.result.cache
 .sass-cache
-.vscode/settings.json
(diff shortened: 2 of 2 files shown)


.distignore
```diff
@@ -10,21 +10,25 @@
...
 _site
+artifacts
 bin
 CHANGELOG.md
 CLAUDE.md
 CODE_OF_CONDUCT.md
 CONTRIBUTING.md
 composer.json
 composer.lock
+coverage
```
obenland: .phpunit.result.cache, .vscode, artifacts, and coverage should already be in `.gitignore` and not checked into the repository.

obenland (approved):

PR #2662, opened 2025-12-18
Add hide_for method to Comment class

pfefferle:
## Proposed changes:

* Add new `hide_for()` method to the `Comment` class that returns post types whose comments should be hidden in the WordPress admin.
* Refactor `comment_query()` and `pre_comment_approved()` methods to use the new helper method.
* Add `activitypub_hide_comments_for` filter to customize which post types have hidden comments.
* Add comprehensive unit tests for the new functionality and filter behavior.

### Other information:

## Testing instructions:

* Run the test suite: `npm run env-test -- --filter=test_hide_for`
* Verify the tests pass and confirm:
  * `ap_post` is the only post type in the default list
  * The filter can add/remove post types
  * The filter affects both comment visibility and auto-approval

### Changelog entry

Changelog entry added manually via `composer changelog:add`.

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -8,7 +8,6 @@
 namespace Activitypub;
 
 use Activitypub\Collection\Actors;
-use Activitypub\Collection\Posts;
 
 /**
  * ActivityPub Comment Class.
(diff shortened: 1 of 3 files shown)


includes/class-comment.php
```diff
@@ -704,9 +703,7 @@ public static function comment_query( $query ) {
 
 		// Do only exclude interactions of `ap_post` post type.
 		if ( \is_admin() ) {
-			if ( \get_option( 'activitypub_create_posts', false ) ) {
-				$query->query_vars['post_type'] = array_diff( \get_post_types_by_support( 'comments' ), array( Posts::POST_TYPE ) );
-			}
+			$query->query_vars['post_type'] = array_diff( \get_post_types_by_support( 'comments' ), self::get_post_types_to_hide_comments_for() );
```
jeherve: Would it be an option to limit that to `ap_post`? What other post types do we want to grab with this change?

pfefferle: If you can verify that this is "only" a problem with `ap_post` and there are no further interference with the friends plugin, sure!

jeherve (dismissed):
This looks good and tests well for me. I like the addition of the filter, it should allow third-parties to hook into that logic if necessary.

jeherve (approved):

PR #2664, opened 2025-12-18
Prevent comments on ap_post when create_posts is disabled

obenland:
## Proposed changes:

* When Reader UI is disabled, the `create_posts` option may also be disabled. However, existing `ap_post` entries can still receive incoming fediverse replies, which would create comments and trigger email notifications.
* This change adds a check in `Interactions::persist()` to reject comments on `ap_post` when the `create_posts` option is disabled, preventing unwanted notifications.

### Other information:

## Testing instructions:

1. Enable Reader UI (which forces `create_posts` on)
2. Follow some accounts and let some `ap_post` posts be created
3. Disable Reader UI
4. Have someone on the fediverse reply to one of the `ap_post` posts
5. Verify that no comment is created and no notification email is sent

### Changelog entry

Changelog entry was added manually via `composer changelog:add`.

### DIFF
--- a/.github/changelog/fix-ap-post-comments-when-create-posts-disabled
+++ b/.github/changelog/fix-ap-post-comments-when-create-posts-disabled
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Prevent comments on ap_post when create_posts option is disabled.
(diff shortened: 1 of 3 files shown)


pfefferle:
I think it is fine to add comments to an `ap_post` if there is one. The root-cause is that we decided to no longer check for the setting to be checked in the Reader PR.

obenland:
Closing in favor of #2667.

PR #2681, opened 2025-12-22
Fix fatal error when fetching remote author in embed handler

maxheadroom:
Prevent "Cannot use object of type WP_Error as array" by initializing $author as an empty array upfront and only populating it if the HTTP request succeeds. This ensures $author is always an array, allowing the code to gracefully fall back when the remote request fails.

Fixes #

## Proposed changes:
* preventing a PHP fatal error if the web finger request does not succeed and thus $author is empty

## Testing instructions:
I believe this happens when Mastodon Posts are embedded in a Blog post via the Embed block and the remote server is too slow to answer.


pfefferle (approved):
I added tests and the missing changelog entry.

Thanks @maxheadroom

PR #2702, opened 2025-12-27
feat: make following table columns filterable

Menrath:
The Following table is not filterable by the default WordPress filters like `"manage_{$post_type}_posts_columns"`. Thats why this PR introduces custom filters. I wonder whether passing the whole table instance as a filter parameter is too much, maybe just `$this->user_id` is sufficient.

### Other information:

I offer to write tests and will provide changelog entries etc., after initial review about this feature idea.

### Background:

The idea came up while porting the "Event Sources" feature of the "Event Bridge For ActivityPub" to the new following structure. I wanted to add a boolean switch to import remote events to the sites calendar, to not do so from all followings but the selected ones. My intuition was to avoid having to manage this feature on two pages (one for adding the following, and a second one for activating existing followings).

- [x] Automatically create a changelog entry

### DIFF
--- a/includes/wp-admin/table/class-blocked-actors.php
+++ b/includes/wp-admin/table/class-blocked-actors.php
@@ -160,34 +160,6 @@ class Blocked_Actors extends \WP_List_Table {
 	}
 
 	/**
-	 * Get columns.
-	 *
-	 * @return array
-	 */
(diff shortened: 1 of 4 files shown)


pfefferle:
Don't forget to add the changelog definition.

PR #2704, opened 2025-12-28
feat: make actor table columns filterable

Menrath:
The Following table is not filterable by the default WordPress filters like `"manage_{$post_type}_posts_columns"`. Thats why this PR introduces custom filters. I wonder whether passing the whole table instance as a filter parameter is too much, maybe just `$this->user_id` is sufficient.

### Background:

The idea came up while porting the "Event Sources" feature of the "Event Bridge For ActivityPub" to the new following structure. I wanted to add a boolean switch to import remote events to the sites calendar, to not do so from all followings but the selected ones. My intuition was to avoid having to manage this feature on two pages (one for adding the following, and a second one for activating existing followings).

@pfefferle I created this new PR in favor of #2702. This time with changelog and correct branch name! Your (contribution) docs improved quite a bit since I last contributed, I must honor that :) 

### More ideas:
I found that more code within the admin tables is quite similar. Maybe using a abstract class for all admin-tables would be the cleanest option? That could also offer to e.g. get rid of the "actor_list_table_key" functions, as the abstract class could enfore setting a key-property in the table class. I guess quite some code is duplicated here, using parent:: might be handy in some places. Or all use the property user_id, etc.


.github/changelog/add-make-actor-table-columns-filterable
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: added
+
+make actor table columns filterable
```
pfefferle:
```suggestion
Make actor table columns filterable.
```

pfefferle (approved):

PR #2722, opened 2026-01-06
Actors: avoid PHP warnings when trying to fetch invalid actor.

jeherve:
Fixes CMF-356

## Proposed changes:

Since `get_post` can return null or an int, it's best to explicitely check for a `WP_Post` return.

This should avoid PHP warnings like this one:

```
Warning: Attempt to read property "post_content" on null in /public_html/wp-content/plugins/activitypub/includes/functions.php on line 114
```

### Other information:

## Testing instructions:

* This should be testable on a WordPress.com sandbox, since that's where the warnings are easily repeatable (see original issue).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/.github/changelog/fix-post-explicit-wp-post-return
+++ b/.github/changelog/fix-post-explicit-wp-post-return
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Actors: avoid PHP warnings when trying to fetch invalid actor.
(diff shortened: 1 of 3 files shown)


jeherve:
Tests are failing for a different root cause, that should get fixed in #2724

pfefferle (approved):
thanks @jeherve

Issue #2735, opened 2026-01-08
Comments on AP Posts: do not hold for moderation


jeherve:
I'm creating this issue to follow-up on #2700, where a bigger issue was revealed:

Comments to `ap_post` post type are not currently always auto-approved, and can be held for moderation. It doesn't make sense for a site owner to have to review and approve comments on conversations of people they follow ; those conversations should appear in their Social Web timeline alongside the others.

Of course, that doesn't mean anything and everything should be displayed in the Social Web timeline. I think the comments should still respect the "Disallowed Comment Keys" field in Discussion settings (`disallowed_keys` option) to allow folks to "mute" terms they do not want to see in their timeline (Related issues: #1589, #2027). 

**This appears to be a bug since #2526 was aiming for those comments to be auto-approved out of the box. **

Maybe the check in https://github.com/Automattic/wordpress-activitypub/blob/bfc5210bc5b27f943542005fc6fb074c54034628/includes/class-comment.php#L771-773 should happen earlier in the method.

PR #2736, opened 2026-01-08
Comments: always exclude comments on ap_post from queries

### LINKED ISSUE #2700: Comments on AP Posts bump count of comments to be moderated in admin bar
This is a follow-up to #2662 and #2667.

The admin bar displays a comment icon, and that icon shows the number of comments that need to be moderated. However, comments on AP posts are not filtered out there:

Image

-- https://github.com/WordPress/wordpress-develop/blob/d96ff0bf873b9c63967b4a81369414c7b3da47a1/src/wp-includes/admin-bar.php#L1040

jeherve:
Fixes #2700

## Proposed changes:

Follow-up to #2662 and #2667

Until now, we were only excluding the ap_post post type from comment queries happening within wp-admin. This is problematic because the admin bar menu item makes a wp_count_comments() call from anywhere on the site where the admin bar is displayed. That includes all frontend pages.

Since we do return early in the cases where we want those comments to appear, we should be able to always exclude them in other situations.

### Other information:

## Testing instructions:

1. Start with a site where you started following other accounts, and where you've received both posts in your inbox, as well as comments on those posts.
2. In a functionality plugin, set `add_filter( 'activitypub_hide_comments_for', '__return_empty_array' );`
3. Load the comment management screen in wp-admin: you should now see the comments on some of the AP posts from the folks you follow.
4. Unapprove 2 of those comments.
5. Remove the custom code you added in step 2.
6. Visit your site's home page.
7. You should now see a counter indicating 2 pending comments in your WordPress admin bar.
8. Checkout this branch.
9. The counter should be set to 0.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/.github/changelog/fix-comment-count-ap-post-comments-admin-bar
+++ b/.github/changelog/fix-comment-count-ap-post-comments-admin-bar
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Admin bar: do not count interactions on posts from folks you follow in the comment bbubble of the admin bar.
--- a/.github/changelog/fix-comment-count-ap-post-comments-admin-bar
+++ b/.github/changelog/fix-comment-count-ap-post-comments-admin-bar
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fixed
 
-Admin bar: do not count interactions on posts from folks you follow in the comment bbubble of the admin bar.
+Admin bar: do not count interactions on posts from folks you follow in the comment bubble of the admin bar.
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -703,9 +703,9 @@ class Comment {
 		}
 
 		/*
-		 * Exclude interactions of `ap_post` post type.
-		 * We do not want to display them on the comment management screen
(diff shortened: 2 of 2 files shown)


pfefferle:
@jeherve this PR seem to break the tests!?

jeherve:
> @jeherve this PR seem to break the tests!?

Yes, I'm looking at that too. That's useful, it seems I'm breaking some expected behavior with that change. I'll see what I can find.

pfefferle:
Yes, it could interfere with the general frontend comment handling.

jeherve:
@pfefferle I don't see an easy way around that problem that won't impact the display of those comments elsewhere. I'm thinking we may be better off addressing #2735 directly instead. What do you think?

pfefferle:
@jeherve makes sense! do you want me to take over or do you want to give it a try!?

jeherve:
If you don't mind, I'd be happy to leave that one to you :)

pfefferle:
No problem! Let's close this then!

PR #2740, opened 2026-01-09
Fix ActivityPub comments being marked as spam by Akismet

### LINKED ISSUE #2735: Comments on AP Posts: do not hold for moderation

pfefferle:
Fixes #2735

## Proposed changes:

* Run `pre_comment_approved` filter at priority 11 (after Akismet's priority 10) so we can override Akismet's spam verdict for authenticated ActivityPub comments
* Fix early return logic to only short-circuit for actually approved comments (`1` or `'1'`), not for `'spam'` or `'trash'` status - previously any truthy value caused an early return

### Other information:

## Testing instructions:

* Install and activate Akismet with an API key
* Receive an ActivityPub comment (like, reply, or announce) from a federated server
* Verify the comment is auto-approved according to your settings, not marked as spam

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fixed ActivityPub comments being marked as spam by Akismet.

</details>

### DIFF
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -29,7 +29,7 @@ class Comment {
 		\add_filter( 'comment_feed_where', array( static::class, 'comment_feed_where' ) );
 		\add_filter( 'get_comment_link', array( self::class, 'remote_comment_link' ), 11, 2 );
 		\add_action( 'pre_get_comments', array( static::class, 'comment_query' ) );
-		\add_filter( 'pre_comment_approved', array( static::class, 'pre_comment_approved' ), 10, 2 );
+		\add_filter( 'pre_comment_approved', array( static::class, 'pre_comment_approved' ), 11, 2 );
 		\add_filter( 'get_avatar_comment_types', array( static::class, 'get_avatar_comment_types' ), 99 );
 		\add_action( 'update_option_activitypub_allow_likes', array( self::class, 'maybe_update_comment_counts' ), 10, 2 );
 		\add_action( 'update_option_activitypub_allow_reposts', array( self::class, 'maybe_update_comment_counts' ), 10, 2 );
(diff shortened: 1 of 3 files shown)


pfefferle:
@jeherve if you maybe have a sec to verify my thoughts!?

pfefferle:
At least co-pilot seems to be fine with the changes ;)

includes/class-comment.php
```diff
@@ -731,7 +731,12 @@ public static function comment_query( $query ) {
 	 * @return int|string|\WP_Error The approval status. 1, 0, 'spam', 'trash', or WP_Error.
 	 */
 	public static function pre_comment_approved( $approved, $comment_data ) {
-		if ( $approved || \is_wp_error( $approved ) ) {
+		/*
+		 * Only return early for already-approved comments, trash, or errors.
+		 * Don't short-circuit on 'spam' - we may want to override Akismet.
+		 * Respect 'trash' since it comes from the WordPress disallowed list.
+		 */
+		if ( 1 === $approved || '1' === $approved || 'trash' === $approved || \is_wp_error( $approved ) ) {
```
jeherve, line 739: Would it make sense to inverse that logic, and instead check if $approved is anything but `spam`?


```suggestion
		if ( 'spam' !== $approved ) {
```

pfefferle: what about "pending" (`hold` or `0`)?

jeherve: ah my bad, I didn't think about those options!

jeherve (approved):

PR #2755, opened 2026-01-13
Fix Follow requests from Pixelfed not being processed

### LINKED ISSUE #2756: Follow requests from Pixelfed are silently ignored
### Quick summary

Users on Pixelfed cannot follow WordPress blogs using the ActivityPub plugin. When clicking "Follow", nothing happens and the follow request is silently ignored.

**Reported via:** https://ruhr.social/@jansenspott/115886782619004163

### Steps to reproduce

1. Have a Pixelfed user attempt to follow a WordPress blog with the ActivityPub plugin enabled
2. Click the "Follow" button on Pixelfed
3. Nothing happens - the follow request is silently ignored
4. The follower never appears in the WordPress followers list

**Root cause:** Pixelfed does not set audience fields (`to`/`cc`/`bcc`) for Follow activities, even when sent to the shared inbox. This causes the ActivityPub plugin to be unable to determine the recipient, resulting in the Follow request being silently discarded.

**Additional context:**
- Multiple users reported this issue
- The issue affects following any WordPress blog from Pixelfed
- Previously working follows may have been established before a change in behavior

### Site owner impact

### Severity

Major

Pixelfed is a popular Fediverse platform for photo sharing. Users being unable to follow WordPress blogs from Pixelfed significantly impacts federation interoperability.

No revenue impact

### Workaround

No known workaround currently available.

### Related PR

This issue is fixed by PR #2755, which adds a fallback to check the activity's `object` field when determining recipients for the shared inbox.

pfefferle:
Fixes #2756

## Proposed changes:
* Add fallback to check the activity's `object` field when determining recipients for the shared inbox
* When no recipients are found through audience fields (`to`/`cc`/`bcc`), extract the actor URI from the `object` field and resolve it to a local user

### Other information:

Pixelfed does not set an audience (`to`/`cc`/`bcc`) for Follow activities, even when sent to the shared inbox. This caused Follow requests to be silently ignored because the recipient could not be determined.

For Follow activities, the `object` field contains the actor being followed, which can be used to identify the local recipient.

## Testing instructions:

1. Have a Pixelfed user send a Follow request to a WordPress user via the shared inbox
2. Verify the Follow request is processed and the follower appears in the followers list

Or run the new tests:
```bash
npm run env-test -- --filter "test_get_local_recipients_follow_without_audience"
npm run env-test -- --filter "test_follow_request_without_audience_via_rest"
```

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fix Follow requests from Pixelfed and other implementations that don't set audience fields.

</details>

### DIFF
--- a/includes/rest/class-inbox-controller.php
+++ b/includes/rest/class-inbox-controller.php
@@ -19,6 +19,7 @@ use function Activitypub\extract_recipients_from_activity;
 use function Activitypub\is_activity_public;
 use function Activitypub\is_collection;
 use function Activitypub\is_same_domain;
+use function Activitypub\object_to_uri;
 use function Activitypub\user_can_activitypub;
 
 /**
@@ -404,6 +405,19 @@ class Inbox_Controller extends \WP_REST_Controller {
 			}
 
 			$user_ids[] = $user_id;
+		}
+
+		// Check for an Actor in the Object field.
+		if ( empty( $user_ids ) ) {
+			$object = object_to_uri( $activity['object'] );
+
+			if ( \is_string( $object ) ) {
+				$user_id = Actors::get_id_by_resource( $object );
+
+				if ( ! \is_wp_error( $user_id ) && user_can_activitypub( $user_id ) ) {
+					$user_ids[] = $user_id;
+				}
+			}
 		}
 
 		return array_unique( array_map( 'intval', $user_ids ) );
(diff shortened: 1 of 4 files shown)


includes/rest/class-inbox-controller.php
```diff
@@ -406,6 +407,19 @@ private function get_local_recipients( $activity ) {
 			$user_ids[] = $user_id;
 		}
 
+		// Check for an Actor in the Object field.
+		if ( empty( $user_ids ) ) {
+			$object = object_to_uri( $activity['object'] );
+
+			if ( \is_string( $object ) ) {
```
jeherve: This was a bit confusing for me at first, because the docblock for `object_to_uri` says it can only return a string. Turns out it can also return null at times. Maybe that would be an opportunity to update that docblock?

jeherve (dismissed):
This looks good to me. I only have a minor comment, but that's not a blocker.

PR #2761, opened 2026-01-13
Defer add_to_outbox to async processing during post save

### LINKED ISSUE #2567: Investigate: Performance impact of @-mention lookups and hashtags on publishing
## Description

Investigate whether @-mention lookups and hashtag processing are causing slowdowns during the post publishing process.

See:

* https://wordpress.org/support/topic/activitypub-7-6-x-slows-site-down-and-does-not-federate-contents/
* https://fosstodon.org/@dusoft/115832047528572137

## Background

When publishing posts, the plugin performs:
- Webfinger lookups for @-mentions to resolve actor URLs
- Hashtag processing and formatting
- Both operations may involve network requests or database queries

These operations could potentially:
- Delay post publication
- Cause timeouts on posts with many mentions/tags
- Impact user experience during publishing

## Investigation Points

1. **@-mention lookups**:
   - Measure time for Webfinger resolution per mention
   - Check if lookups happen synchronously during publish
   - Determine impact of multiple mentions in a single post
   - Check if failed lookups cause delays

2. **Hashtag processing**:
   - Measure time for tag parsing and formatting
   - Check database query performance for tag lookups
   - Determine impact of posts with many hashtags

3. **Overall impact**:
   - Measure total publishing time with/without mentions and tags
   - Identify any blocking operations
   - Check if operations could be moved to background jobs

## Possible Optimizations

If performance issues are found:
- Move Webfinger lookups to background processing
- Cache mention lookups
- Batch process hashtags
- Add configurable timeouts
- Queue federation activities asynchronously

## Success Criteria

- Documented performance measurements for mention/hashtag processing
- Identified bottlenecks (if any)
- Recommendations for optimization (if needed)

pfefferle:
Fixes #2567

## Proposed changes:

* Defer `add_to_outbox()` to async processing via WP Cron instead of running synchronously during post save
* Rename `schedule_post_activity` to `triage` for clarity
* Add `add_to_outbox` method as async handler via `activitypub_add_to_outbox` action
* Apply same async pattern to attachment transitions
* Fall back to synchronous execution when scheduling fails (in tests or when cron is disabled)

### Other information:

## Testing instructions:

* Go to Posts > Add New
* Create a new post and publish it
* Verify the Save button responds quickly
* Go to Tools > Scheduled Actions and verify `activitypub_add_to_outbox` action was scheduled
* Verify the post appears in the outbox after cron runs
* Update a published post and verify Update activity is created
* Trash a federated post and verify Delete activity is created

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Defer outbox processing to async execution to improve publishing performance.

</details>


pfefferle:
@jeherve @Menrath what do you think about that change?

Menrath:
Looks good. As far as I remember https://github.com/Automattic/wordpress-activitypub/issues/1269 was not solved completely for some Event-Plugins, thats why I kept this workaround, so now I could remove it :)

pfefferle:
@Menrath that is even better news!

jeherve (approved):
While this works well, I do worry about the potential impact it may have on sites with a problematic WordPress' cron implementation ; WordPress' cron can be fragile and the performance gains of this PR may not outweigh the potential cron failures that would cause federation to be delayed or to fail.

I don't have a lot of context around #2567. Is this a performance issue that's been reported by some users already, something that's problematic already, or is this something you'd like to address before it becomes a problem?

All that said, I don't want to block this PR, we can certainly give it a try, and see if we see  more reported federation failures in the future.

pfefferle:
> I don't have a lot of context around https://github.com/Automattic/wordpress-activitypub/issues/2567. Is this a performance issue that's been reported by some users already, something that's problematic already, or is this something you'd like to address before it becomes a problem?

Oh yes, missed to add the wordpress.org discussion. Will add it asap.

pfefferle:
@jeherve we can't get rid of CRONs, because (as you already referenced in your comment), the preparing task already slows down publishing. If we think about a follower count above 10, we can not federate in a synchronous way without slowing down the process by minutes or even run in a timeout.

I am happy to revise the whole process, to check if we can improve it, but from my perspective, there is no way to have the plugin run without CRONs.

jeherve:
> If we think about a follower count above 10, we can not federate in a synchronous way without slowing down the process by minutes or even run in a timeout.

In this case, let's go with this approach for now, see how it works for everyone in the wild!

PR #2768, opened 2026-01-14
Add site health check for REST API accessibility

### LINKED ISSUE #2767: Security plugins blocking REST API breaks ActivityPub federation
## Description

Some WordPress security plugins restrict REST API access to authenticated users only, which breaks ActivityPub federation. Remote servers trying to interact with the site receive a 401 error:

```json
{
  "type": "about:blank",
  "title": "rest_login_required",
  "detail": "REST API restricted to authenticated users.",
  "status": 401,
  "metadata": {
    "code": "rest_login_required",
    "message": "REST API restricted to authenticated users.",
    "data": {"status": 401}
  }
}
```

## Affected Plugins

Several security plugins are known to restrict REST API access:

- **[Disable WP REST API](https://wordpress.org/plugins/disable-wp-rest-api/)** - Blocks all unauthenticated REST API requests by default
- **Wordfence** - Has settings that can restrict API access (Firewall Options → Brute Force Protection → "Disable application passwords")
- **Patchstack** - Has "Restrict WP REST API Access to authenticated users only" option
- **Solid Security (iThemes Security)** - Can block API requests
- **All-In-One Security (AIOS)** - May inadvertently block API requests
- **Sucuri Security** - Known to sometimes cause REST API issues

## Impact

When REST API is restricted to authenticated users:
- Remote ActivityPub servers cannot fetch actor profiles
- Follow requests fail
- Replies and interactions from the fediverse are rejected
- WebFinger lookups may fail

## Possible Solutions

1. **Documentation** - Add documentation warning users about this incompatibility and how to whitelist ActivityPub endpoints
2. **Detection** - Add admin notice when REST API restriction is detected
3. **Compatibility** - Investigate if we can hook into these plugins to whitelist our endpoints (e.g., using `rest_authentication_errors` filter priority)

## References

- [Wordfence REST API discussion](https://wordpress.org/support/topic/wordfence-disabled-rest-api/)
- [Disable WP REST API plugin](https://wordpress.org/plugins/disable-wp-rest-api/)

pfefferle:
Fixes #2767

## Proposed changes:

* Add a new Site Health check that detects when security plugins are blocking unauthenticated REST API access
* Uses the `rest_authentication_errors` filter to detect blanket API restrictions
* Does NOT falsely flag AUTHORIZED_FETCH as an issue (that uses HTTP Signatures on specific endpoints, not blanket blocking)
* Shows a critical warning with guidance on which plugins to check (Wordfence, Disable REST API, Patchstack, Solid Security, etc.)

### Other information:

## Testing instructions:

* Install the "Disable WP REST API" plugin and activate it
* Go to Tools → Site Health
* Verify the "REST API is restricted to authenticated users" critical warning appears
* Deactivate the plugin and verify the check passes
* Enable AUTHORIZED_FETCH in ActivityPub settings
* Verify the health check still passes (not falsely flagged)

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Add site health check to detect when security plugins block REST API access.

</details>

### DIFF
--- a/includes/wp-admin/class-health-check.php
+++ b/includes/wp-admin/class-health-check.php
@@ -643,45 +643,68 @@ class Health_Check {
 	}
 
 	/**
-	 * Check if REST API is accessible to unauthenticated requests.
+	 * Check if ActivityPub endpoints are accessible to unauthenticated requests.
 	 *
-	 * This checks if a security plugin is blocking all unauthenticated REST API access
-	 * via the `rest_authentication_errors` filter. This is different from AUTHORIZED_FETCH
-	 * which uses HTTP Signature verification on ActivityPub-specific endpoints.
+	 * Makes an actual HTTP request to the ActivityPub inbox endpoint.
+	 * Only reports errors from security plugins (error titles not starting
+	 * with 'activitypub_').
 	 *
 	 * @return bool|\WP_Error True if accessible, WP_Error otherwise.
 	 */
 	public static function is_rest_api_accessible() {
-		// Temporarily remove our own authentication to test what other plugins do.
-		$current_user = \wp_get_current_user();
-		\wp_set_current_user( 0 );
+		// Test the application actor's inbox endpoint (always available).
+		$actor = Actors::get_by_id( Actors::APPLICATION_USER_ID );
+		$url   = $actor->get_inbox();
 
-		/**
-		 * Simulate what happens when an unauthenticated REST API request comes in.
-		 * Security plugins hook into `rest_authentication_errors` to block access.
-		 *
-		 * Pass `null` to indicate no authentication has been performed yet,
-		 * which is what WordPress does for unauthenticated requests.
-		 */
-		$authentication_error = \apply_filters( 'rest_authentication_errors', null );
+		// Make an unauthenticated request.
+		$response = \wp_remote_get(
+			$url,
+			array(
+				'timeout'   => 5,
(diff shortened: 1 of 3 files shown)


includes/wp-admin/class-health-check.php
```diff
@@ -590,4 +595,116 @@ public static function test_wp_cron() {
...
+		$result['label']          = \__( 'REST API is restricted to authenticated users', 'activitypub' );
+		$result['badge']['color'] = 'red';
+		$result['description']    = \sprintf(
+			'<p>%s</p><p>%s</p>',
+			\__( 'A plugin or custom code is restricting REST API access to authenticated users only. This prevents remote ActivityPub servers from interacting with your site.', 'activitypub' ),
+			$check->get_error_message()
+		);
+		$result['actions']        = \sprintf(
+			'<p>%s</p>',
+			\__( 'Check your security plugin settings (Wordfence, Disable REST API, Patchstack, Solid Security, etc.) and ensure ActivityPub endpoints are accessible to unauthenticated requests.', 'activitypub' )
```
jeherve: I don't know if I would mention plugin names in there, since we'd then have to update that list as more plugins come along or existing plugins get updated. 
Maybe we can editing the message to keep only "security plugin"?

jeherve (commented):
Just a quick comment, but not a blocker.

PR #2773, opened 2026-01-15
Labels: stop recommending the use of labels with [Type] prefix

jeherve:
Part of QAO-204

## Proposed changes:

We're moving away from those labels with the introduction of Linear.

### Other information:

## Testing instructions:

Not much to test here. Check for typos / non-existing labels.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,7 +1,7 @@
 name: Feature Request
 description: Suggest an idea for the ActivityPub plugin!
 title: 'Feature Request:'
-labels: ['[Type] Feature Request']
+labels: ['New Feature']
 type: 'Enhancement'
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,7 +1,7 @@
 name: Feature Request
 description: Suggest an idea for the ActivityPub plugin!
 title: 'Feature Request:'
-labels: ['New Feature']
+labels: ['Enhancement']
 type: 'Enhancement'
 body:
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -1,7 +1,7 @@
 name: Feature Request
 description: Suggest an idea for the ActivityPub plugin!
 title: 'Feature Request:'
-labels: ['[Type] Feature Request']
+labels: ['New Feature']
```
pfefferle: Isn't it `Enhancement`?

jeherve: I think the original idea was to have different templates, one for enhancement and one for feature requests. Here is how it's done in Jetpack for example:
- https://github.com/Automattic/jetpack/blob/trunk/.github/ISSUE_TEMPLATE/enhancement.yml
- https://github.com/Automattic/jetpack/blob/trunk/.github/ISSUE_TEMPLATE/feature-request.yml

This repo doesn't have an enhancement template though. Do you think we should create one, or would you rather we keep only the existing one, but with the Enhancement label? I'm good with both options, maybe even prefering using only one template, it's easier to pick for users I think.

pfefferle: I just learned today to only use Bug, Todo and Enhancement, that is why I asked. I am totally happy with either way, I was just curious!

jeherve: I switched to Enhancement in e9820542c4bb1d8a7d4b2f9559e5424a0d5ef0a9

pfefferle (approved):

PR #2815, opened 2026-01-22
Show notice when trying to follow already-followed account

### LINKED ISSUE #2803: Following: no notice when trying to follow account that I already follow
### Steps to reproduce

1. Go to Users > Following
2. Copy the profile URL of an account you already follow
3. Enter that profile URL in the follow field and submit
    * You'll receive an "Account followed." success message, even though you were already following that person.

Ideally, I think we should display a message saying that the person wasn't added because you're already following them.

I think it's especially important since we don't trigger a new following attempt behind the scenes, there is no matching outbox item since we already follow them.

pfefferle:
Fixes #2803

Screenshot 2026-01-22 at 14 20 45

## Proposed changes:
* Return `WP_Error` with code `activitypub_already_following` from `Following::follow()` when the user is already following the actor
* Display an informational notice "You are already following..." in the UI instead of the misleading "Account followed." success message

### Other information:

## Testing instructions:

1. Go to Users > Following
2. Follow any account
3. Try to follow the same account again
4. Verify you see "You are already following..." info message instead of "Account followed." success message

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Show informational notice when trying to follow an already-followed account.

</details>


includes/wp-admin/table/class-following.php
```diff
@@ -141,6 +141,18 @@ public function process_action() {
...
 
+				// Check if already following before making the request.
+				$actor = Remote_Actors::get_by_uri( $profile );
+				if ( ! \is_wp_error( $actor ) ) {
+					$following = \get_post_meta( $actor->ID, Following_Collection::FOLLOWING_META_KEY, false );
+					$pending   = \get_post_meta( $actor->ID, Following_Collection::PENDING_META_KEY, false );
+					if ( \in_array( (string) $this->user_id, $following, true ) || \in_array( (string) $this->user_id, $pending, true ) ) {
+						/* translators: %s: Account profile that is already being followed */
+						\add_settings_error( 'activitypub', 'followed', \sprintf( \__( 'You are already following &#8220;%s&#8221;.', 'activitypub' ), \esc_html( $profile ) ), 'info' );
+						break;
```
jeherve: Should we display 2 different messages, depending on the follow status? I think it may be interesting to know when you have a pending follow vs. you're already following the person.

PR #2817, opened 2026-01-22
Show warning when user cannot federate replies to fediverse comments

### LINKED ISSUE #2816: Show warning when user without ActivityPub capability replies to fediverse comments
## Problem

When a logged-in user **without the `activitypub` capability** replies to a comment received from the fediverse, their reply is created as a local-only comment and **never federates** to the original author. There's no indication to the user that this is happening.

Currently, the `are_comments_allowed()` method in `Comment` class returns `true` for users who can't federate, showing them a normal WordPress reply link. The reply gets saved locally but silently fails to federate because `should_be_federated()` returns `false` for users without the capability.

This creates a confusing experience where users think they're replying to someone on Mastodon/fediverse, but their reply never reaches the recipient.

## Expected Behavior

When a user without ActivityPub capability tries to reply to a fediverse comment, they should see:
- A warning/notice explaining their reply won't be federated
- Guidance on how to enable federation for their account (if applicable)
- Optionally, still allow the local reply but make it clear it's local-only

## Current Code Flow

1. `Comment::comment_reply_link()` checks `are_comments_allowed()`
2. `are_comments_allowed()` returns `true` if comment is local OR user has `activitypub` capability
3. For fediverse comments + users without capability, it returns `false` → shows remote reply block
4. But for **logged-in users without capability**, it currently returns `false`, showing the remote reply block

Wait, re-reading the code - it actually shows the remote reply block for users without the capability. But the issue is that logged-in users who CAN post comments locally but CAN'T federate see the normal reply link and their replies don't federate.

## Proposed Solution

In `Comment::comment_reply_link()`, add a third case for logged-in users without ActivityPub capability replying to fediverse comments:

1. If user can federate → show normal reply link (reply will federate) ✅
2. If user is not logged in → show remote reply block ✅  
3. **NEW: If user is logged in but can't federate → show warning + reply link**

The warning could say something like:
> "Your reply will be saved locally but won't be sent to the fediverse. To enable federation, [contact your administrator / enable ActivityPub in your profile]."

## Related

- Linear issue CMF-386
- Investigation revealed users without `activitypub` capability can reply but replies don't federate

pfefferle:
Fixes #2816

Screenshot 2026-01-22 at 20 25 56

## Proposed changes:

* Show a warning message instead of the reply link when a logged-in user without ActivityPub capability tries to reply to a fediverse comment.
* Unify `are_comments_allowed()` and `should_be_federated()` to both use `user_can( $user_id, 'activitypub' )` instead of the inconsistent `publish_posts` check.

### Other information:

## Testing instructions:

* Create a comment from the fediverse (e.g., reply to a post from Mastodon)
* Log in as a user WITHOUT `activitypub` capability (e.g., subscriber role)
* View the fediverse comment - should see warning "Your account is not able to reply to fediverse comments."
* Log in as a user WITH `activitypub` capability (e.g., editor role)
* View the same fediverse comment - should see normal reply link
* Log out and view the comment - should see remote reply block

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Show warning instead of reply link when logged-in user cannot federate replies to fediverse comments.

</details>


pfefferle:
@davemart-in @robertbpugh @jeherve I would love to have your feedback on the message. See screenshot? Can you think of a shorter one? Is it fine like it is?

davemart-in:
Thoughts on something like:

> **This reply won’t be sent to the fediverse.**
> [Enable ActivityPub](https://example.net) for your account to reply to fediverse comments.

pfefferle:
Yes! (Sorry for the missing context)

jeherve:
> **This reply won’t be sent to the fediverse.**
> [Enable ActivityPub](https://example.net) for your account to reply to fediverse comments.

Should we add a mention that the author will consequently not receive a notification of your reply? Since we have `$comment` available we could even make that message a bit more dynamic. Something like:

> **Your reply won’t be sent to the fediverse ; $comment_author will not be notified of your reply.**
> [Enable ActivityPub](https://example.net) for your account to reply to fediverse comments.

pfefferle:
@jeherve Ah, I misread the comment of @davemart-in ! The message replaces the link completely, so it is not even possible to send a reply!

I think it makes no sense to have local replies to comments from the fediverse because this would confuse people even more.

jeherve:
That makes sense. We shouldn't show the comment form / offer the option to reply if the person you're chatting with will not know you replied to them.

How about something like this?

"$comment_author is on the Fediverse. To reply to them, [enable ActivityPub for your account](link)."

pfefferle:
@jeherve done!

PR #2837, opened 2026-01-28
Add Following block and shared actor-list components

### LINKED ISSUE #2838: Add Fediverse Following block
## Summary

Add a new block to display the accounts a user follows in the Fediverse, complementing the existing Fediverse Followers block.

## Use Case

Users want to showcase not only who follows them, but also who they follow. This provides a more complete picture of their social connections and helps visitors discover interesting accounts in the Fediverse.

## Proposed Solution

- Create a new "Fediverse Following" block similar to the existing Followers block
- Support the same style variants (Default, Card, Compact)
- Include pagination for large following lists
- Respect social graph privacy settings

## Additional Context

This pairs with the existing Fediverse Followers block to provide a complete social graph display.

pfefferle:
Fixes #2838

## Proposed changes:

* Add new "Fediverse Following" block to display accounts the user follows
* Create shared React components for actor lists (`ActorList`, `ActorItem`, `Pagination`) in `src/shared/actor-list/`
* Add shared SCSS styles with mixins for card and compact style variants
* Add PHP helper method `Blocks::render_actor_list()` for server-side rendering
* Refactor Followers block to use the shared components (reduces code duplication)
* Add Jest mocks for `@wordpress/interactivity` module

### Other information:

## Testing instructions:

* Start wp-env: `npm run env-start`
* Add test data:
  ```bash
  wp-env run cli "wp activitypub add_follower https://mastodon.social/@Gargron"
  wp-env run cli "wp activitypub add_follower https://mastodon.social/@Mastodon"
  wp-env run cli "wp activitypub follow https://mastodon.social/@Gargron"
  wp-env run cli "wp activitypub follow pfefferle@notiz.blog"
  ```
* Accept pending follows for testing:
  ```bash
  wp-env run cli "wp eval '\$pending = Activitypub\Collection\Following::get_pending(0); foreach (\$pending as \$p) { Activitypub\Collection\Following::accept(\$p, 0); }'"
  ```
* Go to http://localhost:8888/wp-admin/
* Create or edit a post/page
* Add the "Fediverse Followers" block - verify it displays followers with avatars
* Add the "Fediverse Following" block - verify it displays following with avatars
* Test both blocks with different style variants (Default, Card, Compact)
* Test pagination by setting "Number of Followers/Following" to 1

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add Fediverse Following block to display accounts the user follows.

</details>


src/following/style.scss
jeherve, line 1: Instead of having 2 separate stylesheets for the following and follower blocks that do the same thing, could we move everything to the `src/shared/actor-list/style.scss` and target the 2 selectors at once?

src/following/render.php
```diff
@@ -0,0 +1,124 @@
+<?php
+/**
+ * Server-side rendering of the following block.
```
jeherve, line 3: Similarly to above, is there more we could extract to avoid duplicating with the followers block? 

At the same time this may not really be a big deal, since we probably won't need a third similar block.

src/following/view.js
jeherve, line 1: Same like above, I think it'd be nice to extract a lot of this to avoid duplication with the followers block.

src/shared/actor-list/actor-item.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param {Object} props                   The component props.
+ * @param {string} props.name              The name of the actor.
+ * @param {Object} props.icon              The icon of the actor.
+ * @param {string} props.url               The URL of the actor.
+ * @param {string} props.preferredUsername The preferred username of the actor.
+ * @return {JSX.Element} The actor component.
+ */
+export function ActorItem( { name, icon, url, preferredUsername } ) {
+	const handle = `@${ preferredUsername }`;
```
jeherve: I initially found it a bit odd to have `preferredUsername` here and not `webfinger`. Since the preferred username doesn't include the instance, it feels like we're missing information when we're looking at the list.
That said that's not a new thing and you were already doing it that way for the Followers block, so I suppose it makes sense to keep doing it that way to remain consistent.

src/following/edit.js
```diff
@@ -0,0 +1,269 @@
...
+						{ settingsUrl
+							? createInterpolateElement(
+									/* translators: <a> is a link to the profile settings page. */
+									__(
+										'The selected user has their social graph hidden. This block will not display following on the frontend. <a>Edit privacy settings</a>',
+										'activitypub'
+									),
+									{
+										/* eslint-disable-next-line jsx-a11y/anchor-has-content -- Content provided by createInterpolateElement */
+										a: <a href={ settingsUrl } target="_blank" rel="noopener noreferrer" />,
```
jeherve: Maybe you could use the `ExternalLink` component here, from `@wordpress/components`?

src/following/edit.js
```diff
@@ -0,0 +1,269 @@
...
+	const settingsUrl = useMemo( () => {
+		if ( ! canEditSettings || ! siteUrl ) {
+			return null;
+		}
+
+		if ( selectedUser === 'blog' ) {
+			return siteUrl + '/wp-admin/options-general.php?page=activitypub&tab=blog-profile';
+		}
+
+		return siteUrl + '/wp-admin/profile.php#activitypub';
```
jeherve, line 145: Should we use a URL that includes the user ID here, like `wp-admin/user-edit.php?user_id=x`, since the notice may be shown for a user that is not the person currently editing the template, and thus going to their own profile page will not be helpful.

Edit: now I see that we have the `canEditSettings` check just above, so we may not end up with this use-case in any scenario.

src/following/edit.js
```diff
@@ -0,0 +1,269 @@
...
+			return [];
+		}
+
+		return usersOptions.filter( ( { value } ) => {
+			// Always keep 'inherit' (Dynamic User) option.
+			if ( value === 'inherit' ) {
+				return true;
+			}
+			// Check blog social graph visibility.
+			if ( value === 'blog' ) {
```
jeherve, line 90: Can the blog user follow others in practice? Does it need to be an option for the Following block?

src/following/edit.js
```diff
@@ -0,0 +1,269 @@
...
+
+				{ ! showHiddenNotice && selectedUser !== 'inherit' && (
+					<ActorList
+						selectedUser={ selectedUser }
+						perPage={ perPage }
+						order={ order }
+						endpoint="following"
+						page={ page }
+						setPage={ setPage }
+						emptyMessage={ __( 'Not following anyone yet.', 'activitypub' ) }
```
jeherve, line 261: I'm thinking the error message could be expanded a bit more, say something like `@user@instance.tld does not follow anyone yet`.

src/following/edit.js
```diff
@@ -0,0 +1,269 @@
...
+						value={ perPage }
+						onChange={ setAttributeWithPageReset( 'per_page' ) }
+						min={ 1 }
+						max={ 10 }
+						__next40pxDefaultSize
+					/>
+				</PanelBody>
+			</InspectorControls>
+
+			<div className={ 'wp-block-activitypub-following ' + className }>
```
jeherve: Pet-peeve of mine, but I think that strings are more readable when using template literals, here and elsewhere in the file (like for the wp-admin URLs):


```suggestion
			<div className={ `wp-block-activitypub-following ${ className }` }>
```

jeherve (commented):
This is looking good, and tests well. I only have minor nitpick remarks.

Mostly, I appreciate that you extracted some things to avoid duplication, but I wonder if maybe you could have extracted even more (see my comments below).


Also, right now I get an API query that returns an error every time I add the block to the editor, because it tries to fetch the accounts the blog user follows (`wp-json/activitypub/1.0/actors/0/following?per_page=5&order=desc&page=1&context=full&_locale=user`). But that user does not follow anyone:

https://github.com/Automattic/wordpress-activitypub/blob/addd7174c2bd31c32ec31f299d847fc3ccc59851/includes/rest/trait-collection.php#L40-L48

src/following/edit.js
pfefferle: this is an option in the following block!

PR #2850, opened 2026-01-31
Don't filter the comment query when `type__not_in` has been set

### LINKED ISSUE #2849: `pre_get_comments` breaks "manual" `WP_Comment_Query->query()` calls with `type__not_in` parameters
### Quick summary

The way `pre_get_comments` is being used it overwrites `type__not_in`:

```php
// Exclude likes and reposts by the ActivityPub plugin.
$query->query_vars['type__not_in'] = self::get_comment_type_slugs();
```

### Steps to reproduce

In the `comments.php` of a theme, add this when the ActivityPub plugin is active:

```php
    global $post;
    $args           = [
            'post_id'      => $post->ID,
            'type__not_in' => 'comment',
    ];
    $comments_query = new WP_Comment_Query;
    $comments       = $comments_query->query( $args );
    var_dump($comments);
```

and you'll see comments in the array 

### Site owner impact

### Severity

Minor

No revenue impact

_No response_

lostfocus:
This should fix #2849 

### Changelog entry

<details>

#### Significance

-   [X] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Don't filter the comment query when type__not_in has been set

</details>


pfefferle:
@lostfocus can you add/fill out the changelog section of the PR template?

pfefferle:
I think instead of returning we would have to merge both arrays!? Otherwise the exclude of the plugin will be ignored.

lostfocus:
I personally think that if someone explicitly writes their query like this, they (hopefully) know what they're doing and they might expect it to include the comments that are usually excluded. But I can see your point as well.

So I'll put the burden of this decision on you 😅 I don't have a problem with adding a merge function, either

pfefferle:
Can you maybe give me some background of how you experienced the issue, to help me understand the usecase?

lostfocus:
I do my own sorting of comments in my comment template. Basically I have all "normal" comments as comments and then I query the remaining comments like this:

```php
global $post;
$args           = [
        'post_id'      => $post->ID,
        'type__not_in' => 'comment',
];
$comments_query = new WP_Comment_Query;
$comments       = $comments_query->query( $args );
var_dump($comments);
```
    
I have some custom logic to determine how, when and if I show each of them. And of course with that query I expect to also get reposts, likes and quotes. (But not comments.)

A good example is this post: https://lostfocus.de/2025/11/24/github-account-handling/ - all comments show up as "pings" a second time, while the one repost doesn't show up at all.

lostfocus:
Actually with a query like this I'd expect to get everything back, but I understand why that's not practical:

```php
global $post;
$args           = [
        'post_id'      => $post->ID,
];
$comments_query = new WP_Comment_Query;
$comments       = $comments_query->query( $args );
var_dump($comments);
```

pfefferle (approved):

PR #2851, opened 2026-01-31
[C2S] Add Client-to-Server ActivityPub API support

### LINKED ISSUE #1255: Accept POST requests in the Outbox endpoint
### What

The spec defines POST requests to the Outbox endpoint for Client => Server communication. Let's enable that endpoint for Clients that want to implement it with Application Passwords.

### Why

We'd be one of the first ActivityPub implementations to offer that feature.

Let's give `@bengo@mastodon.social` a ping when that's done. as he showed interest in it at FOSDEM.

pfefferle:
Fixes #1255

### OAuth 2.0 Foundation
* Implement OAuth 2.0 Authorization Code flow with PKCE for secure third-party client authentication.
* Support RFC 7591 dynamic client registration and Client ID Metadata Document (CIMD) auto-discovery.
* Add OAuth token introspection (RFC 7662) and revocation (RFC 7009).
* Publish Authorization Server Metadata at `/.well-known/oauth-authorization-server`.
* Support Application Passwords as alternative authentication for C2S.

### POST to Outbox
* Add POST to Outbox supporting Create, Update, Delete, Follow, Undo Follow, Like, and Announce activities.
* Notes are created with `status` post format; Articles as regular posts.
* Content goes through a `prepare_content()` pipeline: `wpautop()` → link processing → hashtag processing → HTML-to-Gutenberg-blocks conversion.
* Hashtags in content are automatically saved as WordPress post tags.
* Outbox handlers are thin wrappers delegating to `Collection\Posts` for CRUD operations.

### Inbox & Proxy
* Add GET Inbox for reading received activities with OAuth authentication.
* Add proxy endpoint for fetching remote ActivityPub objects on behalf of authenticated clients.

### Architecture
* Split `Collection\Posts` into `Collection\Posts` (local CRUD for C2S) and `Collection\Remote_Posts` (federated remote posts from S2S).
* Extract outbox handlers into `Handler\Outbox` namespace (separate from S2S inbox handlers).
* Add `Blocks::convert_from_html()` for converting raw HTML into Gutenberg block markup.
* Add `Verification` trait for centralized OAuth + scope authentication checks.
* Add CORS headers for cross-origin C2S client access.

### Connected Applications UI
* Add "Connected Applications" section to user profile page.
* Users can view active OAuth tokens, revoke individual tokens, or revoke all.
* Users can manually register new OAuth clients (name + redirect URI) and receive a client ID.

### Other information:

## Testing instructions:

* Go to Settings → ActivityPub → Advanced and enable "Client-to-Server".
* Verify OAuth endpoints appear in actor JSON (`oauthAuthorizationEndpoint`, `oauthTokenEndpoint`).
* Install an ActivityPub C2S client (e.g. [box](https://github.com/go-ap/box)) and authorize it against your site.
* POST a Note to the outbox and verify a WordPress post is created with `status` format and block markup content.
* POST an Article with `name` field and verify title and excerpt are set.
* Verify hashtags in content are saved as WordPress tags.
* GET the inbox and verify activities are returned.
* Test Follow/Unfollow via outbox POST.
* Test token refresh and revocation.
* Test Like and Announce (boost) via outbox POST.
* Verify Delete removes the corresponding post.
* Go to Profile → Connected Applications and verify you can register a new client and revoke tokens.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Support for ActivityPub Client-to-Server (C2S) protocol, allowing apps like federated clients to create, edit, and delete posts on your behalf.

</details>


mediaformat:
One thing clients will need is a [proxyURL](https://www.w3.org/wiki/ActivityPub/Primer/proxyUrl_endpoint) endpoint, this allows the client to load Actor data from the inbox Activities

This demo illustrates what I mean: https://social.coop/@django/115756317440812767

steve-bate:
I've been updating Flowz and it can now authenticate with Wordpress (with a little custom coding). I've been able to create posts and view an outbox timeline.

However, I've encounted a few issues. Keep in mind that I know a lot more about ActivityPub than about all the OAuth2 variants and related specs. Please correct me if I'm raising non-issues or doing something incorrectly.

**OAuth2 AS metadata**

I initially did some testing with Evan's onepage.pub server. He uses the RFC 8414 endpoint (`.well-known/oauth-authorization-server`) to expose AS metadata. Wordpress appears to use `/wp-json/activitypub/1.0/oauth/authorization-server-metadata`. Am I correct that this is not a standard discovery endpoint? For now, I'm redirecting to the .well-known location in my reverse-proxy. I also needed to add CORS headers so it's available from a browser.

**Actor discovery**

After I have an access token, I'm not sure about the best way to get the actor URI with Wordpress. The onepage.pub server access token is a JWT that can be decoded to get the actor URI. For now, I'm using the Wordpress `oauth/introspect` endpoint which providers the WP username and then I'm constructing the actor URI manually from that (/author/USERNAME/activitypub/). Of course, this isn't a general solution so I'm interested if you have any suggestions. I also needed to inject CORS headers into this request.

**Outbox Activities**

After posting one Note, the outbox had a single page with `totalItems` of 1. That's what I would expect. However, there were actually 4 activities in `orderedItems`. There were 3 Update/Person activities and the Create/Note that I had posted. That was surprising to me.

steve-bate:
One more comment... I'm using a approach where the user enters an ActivityPub server domain (or instance prefix URL) and then Flowz executes the OAuth2 auth flow. I'm assuming that I'll be able to obtain the actor URI based on the auth flow (vs the user manually entering it or entering a handle to be resolved by WebFinger). Is that reasonable?

pfefferle:
@steve-bate 

> I initially did some testing with Evan's onepage.pub server. He uses the RFC 8414 endpoint (.well-known/oauth-authorization-server) to expose AS metadata. Wordpress appears to use /wp-json/activitypub/1.0/oauth/authorization-server-metadata. Am I correct that this is not a standard discovery endpoint? For now, I'm redirecting to the .well-known location in my reverse-proxy. I also needed to add CORS headers so it's available from a browser.

Normally WordPress should URL rewrite the `.well-known/oauth-authorization-server` to `/wp-json/activitypub/1.0/oauth/authorization-server-metadata`! I will check why that does not happen. Maybe a missing flushing.

pfefferle:
> After posting one Note, the outbox had a single page with totalItems of 1. That's what I would expect. However, there were actually 4 activities in orderedItems. There were 3 Update/Person activities and the Create/Note that I had posted. That was surprising to me.

Have you made some changes inside of WordPress before, that triggered the Updates?

Update: Ah, I think the counter might be wrong, because it might exclude the non-public infos! I will check!

steve-bate:
> Have you made some changes inside of WordPress before, that triggered the Updates?

This is a very clean Wordpress install. I just brought it up with Docker, ran the initial setup, created my user, and posted the Note from Flowz. I'm almost sure I didn't do any edits to my user profile in WP.

pfefferle:
> This is a very clean Wordpress install. I just brought it up with Docker, ran the initial setup, created my user, and posted the Note from Flowz. I'm almost sure I didn't do any edits to my user profile in WP.

It is not always easy to find the right triggers in WordPress, so I might track too much 🫣

pfefferle:
The latest push should fix:

- Add `me` parameter to OAuth token and introspect responses with actor URI
  (IndieAuth convention) to help clients discover actor identity after auth
- Skip outbox totalItems override for authenticated requests so C2S clients
  get accurate collection counts matching orderedItems
- Add migration to flush rewrite rules for .well-known/oauth-authorization-server

@steve-bate Can you re-check the endpoints? If the well-known is not working, please go to the permalinks page and simply hit save. You do not have to change anything. This should regenerate the rewrite rules.

pfefferle:
@steve-bate the plugin should also add CORS headers for every endpoint already. Can you list the ones you found, that are still missing?

pfefferle:
@steve-bate found the CORS issue! I messed up the regex 🫣

pfefferle:
the `me` attribute is borrowed by https://indieauth.spec.indieweb.org/#profile-url-response

steve-bate:
> The latest push should fix:
> 
> * Add `me` parameter to OAuth token and introspect responses with actor URI

👍 Works very well. The AP Client API task force should consider "standardizing" this.

> * Add migration to flush rewrite rules for .well-known/oauth-authorization-server
> @steve-bate Can you re-check the endpoints? If the well-known is not working, please go to the permalinks page and simply hit save. You do not have to change anything. This should regenerate the rewrite rules.

👍 The oath-authorization-server endpoint is now working (redirect + CORS). 

However, I'm stilling seeing CORS issues when dereferencing the `/?author=1` actor URI and the `/author/steve/` redirect.

> * Skip outbox totalItems override for authenticated requests so C2S clients
>   get accurate collection counts matching orderedItems

I'm going reset my test instance (clear all the data) and try this again.

pfefferle:
Ah! Forgot about the profile URLs/IDs!!!

steve-bate:
> > Skip outbox totalItems override for authenticated requests so C2S clients
> > get accurate collection counts matching orderedItems

> I'm going reset my test instance (clear all the data) and try this again.

The `totalItems` is matching the length of `orderedItems` now. I notice that after a clean install and no profile changes, I see one Update/Person activity in the outbox. Is that expected? (There's no Create/Person.)

Posting a Note to the outbox works.

pfefferle:
@steve-bate The latest push adds CORS headers to ActivityPub JSON responses served via profile URLs (`/?author=1`, `/author/username/`, etc.). This should resolve the CORS issues you were seeing when dereferencing actor URIs from the browser.

pfefferle:
Addressed all 5 Copilot review comments in commit 8e70b689:

1. **OAuth scope bypass fix** - Don't fall back to Application Passwords when OAuth auth succeeds but scope check fails
2. **SQL cleanup fix** - Changed `AND` to `OR` for transient cleanup query
3. **Cancel button fix** - Now submits form as denial to return proper OAuth error to client
4. **Regex escaping** - Escaped dot in `.well-known` rewrite rule
5. **IPv6 loopback fix** - Added `::1` (unbracketed) to localhost hosts list

pfefferle:
> I see one Update/Person activity in the outbox. Is that expected? (There's no Create/Person.)

I will check this one.

pfefferle:
Addressed 7 of 8 new Copilot comments in commit 3e8ae700:

1. ✅ **Token validation O(1)** - Direct DB lookup by meta_key instead of iterating users
2. ✅ **Token revoke cleanup** - Untrack user when last token is revoked  
3. ✅ **Inbox WP_Error check** - Return error before using invalid $user
4. ⏭️ **C2S setting** - Skipped (design decision: keep C2S always enabled or add setting?)
5. ✅ **OAuth state leak** - Reset $current_token at start of each auth
6. ✅ **Outbox WP_Error check** - Return error before using invalid $user
7. ✅ **Handler return contract** - Added handling for int (outbox ID) and false (not allowed)
8. ✅ **Auth code cleanup** - Only delete expired transients, not in-progress ones

Re: #4 - The PR description mentions an `activitypub_enable_c2s` setting but the current implementation always exposes C2S endpoints. Should we add the setting, or update the PR description to reflect always-on behavior?

steve-bate:
I can dereference the inbox URI, but it seems like I can't dereference the activity or object URI for the items in the inbox. Is this intentional?

I didn't notice it at first because I was using the embedded activities from the inbox response to render the timeline.

pfefferle:
@steve-bate can you make an example? I think Inbox-Activities from remote, need to be queried from the original source?

steve-bate:
> @steve-bate can you make an example? I think Inbox-Activities from remote, need to be queried from the original source?

Inbox collection page: `https://wordpress.stevebate.dev/wp-json/activitypub/1.0/actors/1/inbox?page=1`
One Create/Note in inbox. Can be dereferenced.

Create/Note activity URI: `https://wordpress.stevebate.dev/wp-json/activitypub/1.0/actors/1/inbox?page=1`
This note was sent from wordpress user to another (local).
When dereferenced, this results in 404 and and text/html content although only 'application/json' was requested.

Note URI: `https://wordpress.stevebate.dev/wp-json/activitypub/1.0/objects/c1db5d3f-c434-4a32-8a79-c36dd077d0d2`
When dereferenced, this returns:
```
{"type":"about:blank","title":"rest_no_route","detail":"No route was found matching the URL and request 
method.","status":404,"metadata":{"code":"rest_no_route",
"message":"No route was found matching the URL and request method.","data":{"status":404}}}
```

(Note: the server is on an internal network, you won't be able to access it.)

pfefferle:
Ah, I see!

I haven't thought about internal posts!

pfefferle:
hmmm 🤔

has you client generated: `https://wordpress.stevebate.dev/wp-json/activitypub/1.0/objects/c1db5d3f-c434-4a32-8a79-c36dd077d0d2`? This is no ID that is used by the WordPress plugin!? Do you change the ID according to what the plugin returns after publishing something through the outbox?

steve-bate:
I don’t think so. It’s theoretically possible that the client could set the id of the Note, but Flowz wouldn’t have set a /wp-json/* URI.

EDIT: I checked in a debugger and I'm not setting a Note URI when posting to the outbox. Typically that would be set by the C2S server-side code.

pfefferle:
Ah, I see! This is a race condition in the code. It uses a fake ID instead of the real one! Let me see how I can fix this.

pfefferle:
I generated a dummy ID, to ensure that an object always has an ID, but that was never replaced by the real one 🫣

Should be also fixed now.

steve-bate:
Latest testing....

The actor URIs still don't appear to have CORS headers.

```
curl -I -H 'Accept: application/json' https://wordpress.stevebate.dev/author/alice/
HTTP/1.1 200 OK
Date: Mon, 09 Feb 2026 08:50:50 GMT
Server: Apache/2.4.66 (Debian)
X-Powered-By: PHP/8.3.30
Link: <https://wordpress.stevebate.dev/wp-json/>; rel="https://api.w.org/"
Link: <https://wordpress.stevebate.dev/wp-json/wp/v2/users/2>; rel="alternate"; title="JSON"; type="application/json"
Content-Type: text/html; charset=UTF-8
```

Same for ``https://wordpress.stevebate.dev/?author=2`

I'm having some issues posting to the inbox. I have a pleroma instance that I'm using to post to the wordpress instance. I'm seeing 401 errors from wordpress. This is the related log message. 

`wordpress-1  | 192.168.1.37 - - [09/Feb/2026:08:52:25 +0000] "POST /wp-json/activitypub/1.0/actors/1/inbox HTTP/1.1" 401 818 "-" "Pleroma 2.5.5; https://pleroma.stevebate.dev"`

I didn't find any other logged information related to the issue. I was seeing similar behavior from onepage.pub earlier and I thought maybe there was a bug in the HTTP sig. However, now that it's happening with Pleroma too, it seems like it's not that.

Yesterday, I was able to send a message from one WP user to another by posting to the outbox. That seems to be failing today.

pfefferle:
Are you using the latest commit? I am able to query both the pretty permalink and the `?author={id}` JSON with CORS headers.

I can see the content type is `text/html; charset=UTF-8` in your request!? Does the user you test has the ActivityPub capability? You can check in `/wp-admin/users.php`. You should also check if the blog is set up to even support users: `/wp-admin/options-general.php?page=activitypub&tab=settings` should use either "Author Profiles Only" or "Both author and blog profiles" as "Enable profiles by type".

steve-bate:
> Are you using the latest commit? I am able to query both the pretty permalink and the `?author={id}` JSON with CORS headers.

Commit `138316b366af8e7e822d3424a14af2579b3854c2`
I did a hard reset to the upstream HEAD and I seem to be in sync.

> I can see the content type is `text/html; charset=UTF-8` in your request!? 

That's the WP response content type. However, the response content is actually the text of the actor JSON profile so that content type seems incorrect.

> Does the user you test has the ActivityPub capability? You can check in `/wp-admin/users.php`. 

image

> You should also check if the blog is set up to even support users: `/wp-admin/options-general.php?page=activitypub&tab=settings` should use either "Author Profiles Only" or "Both author and blog profiles" as "Enable profiles by type".

image

pfefferle:
@steve-bate do you need the cors headers for head requests? I think that is the issue with:

`curl -I -H 'Accept: application/json' https://wordpress.stevebate.dev/author/alice/`!?

pfefferle:
@steve-bate have fix it anyways. cors should now be properly set!

steve-bate:
> @steve-bate do you need the cors headers for head requests? I think that is the issue with:
> 
> `curl -I -H 'Accept: application/json' https://wordpress.stevebate.dev/author/alice/`!?

I may have used that as an example, but most of my curl fetches were GET requests. I'll try out the latest changes.

Edit: The `-H` is for adding headers, by the way, the `--head` option does a HEAD request. The request you quoted would have done a GET request. (The `-I` option cause only the headers to be displayed in the response).

steve-bate:
It looks like the CORS issue for the authors is fixed now. 👍 

Is the Wordpress C2S only for posting to the outbox and for inbox access? For example, would you expect an AP reply to a WP post to be acceptable? I was seeing 500 server errors when sending a Note inReplyTo an Article. This was the debug logging...
```
[12-Feb-2026 14:46:57 UTC] [OUTBOX] Failed to add -Activity from: 1 (Error: Invalid Identifier) with Activity: Array
(
    [@context] => https://www.w3.org/ns/activitystreams
    [type] => Create
    [actor] => https://wordpress.stevebate.dev/?author=1
    [object] => Array
        (
            [type] => Note
            [content] => C2S Reply
            [to] => Array
                (
                )

            [attributedTo] => https://wordpress.stevebate.dev/?author=1
            [inReplyTo] => https://wordpress.stevebate.dev/?p=40
            [published] => 2026-02-12T14:46:57Z
        )

    [to] => Array
        (
        )

)
```

The `https://wordpress.stevebate.dev/?p=40` URI was the one in my outbox for a WP post I had created.

pfefferle:
> Edit: The -H is for adding headers, by the way, the --head option does a HEAD request. The request you quoted would have done a GET request. (The -I option cause only the headers to be displayed in the response).

According to the manpage, both send a `HEAD` request in the end:

> -I, --head
>
> (HTTP FTP FILE) Fetch the headers only. HTTP-servers feature the command HEAD which this uses to get nothing but the header of a document. When used on an FTP or FILE URL, curl displays the file size and last modification time only.
>
> Providing [--head](https://curl.se/docs/manpage.html#--head) multiple times has no extra effect. Disable it again with --no-head.
>
> Example:
>
> curl -I https://example.com

That was what confused me too :)

steve-bate:
Ah, I didn't know that -I does a HEAD request. TIL. :-)

pfefferle:
@steve-bate comments should work now too!?

pfefferle:
@steve-bate I think I now fixed most of the issues! I love your toolkit! This helped a lot to understand how a JS-Client behave differently. The `OPTIONS` preflight was a hard one!!!

ThisIsMissEm:
@pfefferle I don't have time to full review, but two questions:
- are you using OAuth authorization server metadata discovery (RFC 8414) and OAuth Protected Resource Metadata (more recently published, allows delegating to an authorization server rather than ever server being an AS)
- did you look at DPoP?

This profile is pretty similar to the [AT Protocol profile](https://atproto.com/specs/oauth), which imo, is really well thought out for decentralized social applications.

You may also want to adopt the new https://datatracker.ietf.org/doc/draft-ietf-oauth-refresh-token-expiration/

which signals not just the expiry of the access token, but also the refresh token and the authorization grant as a whole.

ThisIsMissEm:
Oh! And I'd recommend, as we have done in CIMD, to prefer asymmetrical client authentication mechanisms; DCR obviously is symmetrical, so I'd probably discourage its usage, as it's hard to manage at scale

Edit: Also, unsubscribing from the thread, @mention me if you need me

mediaformat:
@pfefferle I'm testing and getting CORS errors on both User and Blog actor profiles even when requesting with Authorization.

includes/oauth/class-server.php
```diff
@@ -0,0 +1,519 @@
...
+		}
+
+		/*
+		 * Reflect the request Origin instead of using a wildcard to avoid
+		 * leaking private data to arbitrary origins on authenticated endpoints.
+		 */
+		$origin = isset( $_SERVER['HTTP_ORIGIN'] ) ? \esc_url_raw( \wp_unslash( $_SERVER['HTTP_ORIGIN'] ) ) : '';
+		$response->header( 'Access-Control-Allow-Origin', $origin ? $origin : '*' );
+		$response->header( 'Access-Control-Allow-Methods', 'GET, POST, OPTIONS' );
+		$response->header( 'Access-Control-Allow-Headers', 'Content-Type, Authorization' );
```
mediaformat: I think we'll need `Accept` here as well.

includes/oauth/class-server.php
```diff
@@ -0,0 +1,519 @@
...
+			return $namespace . '/oauth/authorize' !== $route;
+		}
+
+		// Proxy endpoint for fetching remote objects.
+		if ( $namespace . '/proxy' === $route ) {
+			return true;
+		}
+
+		// C2S outbox and inbox endpoints.
+		if ( \str_ends_with( $route, '/outbox' ) || \str_ends_with( $route, '/inbox' ) ) {
```
mediaformat: If user is authenticated they must be allowed access to the actor as well as all their collections (inbox, outbox, following, followers), and endpoints.

pfefferle:
> I'm testing and getting CORS errors on both User and Blog actor profiles even when requesting with Authorization.

@mediaformat I can not reproduce this issue. Can you give me an example URL? Does the User have the ActivityPub capability?

mediaformat:
Am able to reproduce with the c2s-toolkit

mediaformat:
@pfefferle I see, it works if I fetch the author URL `https://example.com/author/fulano`, but not if I fetch the author URI `https://example.com/?author=1` likely because there is a redirect!

pfefferle:
You have to add the accept header then there should be no redirect

mediaformat:
@pfefferle narrowing it down further: when its just the accept header it works, when accept AND the authorization header are present it stops working.
```
const data = await fetch(
  uri,
  {
    method: 'GET',
    headers: {
      'Authorization': 'Bearer ' + getAccessToken(),
      'Accept': 'application/activity+json',
    }
  }
);
```

pfefferle:
@mediaformat Do you use the latest version of this branch?

I still can't reproduce it, even with Auth headers!?

ThisIsMissEm:
@pfefferle you may also want to use Rich Authorization Requests for richer control over which resources a client can interact with: e.g., what collections (inbox, outbox, followers, likes, etc) the client can do things to. This is something AT Protocol missed the ball on, and they did this like, dynamic scopes thing which has a whole bunch of parsing logic around a space separated set of strings.

https://datatracker.ietf.org/doc/html/rfc9396

includes/rest/class-proxy-controller.php
```diff
@@ -0,0 +1,153 @@
...
+	/**
+	 * Register routes.
+	 */
+	public function register_routes() {
+		\register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			array(
+				array(
+					'methods'             => \WP_REST_Server::READABLE,
```
mediaformat: This should be a POST, so I think `\WP_REST_Server::CREATABLE`
https://www.w3.org/wiki/ActivityPub/Primer/proxyUrl_endpoint

includes/rest/class-proxy-controller.php
```diff
@@ -0,0 +1,153 @@
...
+					'callback'            => array( $this, 'get_item' ),
+					'permission_callback' => array( $this, 'verify_authentication' ),
+					'args'                => array(
+						'id' => array(
+							'description'       => 'The URI of the remote ActivityPub object to fetch.',
+							'type'              => 'string',
+							'required'          => true,
+							'sanitize_callback' => 'sanitize_url',
```
mediaformat: The `id` will be urlencoded, so I think the sanitize_url will not work directly

mediaformat:
> I still can't reproduce it, even with Auth headers!?

I'm having intermittent issues, even different results depending on Apache or Nginx. For now things are working.

pfefferle:
@mediaformat do you use nginx as main webserver or (reverse) proxy? maybe the headers are not forwarded properly?

includes/collection/class-outbox.php
```diff
@@ -368,6 +399,16 @@ public static function maybe_get_activity( $outbox_item ) {
 			return new \WP_Error( 'invalid_outbox_item', 'Invalid Outbox item.' );
 		}
 
+		// Authenticate via Bearer token for non-REST requests (e.g. permalink access).
+		if ( ! \is_user_logged_in() && ! \wp_is_serving_rest_request() ) {
```
jeherve: Do we need to limit this a bit more, maybe by checking for a role of at least author?

includes/collection/class-posts.php
```diff
@@ -7,613 +7,182 @@
...
-use function Activitypub\object_to_uri;
-use function Activitypub\process_remote_media;
+use function Activitypub\get_content_visibility;
 
 /**
  * Posts collection.
  *
- * Provides methods to retrieve, create, update, and manage ActivityPub posts (articles, notes, media, etc.).
+ * Provides CRUD methods for local WordPress posts created
+ * via ActivityPub (C2S outbox).
```
jeherve: This threw me off for a few seconds, then I saw that it's basically a rename / move.

includes/rest/class-webfinger-controller.php
```diff
@@ -85,8 +85,7 @@ public function get_item( $request ) {
 			$response,
 			$code,
 			array(
-				'Access-Control-Allow-Origin' => '*',
```
jeherve, line 88: Were we adding those headers twice before, once here and once in `add_cors_headers`? Nice catch!

includes/wp-admin/class-admin.php
```diff
@@ -71,13 +74,19 @@ public static function init() {
...
 
 		\add_action( 'wp_dashboard_setup', array( self::class, 'add_dashboard_widgets' ) );
 
 		\add_action( 'wp_ajax_activitypub_moderation_settings', array( self::class, 'ajax_moderation_settings' ) );
 		\add_action( 'wp_ajax_activitypub_blocklist_subscription', array( self::class, 'ajax_blocklist_subscription' ) );
+		\add_action( 'wp_ajax_activitypub_register_oauth_client', array( self::class, 'ajax_register_oauth_client' ) );
+		\add_action( 'wp_ajax_activitypub_delete_oauth_client', array( self::class, 'ajax_delete_oauth_client' ) );
+		\add_action( 'wp_ajax_activitypub_delete_all_oauth_clients', array( self::class, 'ajax_delete_all_oauth_clients' ) );
+		\add_action( 'wp_ajax_activitypub_revoke_oauth_token', array( self::class, 'ajax_revoke_oauth_token' ) );
+		\add_action( 'wp_ajax_activitypub_revoke_all_oauth_tokens', array( self::class, 'ajax_revoke_all_oauth_tokens' ) );
```
jeherve, line 91: In `ajax_moderation_settings` and `ajax_blocklist_subscription` we have a `current_user_can` check, but not for these new ajax handlers. Can we add one?

templates/oauth-authorize.php
```diff
@@ -0,0 +1,129 @@
...
+<form method="post" action="<?php echo esc_url( $form_url ); ?>">
+	<div class="activitypub-oauth-client">
+		<p>
+			<strong>
+			<?php
+			echo wp_kses(
+				sprintf(
+					/* translators: %s: Client name or ID */
+					__( '%s wants to access your account.', 'activitypub' ),
+					'<a href="' . esc_url( $client_id ) . '">' . esc_html( $client_name ?: $client_id ) . '</a>'
```
mediaformat: nit: `client_id` is a url for CIMD apps, but for apps registered with DCR the `client_id` will just be a string. The link here could point to the DCR apps redirect_uris domain...

jeherve: Should this link open in a new tab, to avoid breaking the flow?

includes/functions.php
```diff
@@ -316,3 +337,30 @@ function enrich_content_data( $content, $regex, $regex_callback ) {
...
+	 *
+	 * Sites behind a reverse proxy should hook here to return the real
+	 * client IP from a trusted header (e.g. X-Forwarded-For, CF-Connecting-IP).
+	 *
+	 * @since unreleased
+	 *
+	 * @param string $ip The client IP address from REMOTE_ADDR.
+	 */
+	return \apply_filters( 'activitypub_client_ip', $ip );
+}
```
jeherve, line 393: The filter is a nice way to work around problems on sites with weird setups. I think you could also do something like what we do in Jetpack, where you support an additional set of headers:

https://github.com/Automattic/jetpack/blob/ea5dcc14d22821776d0bf7dfe22c9af73a4a4d19/projects/packages/status/src/class-visitor.php#L22-L42

jeherve (commented):
I won't claim to have thoroughly reviewed all of this PR 😅 But it looks good as it is, and I only have minor comments, below.

includes/collection/class-posts.php
pfefferle: Yes, it made more sense this way and is consistent with the Remote_User pattern then.

pfefferle: Added a comment!

includes/functions.php
pfefferle: Added :)

mediaformat:
Current testing status

- [x] POST a Note to the outbox and verify a WordPress post is created with status format and block markup content.
- [x] POST an Article with name field and verify title and excerpt are set.
- [x] Verify hashtags in content are saved as WordPress tags.
- [x] GET the inbox and verify activities are returned.
- [x] Test Follow
- [x] Test Unfollow via outbox POST.
- [x] Token refresh and revocation.
- [x] Like
- [x] Announce
- [x] Delete removes the corresponding post.
- [x] Go to Profile → Connected Applications and verify you can register a new client and revoke tokens.

Aiming to finish testing by the end of week.

includes/oauth/class-client.php
```diff
@@ -716,13 +716,21 @@ public static function generate_client_secret() {
 	 * @return bool True if valid.
 	 */
 	private static function validate_uri_format( $uri ) {
-		$parsed = \wp_parse_url( $uri );
-
-		if ( ! $parsed || empty( $parsed['scheme'] ) ) {
+		/*
+		 * Extract scheme manually first because wp_parse_url() returns false
+		 * for URIs like "myapp://" (scheme + empty authority, no path).
```
ThisIsMissEm: Small but fun spec note: custom schemes are actually in the format of: `myapp:/path` not `myapp://path`

Yeah, blew my mind too when I first learned of this from Matthieu on the Bluesky team.

That's according to the specification: https://www.rfc-editor.org/rfc/rfc2396#section-3

So `myapp://path/to/thing` isn't actually valid according to the specification since it's missing an authority part. (`path` isn't a valid authority)

includes/oauth/class-client.php
```diff
@@ -716,13 +716,21 @@ public static function generate_client_secret() {
...
+		$scheme = \strtolower( $matches[1] );
+		$parsed = \wp_parse_url( $uri );
+
+		if ( ! $parsed ) {
+			// wp_parse_url fails for "scheme://" — still valid for custom schemes.
+			$parsed = array( 'scheme' => $scheme );
+		}
 
 		// Block dangerous schemes.
 		$blocked_schemes = array( 'javascript', 'data', 'vbscript', 'blob', 'file' );
```
ThisIsMissEm: This isn't necessarily all the unsafe URL schemes.. OWASP has some information here: https://owasp.org/www-community/attacks/xss/

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+				'activitypub_client_id_mismatch',
+				\__( 'Client ID in metadata does not match request.', 'activitypub' ),
+				array( 'status' => 400 )
+			);
+		}
+
+		// Get redirect URIs from metadata or derive from client_id origin.
+		$redirect_uris = array();
+		if ( ! empty( $metadata['redirect_uris'] ) && is_array( $metadata['redirect_uris'] ) ) {
+			$redirect_uris = $metadata['redirect_uris'];
```
ThisIsMissEm, line 229: You're possibly missing validation here

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+	 * @param string $url The client ID URL to fetch.
+	 * @return array|\WP_Error Metadata array or error.
+	 */
+	private static function fetch_client_metadata( $url ) {
+		$response = \wp_safe_remote_get(
+			$url,
+			array(
+				'timeout'     => 10,
+				'headers'     => array(
+					'Accept' => 'application/json, application/ld+json, application/activity+json',
```
ThisIsMissEm: The explicit content type for CIMDs will be `application/cimd+json` though any JSON response should be accepted

ThisIsMissEm: I'd suggest `Accept: application/cimd+json, application/json`

But note: this only was proposed as a change to the internet draft within the last week or two.

includes/oauth/class-client.php
pfefferle: Oh thanks! Updated!

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+		if ( ! is_array( $data ) ) {
+			return new \WP_Error(
+				'activitypub_client_invalid_metadata',
+				\__( 'Invalid client metadata format.', 'activitypub' ),
+				array( 'status' => 400 )
+			);
+		}
+
+		// Normalize ActivityPub Application format to CIMD format.
+		return self::normalize_client_metadata( $data, $url );
```
ThisIsMissEm: If the JSON document does not contain a `client_id` property, throw an error, don't try to fill it in based on Request URI, as this allows for client impersonation through redirects

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+	 */
+	private static function fetch_client_metadata( $url ) {
+		$response = \wp_safe_remote_get(
+			$url,
+			array(
+				'timeout'     => 10,
+				'headers'     => array(
+					'Accept' => 'application/json, application/ld+json, application/activity+json',
+				),
+				'redirection' => 3,
```
ThisIsMissEm: CIMDs explicit prohibit following tedirects, since this allows for client impersonation attacks.

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+		// Normalize ActivityPub Application format to CIMD format.
+		return self::normalize_client_metadata( $data, $url );
+	}
+
+	/**
+	 * Normalize client metadata from various formats to standard format.
+	 *
+	 * Supports:
+	 * - CIMD (Client ID Metadata Document)
+	 * - ActivityPub Application objects
```
ThisIsMissEm, line 320: I would strongly discourage inferring an OAuth Client from an AS2 Application Object.  They just don't give the same information as a CIMD, and this can lead to security issues (for instance mixup or impersonation attacks)

As I've told Evan countless times, just use CIMD, it's an internet draft by the OAuth Working Group, in other words, it gets a tonne of security review from people across the IETF, but especially people with expert knowledge in OAuth

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+	 *
+	 * @param array  $data The raw metadata.
+	 * @param string $url  The client ID URL.
+	 * @return array Normalized metadata.
+	 */
+	private static function normalize_client_metadata( $data, $url ) {
```
ThisIsMissEm: This code may contain vulnerabilities from the AS2 Application Object to CIMD conversion which actually creates vulnerabilities with CIMDs in your implementation.

If you really want to do this conversion, check if `$data` contains a `client_Id` property, and if so, bail.

Your validation logic will (or should) catch things like missing properties that are required (e.g., redirect_uris)

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+			return false;
+		}
+
+		// Public clients don't need secret validation.
+		if ( $client->is_public() ) {
+			return true;
+		}
+
+		// Confidential clients require a valid secret.
+		if ( empty( $client_secret ) ) {
```
ThisIsMissEm, line 409: This is only the case for Dynamic Client Registration and pre-registered clients which allow for symmetrical client authentication methods (e.g., `client_secret_post`, `client_secret_basic`)

For CIMDs, we explicitly require non-symmetrical client authentication methods (e.g., `private_key_jwt` or Client Assertions (I forget the URN for these))

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+	 * Fetch client metadata from URL.
+	 *
+	 * Supports both CIMD JSON format and ActivityPub Application objects.
+	 *
+	 * @param string $url The client ID URL to fetch.
+	 * @return array|\WP_Error Metadata array or error.
+	 */
+	private static function fetch_client_metadata( $url ) {
+		$response = \wp_safe_remote_get(
+			$url,
```
ThisIsMissEm, line 272: Make sure you validate the `$url` is a public URL, i.e., not on .local, .internal, etc and that the host name resolves to a public IP address (e.g., not private use or reserved)

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+			return false;
+		}
+
+		/*
+		 * For auto-discovered clients without redirect_uris, use same-origin policy.
+		 * The redirect_uri must be on the same host as the client_id.
+		 * Custom URI schemes (no host) are not allowed in same-origin fallback
+		 * and must be explicitly registered.
+		 */
+		$redirect_host = \wp_parse_url( $redirect_uri, PHP_URL_HOST );
```
ThisIsMissEm: Don't do this, require redirect URIs

ThisIsMissEm: Further, as implemented, this allows you to be used as an open redirector; See the OAuth Security Best Current Practices document from the IETF as to why this is bad.

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+	 * @param string $allowed_uri  The registered redirect URI.
+	 * @param string $redirect_uri The requested redirect URI.
+	 * @return bool True if they match under loopback rules.
+	 */
+	private static function is_loopback_redirect_match( $allowed_uri, $redirect_uri ) {
+		$allowed_parts  = \wp_parse_url( $allowed_uri );
+		$redirect_parts = \wp_parse_url( $redirect_uri );
+
+		// Must have same scheme.
+		if ( ( $allowed_parts['scheme'] ?? '' ) !== ( $redirect_parts['scheme'] ?? '' ) ) {
```
ThisIsMissEm, line 470: This allows both to omit a scheme.

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+			return false;
+		}
+
+		$allowed_host  = $allowed_parts['host'] ?? '';
+		$redirect_host = $redirect_parts['host'] ?? '';
+
+		// Must have same host.
+		if ( $allowed_host !== $redirect_host ) {
```
ThisIsMissEm, line 478: This allows omitting the hostname.

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+	/**
+	 * Check if two URIs match under RFC 8252 loopback rules.
+	 *
+	 * For loopback addresses (127.0.0.1, localhost), the port is ignored.
+	 *
+	 * @param string $allowed_uri  The registered redirect URI.
+	 * @param string $redirect_uri The requested redirect URI.
+	 * @return bool True if they match under loopback rules.
+	 */
+	private static function is_loopback_redirect_match( $allowed_uri, $redirect_uri ) {
```
ThisIsMissEm, line 465: As written this function has several failure modes which could allow exploits.

You'd also allowing redirects to things with unknown query parameters here, which whilst may be technically allowed, could lead to security issues on localhost

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+	}
+
+	/**
+	 * Get allowed scopes for this client.
+	 *
+	 * @return array The allowed scopes.
+	 */
+	public function get_allowed_scopes() {
+		$scopes = \get_post_meta( $this->post_id, '_activitypub_allowed_scopes', true );
+		return is_array( $scopes ) ? $scopes : Scope::ALL;
```
ThisIsMissEm: Default to no scopes, this currently fails open on access control

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+	 * Get client URI (homepage).
+	 *
+	 * @return string The client URI or empty string.
+	 */
+	public function get_client_uri() {
+		return \get_post_meta( $this->post_id, '_activitypub_client_uri', true ) ?: '';
+	}
+
+	/**
+	 * Get a URL suitable for linking to this client.
```
ThisIsMissEm, line 660: Use client_uri, if it's not present, don't offer a link. The client_id is not really a URI intended for end-users (since it's a JSON response typically, unless content negotiated otherwise)

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+
+		// Block dangerous schemes.
+		$blocked_schemes = array( 'javascript', 'data', 'vbscript', 'blob', 'file' );
+		if ( in_array( $scheme, $blocked_schemes, true ) ) {
+			return false;
+		}
+
+		// Allow http for localhost development.
+		if ( 'http' === $scheme ) {
+			if ( empty( $parsed['host'] ) ) {
```
ThisIsMissEm, line 794: This will allow HTTP for all clients, not just loopback or link-local. You probably don't want that.

includes/oauth/class-client.php
```diff
@@ -0,0 +1,818 @@
...
+				return false;
+			}
+
+			/*
+			 * Include both bracketed and unbracketed IPv6 loopback since parse_url
+			 * may return either format depending on PHP version.
+			 */
+			$localhost_hosts = array( 'localhost', '127.0.0.1', '[::1]', '::1' );
```
ThisIsMissEm: I think there's more forms than these that can point to local resources

includes/oauth/class-client.php
```diff
@@ -0,0 +1,839 @@
...
+				'numberposts' => -1,
+			)
+		);
+
+		foreach ( $post_ids as $post_id ) {
+			\wp_delete_post( $post_id, true );
+		}
+
+		// Also revoke all tokens stored in user meta.
+		Token::revoke_all();
```
ThisIsMissEm, line 841: Probably also revoke the AuthorizationCode's too, even though they should fail to exchange correctly.

includes/oauth/class-scope.php
```diff
@@ -0,0 +1,188 @@
...
+		self::PUSH    => 'Subscribe to real-time event streams',
+		self::PROFILE => 'Edit actor profile',
+	);
+
+	/**
+	 * Default scopes when none are requested.
+	 *
+	 * @var array
+	 */
+	const DEFAULT_SCOPES = array(
```
ThisIsMissEm, line 75: Mastodon actually defaults to the profile scope now, as to prevent access to data without an appropriate scope; this was mainly a UI change, since the API required scopes

includes/oauth/class-server.php
```diff
@@ -0,0 +1,507 @@
...
+			'issuer'                                => \home_url(),
+			'authorization_endpoint'                => $base_url . 'oauth/authorize',
+			'token_endpoint'                        => $base_url . 'oauth/token',
+			'revocation_endpoint'                   => $base_url . 'oauth/revoke',
+			'introspection_endpoint'                => $base_url . 'oauth/introspect',
+			'registration_endpoint'                 => $base_url . 'oauth/clients',
+			'scopes_supported'                      => Scope::ALL,
+			'response_types_supported'              => array( 'code' ),
+			'response_modes_supported'              => array( 'query' ),
+			'grant_types_supported'                 => array( 'authorization_code', 'refresh_token', 'password' ),
```
ThisIsMissEm: Password is deprecated in OAuth 2.1, and has been known to be insecure for a very long time

includes/oauth/class-server.php
```diff
@@ -0,0 +1,507 @@
...
+			'token_endpoint_auth_methods_supported' => array( 'none', 'client_secret_post' ),
```
ThisIsMissEm: There's also `client_secret_basic` but both of these should only be acceptable if the client has a client_secret (i.e., not valid for CIMDs)

includes/oauth/class-server.php
```diff
@@ -0,0 +1,507 @@
...
+			'introspection_endpoint_auth_methods_supported' => array( 'bearer' ),
+			'code_challenge_methods_supported'      => array( 'S256', 'plain' ),
```
ThisIsMissEm: No, don't support PKCE plain, it provides zero security

includes/oauth/class-server.php
```diff
@@ -0,0 +1,507 @@
...
+	 * before this method is called.
+	 *
+	 * @param string $redirect_uri The client's redirect URI.
+	 * @param array  $params       Query parameters to append.
+	 */
+	private static function redirect_to_client( $redirect_uri, $params ) {
+		$url = Sanitize::redirect_uri( \add_query_arg( $params, $redirect_uri ) );
+
+		\nocache_headers();
+		header( 'Location: ' . $url, true, 302 );
```
ThisIsMissEm: You may want to use another status code, see the note here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/302

includes/oauth/class-server.php
```diff
@@ -0,0 +1,495 @@
...
+			'registration_endpoint'                 => $base_url . 'oauth/clients',
+			'scopes_supported'                      => Scope::ALL,
+			'response_types_supported'              => array( 'code' ),
+			'response_modes_supported'              => array( 'query' ),
+			'grant_types_supported'                 => array( 'authorization_code', 'refresh_token', 'password' ),
+			'token_endpoint_auth_methods_supported' => array( 'none', 'client_secret_post', 'client_secret_basic' ),
+			'introspection_endpoint_auth_methods_supported' => array( 'bearer' ),
+			'code_challenge_methods_supported'      => array( 'S256' ),
+			'service_documentation'                 => 'https://github.com/swicg/activitypub-api',
+		);
```
mediaformat, line 260: Apparently we can add a signal for clients supporting CIMD:
`'client_id_metadata_document_supported' => true,`

mediaformat:
> * POST an Article with name field and verify title and excerpt are set.

This did set the Title, but no Excerpt was set in the post Editor
Screenshot 2026-03-08 at 12 57 23 PM

includes/oauth/class-server.php
ThisIsMissEm: not "can" but "MUST" — https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-01.html#name-authorization-server-metada

It's marked as "optional" because all properties in the [Authorization Server Metadata](https://datatracker.ietf.org/doc/html/rfc8414#section-2) are optional, afaik.

pfefferle:
@mediaformat Have you added a summary? Or what do you expect?

mediaformat:
@pfefferle no I didn't add any summary or anything specia, I wasn't expecting anything specific, I may just have interpreted the "excerpt" part too literally. 

Adding type Article makes the post format Standard.

pfefferle:
Yes and the excerpt is generated dynamically on federation! This way the plugin can see if it is a specifically set summary or if it has to be auto generated.

mediaformat:
During OAuth user authorization, when the user is not already logged in, the post login redirect brings the user to the WP dashboard instead of redirecting to the redirect_uris

pfefferle:
I can not reproduce it on neither of my tests systems. Everything works as expected.

Can you share the URL with me? Is it a custom protocol? Have you tested it with localhost?

mediaformat:
> I can not reproduce it on neither of my tests systems. Everything works as expected.

I was testing with a tunnel and it didn't work, but on a hosted install it works fine!

PR #2858, opened 2026-02-02
Revert to synchronous outbox processing with performance improvements

pfefferle:
Fixes #2761

## Proposed changes:

* Cache WebFinger failures for 5 minutes to prevent repeated timeout waits for unreachable servers.
* Reduce HTTP GET timeout from 100s to 15s (still filterable via `activitypub_remote_get_timeout`).
* Add filterable timeout for HTTP POST requests via new `activitypub_remote_post_timeout` filter (30s default).
* Remove async scheduling from Post scheduler - `triage()` and `transition_attachment_status()` now call `add_to_outbox()` directly.
* Remove the `activitypub_add_to_outbox` async handler that was no longer needed.

### Other information:

## Testing instructions:

1. Create a post with no @mentions - should save quickly.
2. Create a post with an @mention to a cached/known actor - should save quickly.
3. Create a post with an @mention to an unreachable server (e.g., `@test@nonexistent.invalid`) - should timeout in ~15s instead of 100s.
4. Re-save the same post with the unreachable mention - should save quickly now (failure is cached for 5 minutes).
5. Verify federation still works correctly by creating a post and checking it appears in the outbox.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Revert to synchronous outbox processing with improved timeout handling and WebFinger error caching.

</details>


includes/class-http.php
```diff
@@ -88,20 +96,24 @@ public static function post( $url, $body, $user_id ) {
...
 	 * @param string   $url    The URL endpoint.
 	 * @param bool|int $cached Optional. Whether the result should be cached, or its duration. Default false.
+	 * @param array    $args   Optional. Additional arguments to customize the request.
+	 *                         - 'headers': Array of headers to override defaults.
+	 *                         - 'signed': Whether to sign the request. Default true.
 	 *
 	 * @return array|\WP_Error The GET Response or a WP_Error.
 	 */
-	public static function get( $url, $cached = false ) {
+	public static function get( $url, $cached = false, $args = array() ) {
```
obenland:
```suggestion
	public static function get( $url, $args = array(), $cached = false ) {
```
`wp_remote_*()` functions [accept a url and args](https://github.com/WordPress/wordpress-develop/blob/6.9/src/wp-includes/http.php#L175-L178), I think it would be worth it to make this API do the same, even with the pre-existing cached arg. For back compat you could check whether $args is a boolean, throw a deprecated argument notice, and update $cached with the $args value.

includes/class-http.php
```diff
@@ -130,29 +142,53 @@ public static function get( $url, $cached = false ) {
...
 
+		$args            = \wp_parse_args( $args, $defaults );
+		$args['headers'] = \wp_parse_args( $args['headers'], $defaults['headers'] );
+
+		// Add HTTP signatures if requested.
+		if ( $args['signed'] ) {
+			$args['key_id']      = Actors::get_by_id( Actors::APPLICATION_USER_ID )->get_id() . '#main-key';
+			$args['private_key'] = Actors::get_private_key( Actors::APPLICATION_USER_ID );
+		}
+		unset( $args['signed'] );
```
obenland: Could we not just sign all our outgoing requests?

includes/class-http.php
```diff
@@ -130,29 +142,53 @@ public static function get( $url, $cached = false ) {
...
-		if ( $code >= 400 ) {
+		if ( \is_wp_error( $response ) || $code >= 400 ) {
 			$response = new \WP_Error( $code, __( 'Failed HTTP Request', 'activitypub' ), array( 'status' => $code ) );
+
+			/*
+			 * Always cache errors to prevent repeated timeout waits.
+			 * - Timeouts and 5xx errors: 1 minute (server may recover quickly).
+			 * - 4xx errors: 15 minutes (client errors are more permanent).
+			 */
+			if ( $code >= 500 || 0 === $code ) {
```
obenland: This could be a good place to reuse our list of retriable http codes.

includes/class-dispatcher.php
```diff
@@ -104,7 +104,7 @@ public static function get_retry_error_codes() {
 		 *
 		 * @param int[] $retry_error_codes The error codes. Default array( 408, 429, 500, 502, 503, 504 ).
 		 */
-		return apply_filters( 'activitypub_dispatcher_retry_error_codes', array( 408, 429, 500, 502, 503, 504 ) );
+		return apply_filters( 'activitypub_dispatcher_retry_error_codes', ACTIVITYPUB_RETRY_ERROR_CODES );
```
obenland, line 107: What's the benefit of saving them in a constant? They're already encapsulated in a single place, in a getter function that's reusable.

pfefferle: But the dispatcher is technically not the right domain for it, I would say!?!

PR #2859, opened 2026-02-02
Enable RFC 9421 signature by default

obenland:
Let's enable it by default. The option has been pretty hidden and we've not received any feedback on the feature so far.
It's been working on activitypub.blog (while it was on), so let's default it to true and see what breaks.

## Proposed changes:
* Enable RFC 9421 (HTTP Message Signatures) by default instead of requiring opt-in.

### Other information:

The existing test coverage applies — this only changes the default value.

## Testing instructions:

* On a fresh install or with the option unset, verify that RFC 9421 signatures are used by default.
* Confirm the toggle in Settings → ActivityPub → Advanced still works to disable RFC 9421 if needed.
* Federation should continue working due to the automatic fallback to Draft Cavage on 4xx responses.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Enable RFC 9421 HTTP Message Signatures by default.

</details>


pfefferle:
I pushed a fix for the failing test, hope you don't mind.

`test_verify_http_signature_returns_verified_key_id` signs through `http_request_args`, so with the new default it signs RFC 9421 and covers `@method`, `@target-uri` and `@authority`. Signing takes those from the URL it is given, verification takes them from `$_SERVER`, and the test hand-builds its request without setting either. A real delivery has both describing the same request, so this only shows up in tests. The draft scheme did not care because its `(request-target)` is rebuilt from the request object.

It needs all four of `REQUEST_METHOD`, `HTTP_HOST`, `REQUEST_URI` and `HTTPS`. I tried it with three and it still failed, the method is the one that is easy to miss.

I set those rather than pinning the option, so the test passes whichever scheme is the default. I checked it both ways, forcing `activitypub_rfc9421_signature` to `1` and to `0`, and it passes for both. Otherwise the assertion stops covering the scheme this PR is switching on.

I left `test_verify_http_signature_with_digest` alone. Setting the option there is not a workaround, it asserts on `Digest` and `digest_mismatch`, which only exist in the draft scheme, so it is picking what it tests. One small thing though: it calls `delete_option()` at the end of the test body, so if an assertion above it fails the option stays set for everything after.

The branch is still behind trunk, which has moved a fair bit since August, so it is worth a rebase before this goes in.

pfefferle (approved):

PR #2860, opened 2026-02-02
Extend soft delete to draft, pending, and private post statuses

pfefferle:
## Proposed changes:

Extends the soft delete functionality (FEP-4f05) to also send Delete activities when a federated post's status is changed to "draft", "pending", or "private". Previously, changing a post to draft would send an Update activity with placeholder content "(This post is being modified)". Now it sends a proper Delete activity and returns a Tombstone, consistent with other soft delete scenarios.

### Changes
* Update scheduler to trigger Delete for draft/pending/private status changes (same as trash)
* Update `is_post_disabled()` to handle draft/pending/private statuses for federated posts
* Mark placeholder content in transformer as deprecated (kept for backwards compatibility)
* Update `invalidate_existing_items()` to invalidate pending Delete activities when Create/Update is added (prevents sending Delete after re-publish)
* Add tests for draft/pending/private soft delete behavior
* Fix handler tests to use published posts

### Soft Delete Behavior
When a federated post is changed to draft/pending/private:
1. Delete activity is sent to followers
2. Post state is set to "deleted"
3. URL is registered in tombstone registry
4. Accessing the ActivityPub URL returns 200 OK with Tombstone (soft delete)

### Re-publish Handling
When a soft-deleted post is re-published:
1. Any pending Delete activity is invalidated (won't be sent)
2. Create activity is sent instead
3. Tombstone is removed from registry

This follows the same pattern as visibility changes to local/private from #2824.

### Known limitations
* **Legacy posts + simultaneous slug change.** Posts that predate the `activitypub_last_post_with_permalink_as_id` migration use their permalink as the ActivityPub ID. If such a legacy post has its slug changed in the *same save* as a soft-delete transition (e.g. publish → draft with a new `post_name`), the Delete targets the new permalink while remote servers cached the original, so the remote copy is not torn down. The trash path mitigates this by caching the pre-transition URL in `_activitypub_canonical_url` via the `wp_trash_post` hook; draft/pending/private/password transitions do not. Narrow blast radius (legacy posts only, and only when slug and status change together) and documented in `Post::get_id()`. Workaround: on a pre-ID-migration site, avoid editing the slug in the same save as the visibility change.

### Other information:

### Soft Delete
1. Create and publish a post (it will be federated)
2. Wait for the post to be federated (check outbox or wait for cron)
3. Edit the post and change status to "Draft", "Pending Review", or "Private"
4. Verify that a Delete activity is created in the outbox
5. Verify the post's `activitypub_status` meta is set to "deleted"
6. Query the post's ActivityPub URL - should return 200 with Tombstone

### Re-publishing (before Delete is sent)
1. Create and publish a post
2. Quickly change to Draft (Delete activity scheduled but not sent)
3. Immediately re-publish the post
4. Verify the Delete activity is invalidated (status changed to publish)
5. Verify a Create activity is pending

### Re-publishing (after Delete is sent)
1. Take a soft-deleted post (status changed to Draft/Pending/Private)
2. Wait for Delete activity to be sent
3. Publish the post again
4. Verify a Create activity is created in the outbox
5. Verify the tombstone is removed

### Changelog entry

Changelog entry has been added manually.


pfefferle:
Blocked by: https://github.com/mastodon/mastodon/issues/37730

PR #2861, opened 2026-02-02
Add bot account support for blog and user profiles

kraftbj:
## Proposed changes:

Being a good Fediverse citizen means being transparent about automated accounts. Mastodon and other platforms display a "BOT" badge on profiles that identify as bots or automated accounts, helping users understand they're interacting with automated content.

This PR adds the ability to mark blog and user profiles as bot accounts:

* Adds a "Bot Account" checkbox to the Blog Profile settings (Settings → ActivityPub → Blog Profile)
* Adds a "Bot Account" checkbox to user profile settings (Users → Profile → ActivityPub section)
* When enabled, changes the Actor `type` from `Person`/`Group` to `Service`, which Fediverse platforms recognize as a bot
* Respects the existing relay mode setting (which already implies bot status)

This is useful for:
- Blogs that primarily syndicate/aggregate content automatically
- Author accounts that post via automation (IFTTT, Zapier, etc.)
- Any account that wants to be transparent about its automated nature

This branch is currently running on [kraft.beer](https://kraft.beer) for the `@beers@kraft.beer` account.

## Screenshots
Screenshot 2026-02-02 at 17 03 24

Screenshot 2026-02-02 at 17 02 54

### Other information:

## Testing instructions:

1. Go to **Settings → ActivityPub → Blog Profile**
2. Look for the "Bot Account" checkbox near the bottom of the settings
3. Enable it and save
4. Visit your blog's actor endpoint (e.g., `/?author=0`) and verify `"type": "Service"` appears in the JSON
5. Go to **Users → Profile** and scroll to the ActivityPub section
6. Enable the "Bot Account" checkbox for a user and save
7. Visit that user's actor endpoint and verify `"type": "Service"` appears
8. If relay mode is enabled, verify the blog profile shows a notice that bot status is already implied

### Changelog entry

<details>

#### Significance

- [x] Minor

#### Type

#### Message

Add bot account option for blog and user profiles. When enabled, the profile displays with a "BOT" badge on Mastodon and other Fediverse platforms.

</details>


pfefferle:
I generally like the idea but it might break things! I am not sure if this should be a user facing setting or if it should be done through filters!?

Setting the Blog-User to a `Service` instead of a `Group` will break the "[FEP-1b12: Group federation](https://codeberg.org/fediverse/fep/src/branch/main/fep/1b12/fep-1b12.md)" for example.

kraftbj:
Would it make sense for a single-user site that has a blog-user to be able to set the tag, but a multiuser site not be able to?

I'm also okay making this only a filter for now and leaving the UI for another PR/another day.

kraftbj:
Good point on the Group → Service concern — that would break FEP-1b12 group federation.

I pushed a commit that scopes the blog bot option to single-user mode only:

- **Single-user mode** (blog type is `Person`): bot checkbox is shown and `Person → Service` works safely
- **Multi-user mode** (blog type is `Group`): bot checkbox is hidden and the setting is ignored in `get_type()`, preserving FEP-1b12 compliance
- **User profiles**: unchanged — `Person → Service` is always safe
- **Mode switching**: if a site goes from single-user to multi-user, the setting stays in the database but is ignored in both UI and code. Switching back restores it.

One question — when you mentioned this might be better handled through filters rather than a user-facing setting, were you thinking about the whole bot feature (including user profiles), or specifically the blog profile? For user profiles, a simple checkbox feels like the right UX since there's no protocol side effect. Curious about your thinking there.

pfefferle:
What about using filters instead and add an example to the snippets: https://github.com/Automattic/wordpress-activitypub/tree/trunk/snippets???

kraftbj:
@pfefferle like this?

pfefferle (approved):

PR #2873, opened 2026-02-04
Hide admin REST endpoints from API index

pfefferle:
## Proposed changes:
* Add `show_in_index => false` to admin actor REST endpoints (unfollow, block, follow) so they don't appear in the `/wp-json/` REST API discovery index.

### Other information:

## Testing instructions:

* Visit `http://localhost:8888/wp-json/activitypub/1.0/`
* Verify that no `/admin/actors/` routes appear in the response
* The endpoints still function normally when called directly

### Changelog entry

Changelog entry already committed to branch.


obenland:
Nice

PR #2915, opened 2026-02-13
Fix visibility default for old posts to respect 'Do not federate' setting

### LINKED ISSUE #2618: Updated post (already published) published to activity feed
### Quick summary

Editing an old post (previously published - in this case ~5 years ago) causes it to be "published" again to activitypub followers. 

Not certain what the Activity Pub protocol says about edits to previously published posts - maybe this is because this post was not published to ActivityPub?

Either way I'd think the user expectation is that minor changes to an old post (which does not change its publish date) would not result in it being sent to existing feed subscribers / followers. 

### Steps to reproduce

Edit an old post / update some content in it. 

View your activity feed externally (in my case I keep a Mastodon account that follows my WordPress blog for testing). 

### Site owner impact

### Severity

Minor

_No response_

_No response_

pfefferle:
## Description

Fixes #2618

When editing posts older than 1 month that have never been federated, the visibility now defaults to 'local' (do not federate) instead of 'public'. This matches the intended behavior documented in the editor plugin and prevents unintentional federation of old posts.

## Problem

Previously, the JavaScript editor plugin attempted to set the default visibility to 'local' for old posts, but this was not enforced server-side when posts were saved. This meant that:
- Users could edit an old post (e.g., from 5 years ago)
- Even with the "Do not federate" setting visible in the sidebar
- The post would still be sent to ActivityPub followers upon saving
- This happened because the PHP-side visibility check had no concept of "old post defaults"

## Solution

Updated `get_content_visibility()` in `includes/functions-post.php` to apply the same default logic as the JavaScript editor plugin:
- Old posts (>30 days) without explicit visibility meta now default to 'local' instead of 'public'
- Posts that were already federated are not affected and remain public
- Explicit visibility settings are always respected
- The check uses `metadata_exists()` to distinguish between "not set" and "explicitly set to public"

## Testing

Added comprehensive tests in `tests/phpunit/tests/includes/class-test-functions-post.php`:
- Old posts without visibility meta default to 'local'
- New posts without visibility meta default to 'public'
- Explicit visibility settings are respected for old posts
- Federated old posts remain public

All tests pass.

## Test plan

1. Create or edit a post that is more than 1 month old
2. Check the visibility setting in the Fediverse sidebar panel
3. Save the post without changing the visibility
4. Verify the post is not sent to ActivityPub followers (check your Mastodon feed)
5. Edit the same post and explicitly set visibility to "Public"
6. Save and verify it IS sent to followers

## Changelog

- [x] Added to `.github/changelog/`

### DIFF
--- a/.github/changelog/2618-from-description
+++ b/.github/changelog/2618-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix visibility default for old posts to prevent unintentional federation when editing posts older than 1 month.
--- a/includes/functions-post.php
+++ b/includes/functions-post.php
@@ -283,6 +283,20 @@ function get_content_visibility( $post_id ) {
 
 	if ( in_array( $visibility, $options, true ) ) {
 		$_visibility = $visibility;
+	} elseif ( ! \metadata_exists( 'post', $post->ID, 'activitypub_content_visibility' ) ) {
+		// Apply default visibility logic for posts without explicit visibility.
+		// Posts older than 1 month default to 'local' (not federated).
+		$object_state = get_wp_object_state( $post );
+
+		// If post was already federated, keep it public.
+		if ( ACTIVITYPUB_OBJECT_STATE_FEDERATED !== $object_state ) {
+			$post_timestamp = \strtotime( $post->post_date_gmt );
+			$one_month_ago  = time() - ( 30 * DAY_IN_SECONDS );
+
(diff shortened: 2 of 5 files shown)


pfefferle:
Closing: trunk already handles this correctly. The Classic Editor pre-selects 'Do not federate' for old posts via get_default_visibility(), and save_meta_data() persists it on save. The block editor's useEffect syncs the default to meta as well. No changes needed.

PR #2917, opened 2026-02-13
Fix visibility default overriding explicit user choices on old posts

### LINKED ISSUE #2618: Updated post (already published) published to activity feed

pfefferle:
## Summary

Fixes a bug where editing an old post (>30 days) with visibility explicitly set to "Public" would incorrectly have its visibility changed to "Do not federate" (local) when saving.

## Root Cause

The `ACTIVITYPUB_CONTENT_VISIBILITY_PUBLIC` constant is stored as an empty string (`''`). The visibility default logic was checking if the visibility meta value was truthy (`if ( $saved_visibility )` in PHP, `if ( meta?.activitypub_content_visibility )` in JS), which treats an empty string as falsy. This caused the code to think "Public" visibility was not set, and it would apply the age-based default (local for posts >30 days old).

### PHP (Classic Editor)
- Updated `get_default_visibility()` in `integration/class-classic-editor.php` to use `metadata_exists()` instead of a truthy check
- This properly distinguishes between "not set" (no meta key) and "set to public" (empty string meta value)

### JavaScript (Block Editor)
- Updated `getDefaultVisibility()` in `src/editor-plugin/utils.js` to check if the key exists in the meta object using the `in` operator
- Updated the `useEffect` in `src/editor-plugin/plugin.js` to properly detect if visibility is set
- Returns `'public'` (or empty string normalized to `'public'`) when the meta key exists with an empty string value

### Tests
- Added 5 new test cases in `tests/phpunit/tests/integration/class-test-classic-editor.php`:
  - `test_get_default_visibility_respects_explicit_public()` - Regression test for issue #2618
  - `test_get_default_visibility_defaults_old_posts_to_local()` - Verifies default behavior for old posts
  - `test_get_default_visibility_respects_explicit_local()` - Verifies explicit local choice is preserved
  - `test_get_default_visibility_defaults_new_posts_to_public()` - Verifies default behavior for new posts
  - `test_get_default_visibility_federated_posts_default_to_public()` - Verifies federated posts stay public
- Updated JavaScript tests in `src/editor-plugin/__tests__/plugin.test.js`:
  - Added `test('respects explicit public visibility (empty string) on old posts')` - Regression test
  - Updated sync logic tests to match new behavior

## Test Plan

- [x] All existing tests pass
- [x] New test cases pass
- [x] PHP linting passes (`composer lint`)
- [x] JavaScript linting passes (`npm run lint:js`)
- [x] Manual testing:
  - Create an old post (>30 days) without setting visibility → defaults to "Do not federate" ✓
  - Edit that post and explicitly set to "Public" → saves as Public ✓
  - Re-open the post editor → visibility stays "Public" (not changed to "Do not federate") ✓

## Changelog

**Type:** Fixed
**Significance:** Patch
**Message:** Fix visibility default overriding explicit user choices on old posts.

Fixes #2618


pfefferle:
Closing: trunk already handles this correctly. See #2915 for details.

PR #2926, opened 2026-02-16
Refactor: Extract Application from the shared actor pipeline

pfefferle:
Fixes #

## Proposed changes:

* Extract the Application actor from the shared actor pipeline into a standalone utility class (`\Activitypub\Application`) and self-contained REST controller.
* The Application is no longer treated as a "virtual user" with ID `-1` flowing through `Actors::get_by_id()`. It exists only as a JSON-LD document and HTTP signing identity.
* Remove `Actors::APPLICATION_USER_ID` constant and all `-1` guard clauses from Mailer, Follow handler, CLI, Health Check, Outbox, Stream integration, etc.
* Add migration to rename keypair option from `activitypub_keypair_for_-1` to `activitypub_application_keypair`, with legacy fallback.
* Deprecate `\Activitypub\Model\Application` with `_deprecated_class()` notice pointing to the new classes.
* Add dedicated `@application` rewrite rule so the pretty URL resolves outside the actor routing system.
* Fix `WP_Query` parameter (`number` → `posts_per_page`) and guard `strtotime()` against empty `post_date_gmt`.
* Explicitly Reject Follow requests aimed at the Application actor: `Follow::reject_application_follow()` sends a directly-delivered Reject signed with the Application key (`Http::post()` accepts `null` as user ID for that), so remote follow requests don't sit "pending" forever.
* Retire pending pre-upgrade `application` outbox items instead of misattributing them to the Blog actor (they share `post_author` 0).
* Reserve the `application` identifier in the blog-identifier sanitizer to avoid handle collisions with the Application actor.
* Deduplicate `get_icon()`/`get_published()` into the `Application` utility class (controller and deprecated model delegate) and tune the published-date `WP_Query`.
* Pass a hydrated `Generic_Object` to the `activitypub_activity_object_array` filters instead of `null`, keeping the documented filter contract.
* Fix a fatal in the Fediverse Stats block fallback that still referenced the removed `APPLICATION_USER_ID` constant.

## Why handle the Application separately?

This is the first step of a larger effort to move the local actor model off magic IDs (`0` for the blog actor, `-1` for the application actor) and onto real, WordPress-native identity. The blog/group actor and the application actor need **different** treatments, and this PR settles the easy half so the harder blog-actor work can focus on a single problem.

The application actor has **no actor surface to preserve**:

* no `post_author` rows (it authors no content),
* no outbox content,
* no followers / following,
* no inbox recipient entries,
* no images / extra fields,
* no profile UI.

It exists only as a JSON-LD document and an HTTP-signing identity for signed fetches / secure mode. Because there is nothing to migrate and nothing that needs a `wp_users` row, the cheapest correct move is to take it out of the actor abstraction entirely (this PR) rather than give it a real user or keep it as a resolver alias.

The **blog/group actor is the opposite case** and is intentionally **not** touched here: it owns content (`post_author`), followers, inbox entries, images, and external `/actors/0/...` URLs that must stay resolvable. That actor is the subject of separate work and will move toward a real `WP_User` with `0` kept only as a compatibility alias. Rule of thumb established by this PR: *extraction* for identities with no actor surface (application); *alias-then-real-user* for identities that own content and must stay resolvable (blog).

### Other information:

## Testing instructions:

* Verify `https://example.com/wp-json/activitypub/1.0/application` returns a valid Application actor with `publicKey`, `inbox`, `outbox`, and proper `@context`.
* Verify `https://example.com/@application` serves the Application actor JSON directly (an internal rewrite to the `/wp-json/activitypub/1.0/application` endpoint; it resolves in place, it does not issue an HTTP redirect).
* Verify WebFinger for `acct:application@example.com` returns correct data.
* On an existing installation, verify the Application keypair is preserved after upgrade (keys should not change).
* Verify HTTP Signature verification still works for outbound GET requests (e.g., fetching remote actor profiles).
* From a remote instance (e.g. Mastodon), try to follow `application@example.com` — the follow request should be explicitly rejected instead of staying "pending".
* Try to set the blog profile ID to `application` in the settings — it should be refused as reserved.
* Run full PHPUnit suite — all tests should pass.

### Changelog entry

<details>

#### Significance

-   [x] Major

#### Type

#### Message

Improve the internal handling of the Application actor used for server-to-server requests.

</details>


jeherve (commented):
> Verify https://example.com/@application redirects/resolves to the Application endpoint.

The json gets served, but the redirect doesn't happen for me.

> On an existing installation, verify the Application keypair is preserved after upgrade (keys should not change).

They do change for me when I switch from one branch to the other. 

******

Everything else lines up and looks good for me right now.

pfefferle:
@jeherve Thanks for testing! Two follow-ups:

**Keypair changing:** Fixed in the latest push. There's now a runtime legacy fallback that reads the old `activitypub_keypair_for_-1` option when the new `activitypub_application_keypair` isn't set yet, so the key survives an upgrade even before the rename migration has run. One caveat worth flagging: switching *back* to trunk will still regenerate, because trunk doesn't know about the new option name. That's a dev branch-hopping artifact, not something a real upgrade hits.

**`@application` redirect:** That one's working as designed. Unlike author actors, the Application is a system actor with no human-facing profile page, so there's nothing to redirect a browser to. The rewrite rule resolves `@application` directly to the REST endpoint and serves the JSON-LD. So "JSON served, no redirect" is the expected behavior here.

Mind giving the keypair path another spin after the latest commits?

jeherve (dismissed):
This is looking good to me!

Issue #2927, opened 2026-02-16
Prevent unnecessary outbox activities from third-party plugin post updates


pfefferle:
## Problem

Third-party plugins (editorial calendars, social schedulers, content converters, hosting sync tools) frequently call `wp_update_post()` on published posts without meaningfully changing the content. Each call fires `wp_after_insert_post`, which creates an outbox entry — even when the post content hasn't actually changed.

This results in:
- **Excessive Update activities** sent to followers for unchanged posts
- **Spurious Delete activities** when plugins briefly transition post status
- Outbox flooding at rates of several activities per minute during bulk operations
- Old posts (years old) being re-federated unnecessarily

### Contributing factors

1. **Plugins calling `wp_update_post()` remotely or in batch** — e.g., cloud-based editorial calendars syncing post metadata, content migration tools converting classic posts to blocks, search-and-replace operations
2. **Non-deterministic `the_content` output** — plugins injecting HTML with random values (nonces, embed secrets, random player IDs) into `the_content` filter output, making content appear different on every render even when the actual post content is identical. Examples:
   - WordPress core's `get_post_embed_html()` generates a fresh `#?secret=` on every call
   - Podcast player plugins inject random `player_id` values via `wp_rand()`
   - Subscription/paywall blocks rendering dynamic URLs
3. **No deduplication gate** — activity type is currently determined based solely on post status transitions, without checking whether the federated content actually changed

## Proposed solution

Add a lightweight content-change detection step before creating outbox entries for Update activities. When an Update is determined, compare a hash of the relevant post properties against the previously federated version before writing to the outbox.

### Considerations

- **Latency budget is tight** — this runs on every `wp_after_insert_post`, so the check must be fast (hash comparison, not full object transformation)
- **Storage** — could store a content hash in post meta (e.g., `_activitypub_content_hash`) that gets set when an activity is actually dispatched
- **What to hash** — should include the fields that matter for federation: `post_content`, `post_title`, `post_excerpt`, `post_status`, key meta fields. Should NOT include rendered `the_content` output (which is non-deterministic)
- **Create vs Update** — this check should only apply to Update activities; Create and Delete should proceed unconditionally
- **Escape hatch** — provide a filter so plugins can bypass the check if needed

### Alternative approaches

- Rate-limiting outbox entries per post (e.g., max 1 Update per post per N minutes)
- Debouncing via short-lived transients before scheduling the outbox write
[...]

PR #2929, opened 2026-02-16
Fix purge methods to handle large collections without OOM or timeout

pfefferle:
Related #2927

## Proposed changes:
* Fix `Outbox::purge()`, `Inbox::purge()`, and `Posts::purge()` to process deletions in batches of 100 with a 30-second time guard, preventing OOM and timeout on large collections.
* Fix `wp_count_posts()` threshold check to count all post statuses (not just `publish`), since outbox items start as `pending`.
* Add `MAX_ITEMS` hard cap (5000) per collection — when exceeded, purge oldest items regardless of age, then re-apply date filter once back under the cap.
* Add early return when `$days <= 0` to prevent accidental purge-everything behavior.

### Other information:

Existing 24 purge tests all pass. The batching and overflow logic is exercised by the existing test suite.

## Testing instructions:

* Run `npm run env-test -- --filter=test_purge` — all 24 tests should pass.
* To manually verify the overflow behavior: create more than 5000 outbox items and confirm purge deletes oldest first regardless of retention period.
* To verify batching: add logging to `Outbox::purge()` and confirm it queries in batches of 100.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fix collection purge methods to batch deletions and enforce a hard item cap, preventing silent OOM failures on large sites.

</details>

### DIFF
--- a/.github/changelog/2929-from-description
+++ b/.github/changelog/2929-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix collection purge methods to batch deletions and enforce a hard item cap, preventing silent OOM failures on large sites.
--- a/.github/changelog/2929-from-description
+++ b/.github/changelog/2929-from-description
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fixed
 
-Fix collection purge methods to batch deletions and enforce a hard item cap, preventing silent OOM failures on large sites.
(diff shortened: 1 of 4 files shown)


jeherve (approved):
This looks good, should be another step in the right direction.

PR #2932, opened 2026-02-17
Delete superseded outbox items instead of publishing them

pfefferle:
Related #2927

## Proposed changes:
* Superseded pending outbox items were previously marked as `publish`, which is semantically incorrect — they were never actually federated. This changes them to be hard-deleted instead, keeping the outbox clean.
* Rename `invalidate_existing_items` to `delete_superseded_items` with improved PHPDocs and parameter naming.
* Update tests to assert deletion (`assertFalse`) instead of status change (`assertEquals 'publish'`).

### Other information:

## Testing instructions:

* Create a post and verify an outbox item is created with `pending` status.
* Update the same post quickly (before federation completes) and verify the first outbox item is deleted (not published), and a new pending item exists.
* Delete the post and verify all pending outbox items for that post are removed.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Superseded outbox activities are now removed instead of kept, reducing clutter in the outbox.

</details>

### DIFF
--- a/.github/changelog/2932-from-description
+++ b/.github/changelog/2932-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Superseded outbox activities are now removed instead of kept, reducing clutter in the outbox.
--- a/includes/collection/class-outbox.php
+++ b/includes/collection/class-outbox.php
@@ -116,25 +116,32 @@ class Outbox {
 	}
 
 	/**
-	 * Delete pending outbox items that have been superseded by a newer item
-	 * with the same activity type and object ID.
+	 * Delete pending outbox items that have been superseded by a newer item.
+	 *
+	 * For most activity types, only items with the same type and object ID are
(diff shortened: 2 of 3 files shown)


jeherve:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jeherve (approved):
This looks good to me, although it doesn't fully address the problem we discussed in #2927, it would definitely cut through some of the noise.

pfefferle:
@jeherve I also thought about comparing the full JSON with the JSON before, but WordPress adds random hashes to embeds, so it is not simply comparing two strings or signatures, but needs to have some logic to strip these kind of randomized data.

And fully relying on such a mechanism is too fragile, because there are also a lot of unknowns like for example SEO tools that dynamically change content to make the site more attractive to visitos.

PR #2971, opened 2026-02-24
Add pre-publish panel suggesting post formats for federation

### LINKED ISSUE #2903: Improve detection and signaling of image-focused posts for federation
## Context

As the German idiom goes, "vor der eigenen Tür kehren" (sweep in front of one's own door), before asking image-focused platforms to add `Article` support, we should first improve how we detect and represent image-focused posts on our end.

- https://github.com/VernissageApp/VernissageServer/pull/442
- https://github.com/pixelfed/pixelfed/pull/6458

Image-focused Fediverse platforms like Vernissage and Pixelfed only handle `Note` objects with image attachments. They shouldn't need to add special handling for `Article` just because WordPress sends blog posts that way. If a WordPress post is essentially a collection of images with minimal text, we should be smart enough to detect that and send it as a `Note` with attachments instead.

## Current behavior

The plugin determines the object type (`Article` vs `Note`) based on WordPress post formats:

- **No post format / standard** → `Article`
- **Post formats like `aside`, `gallery`, `image`, `video`, etc.** → `Note`
- **No title** → `Note`
- **Pages** → `Page`

This means a photo-heavy blog post with a title and no explicit post format is always sent as an `Article` — even if the content is 90% images. These posts are invisible on image-focused platforms because they don't process `Article` objects.

## Ideas to explore

1. **Content-based heuristics** — Analyze the post content to determine if images are the dominant element. If a post is mostly images with little text, send it as a `Note` regardless of post format.

2. **Image-to-text ratio detection** — Calculate the ratio of image blocks/elements to text content. Posts exceeding a threshold could be treated as image-focused.

3. **Gallery and image block detection** — Posts consisting primarily of `core/gallery`, `core/image`, `jetpack/tiled-gallery`, or `jetpack/slideshow` blocks could be detected as image-focused content.

4. **Better attachment metadata** — Ensure image attachments include rich metadata (dimensions, focal point, alt text) so that platforms that do display them can do it well.

5. **Make the detection filterable** — Provide a filter so themes and plugins (e.g., photo blogging plugins) can signal that their posts are image-focused.

6. **Increase default attachment limit** — `activitypub_max_image_attachments` defaults to 4, which may be too low for photo-heavy posts.

## Related

- The `filter_media_by_object_type` filter already exists for customizing media output per post type
- WordPress post formats (`image`, `gallery`) already exist but are rarely used by themes
- FEP-b2b8 discusses Article handling: https://codeberg.org/fediverse/fep/src/branch/main/fep/b2b8/fep-b2b8.md

pfefferle:
Fixes #2903

Screenshot 2026-02-24 at 13 43 03

## Proposed changes:
* Add a pre-publish panel that analyzes post content and suggests a better post format when the object type setting is "Post Format". This helps image/video/audio posts get shared as Notes instead of Articles, making them visible on media-focused platforms like Pixelfed and Vernissage.
* Pass `objectType` and `noteLength` from PHP to the editor JS via `_activityPubOptions`.
* Raise `ACTIVITYPUB_NOTE_LENGTH` default from 400 to 500 characters.
* New `src/pre-publish-panel/` package with detection utilities, component, and tests.

### Other information:

## Testing instructions:

* Go to Settings → ActivityPub and set "Object Type" to "Post Format" (this is the default).
* Add an image block (or gallery, video, audio) with a short caption.
* Click "Publish" — the pre-publish panel should show a "Fediverse ⁂" section suggesting the appropriate format (e.g., Image, Gallery).
* Click the "Set format to Image" button — the suggestion disappears as the format is applied.
* Verify that long text-only posts (>500 chars) do not trigger a suggestion.
* Verify that posts with an explicitly set post format do not trigger a suggestion.

### Changelog entry

<details>

#### Significance

-   [x] Minor

#### Type

#### Message

Add a pre-publish suggestion that recommends a post format for better compatibility with media-focused Fediverse platforms.

</details>


src/pre-publish-panel/block.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "pre-publish-panel",
```
jeherve: I'm not used to seeing Gutenberg plugins with a `block.json` declaration so I'm not sure what's the standard there, but Copilot's review [here](https://github.com/Automattic/wordpress-activitypub/pull/2971/changes#r2846922150) seems to make sense: there may be other plugins in the repo that would use `pre-publish-panel` as name. Could we switch to `actvitypub/pre-publish-panel`?

src/pre-publish-panel/block.json
```diff
@@ -0,0 +1,8 @@
...
+	"title": "Pre-Publish Panel: not a block, but block.json is very useful.",
```
jeherve: I see now that you've done that for other plugins as well, but since this info is displayed on the plugin page on WordPress.org, maybe we should be more descriptive?

Image

src/pre-publish-panel/utils.js
```diff
@@ -0,0 +1,210 @@
...
+ * Recursively analyzes blocks and returns content statistics.
+ *
+ * Walks all blocks (including inner blocks) and counts media elements
+ * and text content to help determine the best post format.
+ *
+ * @param {Array} blocks The blocks to analyze.
+ *
+ * @return {Object} Content statistics with imageCount, galleryCount, videoCount, audioCount, textLength, textBlockCount.
+ */
+export const analyzeBlocks = ( blocks ) => {
```
jeherve, line 75: I could see something like this becoming resource-heavy on super-long or intricate posts ; I don't know if we should worry about it.

It makes me wonder if a utility like that already exists in Core ; it seems like it would be useful. This is why [the block parser](https://make.wordpress.org/core/2025/11/19/introducing-the-streaming-block-parser-in-wordpress-6-9/) was created (although it may not be the best fit here).

pfefferle: from Claude: the analysis is already memoized with `useMemo` and is just a simple property-reading tree walk. I don't think it needs further optimization, but let me know if you'd like me to add a safeguard or reply on the PR.

pfefferle:
@jeherve follow up to improve block.json: https://github.com/Automattic/wordpress-activitypub/pull/2980

jeherve (approved):

PR #3025, opened 2026-03-10
Fix Reactions block alignment in block themes

### LINKED ISSUE #3023: Reactions block not aligned in front-end
### Quick summary

After update, reactions block not aligned.

### Steps to reproduce

Since the update to version 8.0.0, the reactions block is not left-aligned to the content group or column in single posts, regardless of whether the action buttons are activated or deactivated. It remains fully left-aligned.

The issue is NOT observed in the back-end theme editor.

I can confirm the issue in both Safari and Chrome browsers.

Image
Image

Site statistics:
WordPress Version: 6.9.1
Active Theme: TT4 1.4

### Site owner impact

### Severity

Minor

_No response_

_No response_

pfefferle:
Fixes #3023

## Proposed changes:

* Remove `layout` block support and replace with CSS self-constraining (`max-width` + `margin: auto`).
* Remove `__unstableLayoutClassNames` prop from the editor component since layout support is removed.

### Root cause analysis

When a user edits a template containing the Reactions block (e.g., toggling "Show action buttons"), WordPress saves the block with its `save()` wrapper HTML into the template markup. From that point, WordPress's `wp_render_layout_support_flag()` attempts to apply layout classes by comparing the saved wrapper HTML against the rendered output using `str_contains()`.

The comparison fails because WordPress Core produces CSS classes in **different orders** on the client vs. server:

- **JS (`useBlockProps.save()`)**: generated classname first, then className → `"wp-block-activitypub-reactions is-style-facepile"`
- **PHP (block supports)**: `custom-classname` support runs before `generated-classname` support → `"is-style-facepile wp-block-activitypub-reactions"`

This is a Core-level issue in how block supports are registered — the PHP support registration order doesn't match the JS class concatenation order. The `str_contains()` substring check then fails because the needle (`wp-block-activitypub-reactions is-style-facepile`) doesn't appear as a substring in the haystack (`is-style-facepile wp-block-activitypub-reactions ...`).

### Why CSS self-constraining

We investigated fixing the class order via `get_block_wrapper_attributes()` but the mismatch originates in Core's block support registration order, not in how we call the function. Since we can't fix that on our side, CSS self-constraining is the pragmatic solution:

1. Works identically whether the block is auto-inserted via `blockHooks` or saved in template markup.
2. Properly respects `alignwide` and `alignfull` via targeted `max-width` overrides.
3. Doesn't depend on WordPress Core's internal layout class application mechanism.

This should also be reported as a Core bug — any dynamic block with a render.php that uses `get_block_wrapper_attributes()` is affected by this class-order mismatch.

### Other information:

## Testing instructions:

1. Use a block theme (e.g., Twenty Twenty-Four).
2. Ensure the Reactions block is present after post content (it auto-inserts via `blockHooks`).
3. Edit the template in the Site Editor — toggle any block setting (e.g., "Show action buttons") and save.
4. View a post on the frontend and verify the Reactions block is centered/aligned with the post content.
5. Test `alignwide` and `alignfull` alignment options still work correctly.
6. Verify auto-hooked blocks (never edited in Site Editor) still get the correct default style.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fix Fediverse Reactions block not aligning with post content in block themes.

</details>


jeherve (commented):
I worry that the CSS approach will prove less flexible and less compatible with a wide variety of themes. I think it's better to let Core and the themes handle alignment and layout in general, whenever possible. Could we keep `layout`, but fix the output from `get_block_wrapper_attributes()` instead?

> these two functions produce CSS classes in different orders:
>  useBlockProps.save(): "wp-block-activitypub-reactions is-style-facepile"
>  get_block_wrapper_attributes(): "is-style-facepile is-style-facepile wp-block-activitypub-reactions"

This feels like something that could potentially be fixed in Core?

pfefferle:
@jeherve Good point — we investigated several approaches to keep `layout` support working.

### What we tried

1. **Not passing `className` to `get_block_wrapper_attributes()`** — PHP block supports still produces classes in the wrong order because `custom-classname` support runs before `generated-classname` in Core's support registration. The order is baked into Core, not controllable from our side.

2. **`save() { return null; }` (fully dynamic block)** — Core's layout filter doesn't inject layout classes for fully dynamic blocks at all. Added a v3 deprecation, but layout classes were simply never applied.

3. **Keeping `layout` support for editor alignment controls** — The wide/full alignment toolbar only appears when the block is inside a constrained layout parent (e.g., inside post-content). Since `blockHooks` places the block *after* post-content (as a sibling at the template level), alignment controls don't show regardless.

### Why CSS self-constraining

Looking at the other blocks in this plugin, **none of them use `layout` support** — follow-me, followers, following, extra-fields all work fine without it. The `layout` support was never actually functioning for the reactions block due to the class-order mismatch, so removing it has zero regression impact.

The CSS approach (`max-width: var(--wp--style--global--content-size)` + `margin: auto`) reliably handles alignment regardless of:
- Whether the block is auto-inserted via `blockHooks` or saved in template markup
- Class ordering between JS and PHP
- Parent layout context

The underlying class-order mismatch between `useBlockProps.save()` and PHP block supports is a Core-level bug that affects any dynamic block with `render.php` + `layout` support. Worth filing on Trac separately.

PR #3035, opened 2026-03-14
Add Posts and Replies block for author archives

### LINKED ISSUE #1213: Feature Request: Posts and replies block
### What

Mastodon and other Fediverse software make it easy to browse someone's profile and see what they've posted:

- Posts and boosts
- Replies

Image

Right now, this is not possible on a WordPress site. On your profile (e.g. your author archive page on your site, you can only see a list of your recent posts.

It would be nice if the ActivityPub plugin came with a block you could use in the author archive template of a block-based theme, to display both your recent posts and your recent replies, in separate tabs. 

### Why

One of the advantages of using WordPress as your Fediverse software is that you can customize the look of your site as much as you want. You're not limited to the Mastodon interface, you're free to customize things and make your Fediverse profile look unique.

Adding more blocks in the ActivityPub plugin that will bring even more flexibility to one's profile is a great way to further show off that flexibility WordPress offers.

### How

Since the ActivityPub plugin now offers a Federated reply block, we should be able to detect posts that are replies based on the presence of that block in the post content. We can then make 2 different WP queries to get all posts, and all replies.

pfefferle:
Fixes #1213

## Proposed changes:
* Add a new `activitypub/posts-and-replies` block that renders a tabbed interface on author archive pages, similar to Mastodon's profile view.
* The "Posts" tab excludes replies (posts containing the `activitypub/reply` block) while "Posts & Replies" shows all posts.
* Uses the Interactivity API for instant client-side tab switching with both tab contents pre-rendered server-side.
* Updates the Fediverse author archive template to use the new block instead of `core/query`.

### Other information:

## Testing instructions:

* Start wp-env with `npm run env-start`.
* Go to the Site Editor and verify the "Author Archive (Fediverse)" template now includes the `Posts and Replies` block.
* Visit an author archive page (e.g., `/author/admin/`).
* Verify two tabs render: "Posts" and "Posts & Replies".
* Click between tabs — switching should be instant (no page reload).
* Create a post that starts with the `activitypub/reply` block (Federated Reply).
* Verify the "Posts" tab excludes that reply post, while "Posts & Replies" shows it.
* Verify pagination works correctly with pretty permalinks.
* Verify screen reader announces tab selection changes (aria-selected updates dynamically).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add a "Posts and Replies" block for author archives that shows tabs to filter between posts and replies, similar to Mastodon's profile view.

</details>


src/posts-and-replies/render.php
```diff
@@ -0,0 +1,299 @@
...
+	$author = \get_queried_object();
+	if ( $author instanceof \WP_User ) {
+		$author_id = $author->ID;
+	}
+}
+
+// If not on an author archive, bail.
+if ( ! $author_id ) {
+	return;
+}
```
jeherve: I don't think I would limit the block to author archive pages in general ; I think it should be available when inserted in any page. That will allow folks to use it on a static home page for example. I could imagine someone having different sections on their home page, one with their latest blog posts, and then downer a more "social" section, with multiple Fediverse elements.

src/posts-and-replies/render.php
```diff
@@ -0,0 +1,299 @@
...
+$base_args = array(
+	'author'         => $author_id,
+	'post_type'      => 'post',
+	'post_status'    => 'publish',
+	'posts_per_page' => $attributes['postsPerPage'],
+	'paged'          => $current_page,
+);
+
+/**
+ * Filter to exclude posts that contain the activitypub/reply block.
```
jeherve: I wonder if it would be easier to test and maintain if the different query manipulation functions were moved to a dedicated class, so the render file can remain as lean as possible, and handle just the render?

PR #3041, opened 2026-03-16
Fix stale avatar URLs causing 404s

### LINKED ISSUE #3038: Stale avatar URLs cause 404s when remote actors change their profile pictures
When a remote Fediverse actor changes their avatar, the locally stored actor JSON (post_content of `ap_actor` CPT) retains the old icon URL. This causes 404s for avatars displayed in the reactions block and comments.

The `activitypub_update_remote_actors` cron runs hourly but only processes a small batch (default 50) of actors sorted by oldest post_modified. With a large number of actors (e.g. 21,525 on my site), it takes many days/weeks to cycle through all of them. During that time:

1. Actor JSON icon.url points to a URL that no longer exists on the remote server
2. Remote_Actors::get_avatar_url() reads this stale URL from stored actor data
3. Avatar::maybe_cache() stores the stale remote URL in _activitypub_avatar_url meta (caching fails because the remote file is 404/403)
4. The avatar displays as broken/default

Steps to reproduce:

1. Have a site with many remote actors (thousands)
2. Wait for some of those actors to change their Mastodon/Fediverse avatar
3. Observe that their avatars on your site show 404 or fall back to the default mp.jpg

Expected behavior:

Avatar updates should be detected and cached in a timely manner, regardless of the total number of actors.

Possible solutions:

- Increase the batch size for activitypub_update_remote_actors or make it configurable
- When Avatar::maybe_cache() fails to download a remote URL (404/403), trigger an actor data refresh instead of storing the broken URL in meta
- Add a filter/hook to allow site owners to configure refresh priority (e.g. prioritize actors who have recently interacted)

Environment:

- WordPress ActivityPub plugin version: 8.0.0
- ~21,500 remote actors
- Avatars served via the reactions block ()

### Steps to reproduce

Steps to reproduce:

1. Have a site with many remote actors (thousands)
2. Wait for some of those actors to change their Mastodon/Fediverse avatar
3. Observe that their avatars on your site show 404 or fall back to the default mp.jpg

### Site owner impact

### Severity

Minor

No revenue impact

Add a daily cron in your theme to clear _activitypub_avatar_url meta entries that point to remote (non-local) URLs, forcing the plugin to re-download and cache them on next access.

pfefferle:
Fixes #3038

## Proposed changes:
* Refactor `Avatar::maybe_cache()` to use the existing `File::get_or_cache()` filesystem-based cache infrastructure instead of a custom `_activitypub_avatar_url` post meta shortcut.
* The old approach permanently cached broken remote URLs in meta when downloads failed, causing persistent 404s until the actor was manually updated. With 21,500+ actors and only 5-50 processed per cron hour, stale avatars could persist for days/weeks.
* The new approach matches how `Media::maybe_cache()` already works: check filesystem for cached file, download if not found, return original URL on failure. No broken URL gets persisted, so the next access retries.
* Legacy `_activitypub_avatar_url` meta is cleaned up on actor update for backwards compatibility.

### Other information:

## Testing instructions:

* Verify avatar caching still works: visit a post with federated comments — remote actor avatars should display correctly.
* Verify that a failed avatar download does NOT permanently cache a broken URL — the original remote URL should be returned as fallback, and the next page load retries.
* Verify that updating an actor (e.g., via the cron or manually) clears the cached avatar files so re-download happens on next access.
* Run `npm run env-test -- --filter=Avatar` — all 26 tests should pass.
* Run `npm run env-test -- --filter=Remote_Actors` — all 33 tests should pass.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fix remote actor avatars getting stuck on broken URLs when the original image becomes unavailable.

</details>

### DIFF
--- a/.github/changelog/3041-from-description
+++ b/.github/changelog/3041-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix remote actor avatars getting stuck on broken URLs when the original image becomes unavailable.
--- a/includes/cache/class-avatar.php
+++ b/includes/cache/class-avatar.php
@@ -88,18 +88,17 @@ class Avatar extends File {
 	}
 
 	/**
-	 * Clear the cached avatar URL meta when an actor is updated.
+	 * Clear the cached avatar when an actor is updated.
 	 *
-	 * This allows lazy re-caching of the avatar on next access,
-	 * ensuring updated avatars are fetched.
+	 * Invalidates cached files so the avatar is re-downloaded on next access.
 	 *
 	 * @param int $post_id The actor post ID.
 	 */
 	public static function clear_avatar_meta( $post_id ) {
-		// Invalidate cached files.
+		// Invalidate cached files so next access re-downloads.
 		self::invalidate_entity( $post_id );
 
-		// Clear the meta so get_avatar_url() will re-cache on next access.
+		// Clean up legacy meta from previous versions.
 		\delete_post_meta( $post_id, '_activitypub_avatar_url' );
 	}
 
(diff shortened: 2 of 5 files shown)


rollecode:
Awesome, thank you!

PR #3045, opened 2026-03-17
Block non-public posts from ActivityPub content negotiation

pfefferle:
## Proposed changes:

* Block draft, pending, future, and trash posts from being served via ActivityPub content negotiation. Previously only `private` status was checked, allowing unauthenticated access to non-public post data via `?p=<id>` with an ActivityPub Accept header.
* Only allow `publish` status (and `inherit` for attachments) through `is_post_disabled()`.
* Previously federated posts that change status remain accessible so the scheduler can create the correct Delete/Update activities.
* Handle `publish → pending` transitions in the scheduler (sends Update with stripped content, like draft).
* Strip content/summary/attachments from all non-published posts in the transformer (previously only drafts).

### Other information:

## Testing instructions:

* Create a post and leave it as draft, pending, or schedule it for the future
* Request `?p=<post_id>` with an `Accept: application/activity+json` header
* Verify you get a 404 / no ActivityPub data instead of the post content
* Publish the post, then move it back to draft or pending
* Verify the post content is stripped (shows "This post is being modified")
* Verify that trashing a previously published post still serves a tombstone correctly

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

-   [x] Security - in case of vulnerabilities

#### Message

Prevent non-public posts (drafts, scheduled, pending review) from being accessible via ActivityPub.

</details>

### DIFF
--- a/.github/changelog/3045-from-description
+++ b/.github/changelog/3045-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: security
+
+Prevent non-public posts (drafts, scheduled, pending review) from being accessible via ActivityPub.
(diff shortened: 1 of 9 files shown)


jeherve (approved):
This looks good and tests well for me. 🚢

PR #3090, opened 2026-03-24
Improve pre-publish panel with clearer messages and confirmation

### LINKED ISSUE #3085: Pre-publish panel suggestions for federation missing help text or confirmation
I encountered the new federation pre-publish panel for the first time while helping a family member post. The panel was added here: [https://github.com/Automattic/wordpress-activitypub/pull/2971](<https://github.com/Automattic/wordpress-activitypub/pull/2971>)

There are variations on the text but the one I saw read "`This post contains multiple images. Setting the format to Gallery will share it as a media post, making it visible on platforms like Pixelfed.`"

1. There's no link to a support page explaining what this does. For example, is this going to change my post?
2. After clicking the button, the panel disappears. It's not clear what was applied. I'd expect to see the current format or another indicator of change.

pfefferle:
Fixes #3085

| Screenshot 2026-03-24 at 09 53 31 | Screenshot 2026-03-24 at 09 53 48 |
|---|---|

## Proposed changes:
* Reword suggestion messages to clarify that changing the post format only affects how the post appears on the Fediverse, not the post on the WordPress site itself.
* Show a confirmation panel ("This post will be shared as Gallery on the Fediverse.") after a format is applied, instead of silently hiding the panel.

### Other information:

Existing tests cover the suggestion logic. The confirmation state is a simple render-path change based on `postFormat` already being tested.

## Testing instructions:

* Go to Settings → ActivityPub and ensure "Object Type" is set to "Post Format" (the default).
* Add two image blocks with a short caption.
* Click "Publish" — the pre-publish panel should show a "Fediverse ⁂" section suggesting Gallery format with the updated message mentioning it won't affect the site.
* Click "Set format to Gallery" — the panel should now show "This post will be shared as Gallery on the Fediverse." instead of disappearing.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Improve the pre-publish format suggestion panel with clearer messages and a confirmation after applying a format.

</details>


src/pre-publish-panel/plugin.js
```diff
@@ -33,6 +33,22 @@ const PrePublishPanel = () => {
...
 	if ( ! suggestion ) {
+		// Show confirmation when the user has a non-default format applied.
+		if ( postFormat && postFormat !== 'standard' ) {
+			const formatLabel = FORMAT_LABELS[ postFormat ] || postFormat;
+			return (
+				<PluginPrePublishPanel title={ __( 'Fediverse ⁂', 'activitypub' ) } initialOpen>
+					<p>
+						{ sprintf(
+							/* translators: %s: The current post format name (e.g., "Image", "Gallery", "Video"). */
+							__( 'This post will be shared as %s on the Fediverse.', 'activitypub' ),
```
allilevine: The confirmation message `"This post will be shared as %s on the Fediverse."` reads awkwardly without an article — e.g., "shared as Gallery" instead of "shared as a Gallery."

Adding `"a"` doesn't work universally either, since Audio and Image need `"an"` not `"a"` — and hardcoding the article per format is fragile for translators.

Suggestion: rephrase to avoid the article entirely:

```suggestion
							__( 'This post will be shared in %s format on the Fediverse.', 'activitypub' ),
```

This works cleanly for all formats and is more translation-friendly.

src/pre-publish-panel/plugin.js
```diff
@@ -33,6 +33,22 @@ const PrePublishPanel = () => {
 	}
 
 	if ( ! suggestion ) {
+		// Show confirmation when the user has a non-default format applied.
+		if ( postFormat && postFormat !== 'standard' ) {
```
allilevine: This is an edge case, but if the user sets their post format to something like `aside` via the standard WordPress format picker, this condition would match. For those, `FORMAT_LABELS[ postFormat ]` is `undefined`, so the fallback shows the raw slug — e.g., "This post will be shared as aside on the Fediverse" (lowercase, untranslated).

Image

Consider restricting this to formats in `FORMAT_LABELS`:

```suggestion
		if ( postFormat && postFormat !== 'standard' && FORMAT_LABELS[ postFormat ] ) {
```

That way the confirmation only appears for the five formats the plugin actually maps to ActivityPub object types.

allilevine (commented):
This is much clearer, thank you for the changes! I suggested two adjustments.

allilevine (approved):
Looks great! ✅ 

Image
Image

pfefferle:
Thanks @allilevine :)

PR #3094, opened 2026-03-25
Cap remote recipient fetches per incoming activity

pfefferle:
## Summary

- Limits the number of outbound HTTP requests triggered by remote recipient URLs in `to`/`cc`/`bcc` fields of incoming activities.
- Defaults to 10 remote fetches per activity, filterable via `activitypub_max_remote_recipient_fetches`.
- Skips the cap entirely for recipients we already know:
  - The actor's followers collection URL (read from the cached actor profile).
  - Any URL already cached as a `Remote_Actors` post (it's a known actor, not a collection — no local recipients to resolve).
- Normalizes `actor` and `followers` through `object_to_uri()` so peers that serialize either field as an inline object are still matched correctly.
- Prevents abuse where a crafted activity with many unknown remote recipients could trigger unbounded outbound requests from the server.

## Test plan

- [ ] Send an activity with fewer than 10 remote recipients in `to`/`cc`/`bcc` — all should be resolved as before.
- [ ] Send an activity with more than 10 remote recipients — only the first 10 unknown URLs should trigger outbound fetches, the rest should be skipped.
- [ ] Verify local (same-domain) recipients are unaffected by the cap.
- [ ] Send an activity addressed to many recipients your instance has already cached — none of them should consume a fetch slot.
- [ ] Send an activity with the actor's followers collection URL in `cc` — followers should be resolved without an HTTP fetch, even if the cap is already reached.
- [ ] Verify cross-domain followers URLs (e.g., WordPress.com where actor and collection domains differ) are still detected from cached metadata.
- [ ] Verify peers that serialize `actor` or `followers` as an inline object (rather than a string IRI) are handled correctly.


jeherve (commented):
This is looking good.

One question that came up was whether it’s too risky to drop incoming Activities because of large audience lists. From what I can tell, this patch doesn’t actually drop the Activity itself; it only stops doing additional remote recipient fetches once the cap is reached. Same-domain recipients still work, cached remote actors still short-circuit, and cached followers URLs can still resolve locally. That feels like a reasonable tradeoff to me.

That said, there are a few nitpicks that I would highlight:

- In `includes/rest/class-inbox-controller.php`, the new inline-object normalization path for `actor` / `followers` still isn’t covered by tests. The new PHPUnit cases all use string IRIs, so the main interoperability improvement described in the PR isn’t actually verified yet. Maybe worth adding some tests?
- In that same file, I’d also like a test for the legacy `_activitypub_actor_json` fallback in `get_cached_followers_url()`. That fallback is now important for the cached followers shortcut, we don't want that to start failing in the future.
- There’s also a small perf regression here: we now look up the same remote actor multiple times in one request (`get_cached_followers_url()`, public follower resolution, and again in the followers-URL branch). I’m wondering if we should pass the actor post through, or memoize it in-request, so we only do that lookup once.
- Small doc cleanup: the new `activitypub_max_remote_recipient_fetches` filter should get an `@since unreleased` tag.

Maybe something else worth considering: should we log a message when the cap is reached? That may help folks (and ourselves) with some debugging.

jeherve (approved):
This looks good to me. 🚢

PR #3110, opened 2026-03-26
Fix Update handler using stale local actor data instead of activity payload

akirk:
## Proposed changes:

I noticed that a friend's avatar on my Friends plugin timeline was stale — the `ap_actor` still had their old Mastodon avatar URL (`37df032a5951b96c.jpg`) while Mastodon had already switched to a new one (`388649acb2026701.webp`). 

The Friends plugin had the correct avatar because it updates it when receiving an `Update` Person activity. But the `ap_actor` post in the ActivityPub plugin was never updated, even though the same `Update` activity was delivered.

The root cause is in `update_actor()`: it calls `get_remote_metadata_by_actor()` which calls `fetch_by_uri()`, which **returns the stale locally cached copy** when the actor already exists locally (line 262-265 of `class-remote-actors.php`). The `$cached = false` parameter is not passed down the call chain and is thus ignored.

The `Update` activity already contains the fresh actor data in `$activity['object']` — so this PR uses it directly instead of re-fetching.

### Other information:

## Testing instructions:

1. Have a remote actor (e.g. a Mastodon user) already stored locally as an `ap_actor`
2. Have that user change their avatar on Mastodon
3. Mastodon sends an `Update` Person activity
4. **Before this fix:** the `ap_actor` post_content and avatar remain stale
5. **After this fix:** the `ap_actor` is updated with the fresh data from the activity

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fix Update handler using stale local actor data instead of the activity payload

</details>

### DIFF
--- a/.github/changelog/3110-from-description
+++ b/.github/changelog/3110-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix Update handler using stale local actor data instead of the activity payload
--- a/includes/handler/class-update.php
+++ b/includes/handler/class-update.php
@@ -139,18 +139,17 @@ class Update {
 		if ( ! \is_array( $actor ) || ! isset( $actor['id'] ) ) {
 			$object = Http::get_remote_object( $activity['actor'], false );
 
-			if ( \is_wp_error( $object ) || ! \is_array( $object ) ) {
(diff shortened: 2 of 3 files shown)


pfefferle:
Oh! Good catch!

pfefferle (approved):

Issue #3026, opened 2026-03-10
Post meta being set on new posts breaks Gutenberg's starter pattern modal


nate-allen:
### Quick summary

The plugin calls `setMeta()` inside `useEffect` hooks when the editor loads, which marks brand-new posts as "dirty" before the user has interacted with anything. This prevents Gutenberg's "Choose a pattern" starter pattern modal from appearing.

Gutenberg's `StartPageOptions` component only shows the modal when a post is both empty and clean:

```js
const isFreshPage = !isEditedPostDirty() && isEditedPostEmpty();
```

Because ActivityPub eagerly writes default meta values on mount, `isEditedPostDirty()` returns `true` and the modal is skipped.

### Steps to reproduce

1. Activate a block theme (e.g., Twenty Twenty-Five)
2. Register a starter pattern for posts in a plugin or theme:
    ```php
    register_block_pattern( 'my/starter', array(
        'title'      => 'Post starter',
        'blockTypes' => array( 'core/post-content' ),
        'postTypes'  => array( 'post' ),
        'content'    => '<!-- wp:paragraph --><p>Hello</p><!-- /wp:paragraph -->',
    ) );
    ```
3. With ActivityPub deactivated, create a new post -> "Choose a pattern" modal appears
4. Activate ActivityPub, create a new post -> modal does **not** appear
5. Deactivate ActivityPub, create a new post -> modal appears again 

You can verify in the browser console on a new post:

```js
wp.data.select('core/editor').isEditedPostDirty()
// true (should be false on a brand new post)
```

### Site owner impact

Fewer than 20% of the total website/platform users

### Severity

Moderate

### What other impact(s) does this issue have?

No revenue impact

### If a workaround is available, please outline it here.

Remove the `useEffect` hooks that eagerly sync defaults and instead handle defaults at read time or in PHP.

**Option 1: Add default values to `register_post_meta` in PHP** (`includes/class-post-types.php`)

```php
\register_post_meta(
    $post_type,
    'activitypub_interaction_policy_quote',
    array(
        'type'         => 'string',
        'single'       => true,
        'show_in_rest' => true,
        'default'      => \get_option( 'activitypub_default_quote_policy', 'anyone' ),
        // ... sanitize_callback
    )
);

\register_post_meta(
    $post_type,
    'activitypub_content_visibility',
    array(
        'type'         => 'string',
        'single'       => true,
        'show_in_rest' => true,
        'default'      => '', // empty string = public
[...]

PR #3126, opened 2026-04-01
Add stats block with shareable image generation

### LINKED ISSUE #3074: Fediverse Wrapped: Generate a shareable image (sharepic) for year-end stats
## What

Generate a visually appealing sharepic (shareable image) as part of the Fediverse Wrapped feature, so users can show off their annual blog stats — similar to how everyone shares their Spotify Wrapped at the end of the year.

## Why

Having stats in the dashboard is great, but users want to *share* them. A downloadable/shareable image makes it easy to post year-end blog stats on Mastodon, in a blog post, or anywhere else.

## Context

Requested in [https://github.com/Automattic/wordpress-activitypub/issues/2597#issuecomment-4085357025](<https://github.com/Automattic/wordpress-activitypub/issues/2597#issuecomment-4085357025>) by @kaffeeringe:

> I love the stats-feature. But it would be cool to have a generated sharepic at the end of the year. So I can show off my blog stats while everybody posts their Spotify Wrapped. ;-)

Related to [CMF-329](https://linear.app/a8c/issue/CMF-329/feature-request-fediverse-wrapped-annual-blog-summary-converted-to) (Fediverse Wrapped).

pfefferle:
Fixes #3074

## Proposed changes:

* Add `activitypub/stats` block that displays annual Fediverse statistics as a styled card on the site, showing posts federated, engagements, follower growth, top supporter, and top posts.

Screenshot 2026-04-01 at 12 53 39

* When shared via ActivityPub, the block HTML is stripped from the content and replaced with a server-generated 1200x630 image attached to the activity, so it appears as a shareable image on Mastodon, Pixelfed, and other platforms.

Screenshot 2026-04-01 at 12 53 50

* Generated images are cached in `uploads/activitypub/stats/{user_id}/` using `WP_Image_Editor` for WebP optimization. Cached images are served as direct file URLs, bypassing PHP on subsequent requests.

Screenshot 2026-04-01 at 12 54 20

* Two REST endpoints: `/stats/image/{user_id}/{year}` serves the image binary (generates and caches on first hit), `/stats/image-url/{user_id}/{year}` returns the resolved URL as JSON (cached file URL or REST endpoint).
* The share image automatically adopts colors from the site's Global Styles palette and resolves fonts from the active theme or Font Library.
* Editor sidebar includes a "Share Image" panel that fetches and displays the resolved image URL with a copy button and preview link. Hidden when GD is unavailable.

Screenshot 2026-04-01 at 12 54 11

* Block supports native color, border, spacing, and typography settings. Border color is passed through to inner elements. Default colors are derived from the site's theme via CSS custom properties.
* Uses the actor's webfinger identifier in both the block and the share image.
* Image caching can be disabled via the `activitypub_cache_stats_image_enabled` filter. The image URL can be routed through a CDN or image proxy like Photon via the `activitypub_stats_image_url` filter.
* Graceful degradation when GD is unavailable: endpoints return 501, sidebar panel is hidden, no broken attachments.
* All URLs work with both pretty and plain permalinks.

### Other information:

## Testing instructions:

* Go to WP-CLI and run `wp activitypub stats collect --year=2025 --force && wp activitypub stats compile --year=2025` to seed stats data.
* Go to the block editor and insert the "ActivityPub Stats" block into a post.
* Verify the card renders with stats, engagement breakdown, details, and top posts.
* Go to the block sidebar, open the "Share Image" panel, verify the URL points to a cached file in uploads, and the copy button and preview link work.
* Go to the cached image URL directly in the browser and confirm it loads without hitting PHP.
* Go to the REST image endpoint directly and confirm it serves the cached image.
* Go to the post URL with `?activitypub` appended and verify the image is attached (as a direct uploads URL) and the block HTML is stripped.
* Go to Settings > Permalinks, switch to "Plain", and verify the share image URL and block still work.
* Go to Appearance > Themes, switch themes, and verify the share image adapts colors and fonts.
* Go to the block editor, select a user with no stats, and verify saving does not fail.
* Verify the Share Image panel is hidden when GD is unavailable (can test by filtering `activitypub_cache_stats_image_enabled` to false).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Add a stats block that displays annual Fediverse statistics as a card on the site and as a shareable image on the Fediverse, with automatic color and font adoption from the site's theme.

</details>


src/stats/render.php
```diff
@@ -0,0 +1,239 @@
...
+	<?php
+	// phpcs:ignore WordPress.Security.EscapeOutput
+	echo $wrapper_html;
+	?>
+	data-year="<?php echo esc_attr( $stats_year ); ?>"
+>
+			<div class="activitypub-stats__header">
+				<h2 class="activitypub-stats__title"><?php echo \esc_html( $title_text ); ?></h2>
+				<?php if ( $actor_webfinger ) : ?>
+					<p class="activitypub-stats__subtitle"><?php echo \esc_html( $actor_webfinger ); ?></p>
```
jeherve: Should this prepend `@`?

includes/cache/class-stats-image.php
```diff
@@ -0,0 +1,668 @@
...
+
+		static::get_filesystem()->move( $tmp_file, $dest_path, true );
+
+		// Optimize via WP_Image_Editor (handles WebP conversion).
+		$optimized = static::optimize_image( $dest_path, self::WIDTH );
+
+		return $optimized;
+	}
+
+	/**
```
jeherve, line 258: Looks like this was a left-over.

src/stats/render.php
```diff
@@ -0,0 +1,239 @@
...
+									?>
+								</span>
+							</li>
+						<?php endforeach; ?>
+					</ol>
+				</div>
+			<?php endif; ?>
+
+			<div class="activitypub-stats__footer">
+				<span class="activitypub-stats__branding"><?php echo \esc_html( $site_name ); ?> · <?php \esc_html_e( 'Powered by ActivityPub', 'activitypub' ); ?></span>
```
jeherve: I wonder if we should say "the ActivityPub plugin" instead of "ActivityPub"?

src/stats/edit.js
```diff
@@ -0,0 +1,144 @@
...
+			return;
+		}
+		apiFetch( { url: endpoint } )
+			.then( ( response ) => setImageUrl( response.url || '' ) )
+			.catch( () => setImageUrl( '' ) );
+	}, [ selectedUser, displayYear ] );
+
+	useEffect( () => {
+		fetchImageUrl();
+	}, [ fetchImageUrl ] );
```
jeherve, line 82: This seems to trigger 2 requests on every block load in the editor: once with `selectedUser` undefined (falling back to `'blog'`), then again when the `useEffect` on line 67-72 when we get `selectedUser` from `usersOptions`.

includes/cache/class-stats-image.php
```diff
@@ -0,0 +1,668 @@
...
+	 * @return string The hash string.
+	 */
+	private static function get_hash() {
+		$parts = array(
+			\get_stylesheet(),
+			\wp_get_theme()->get( 'Version' ),
+		);
+
+		return \md5( \wp_json_encode( $parts ) );
+	}
```
jeherve: I wonder if we should include the compiled date in the cache key, to ensure that if stats are recompiled (via CLI `--force`, or when the monthly cron updates), the cached image doesn't stay there as stale?

jeherve (commented):
Nice work! If we can remind folks that the block exists when the end of the year approaches, I'm sure it will be super popular!

Here are a few notes below.

*******

When loading the block in the editor, I see multiple requests to `/undefined/actors/{user_id}` that all 404. This seems from `src/shared/use-user-options.js:49` — the `useEffect` that checks if the current user has ActivityPub capability fires before `window._activityPubOptions` is loaded, so `namespace` is `undefined`.

******

I'm not super clear on the stat collection part, so correct me if I'm wrong.

Right now, should we add a one-time backfill on upgrade (via the migration class) so past months in the current year, or the whole last year gets populated? Otherwise everyone will get an empty stats blockk when they try it out.

On my end, I used the CLI tool to collect data, but I think that when I specify `--year`, e.g. `wp activitypub stats collect --year=2025`, only the current month (i.e. April 2025) is collected, vs. all 12 months of 2025. 

I'm think it would make sense that when `--year` is provided without `--month`, it should loop through all months of that year (or at least up to the current month if it's the current year).

********

The stats image is rendered as PNG via `imagepng()`, then converted to WebP by the inherited `optimize_image()`. That pipeline makes sense for caching remote images, but I'm not sure we should use WebP for an image that will be shared on different social networks / apps / clients. WebP doesn't yet have the wide support png has. I'm thinking we could skip that conversion for stats images meant to be shared.

********

When no stats exist, the block outputs nothing on the front end. The editor shows a placeholder message, but a published page just has a blank gap where the block should be. I'm wondering if we should either hide the block wrapper entirely (so no gap), or show a graceful fallback message.

*******

Something seems off with the image generation and the background color. Check this small screencast (also note the padding issues in the editor):

https://github.com/user-attachments/assets/4293cb87-c3c2-4e4e-a754-c63bec188f7c

pfefferle:
@jeherve 

> Nice work! If we can remind folks that the block exists when the end of the year approaches, I'm sure it will be super popular!

I thought about creating a post-template that shows up on january (or on december).

Like I did by accident in https://github.com/Automattic/wordpress-activitypub/issues/3026

pfefferle:
The backfill is already implemented and wired into the migration system. But it only triggers for upgrades from < 7.9.0. Since this branch hasn't been released yet, anyone installing the plugin for the first time (or upgrading from a version >= 7.9.0) won't get the backfill.

pfefferle:
@jeherve you can change the padding/margin in the block settings.

have you changed the colors of the block? is the issue that the block does not use the background of the theme, or is it that the image does not adapt to the changes of the block!?

The idea was, that both should use the colors of the theme if possible.

jeherve:
> have you changed the colors of the block? is the issue that the block does not use the background of the theme, or is it that the image does not adapt to the changes of the block!?

I didn't change the colors of the block now. I think it's a good idea to try to use the theme's colors as much as possible, but as you can see that doesn't quite happen on my site, because the variables are not present in my theme. In the editor it falls back to the default color you set for the background (white), but the generated image does not inherit that color.

pfefferle:
But why do they work on the image then!?

PR #3129, opened 2026-04-01
Fix comments on remote posts being held in moderation

pfefferle:
## Proposed changes:
* Move the `ap_post` auto-approval check in `pre_comment_approved` before the `comment_previously_approved` and `protocol` gates.
* Previously, comments on remote posts (`ap_post`) could be held in moderation if WordPress Discussion settings (like "Comment must be manually approved") or anti-spam plugins set the approval status to `0` and the `comment_previously_approved` option was not `'1'`. The `ap_post` check was unreachable in that case.
* Comments on `ap_post` types should always be auto-approved since they are not visible in the WP admin comment moderation screen.

### Other information:

## Testing instructions:

* Set up a site that follows remote actors (so `ap_post` items exist).
* Change WordPress Discussion settings: uncheck "Comment author must have a previously approved comment" and/or enable "Comment must be manually approved".
* Trigger a remote reply to an `ap_post`.
* Verify the comment is auto-approved and not held in moderation.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Fix comments on remote posts being incorrectly held in moderation.

</details>


jeherve (approved):
It works well in my tests since yesterday. 🚢

PR #3188, opened 2026-04-16
Fix blog actor outbox activity handling

### LINKED ISSUE #3187: Blog actor outbox leaks private Accept activities to unauthenticated requests
### Quick summary

When using blog actor mode (`user_id=0`), the outbox REST endpoint skips the activity type and visibility filters for unauthenticated requests. This causes private `Accept` activities (follow confirmations) to appear in the public outbox alongside regular `Create` activities.

### Steps to reproduce

1. Enable blog actor mode (Settings → ActivityPub → Blog profile)
2. Gain at least one follower (creates `Accept` outbox entries with `activitypub_content_visibility = private`)
3. As an unauthenticated visitor, fetch `https://example.com/wp-json/activitypub/1.0/actors/0/outbox?page=1`
4. `Accept` activities are visible in the response, despite being marked private

**Root cause** in `includes/rest/class-outbox-controller.php`, method `get_items()`:

```php
if ( get_current_user_id() !== $user_id && ! current_user_can( 'activitypub' ) ) {
    // activity type + visibility filters applied here
}
```

For the blog actor, `$user_id = 0`. For unauthenticated requests, `get_current_user_id()` also returns `0`. The condition `0 !== 0` evaluates to `false`, so the filters are never applied for the blog actor.

**Expected behavior**: Private `Accept` activities should be filtered from the public outbox response, regardless of actor type.

**Actual behavior**: All outbox entries (including private `Accept` activities) are returned to unauthenticated users when querying the blog actor's outbox.

### Impact

- **Severity**: Minor
- **Site owner impact**: Fewer than 20% of the total website/platform users
- **Revenue impact**: No revenue impact

Mastodon and other Fediverse software ignore `Accept` activities in outbox fetches, so there is no functional breakage. However it is technically information leakage (follow timestamps are exposed) and it inflates the outbox page count unnecessarily.

### Workaround

Manually delete `Accept` outbox entries via SQL after they have been processed, or add a `rest_activitypub_outbox_query` filter in a code snippet to exclude them.

### Environment

- WordPress 6.9.4
- ActivityPub plugin v8.0.2
- Blog actor mode
- PHP 8.3.29

pfefferle:
Fixes #3187

## Proposed changes:
* Fix blog actor outbox (`user_id=0`) exposing private `Accept` activities to unauthenticated visitors. When `get_current_user_id()` returns `0` for an anonymous visitor, the condition `0 !== 0` evaluated to `false`, skipping the activity type and visibility filters entirely. Added `is_user_logged_in()` check so filters always apply for logged-out users.

### Other information:

## Testing instructions:

* Enable blog actor mode (Settings → ActivityPub → Blog profile)
* Gain at least one follower (creates `Accept` outbox entries with `activitypub_content_visibility = private`)
* As an unauthenticated visitor, fetch `https://example.com/wp-json/activitypub/1.0/actors/0/outbox?page=1`
* Verify that `Accept` activities are no longer visible in the response

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

-   [x] Security - in case of vulnerabilities

#### Message

Fix blog actor outbox exposing private activities to unauthenticated visitors.

</details>


Kernel-Error:
Thanks for the quick fix! The `is_user_logged_in()` guard is exactly the approach I had in mind too when filing the issue. Looks good to me.

pfefferle:
Thanks a lot for reviewing the changes @Kernel-Error

PR #3194, opened 2026-04-17
Fix Reader view crash and infinite scroll on WP 6.9

pfefferle:
Fixes the blank Reader page and broken infinite scroll on current WordPress versions. Three independent problems were tangled together; untangling them took some yak-shaving (eslint flat config, stylelint DS-token rule, npm hoisting for `@base-ui/react`), but the actual fixes are small.

## Proposed changes

- **Drop `@wordpress/views` from the bundle via tree-shaking.** WP 6.9's `privateApis` allowlist no longer includes `@wordpress/views`, so the package's barrel export throws at module-eval time and kills the whole Reader app. Marking `@wordpress/views` as `sideEffects: false` in `webpack.config.js` lets webpack eliminate the unused `useViewConfig` → `lock-unlock` chain; the opt-in call is gone from the bundle and the app boots clean.
- **Wire feed infinite scroll to dataviews 14's new API.** dataviews 14 drives infinite scroll via `view.startPosition` and dropped the `paginationInfo.infiniteScrollHandler` callback. `feed/stage.tsx` now seeds `startPosition: 1` and translates incoming `startPosition` changes into our page-based loader; the old handler prop is removed.
- **Give the feed scroll container an overflow.** DataViews attaches its scroll listener to `.dataviews-layout__container`, which gets `overflow: auto` from the dataviews stylesheet. The followers route imports that stylesheet; the feed route didn't, so the listener attached to a non-scrolling element. Added the `@use` import to `src/app/routes/feed/style.scss`.

Also included (fallout from the `wp-scripts` 32 / `dataviews` 14 bumps that exposed these bugs):

- Bumped `@wordpress/scripts` to 32.0.0 and `@wordpress/dataviews` to 14.1.0 (the two outstanding Dependabot majors).
- Added `@base-ui/react` as a direct devDep. dataviews 14 imports from it via `@wordpress/ui` but npm doesn't hoist it to where webpack resolves the `/wp` subpath bundle.
- Migrated ESLint config to flat config (`eslint.config.cjs`). **ESLint 9 silently ignores `.eslintrc.*` and `.eslintignore`**, so every project override (JSX types, test-file allowances, `assets/` ignore) was being dropped without warning once wp-scripts 32 landed.
- Added `.stylelintrc.json` disabling the new `plugin-wpds/no-unknown-ds-tokens` / `no-setting-wpds-custom-properties` rules; our `--wpds-*` tokens come from outside the WPDS internal allowlist.

### Other information

- [x] Changelog entry added (`fix/reader-view` - patch / fixed).
- [ ] No new tests — this fixes boot-time crashes and the scroll-container CSS contract, both covered end-to-end by manual reader exercise. The existing 238-test jest suite still passes.

## Testing instructions

1. `npm install && npm run build && npm run env-start`.
2. Seed a batch of feed items so there's something to scroll — at least `perPage + 1 = 21`, more fun at 40+:

   ```sh
   for i in $(seq 1 45); do
     npx wp-env run cli wp post create --post_type=ap_post \
       --post_title="Test Post $i" \
       --post_content="Content $i" \
       --post_status=publish --porcelain
   done
   # Attach each to both possible active actors.
   npx wp-env run cli bash -c '
     for id in $(wp post list --post_type=ap_post --format=ids); do
       wp post meta update $id _activitypub_user_id 0 > /dev/null
       wp post meta add    $id _activitypub_user_id 1 > /dev/null
     done
   '
   ```

3. Visit `http://localhost:8888/wp-admin/admin.php?page=activitypub-social-web`. The page should render the feed with 20 items, no console errors.
4. Scroll the feed to the bottom. More items should load in batches until all 45 are visible.
5. Visit the Followers page too — it should still render fine (dataviews 14 regression check).
6. `npm run lint:js` — clean. `npm run lint:css` — clean.  `npm run test:unit` — 238 tests pass.

### Changelog entry

Entry already committed; no action required from the workflow.

<details>

#### Significance

- [x] Patch

#### Type

- [x] Fixed

#### Message

Fix Reader feed failing to load with newer WordPress versions.

</details>


pfefferle:
@jeherve can you check if that fixes the reader for you?

PR #3196, opened 2026-04-18
Fix array_keys(null) fatal in get_comment_type_slugs()

### LINKED ISSUE #3195: Fatal: array_keys(null) in Comment::get_comment_type_slugs() when no custom comment types registered
## Bug

`Activitypub\Comment::get_comment_types()` returns `global $activitypub_comment_types` directly. Before any call to `Activitypub\add_comment_type()`, that global is uninitialized (`null`). `get_comment_type_slugs()` then passes it straight to `array_keys()`, which fatals on PHP 8+:

```
PHP Fatal error: Uncaught TypeError: array_keys(): Argument #1 ($array) must be of type array, null given
  in .../activitypub/includes/class-comment.php:615
```

The `did_action( 'init' )` guard at the top of `get_comment_type_slugs()` covers the "called too early" case but not the "init has run and nobody registered a type" case.

## Stack trace (abridged)

```
#0 class-comment.php(615): array_keys(null)
#1 scheduler/class-comment.php(64): Activitypub\Comment::get_comment_type_slugs()
#2 class-wp-hook.php(341): Activitypub\Scheduler\Comment::schedule_comment_activity()
#3 class-wp-hook.php(365): WP_Hook->apply_filters()
#4 plugin.php(522): WP_Hook->do_action()
#5 wp-includes/comment.php(1898): do_action('wp_transition_comment_status', ...)
#6 wp-includes/comment.php(2582): wp_transition_comment_status(...)
#7 third-party plugin calls wp_set_comment_status(...)
```

## Reproduction

Any third-party plugin that calls `wp_set_comment_status()` on a custom `comment_type` triggers this when no ActivityPub comment types are registered on the site. In our case (GatherPress), a magic-link email verifies an RSVP by transitioning a `gatherpress_rsvp`-typed comment from `hold` → `approve`. First click 500s; subsequent requests no-op the transition, so the crash only surfaces on the first legitimate approve.

A minimal reproduction:

```php
add_action( 'init', function () {
    $comment_id = wp_insert_comment( array(
        'comment_post_ID' => 1,
        'comment_type'    => 'custom_type',
        'comment_approved' => '0',
    ) );
    wp_set_comment_status( $comment_id, 'approve' );
} );
```

## Suggested fix

Have `get_comment_types()` always return an array:

```php
public static function get_comment_types() {
    global $activitypub_comment_types;

    return is_array( $activitypub_comment_types ) ? $activitypub_comment_types : array();
}
```

That also makes `is_registered_comment_type()` and `get_comment_type()` safe against the same state.

## Versions

- ActivityPub: trunk (class-comment.php:615 as of this issue)
- WordPress: 6.x
- PHP: 8.x

mauteri:
Fixes #3195

## Proposed changes:

* Coerce `Comment::get_comment_types()`'s return to an `array` so callers never see the raw `$activitypub_comment_types` global when it's still `null`. Before this change, `Comment::get_comment_type_slugs()` passed that `null` straight to `array_keys()`, which is a `TypeError` under PHP 8+ and takes down any request that reaches it. The `did_action( 'init' )` guard in `get_comment_type_slugs()` covers the "called too early" case but does nothing when `init` has run and no one has registered a comment type — which is the state on every request until `add_comment_type()` is first called.
* The crash is reachable from any plugin that fires `wp_transition_comment_status` on a custom comment type. In our case (GatherPress) a magic-link email verifies an RSVP by transitioning a `gatherpress_rsvp`-typed comment from `hold` → `approve`, and the first click on the link 500s. A one-line `(array)` cast on the global — as suggested in the issue — fixes both `get_comment_types()` and its downstream callers (`get_comment_type_slugs()`, `is_registered_comment_type()`, `get_comment_type()`), which already handled an empty-array return correctly but assumed the value was ever an array in the first place.

### Other information:

Added `Test_Comment::test_get_comment_types_returns_array_when_global_is_null` — backs up `$activitypub_comment_types`, sets it to `null`, and asserts both `get_comment_types()` and `get_comment_type_slugs()` return `array()` instead of fataling. Restores the global in a `finally` so the rest of the test suite is unaffected. Without a test anchored on the `null` case, a future "simplify" refactor could easily drop the cast and reintroduce the fatal.

## Testing instructions:

1. With a clean environment where no plugin has called `add_comment_type()` yet, insert a comment whose `comment_type` is anything custom (e.g. `'custom_type'`) and then call `wp_set_comment_status( $comment_id, 'approve' )`. Without this PR, the transition fires `wp_transition_comment_status`, hits `Activitypub\Scheduler\Comment::schedule_comment_activity()`, and fatals on `array_keys( null )`. With this PR, the transition completes cleanly.
2. Run the PHPUnit suite:
   ```
   composer test
   ```
   The new `test_get_comment_types_returns_array_when_global_is_null` case should pass. Reverting just the one-line change in `includes/class-comment.php` should make it fail with a `TypeError`, confirming the test actually guards the regression.
3. Sanity-check that sites where comment types *do* get registered still behave identically. Load a site in a browser where your theme/plugin stack calls `Activitypub\add_comment_type()` during `init`; every ActivityPub feature that depended on `get_comment_types()` / `get_comment_type_slugs()` should render exactly as before because `(array) $already_an_array === $already_an_array`.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fix a fatal `array_keys(null)` in `Comment::get_comment_type_slugs()` that could take down any request where a third-party plugin transitioned a custom comment type before `add_comment_type()` had been called.

</details>

### DIFF
--- a/.github/changelog/3196-from-description
+++ b/.github/changelog/3196-from-description
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix a fatal `array_keys(null)` in `Comment::get_comment_type_slugs()` that could take down any request where a third-party plugin transitioned a custom comment type before `add_comment_type()` had been called.
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -581,7 +581,7 @@ class Comment {
 	public static function get_comment_types() {
 		global $activitypub_comment_types;
 
-		return $activitypub_comment_types;
+		return (array) $activitypub_comment_types;
 	}
 
 	/**
--- a/tests/phpunit/tests/includes/class-test-comment.php
+++ b/tests/phpunit/tests/includes/class-test-comment.php
@@ -553,9 +553,10 @@ class Test_Comment extends \WP_UnitTestCase {
 	/**
 	 * Regression: get_comment_types() must coerce null to an empty array.
 	 *
-	 * $activitypub_comment_types is null until add_comment_type() runs at
-	 * least once, which isn't guaranteed on every request. array_keys( null )
-	 * fatals under PHP 8+, so callers must see an array either way.
+	 * $activitypub_comment_types can be null until comment types are registered
+	 * (for example via Comment::register_comment_types()), which isn't
+	 * guaranteed on every request. array_keys( null ) fatals under PHP 8+,
+	 * so callers must see an array either way.
 	 *
 	 * @covers ::get_comment_types
 	 * @covers ::get_comment_type_slugs
--- a/tests/phpunit/tests/includes/class-test-comment.php
+++ b/tests/phpunit/tests/includes/class-test-comment.php
@@ -551,6 +551,36 @@ class Test_Comment extends \WP_UnitTestCase {
 	}
 
 	/**
+	 * Regression: get_comment_types() must coerce null to an empty array.
+	 *
+	 * $activitypub_comment_types is null until add_comment_type() runs at
+	 * least once, which isn't guaranteed on every request. array_keys( null )
+	 * fatals under PHP 8+, so callers must see an array either way.
+	 *
+	 * @covers ::get_comment_types
+	 * @covers ::get_comment_type_slugs
+	 */
+	public function test_get_comment_types_returns_array_when_global_is_null() {
+		global $activitypub_comment_types;
+
+		$backup                    = $activitypub_comment_types;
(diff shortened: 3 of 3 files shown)


pfefferle (approved):
Thanks @mauteri 😊

PR #3036, opened 2026-03-14
Add Posts and Replies block using query_loop_block_query_vars

pfefferle:
Related: #1213

Alternative to #3035 — uses a different architecture.

## Proposed changes:
* Add a lightweight `activitypub/posts-and-replies` block that renders a tab bar on any archive page (hidden from ActivityPub requests and feeds).
* Instead of running custom queries, it hooks into the main query via `pre_get_posts` and a one-shot `posts_where` filter, so `core/query` blocks with `inherit: true` automatically pick up the filtering.
* The "Posts" tab (default) adds a `posts_where` clause to exclude posts containing the `activitypub/reply` block. The "Posts & Replies" tab shows all posts unfiltered.
* Tab state is managed via a URL parameter (`?filter=posts-and-replies`), so pagination and Query Loop work naturally.
* Users keep full control over the post template layout in the Site Editor since `core/query` is preserved as-is.
* The `posts_where` filter removes itself after the first execution, so widget queries, related posts, and other secondary queries on the same page are not affected.

### Comparison with #3035
| | #3035 (self-contained) | This PR (query filter) |
|---|---|---|
| Query | Custom `WP_Query` in render.php | Hooks into `core/query` via `pre_get_posts` + `posts_where` |
| Layout control | Fixed (title + date + excerpt) | Full Site Editor customization |
| Tab switching | Instant (Interactivity API, pre-rendered) | Page reload (URL parameter) |
| Pagination | Shared across tabs | Native, per-tab |

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Start wp-env with `npm run env-start`.
* Go to the Site Editor → Templates → "Author Archive (Fediverse)".
* Verify the `Posts and Replies` block appears above the Query Loop block.
* Visit an author archive page (e.g., `/author/admin/`).
* Verify two tabs render: "Posts" (active by default) and "Posts & Replies".
* Create a post that starts with the `activitypub/reply` block (Federated Reply).
* On the "Posts" tab, verify the reply post is excluded.
* Click "Posts & Replies" tab — page reloads with `?filter=posts-and-replies` and the reply post appears.
* The block can also be placed on other archive pages (category, tag, date, home) — not limited to author archives.
* Verify pagination works correctly on both tabs.
* Customize the Query Loop post template in the Site Editor (e.g., add featured image) — verify customizations appear on the frontend.
* Verify that secondary queries (widgets, related posts) are **not** affected by the reply filter.

### Changelog entry

[...]

PR #3209, opened 2026-04-21
Fix reply posts disappearing from front page and admin list

### LINKED ISSUE #3208: Reply posts don't show up on the front page or in the posts view
### Quick summary

There's no toggle to stop this behaviour, but posts with the reply block are like Schrodinger's pos,t both existing and not existing at the same time. They don't appear on my blog, they don't exist in the "posts" view in the admin area, but the next/previous links in my theme still link to the posts.

### Steps to reproduce

1. Post something with a reply block in it.
2. Wonder where it has gone when you go looking for it.

### Site owner impact

### Severity

Critical

No revenue impact

I documented the problem here: https://node.lordmatt.co.uk/2026/04/21/posts/ive-hit-the-wildest-bug-in-wordpress/

I have checked cache, double and triple checked the database. Finally I figured out that reply posts are hidden as if they do not exist.

jeherve:
Fixes #3208.

## Proposed changes:

Hey! This scopes `Blocks::filter_query_loop_vars()` so the `post_content NOT LIKE '%<!-- wp:activitypub/reply%'` clause only gets attached when the current frontend request actually renders the `activitypub/posts-and-replies` tab block.

- Admin and feed queries are now always skipped.
- Block presence is detected via `has_block()` on the static front page / posts page (classic / hybrid setups) and on the likely block template for the current query type (home, front-page, archive, author, date, search, index).
- `candidate_template_slugs()` is a small private helper mapping `WP_Query` conditionals to the block template slugs WordPress would resolve first.

### Why

Shipped in 8.1.0 (#3036) to back the Posts & Replies tab block, the hook runs on every main non-singular `pre_get_posts` and appends the `NOT LIKE` clause via `posts_where`. That hides reply-block posts from both the wp-admin post list view and the front page / archives on sites that don't use the tab block at all — Lord Matt describes it as [Schrödinger's post](https://node.lordmatt.co.uk/2026/04/21/posts/ive-hit-the-wildest-bug-in-wordpress/).

Of note, Friends already worked around the same filter in [akirk/friends@d4639db](https://github.com/akirk/friends/commit/d4639db) after measuring a ~4.7s TTFB regression from the full table scan, so scoping this fixes the performance side of things too.

### Other information:

## Testing instructions:

1. On a site running `trunk`, publish a post containing the Federated Reply block. The easiest way is to open the editor with `?in_reply_to=https://mastodon.social/@some/post` in the URL; the reply-intent plugin inserts the block for you.
2. Without this PR, the post disappears from both the front page listing and `wp-admin → Posts`. With this PR, it appears in both places.
3. Drop the "Posts and Replies" tab block onto the page that renders your blog — the `home.html` template in a block theme, or the page assigned as "Posts page" in Settings → Reading. The default "Posts" tab should still hide the reply-block post on the frontend; the "Posts & Replies" tab should show it.
4. `wp-admin → Posts` should list everything regardless of where the tab block is placed.

### Changelog entry

Manually added at `.github/changelog/fix-posts-and-replies-filter-scope`; no need to check the box below.

### DIFF
--- a/.github/changelog/fix-posts-and-replies-filter-scope
+++ b/.github/changelog/fix-posts-and-replies-filter-scope
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Stop hiding posts that contain a federated reply block from the main blog listing and the admin post list on sites that do not use the Posts and Replies block.
--- a/build/posts-and-replies/render.php
+++ b/build/posts-and-replies/render.php
@@ -17,11 +17,15 @@ if ( is_activitypub_request() || \is_feed() ) {
 	return;
 }
 
-// Determine active tab from URL parameter.
+/*
+ * Determine active tab from URL parameter. Default to "Posts & Replies" so
+ * the reply-exclusion filter only attaches on an explicit opt-in via
+ * ?filter=posts (clicked from the tab bar below).
+ */
 // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-$active_tab = isset( $_GET['filter'] ) ? \sanitize_key( $_GET['filter'] ) : 'posts';
+$active_tab = isset( $_GET['filter'] ) ? \sanitize_key( \wp_unslash( $_GET['filter'] ) ) : 'posts-and-replies';
 if ( ! \in_array( $active_tab, array( 'posts', 'posts-and-replies' ), true ) ) {
-	$active_tab = 'posts';
+	$active_tab = 'posts-and-replies';
 }
 
 $current_url = \remove_query_arg( array( 'filter', 'paged' ) );
--- a/includes/class-blocks.php
+++ b/includes/class-blocks.php
@@ -1259,15 +1259,16 @@ class Blocks {
 	/**
 	 * Filter the main query to exclude replies.
 	 *
-	 * When the "Posts" tab is active (default), adds a WHERE clause to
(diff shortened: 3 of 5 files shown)


pfefferle:
Reworked this to drop the template-probing entirely (d433e76).

**New approach:** the `posts_where` filter only attaches when `$_GET['filter'] === 'posts'` is explicit. The tab block's default tab is now "Posts & Replies", so the only way `?filter=posts` appears in a URL is from a click on the tab bar rendered by the block itself. No block-presence detection needed.

**What this means for Copilot's review comments:**
- `current_request_has_posts_and_replies_block()` + `candidate_template_slugs()` are gone, so the hierarchy-coverage concerns (`category-{slug}`, `tag-{slug}`, plugin namespace, etc.) and the template-part limitation no longer apply.
- The `is_feed()` branch is still there and now covered by a dedicated test.
- No `@since unreleased` tags remain.

**UX change to flag:** the active-by-default tab flips from "Posts" to "Posts & Replies". Visitors who land on a page with the block see all posts by default and can click "Posts" to hide reply-block posts. That is arguably friendlier (no silent hiding) but it is a behavior change from 8.1.0. Will update the PR title/description next.

**Diff:** net -93 lines. `Test_Blocks` 59/59 green, PHPCS clean.

pfefferle (approved):

PR #3211, opened 2026-04-22
Harden the reactions API response against unsanitized remote data

pfefferle:
Fixes #

## Proposed changes:

`Post_Controller::get_reactions` echoed `comment_author` and `comment_author_url` straight from the database into the JSON response, with `html_entity_decode()` actively undoing WordPress's entity encoding. Our React frontend happens to render these as text nodes (safe), but the API is public — any non-React consumer that renders `reaction.name` through `innerHTML` would be XSS-vulnerable via a stored remote actor name, and a stored `javascript:` URL would pass straight through.

* Strip HTML tags from the name before decoding entities, so a stored name cannot introduce markup into the response regardless of what bypassed `kses` at write time.
* Run `comment_author_url` through `esc_url()` so non-HTTP(S) schemes (e.g. `javascript:`) are rejected.
* Apply `esc_url()` to the gravatar URL for symmetry.

### Other information:

`test_get_reactions_sanitizes_remote_actor_metadata` stores a comment with `` as the author name and `javascript:alert(1)` as the URL, then asserts the response contains neither — and that legitimate entity-encoded text (`&amp;friends` → `&friends`) still decodes correctly.

## Testing instructions:

* `npm run env-test -- --filter=Post_Controller` — 21/21 pass including the new test.
* Manual: `curl /wp-json/activitypub/1.0/posts/{id}/reactions` on a post that has a like/repost from a remote actor. The response should contain the actor's display name as plain text and the URL as an `http(s)://` URL or an empty string.

### Changelog entry

<details>

#### Significance

- [x] Patch

#### Type

- [x] Fixed

#### Message

Harden the reactions API response so stored author names and URLs cannot introduce markup or non-HTTP schemes into the JSON output.

</details>

### DIFF
--- a/.github/changelog/fix-reactions-api-response-sanitization
+++ b/.github/changelog/fix-reactions-api-response-sanitization
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Harden the reactions API response so stored author names and URLs cannot introduce markup or non-HTTP schemes into the JSON output.
--- a/includes/rest/class-post-controller.php
+++ b/includes/rest/class-post-controller.php
@@ -165,15 +165,13 @@ class Post_Controller extends \WP_REST_Controller {
 						 * Decode entities first so a stored pseudo-tag like
 						 * `&lt;img&gt;` becomes a real `<img>` for the next
 						 * step to remove, then strip any tags so the JSON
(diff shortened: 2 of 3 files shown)


jeherve (approved):
LGTM!

PR #3218, opened 2026-04-23
Add activitypub_supported_post_types filter

kraftbj:
See DOTCOM-16875.

## Proposed changes:

* Extracts the `activitypub` post-type support registration from `Post_Types::init()` into a dedicated `register_supported_post_type_feature()` method.
* Adds a new `activitypub_supported_post_types` filter that runs after the `activitypub_support_post_types` option is read and before `add_post_type_support()` is called, letting integrators augment or replace the stored admin selection.
* Casts the filter return to an array so a misbehaving integrator returning `null`/scalar does not fatal.

The admin UI at Settings → ActivityPub → "Post Types" continues to reflect the stored option, not the filtered value. Integrations that read `activitypub_support_post_types` directly (Yoast SEO, Enable Mastodon Apps, REST cache, Statistics) are intentionally left alone — this filter governs `post_type_supports( …, 'activitypub' )` registration only, which is the path every federation decision already uses.

### Other information:

Five new PHPUnit tests in `Test_Post_Types` cover: default registration from the stored option (no filter), the filter receiving the stored option value, the filter adding a post type, an empty array skipping registration, and the `(array)` cast guarding non-array returns. A shared `tear_down()` restores the option and `post` support to prevent cross-test leakage.

## Testing instructions:

* Check out this branch and run `npm run env-test -- --filter=Test_Post_Types` — all tests pass.
* Sanity-check with WP-CLI:
  ```bash
  wp shell
  >>> add_filter( 'activitypub_supported_post_types', fn( $t ) => array_merge( $t, [ 'page' ] ) );
  >>> \Activitypub\Post_Types::register_supported_post_type_feature();
  >>> var_dump( post_type_supports( 'page', 'activitypub' ) ); // bool(true)
  ```
* Confirm Settings → ActivityPub → "Post Types" checkboxes still reflect the stored option only (filter additions are not shown as checked).

### Changelog entry

<details>

#### Significance

- [x] Minor

#### Type

#### Message

Added the `activitypub_supported_post_types` filter so integrators can control which post types federate.

</details>

### DIFF
--- a/.github/changelog/add-supported-post-types-filter
+++ b/.github/changelog/add-supported-post-types-filter
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added the `activitypub_supported_post_types` filter so integrators can control which post types federate.
--- a/includes/class-post-types.php
+++ b/includes/class-post-types.php
@@ -48,8 +48,42 @@ class Post_Types {
 		\add_filter( 'add_post_metadata', array( self::class, 'prevent_empty_post_meta' ), 10, 4 );
 		\add_filter( 'update_post_metadata', array( self::class, 'prevent_empty_post_meta' ), 10, 4 );
 
-		// Add support for ActivityPub to custom post types.
-		foreach ( \get_option( 'activitypub_support_post_types', array( 'post' ) ) as $post_type ) {
+		self::register_supported_post_type_feature();
+	}
+
+	/**
+	 * Registers the `activitypub` feature for each supported post type.
+	 *
+	 * Reads the admin-selected post types from the `activitypub_support_post_types`
+	 * option and applies the `activitypub_supported_post_types` filter before calling
+	 * `add_post_type_support()`. The admin UI continues to reflect the stored option,
+	 * not the filtered value.
+	 *
+	 * Note: integrations that read the `activitypub_support_post_types` option directly
+	 * (rather than checking `post_type_supports( $type, 'activitypub' )`) are not
+	 * affected by the filter.
+	 *
+	 * @since unreleased
+	 */
+	public static function register_supported_post_type_feature() {
+		$post_types = \get_option( 'activitypub_support_post_types', array( 'post' ) );
+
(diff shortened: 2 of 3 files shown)


pfefferle:
@kraftbj why not simply use `\add_post_type_support( $your_custom_post_type, 'activitypub' );`???

kraftbj:
Thanks, fair question. `add_post_type_support()` is the right tool when a plugin wants to opt its own CPT in.

With FOSSE bundling AP and AT, I want to expose a single simpler settings UI over both so initial users configure one screen instead of two. That means FOSSE's option has to be the source of truth, including for removals. If a user unchecks `post` in FOSSE, AP needs to stop federating `post`. `add_post_type_support` only covers the add side.

I want to avoid two sources of truth (FOSSE option plus AP's still-editable settings screen), and hesitate to just write `activitypub_support_post_types` from FOSSE for the same reason.

A runtime filter is what I landed on because it matches what we already do for object type (FOSSE hooks `activitypub_post_object_type` to project one option into AP). That said, if there's an existing hook I missed or a cleaner pattern for projecting a wrapper's option into AP, I'd rather use that than add a new filter. Open to ideas.

pfefferle:
I still think it is fine to `add_/remove_post_type_support()` because it is literally what the plugin does, but you could also hook into the `option_activitypub_supported_post_types` filter, which is basically the same as the filter you introduced!?

kraftbj:
Thanks, that makes sense. The stock `option_activitypub_support_post_types` filter gets us there without a new hook, and `add_post_type_support()` is the right tool when a plugin owns its own CPT.

For FOSSE I think we'll skip projecting a parallel option and just read `activitypub_support_post_types` directly — one-way projection into Atmosphere's filter, AP's admin stays the source of truth. Keeps the UI honest and avoids the two-sources-of-truth problem that pushed me to open this in the first place.

Closing in favor of that approach. Appreciate the direction.

PR #3281, opened 2026-05-12
Allow site admins to act as the blog actor via C2S

### LINKED ISSUE #3276: C2S: Blog actor has no working Client-to-Server path
## Summary

The Client-to-Server (C2S) API does not work for the blog actor (`BLOG_USER_ID = 0`). A client cannot authorize as it, cannot obtain a token for it, and cannot pass ownership checks against it. The OAuth token response also advertises `me = <blog actor URI>` in some cases, which misleads clients into believing they can act as the blog actor when they cannot.

## Why it's broken

Three interlocking issues, not just one:

1. **No identity to authorize as.** `wp-login.php?action=activitypub_authorize` (`includes/oauth/class-server.php:268`) requires `is_user_logged_in()`. The blog actor is virtual (`user_id = 0`); nobody can log in as it.
2. **No place to store the token.** `Token::create()` calls `update_user_meta( $user_id, ... )` (`includes/oauth/class-token.php:123`), which silently no-ops for `user_id = 0` since there is no `wp_users` row.
3. **Ownership check is identity-equality.** `verify_owner` (`includes/rest/trait-verification.php:194`) does `get_current_user_id() === (int) $user_id`. The blog actor has no "owner", so equality is the wrong relation.

Each layer assumes a 1:1 mapping between WP user and AP actor. The blog actor breaks that assumption.

The misleading `me` fallback lives at `includes/oauth/class-token.php:147-151`: when the consenting WP user has no user actor (e.g., user actors disabled for them), the token response labels their `me` as the blog actor URI, but the underlying token still belongs to that WP user, and any subsequent `POST /actors/0/outbox` is rejected by `verify_owner`.

## Design constraints

- Must not regress single-user/blog-only sites.
- Should not require giving the blog actor a real `wp_users` row (huge migration; touches every `user_id = 0` branch).
- Must produce an audit trail — "the blog posted X" needs to map back to which human did it.
- Should reuse existing OAuth machinery (RFC 7591/7636, scopes), not invent a parallel auth path.
- Multi-author sites: more than one WP user may legitimately post as the blog.

### A. Delegated authority (recommended, smallest change)

Treat the blog actor as a *resource* a real WP user can act on, gated by capability.

- **Token data:** Add `actor_id` field (default = `user_id`, can be `BLOG_USER_ID` if scope granted).
- **Scope:** New scope (e.g. `blog` or `actor:blog`) the consent screen surfaces only to users with a new `activitypub_post_as_blog` capability (granted by default to admins on blog-enabled sites).
- **Authorize UI:** When the client requests the blog scope, show \"post as &lt;site name&gt;\" alongside \"post as &lt;user name&gt;\".
- **`verify_owner`:** Replace identity equality with \"does this token authorize acting on `\$user_id`?\":
  ```php
  \$token_actor = Token::get_actor( \$request );
  if ( (int) \$token_actor === (int) \$user_id ) { return true; }
  ```
- **`validate_ownership` in outbox controller (`includes/rest/class-outbox-controller.php:553`):** Compare submitted `actor` against `Token::get_actor()`'s URI, not `get_current_user_id()`'s actor.

**Pros:** No DB migration. Existing tokens unaffected. Audit trail intact (token still belongs to a WP user; we just know they were acting *as* the blog).
**Cons:** Token semantics get a new dimension. Two real users could both post as blog — arguably desired.

### B. Per-token actor selection (cleanest long-term)

Generalization of A: every token records an explicit `actor_id`, regardless of whether it is user or blog. Drop the implicit \"actor = WP user\" assumption entirely. Authorize UI becomes a dropdown of actors the consenting user is allowed to post as.

**Pros:** Conceptually correct. Drops in cleanly if/when other actor types need C2S. Multi-author shared blog actor becomes natural.
**Cons:** Bigger blast radius — touches verification, ownership, possibly the proxy controller. More tests.

### C. Promote the blog actor to a real WP user

Kill the virtual blog user; create an actual hidden `wp_users` row, apply role/cap to control who can publish through it.

**Pros:** OAuth, capabilities, post authorship — everything just works.
**Cons:** Migration is the boss fight. Every `BLOG_USER_ID === 0` check, every `'author' => 0` query, every `update_user_meta( 0, ... )` either changes meaning or needs migration. This is a major-version concern, not a fix.

### D. Stopgap: stop advertising `me`

Fix the `me` fallback at `includes/oauth/class-token.php:147-151`. If the consenting user has no user actor and the token is not authorized to act as the blog, return `me = null` or refuse the token. No new functionality, but stops promising what we cannot deliver.

**Pros:** A few lines, correct today.
**Cons:** Does not enable blog C2S — it just stops claiming to.

## Recommendation

1. **Ship D immediately** so we stop misleading clients while the bigger work lands.
2. **Then build A.** It unlocks blog C2S without a migration and slots into the existing OAuth flow:
   - Add `actor_id` to token data (default `= user_id`, backward-compatible).
   - Add blog scope + `activitypub_post_as_blog` capability check.
   - Extend authorize form: when client requests blog scope and user has the cap, surface \"as the site\" option.
   - Centralize \"what actor does this token act as?\" in `Token::get_actor( \$request )`.
   - Replace `verify_owner` equality check and `validate_ownership` actor check with that helper.
   - Tests: `POST /actors/0/outbox` with admin-issued blog-scoped token returns 201; same with user-scoped token returns 403; non-admin requesting blog scope sees no blog option in consent.
3. **Treat B as a future refactor**, not a precondition — A is a strict subset.
4. **Treat C as its own initiative** tied to the broader actor-architecture rethink; do not let it block C2S for the blog actor for another release cycle.

pfefferle:
Fixes #3276

## Proposed changes:

* Add a capability-gated bypass in `verify_owner()`: users with `manage_options` (administrators by default) pass the ownership check for `Actors::BLOG_USER_ID`. The existing identity-equality check could never satisfy the blog actor because it has no `wp_users` row.
* Wrap the capability check in a new `activitypub_user_can_act_as_blog` filter so integrations can broaden (or narrow) the allow-list — for example, granting editors on multi-author sites the right to publish through the site's blog account.

This is the minimal-viable fix from the issue's option A — admins authorize as themselves via the existing OAuth flow, the token is stored against their real `user_id`, and this patch is the only ownership-check change needed. No DB migration, no new scope, no token-schema changes.

Out of scope (separate follow-ups):
- The `me = <blog URI>` advertising issue (option D in the ticket) — once admins can legitimately act as the blog, the response is correct for them; the misleading case for non-admins without user actors is still open.
- Full per-token `actor_id` plumbing (option B).
- Promoting the blog actor to a real `wp_users` row (option C).

### Other information:

## Testing instructions:

* Enable the blog actor: set Actor Mode to "Blog and User" (or "Blog only") under Settings → ActivityPub.
* Register an OAuth client and authorize it while logged in as an administrator.
* `POST /wp-json/activitypub/1.0/actors/0/outbox` with the resulting bearer token and a valid Create activity — request should be accepted (201).
* Repeat the OAuth flow as a non-admin (e.g. author). The token will issue, but `POST /actors/0/outbox` should be rejected (403, `activitypub_forbidden`).
* Add `add_filter( 'activitypub_user_can_act_as_blog', '__return_true' );` in a mu-plugin and retry as the non-admin — request should now be accepted.

### Changelog entry

<details>

#### Significance

- [x] Minor

#### Type

- [x] Added

#### Message

Allow site administrators to post from third-party apps on behalf of the site's blog account.

</details>

### DIFF
--- a/.github/changelog/blog-actor-c2s-admin
+++ b/.github/changelog/blog-actor-c2s-admin
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Allow site administrators to perform Client-to-Server actions on behalf of the site's blog actor.
--- a/.github/changelog/blog-actor-c2s-admin
+++ b/.github/changelog/blog-actor-c2s-admin
@@ -1,4 +1,4 @@
 Significance: minor
 Type: added
 
-Allow site administrators to perform Client-to-Server actions on behalf of the site's blog actor.
+Allow site administrators to post from third-party apps on behalf of the site's blog account.
--- a/.github/changelog/fix-anonymous-blog-actor-ownership
+++ b/.github/changelog/fix-anonymous-blog-actor-ownership
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Prevent the site's follower and following lists from being visible to logged-out visitors when the social graph is set to private.
--- a/.github/changelog/fix-blog-actor-outbox-permalink-privacy
+++ b/.github/changelog/fix-blog-actor-outbox-permalink-privacy
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Prevent private outbox items authored by the site account from being visible to logged-out visitors at their permalink URLs.
--- a/includes/collection/class-outbox.php
+++ b/includes/collection/class-outbox.php
@@ -14,6 +14,7 @@ use Activitypub\Webfinger;
 
 use function Activitypub\add_to_outbox;
 use function Activitypub\object_to_uri;
+use function Activitypub\user_can_act_as_blog;
 
 /**
  * ActivityPub Outbox Collection
@@ -431,9 +432,25 @@ class Outbox {
 			\Activitypub\OAuth\Server::authenticate_oauth( null );
 		}
 
-		// Allow the author to view their own outbox items regardless of visibility.
-		if ( \get_current_user_id() === (int) $outbox_item->post_author ) {
-			return self::get_activity( $outbox_item );
+		/*
+		 * Allow the author to view their own outbox items regardless of visibility.
+		 * The `is_user_logged_in()` guard prevents anonymous visitors from matching
+		 * the blog actor's items (where both `get_current_user_id()` and `post_author`
+		 * are `0`), which would otherwise expose private activities at their permalink.
+		 *
+		 * Users authorized to act as the blog actor are treated as the author of
+		 * blog-actor items so they can read the same private outbox they can post to.
+		 */
+		if ( \is_user_logged_in() ) {
+			$author = (int) $outbox_item->post_author;
+
+			if ( \get_current_user_id() === $author ) {
+				return self::get_activity( $outbox_item );
+			}
+
+			if ( Actors::BLOG_USER_ID === $author && user_can_act_as_blog() ) {
+				return self::get_activity( $outbox_item );
+			}
 		}
 
 		// Check if Outbox Activity is public.
--- a/includes/functions-user.php
+++ b/includes/functions-user.php
@@ -158,6 +158,11 @@ function user_can_act_as_blog() {
 	 * Defaults to true for users with the `manage_options` capability (administrators).
 	 * Filter to broaden the allow-list, for example to editors on multi-author sites.
 	 *
+	 * Security note: returning a static `true` (e.g. via `__return_true`) grants
+	 * EVERY authenticated user the right to post as, read private outbox items of,
+	 * and view stats for the blog actor. Always inspect the current user inside
+	 * the callback (`current_user_can()`, role, allowlist) before returning `true`.
+	 *
 	 * @since unreleased
 	 *
 	 * @param bool $can_act_as_blog Whether the current user can act as the blog actor.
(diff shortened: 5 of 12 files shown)


pfefferle:
Addressing the Copilot review pass:

1. **`@since unreleased`** — This is intentional. Per `AGENTS.md`, all new `@since` / `@deprecated` tags must use the literal string `'unreleased'` so the release script can replace it with the next plugin version when the release is cut. Keeping as-is.

2. **Anonymous `user_id=0` bypass on identity-equality** — Real pre-existing bug, fixed in 849dab415. Trace:
   - `class-followers-controller.php` and `class-following-controller.php` use `verify_signature` as their permission callback (not `verify_authentication`).
   - For sites without authorized fetch, anonymous GETs satisfy that callback.
   - The handler then calls `$this->show_social_graph( $request )`, which falls back to `verify_owner()`.
   - `\get_current_user_id() === (int) $user_id` evaluates `0 === 0` for anonymous requests against the blog actor, returning `true` and exposing the social graph even when the admin set it to private.
   
   Added an explicit `\is_user_logged_in()` guard in `verify_owner()` so anonymous callers fall through to the 403, plus a regression test (`test_verify_owner_anonymous_cannot_act_as_blog`) and a `patch` / `fixed` changelog entry covering this user-facing fix.

PR #3293, opened 2026-05-13
Move tombstone storage off the autoloaded options row

### LINKED ISSUE #3290: Move ActivityPub tombstone URL storage out of options to prevent alloptions criticals
`activitypub_tombstone_urls` can grow into a near-1 MB `wp_options` row. On [WordPress.com](<http://WordPress.com>), every option row is loaded into alloptions, so this can trigger `a8c-alloptions` critical alerts even if the option is stored with normal WordPress defaults.

Evidence:

* One affected site has 8,172 tombstone entries totaling 955,643 B.
* The same site’s serialized alloptions size was 983 KB during the seven-day review.
* Current ActivityPub code stores the tombstone registry in a single option.
* `Tombstone::bury()` appends normalized URLs and de-duplicates the array.
* `Tombstone::exists_local()` checks membership by loading the same option.
* Delete handling can add multiple URLs for one object through `maybe_bury()`.

Related prior work:

* `CMF-429` fixed a tombstone behavior bug where SEO plugin saves could unbury soft-deleted posts. That confirms tombstone registry behavior is product-significant, but it does not address storage growth or alloptions pressure.
* Prior ActivityPub alloptions work reduced stored data shape in another path by using lighter IDs and locks.

Implementation direction:

* Add a `Tombstones` collection/storage layer in ActivityPub instead of storing every tombstone in one option.
* Use WordPress-native storage that is not alloptions-loaded. One concrete approach: a private `ap_tombstone` post type, with the normalized URL hash as the lookup key and the normalized URL stored as metadata for debugging/migration.
* Update `Tombstone::exists_local()`, `Tombstone::bury()`, and `Tombstone::remove()` to read/write through that storage layer.
* Add a migration path that processes the existing `activitypub_tombstone_urls` option in chunks, writes tombstone records, and deletes the legacy option after verification.
* Decide whether tombstones are permanent or whether the product can support an explicit retention window. If a retention window is acceptable, document it and add scheduled cleanup.

Acceptance criteria:

* Sites with many deletes cannot grow `activitypub_tombstone_urls` near the 1 MB alloptions critical threshold.
* Local tombstone checks still work for retained or migrated entries.
* Soft-deleted posts continue to serve tombstone responses where expected.
* Any cleanup or migration path preserves product behavior or documents the retained window clearly.

Notes:

* This is ready for a bug/investigation issue with an implementation direction.
* It still needs the owning team to decide storage semantics: permanent registry or explicit retention window.
* Do not silently delete or trim the live option without Fediverse/Creators signoff.

pfefferle:
Fixes #3290

## Proposed changes:

* Move the local tombstone registry from the autoloaded `activitypub_tombstone_urls` option (which can grow toward the 1 MB alloptions critical threshold on busy sites) onto a private `ap_tombstone` custom post type. Each tombstone is a single post: `post_name` is `md5( normalize_url( \$url ) )` for an indexed O(1) slug lookup, `guid` keeps the original URL for human-readable debugging.
* Keep the public API (`Tombstone::bury / remove / exists_local`) unchanged, so all existing callers (`class-router.php`, the delete/create handlers, `class-followers.php`, the admin UI, etc.) need no changes.
* Add a chunked migration cron that drains the legacy option into the CPT and then deletes the option. The migration is lock-guarded against concurrent workers, with stale-lock recovery so a PHP fatal mid-chunk does not permanently stall it. `Tombstone::exists_local()` falls back to the legacy option during the transition so no tombstone disappears while the migration is in flight.
* Add a daily purge with a configurable retention window. Default is 90 days, override via the `activitypub_tombstone_retention_days` filter (set to 0 or negative to disable).

### Hardening pass

A second pass tightened three real issues:

- **Migration halts when a batch makes zero progress.** Previously a persistently-failing `wp_insert_post()` would keep the scheduler in an unbounded retry loop. Failed URLs stay in the legacy option, which `exists_local()` still consults, so no tombstone is lost; an admin can re-trigger the migration after fixing the underlying cause.
- **Migration writes the legacy option with `autoload => false`.** Relief on `alloptions` kicks in on the first batch rather than waiting for the option to be fully drained.
- **Uninstall now calls a new `Tombstone::delete_all()`** so the CPT rows do not leak past uninstall.

### Lookups: keyed by `post_name`

The MD5 of the normalized URL is unique per URL and `bury()`'s existence check makes duplicate calls a no-op, so `find_post_ids_by_url()` is a single exact match by `post_name` with `LIMIT 1`. Used by `bury` (existence check), `remove`, and `exists_local`.

### `bury()` failure surface

`bury()` fires `activitypub_tombstone_bury_failed` when `wp_insert_post()` returns a `WP_Error` or 0. Useful as a monitoring hook — the downstream consequence of a silent failure is benign (the request path serves the normal not-found response instead of a 410).

### Other information:

Tests cover:

- `test_delete_all` — bury three URLs + populate legacy option → `delete_all()` → assert everything gone.
- `test_purge_deletes_old_tombstones` / `test_purge_disabled_when_retention_zero` — exercise the retention sweep and the disable filter.
- `test_migrate_tombstones_to_cpt_halts_on_no_progress` — forces every insert to fail via `wp_insert_post_empty_content`, asserts the return is `null` (scheduler halts) and the legacy option still holds the URLs.
- Plus the existing `test_bury`, `test_remove`, `test_bury_duplicate_urls`, `test_exists_local_*`, and migration suite — all still passing.

## Testing instructions:

* Activate the plugin on a site that already has entries in `activitypub_tombstone_urls`.
* Wait for WP-Cron to fire `activitypub_tombstone_migrate` (or run `wp cron event run activitypub_tombstone_migrate`). The migration drains in batches of 500 URLs, with a default pause of 30 seconds between batches between (tunable via the `activitypub_scheduler_async_batch_pause` filter), until the legacy option is drained and deleted.
* Confirm that previously soft-deleted local posts still return 410 / `Tombstone` responses both during and after migration.
* Wait 24h (or set `activitypub_tombstone_retention_days` to a small value and run `wp cron event run activitypub_tombstone_purge`) to confirm expired tombstones are purged from the CPT.
* `wp post list --post_type=ap_tombstone --format=count` to inspect how many tombstones live in the new store.

### Changelog entry

Already included in `.github/changelog/fix-tombstone-storage`.


pfefferle:
Pushed two follow-ups since the last round:

**[ef9e5dc](https://github.com/Automattic/wordpress-activitypub/pull/3293/commits/ef9e5dc02)** — Match the `http://`-prefixed `guid` stored by `esc_url()`.

`normalize_url()` strips the scheme intentionally (so `http`, `https`, and `www.` variants share one tombstone), but `wp_insert_post()` runs `esc_url()` on `guid` and prefixes the schemeless value with `http://` before persisting it. The previous `find_post_ids_by_url()` queried for the bare canonical string and never matched anything stored — that's why `Test_Tombstone::test_remove` and the `Test_Create::test_maybe_unbury_*` suite were failing. Now the lookup uses `'http://' . $normalized` so it matches the actual on-disk form. Updated the two fixture-rows in the new self-heal tests likewise (they go through `$wpdb->insert`, which doesn't apply `esc_url`).

Considered going further and stripping the host as well (for domain-rename resilience), but `normalize_url()` is also used in `Actors::get_id_by_various()` to compare a URI against `site_url()` — dropping the host would collapse every URI to root-path-only and resolve every actor to the Blog actor. Out of scope.

**[027750a](https://github.com/Automattic/wordpress-activitypub/pull/3293/commits/027750aea)** — Address Copilot review feedback.

- `exists_local()` now matches by `guid` via the shared `find_post_ids_by_url()` helper instead of `get_page_by_path()`. Catches orphan rows whose `post_name` drifted to `<hash>-2` from a prior concurrent insert (which the slug-only lookup would silently miss), and skips the full `WP_Post` hydration on every router hit.
- `migrate_tombstones_to_cpt()` now writes the shrinking legacy option with `autoload => false`. The whole point of the migration is to take the option off `alloptions` — flip the flag immediately rather than wait for the option to be fully drained.

Two earlier Copilot comments (`remove()` legacy cleanup, `purge()` reschedule docblock) were on the commit before the hardening pass and are stale — current code already matches what those comments asked for.

Full PHPUnit suite: 2555 tests, 6225 assertions, all tombstone-touched suites pass. The one outlier (`Test_Migration::test_add_default_extra_field`) is the pre-existing failure called out in the original review — unrelated to tombstones.

PR #3294, opened 2026-05-13
Prevent duplicate Powered by extra fields on migration rerun

pfefferle:
Fixes #3280

## Proposed changes:

* `Migration::add_default_extra_field()` previously did an unconditional `wp_insert_post()` for every user with the `activitypub` capability plus one for the blog actor. If the migration ever ran more than once — concurrent requests during initial activation, plugin re-install after uninstall, or any other path that resets `activitypub_db_version` — duplicate "Powered by WordPress" entries accumulated. The user-reported symptom (a field reappearing after manual deletion) traces to surviving duplicates being re-rendered on the next admin request.
* Guard each insert with a small `has_powered_by_extra_field( $author_id, $post_type )` helper. If a matching entry already exists, skip the insert.
* Added a new test asserting `add_default_extra_field` is idempotent (call twice, get one entry).

### Other information:

- [x] Have you written new tests for your changes, if applicable?

## Testing instructions:

* On a fresh install of `trunk`, plugin activation creates one "Powered by WordPress" extra field per user + one for the blog actor.
* Delete `activitypub_db_version` from the options table (simulating a re-install / racey activation).
* Reload an admin page (re-triggers `Migration::maybe_migrate()` → `add_default_settings()`).
* Confirm no new "Powered by WordPress" entries are added (compare counts in `Profiles → Extra Fields` and `Settings → Blog Profile`).
* Repeat the reload — still no duplicates.

### Changelog entry

Already included in `.github/changelog/fix-powered-by-duplicate`.

Issue #3280, opened 2026-05-12
**Bug report: "Powered by WordPress" extra field duplicates randomly on post edits**


vincentbreton-auteur:
## Description

The ActivityPub plugin automatically creates a "Powered by WordPress" entry in the user's extra fields (stored as `ap_extra_field` post type). This field is not created by the user and cannot be permanently removed: after deleting it manually, it reappears - seemingly triggered by post edits, though the behavior is not consistent.

### Steps to reproduce

The ActivityPub plugin automatically creates a "Powered by WordPress" entry in the user's extra fields (stored as `ap_extra_field` post type). This field is not created by the user and cannot be permanently removed: after deleting it manually, it reappears - seemingly triggered by post edits, though the behavior is not consistent.

### Site owner impact

Fewer than 20% of the total website/platform users

### Severity

Moderate

pfefferle:
Hey @vincentbreton-auteur 👋

do you use the trunk version of the plugin or simply the one that can be installed from the WordPress.org repo?

Do you have tools that cleanup the dp or do you drop `activitypub_db_version` for test purposes?

vincentbreton-auteur:
I use the standard version from the WordPress.org repository. I don't use any database cleanup tools and I don't drop activitypub_db_version . I'm just running the plugin in a normal production setup, without any testing procedures

PR #3295, opened 2026-05-13
Bail from maybe_migrate when concurrent worker holds the lock

pfefferle:
## Proposed changes:

* `Migration::lock()` returns `true` on a successful claim and the existing lock timestamp otherwise (see its docblock and the existing `test_lock_acquire_new` / `test_lock_get_existing` tests). The contract has always been there; `Migration::maybe_migrate()` just wasn't using it — it called `self::lock()` and threw the return value away. Two concurrent requests could therefore both pass `is_locked()`, both call `lock()`, and only one would actually insert the lock row — but both would proceed into the migration body. Every migration callback then runs twice.
* Fix: bail when `lock()` returns anything other than `true`. The `is_locked()` fast-path stays as an early return + stale-lock recovery; the new check closes the time-of-check / time-of-use race that lives inside it.
* This is the deeper companion to #3294, which made the "Powered by WordPress" default-field provisioning idempotent at the per-author level. Together, the two fixes mean:
  * #3294 prevents duplicates from any cause, even if migrations rerun (defense in depth).
  * This PR prevents migrations from rerunning in the first place when a concurrent worker is already in flight.

### Other information:

## Testing instructions:

* Run the new test `test_maybe_migrate_bails_when_lock_acquired_by_concurrent_worker`. It pre-inserts a lock row, hides it from `get_option` via the `option_*` filter (simulating the race window where `is_locked()` failed to see it), and asserts that `maybe_migrate()` does not provision any defaults and does not set `activitypub_db_version`.
* Without the fix, the same test fails because `maybe_migrate` ignores `lock()`'s return value and runs the full migration body.

### Changelog entry

Already included in `.github/changelog/fix-migration-lock-race`.

### DIFF
--- a/.github/changelog/fix-migration-lock-race
+++ b/.github/changelog/fix-migration-lock-race
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Prevent migrations from running twice when concurrent requests race through the upgrade check.
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -117,7 +117,14 @@ class Migration {
 			return;
 		}
 
-		self::lock();
+		/*
+		 * Atomic claim. lock() returns true only when this caller's INSERT IGNORE
+		 * actually inserted the row. If a concurrent worker won the race (lock
+		 * returns the existing timestamp), bail so we don't double-run migrations.
+		 */
+		if ( true !== self::lock() ) {
+			return;
+		}
 
 		$version_from_db = self::get_version();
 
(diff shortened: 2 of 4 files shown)


pfefferle:
Closing — see #3294's closing comment. The migration-lock race is a real concern in theory, but it isn't what produces the user-reported bug in #3280. The actual bug is the missing `activitypub_default_extra_fields` flag in `Migration::add_default_extra_field`. Replacing with a single focused PR.

PR #3303, opened 2026-05-15
Make C2S Undo of Follow work end-to-end

pfefferle:
Two tightly related fixes that together make POSTing an `Undo` of a `Follow` to the outbox API actually work the way clients expect.

## Proposed changes

1. **`Following::unfollow()` returns the Undo outbox ID instead of the WP_Post of the remote actor** (`int|WP_Error`). The controller's existing `is_int( $result ) && $result > 0` branch then handles the response correctly. Previously the controller looked up the Undo by the remote actor's local URL, found nothing, and returned 500.
2. **`Undo::handle_undo` accepts an inlined Follow without an `id`.** ActivityStreams 2.0 doesn't require an `id` on every embedded activity; spec-valid clients (e.g. Reader Fediverse, Mastodon's own admin UI when unfollow happens before the original Follow id is cached) inline the Follow and identify it by `(type, actor, object)`. Mastodon and others fall back to matching on the inner Follow's target — see `mastodon/app/lib/activitypub/activity/undo.rb#undo_follow`. Mirror that fallback here, otherwise the handler bails and the controller writes a raw outbox entry without removing local following state.

## Why

A client POSTs `{ \"type\": \"Undo\", \"object\": { \"type\": \"Follow\", \"actor\": \"<me>\", \"object\": \"<them>\" } }` to `/wp-json/activitypub/1.0/actors/{user_id}/outbox`. With this branch:

- On `trunk`, both shapes (with and without an embedded `id`) fail. The `id` shape returns 500 (`\"Failed to add activity to outbox\"`). The id-less shape returns 201, but the local following meta is **not** removed and no `Outbox::undo()` is called — the Undo just federates outward into the void.
- On this branch, both shapes return 201, create a real Undo outbox entry tied to the original Follow, and remove the local following meta. The C2S response activity round-trips cleanly.

## Other information

Three new tests + one existing test updated:

- `Test_Undo::test_handle_undo_follow_returns_outbox_id` — unit test for the handler returning the new Undo outbox ID (covers the int return contract).
- `Test_Undo::test_handle_undo_follow_without_id` — unit test for the id-less embedded Follow fallback.
- `Test_Outbox_Controller::test_c2s_undo_follow_returns_201` — end-to-end REST test posting an Undo→Follow with an embedded `id` to the outbox endpoint.
- `Test_Outbox_Controller::test_c2s_undo_follow_without_id_returns_201` — end-to-end REST test for the id-less shape.
- `Test_Following::test_unfollow_removes_user` — updated to assert the new return contract and to publish the Follow outbox (mirrors normal post-Accept state, since `Following::unfollow`'s `WP_Query` defaults to `publish`-only).

## Testing instructions

1. Pull this branch and run the PHPUnit suite — `Test_Following`, `Test_Undo`, and `Test_Outbox_Controller` should all pass on PHP 7.4 / 8.3 / 8.5.
2. To reproduce manually on `trunk` vs. this branch:
   - As an `activitypub`-capable user with at least one Follow in their outbox, POST one of:
     - `{ \"type\": \"Undo\", \"actor\": \"<me>\", \"object\": { \"id\": \"<follow-guid>\", \"type\": \"Follow\", \"actor\": \"<me>\", \"object\": \"<them>\" } }` (id present)
     - `{ \"type\": \"Undo\", \"actor\": \"<me>\", \"object\": { \"type\": \"Follow\", \"actor\": \"<me>\", \"object\": \"<them>\" } }` (id absent)
   - `trunk`: first form returns 500, second returns 201 but local meta unchanged.
   - This branch: both return 201, local following meta removed, Undo outbox entry created.

### Changelog entry

Already added at `.github/changelog/fix-unfollow-return-outbox-id`.


pfefferle:
Pushed [8ad434dcb](https://github.com/Automattic/wordpress-activitypub/pull/3303/commits/8ad434dcb) addressing all three Copilot comments from the latest review:

1. **Id-less Follow fallback ignored the embedded `actor`** (`class-undo.php`) — real security gap. Now validates that `embedded_actor === Actors::get_by_id( $user_id )->get_id()` before calling `unfollow()`. A mismatched actor returns `WP_Error('activitypub_forbidden', 403)` instead of mutating local follow state. Added a regression test for the rejection path.
2. **`Following::unfollow()` couldn't find pending Follows** (`class-following.php`) — meta_query defaulted to publish-only, so a user POSTing Follow then Undo→Follow before the remote Accept arrived found no Follow outbox and silently fell through. The query now matches `[ 'publish', 'pending' ]`.
3. **Test hardcoded actor URL** (`class-test-undo.php`) — replaced `'https://example.org/?author=...'` with `Actors::get_by_id( $this->user_id )->get_id()`, so it now exercises the real actor-validation logic added in (1).

Out of scope and explicitly deferred: extending the id-less fallback to `Block` activities. Captured in our review discussion; happy to ship as a follow-up if useful.

PR #3327, opened 2026-05-19
Add blurhash term to JSON-LD context

kraftbj:
## Proposed changes:

* Adds one entry to the inner array of `Activity::JSON_LD_CONTEXT`:

  ```php
  'blurhash' => 'toot:blurhash',
  ```

  This maps the `blurhash` term to Mastodon's `toot:` namespace (`http://joinmastodon.org/ns#`) so outbound activities that include a `blurhash` property on image attachments serialize as strictly correct JSON-LD.

### Why

Downstream emitters are starting to ship `attachment[].blurhash` on outbound image activities so receivers (Mastodon, Pixelfed) can paint a colored-blur placeholder while images load. Real-world consumers read the property by name regardless of context, so it works today without this change — but the JSON-LD path wants the term mapped.

This patch is pure parity with Mastodon: Mastodon coined the property under the `toot:` namespace and emits exactly this mapping in its own outbound context. No new dependencies, no behavior change for code that doesn't already populate the property.

Use case driving the request: https://github.com/Automattic/fosse/pull/159 (FOSSE adds `blurhash` to outbound image attachments).

### References

* Mastodon vocab namespace: `http://joinmastodon.org/ns#`
* Blurhash spec: https://blurha.sh

### Other information:

No tests added — the change is a single entry in a class constant. The constant is already covered indirectly by the collections controller tests (`Base_Object::JSON_LD_CONTEXT` shape assertions), and adding a term doesn't change any existing assertions because those test a different class's constant.

## Testing instructions:

* Inspect an outbound Activity payload (e.g. via the Outbox or by sending a Note with an image attachment to a remote actor) and confirm `@context[1].blurhash` is present and equals `"toot:blurhash"`.
* Confirm the rest of the context (`https://www.w3.org/ns/activitystreams`, `toot`, `QuoteRequest`) is unchanged.
* Optional: emit an attachment with `blurhash` set on it and verify a JSON-LD validator no longer flags the property as unmapped.

### Changelog entry

A changelog entry is included in this PR at `.github/changelog/add-jsonld-blurhash-term` (significance: patch, type: changed). The "Skip Changelog" label is not needed.


kraftbj:
Fix-up commit (368552ea) — second adversarial review pointed out the term was on `Activity::JSON_LD_CONTEXT`, but production envelopes serialize the activity's object and `Activity::get_json_ld_context()` resolves via the object's class constant (`Base_Object::JSON_LD_CONTEXT`) whenever an object is present. Applied the same `toot` namespace and `blurhash` => `toot:blurhash` mapping there too, and kept the original entry on `Activity` as belt-and-suspenders for the no-object fallback path.

pfefferle (approved):

PR #3362, opened 2026-06-02
Hide follower/following counts when the social graph is private

pfefferle:
## Proposed changes:

The followers and following REST collections emitted `totalItems` (the follower/following count) unconditionally, while only the `orderedItems` list was gated behind `show_social_graph()`. So when a user hid their social graph, a non-owner could still read the count via `totalItems`.

This gates `totalItems` on the same `show_social_graph()` check (emitting `0` when hidden) and reuses the result for `orderedItems`. Because `show_social_graph()` returns true for the resource owner, owners still see the real count. The deprecated `activitypub_rest_following` filter path is gated too, so it cannot re-leak the count for sites still using that hook.

### Other information:

Regression tests for both endpoints: a hidden graph returns `totalItems: 0` and no list to a non-owner; and a dedicated test confirms the deprecated `activitypub_rest_following` filter cannot re-leak the count. Verified by negative control — both fail without the fix.

## Testing instructions:

* Enable "hide social graph" for an actor (or the blog actor via the site setting).
* As an unauthenticated request, fetch `?rest_route=/activitypub/1.0/actors/{id}/followers` (and `/following`).
* Confirm `totalItems` is `0` and no `orderedItems` are present.
* Confirm the owner (logged in, viewing their own) still sees the real count and list.

### Changelog entry

Changelog entry added in `.github/changelog/fix-hidden-social-graph-total`.

### DIFF
--- a/.github/changelog/fix-hidden-social-graph-total
+++ b/.github/changelog/fix-hidden-social-graph-total
@@ -1,0 +1,4 @@
+Significance: patch
+Type: security
+
+Hide the follower and following counts when the social graph is set to private, not just the lists.
--- a/includes/rest/class-followers-controller.php
+++ b/includes/rest/class-followers-controller.php
@@ -191,11 +191,17 @@ class Followers_Controller extends Actors_Controller {
 
 		$data = Followers::query( $user_id, $per_page, $page, array( 'order' => \ucwords( $order ) ) );
 
+		/*
+		 * When the social graph is hidden, expose neither the list nor the count: a
+		 * non-owner would otherwise still learn the follower total via totalItems.
+		 */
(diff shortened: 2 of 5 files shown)


pfefferle:
Closing after checking the Fediverse norm. Mainline Mastodon intentionally keeps `totalItems` on a hidden followers/following collection (`restrict_fields_to` returns `id`, `type`, `total_items`) and only hides the identities — it 403s `?page=` requests and omits the list. The plugin already hides the list when the social graph is hidden, so the count being visible matches the ecosystem; zeroing `totalItems` would diverge from what every other server exposes. Hiding the count is a non-standard glitch-soc fork feature, not mainline behavior. Closing as by-design.

PR #3372, opened 2026-06-05
Fix fatal error in Stream connector on new follower

pfefferle:
## Proposed changes:

Fixes a **fatal error on every inbound Follow** for any site running the [Stream](https://wordpress.org/plugins/stream/) plugin.

Since 7.5.0, the `activitypub_handled_follow` hook passes its user IDs as an array (`int[]`). The Stream integration's `callback_activitypub_handled_follow()` still treated that argument as a scalar and forwarded the array straight into `WP_Stream\Connector::log()`'s `$user_id` parameter. Stream then built a `WP_User` from the array:

```
PHP Fatal error: Uncaught TypeError: trim(): Argument #1 ($string) must be of type string, array given
  …/class-wp-user.php(160): WP_User::get_data_by('login', Array)
  …/stream/classes/class-log.php(78): WP_User->__construct(0)
  …/activitypub/integration/stream/class-connector.php(122): WP_Stream\Connector->log(…, Array)
  …/activitypub/includes/handler/class-follow.php(83): do_action('activitypub_handled_follow', …)
```

The fix normalizes the user IDs to a single integer in the callback (`\is_array( $user_ids ) ? \reset( $user_ids ) : $user_ids`), mirroring the pattern already used by every other consumer of this hook (`Follow::queue_accept`, `Mailer::new_follower`, `Event_Stream::signal_inbox`). I audited all consumers of the `(array) $user_ids` hooks — the Stream follow callback was the only one mishandling the array.

### Other information:

- [x] Have you written new tests for your changes, if applicable? _(Updated the Stream connector tests to pass the array the hook actually emits; they fail without the fix.)_

## Testing instructions:

* With the Stream plugin active, have a remote account follow your site (or fire `do_action( 'activitypub_handled_follow', $activity, array( 1 ), true, $remote_actor )`).
* Before this change: a fatal `TypeError` in `class-wp-user.php`. After: the follow is logged to Stream as "New Follower" with no error.
* `npm run env-test -- --filter='Follow|Stream'` → all green.

### Changelog entry

A changelog entry is already included in this PR (`.github/changelog/fix-stream-follower-fatal`, patch / fixed), so the automatic entry box is left unchecked.

### DIFF
--- a/.github/changelog/fix-stream-follower-fatal
+++ b/.github/changelog/fix-stream-follower-fatal
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix a fatal error when receiving a new follower while the Stream plugin is active.
--- a/integration/stream/class-connector.php
+++ b/integration/stream/class-connector.php
@@ -111,14 +111,14 @@ class Connector extends \WP_Stream\Connector {
 	/**
 	 * Callback for activitypub_handled_follow.
 	 *
-	 * @param array         $activity The ActivityPub activity data.
-	 * @param int|int[]     $user_ids The local user ID(s) of the followed actor(s).
-	 * @param mixed         $state    Status or WP_Error object indicating the result of the follow handling.
-	 * @param \WP_Post|null $context  The WP_Post object representing the remote actor/follower.
+	 * @param array              $activity The ActivityPub activity data.
+	 * @param int|int[]          $user_ids The local user ID(s) of the followed actor(s).
+	 * @param bool               $success  Whether the follow was handled successfully.
+	 * @param \WP_Post|\WP_Error $context  The remote actor/follower, or WP_Error on failure.
 	 */
-	public function callback_activitypub_handled_follow( $activity, $user_ids, $state, $context ) {
-		// The hook passes an array of user IDs; Stream's log() expects a single integer ID.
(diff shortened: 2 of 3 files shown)


pfefferle:
This likely resolves the user report at https://wordpress.org/support/topic/follow-activity-reverts-from-unfollow-to-confirm-request/.

**Why the symptom matches:** the `TypeError` here is uncaught, so the inbox `POST` returns **HTTP 500**. The remote server never receives a clean response for the Follow, so the Accept handshake never completes and the follow button reverts from "Unfollow" back to "Confirm request." Production logs from an affected (Stream-active) site show this exact fatal firing on `activitypub_handled_follow`.

**Caveat:** this only explains the report if the affected site has the [Stream](https://wordpress.org/plugins/stream/) plugin active — the fatal originates in the Stream connector. Sites without Stream would need a different explanation.

pfefferle:
Thanks @copilot — both addressed in 260fa27:

1. **Docblock** — corrected: the third argument is the boolean `$success` flag, and `$context` is `\WP_Post|\WP_Error` (the callback already guards with `is_wp_error()`), not `\WP_Post|null`.
2. **Integer coercion** — the resolved user ID is now cast to `int`, so Stream's `log()` (which constructs a `WP_User`) always receives a scalar even if the user IDs array is unexpectedly empty.

CI is green across PHP 7.4 / 8.3 / 8.5.

PR #3374, opened 2026-06-08
Federate comments only when the parent post is federated

pfefferle:
## Proposed changes:

Comments are now federated only when their parent post is federated.

Previously, `Comment::should_be_federated()` decided whether to federate a comment based on the comment author alone, without considering the parent post. This change adds a parent-post check so a brand-new comment is federated only when the post it belongs to is itself federated.

- Adds `is_post_federated()` in `includes/functions-post.php`: a post is federated when its stored federation state is `federated` **and** it is still publicly queryable (`is_post_publicly_queryable()` — post type enabled for ActivityPub, public status, not password-protected, allowed content visibility, and respecting the `activitypub_is_post_publicly_queryable` filter).
- `Comment::should_be_federated()` now gates brand-new comments on `is_post_federated()`. Comments that were already sent (`was_sent()`) are unaffected, so their `Update` and `Delete` activities continue to federate and keep remote copies in sync.

Combining the stored state with the live queryability check means a `federated` state left over from a later visibility or status change does not cause a comment to be federated.

### Other information:

## Testing instructions:

* Create a post that is federated, add an approved comment from an ActivityPub-enabled user, and confirm a `Create` is added to the outbox as before.
* Set a post to a private status (or local content visibility), add an approved comment from an ActivityPub-enabled user, and confirm no comment activity is added to the outbox.
* Confirm an already-federated comment still produces `Update`/`Delete` activities after its parent post changes status.
* `npm run env-test -- --filter='Test_Comment|Test_Functions_Post'` — all green.

### Changelog entry

A changelog entry is included in this PR (`.github/changelog/fix-comment-federation-parent-state`), so the automatic entry box is left unchecked.

### DIFF
--- a/.github/changelog/fix-comment-federation-parent-state
+++ b/.github/changelog/fix-comment-federation-parent-state
@@ -1,0 +1,4 @@
+Significance: minor
+Type: security
+
+Only share comment replies in the Fediverse when the post they belong to is itself federated, so replies on private or non-federated posts stay private.
--- a/includes/class-comment.php
+++ b/includes/class-comment.php
@@ -299,6 +299,17 @@ class Comment {
 			return false;
 		}
 
+		/*
+		 * Do not federate brand-new comments on a post that is not federated itself
+		 * (e.g. a private post, a post switched to local visibility, or a non-ActivityPub
+		 * post type). This prevents leaking replies on content the post type's read rules
+		 * would otherwise protect. Comments that were already sent are allowed through so
+		 * their Update and Delete activities can still federate (and tear down remote copies).
+		 */
+		if ( ! self::was_sent( $comment ) && ! is_post_federated( $comment->comment_post_ID ) ) {
+			return false;
+		}
+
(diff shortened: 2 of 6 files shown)


pfefferle:
Thanks @copilot — both addressed in 506e6744e:

1. **Data-provider null static** — `self::$comment_post_ID` is now initialized to `0`. PHPUnit evaluates data providers before `set_up()`, so a `null` static made `isset()` false and the substitution in `test_no_activity_scheduled()` silently skipped, building comments with an invalid post ID. Initializing to `0` keeps the provider output non-null so the substitution picks up the real federated post assigned in `set_up()`.
2. **Filter cleanup on failure** — the `activitypub_is_post_publicly_queryable` filter add/assert/remove is now wrapped in `try/finally`, so the filter cannot leak into subsequent tests if the assertion fails.

CI remains green across PHP 7.4 / 8.3 / 8.5.

PR #3385, opened 2026-06-10
Backfill missing inbox actor from HTTP signature keyId

pfefferle:
## The problem

Incoming `FeatureRequest` activities (FEP-7aa9, "canFeature") from Mastodon are silently rejected. They never get stored in the inbox, and the `Accept`/`Reject` we are supposed to send back is never queued. From the curator's side it looks like our server simply ignored the request.

The root cause is an addressing mismatch. Mastodon's `FeatureRequestSerializer` emits only `id`, `type`, `object`, and `instrument`, and signs the delivery with the curator's key. It does **not** put an `actor` in the activity body, the requesting actor is conveyed solely through the HTTP signature keyId. This is intentional per FEP-7aa9, and FeatureRequest is the only request type Mastodon serializes this way (QuoteRequest, for example, includes `actor`).

Our inbox routes declare `actor` as a **required** REST parameter, and `Feature_Request::validate_object()` also requires it. WordPress validates required parameters in `WP_REST_Server::dispatch()` *before* the `permission_callback` (where signature verification, and therefore keyId resolution, happens). So the request is rejected with `400 rest_missing_callback_param` before it ever reaches storage, the activity dispatch, or the handler. No inbox item, no Accept.

## Proposed changes:

* Add a `rest_pre_dispatch` filter, `Server::maybe_add_actor_from_signature()`, scoped to actor-less `FeatureRequest` POSTs on inbox endpoints. It reads the keyId from the signature, derives the actor URI (fragment stripped, matching the convention `Remote_Actors::get_public_key()` already uses), and adds it to the request.
* The actor is injected with `set_param()`, **not** by rewriting the request body. Inbox POSTs read JSON parameters first (`Server::request_parameter_order()`), so the value is visible to both required-parameter validation and the handler via `get_json_params()`, while the raw body stays byte-identical.
* `rest_pre_dispatch` is the only hook that fires before required-parameter validation, so by the time `actor` is checked it is present. No changes to route args, validators, or the handler.
* Extract the keyId parsing into a shared `Signature::get_key_id()` and reuse it in `verify_key_id()` (which previously parsed the same headers inline).

### Why the body is not rewritten

An earlier revision rewrote the request body to inject the actor. That broke HTTP Signature verification: `Http_Signature_Draft::verify_content_digest()` recomputes `base64( sha256( body ) )` and compares it to the signed `Digest` header, so any change to the body (which Mastodon always digests) turns into a `401 digest_mismatch`. Injecting via `set_param()` leaves the body untouched, so the digest still verifies. A unit test asserts the raw body is unchanged after the filter runs.

### Why this is safe

The injected actor is not trusted on its own. After injection, the request still goes through `Signature::verify_http_signature()` (proving the sender holds the private key for that keyId) and `verify_key_id()` (binding key host to actor host). Because the actor is derived *from* the keyId, the host check is satisfied by construction, and an attacker cannot inject a different actor than the key they actually control.

Crucially, the keyId is the one the request is *actually verified against*. `Signature::get_key_id()` mirrors `verify_http_signature()`'s selection: it takes the keyId from `Signature-Input` when present (RFC 9421, which the verifier prefers) rather than a possibly-spoofed draft `Signature` header, and it refuses to backfill when the signature carries more than one keyId, since the verifier accepts whichever label validates and the verifying key cannot be known in advance. That closes the gap where an actor could be injected from a key that did not sign the request.

The change is purely additive and narrowly scoped: only actor-less FeatureRequests are affected, and a well-formed activity that already carries an `actor` is left untouched.

### Other information:

Unit tests cover the filter in isolation (backfill from both signature header formats, raw-body-unchanged/digest-preserved, existing-actor passthrough, non-FeatureRequest type, non-inbox route, missing signature, non-POST, short-circuit), the shared `Signature::get_key_id()` helper, and an integration test that drives a signed actor-less `FeatureRequest` through the full REST stack and asserts the inbox item is stored and the `Accept` is queued, addressed to the keyId-derived curator.

## Testing instructions:

* From a Mastodon account, feature/endorse a post authored by a user on the test site (or replay a signed `FeatureRequest` to `/wp-json/activitypub/1.0/users/{id}/inbox` with an `actor`-less body and valid `Signature` + `Digest` headers).
* Confirm the request returns `202` instead of `400`.
* Confirm an inbox item is recorded for the activity.
* With the feature policy set to allow it, confirm an `Accept` is queued in the outbox, addressed to the requesting actor.

The changelog entry is already included in this branch.

### DIFF
--- a/.github/changelog/fix-inbox-actor-from-key-id
+++ b/.github/changelog/fix-inbox-actor-from-key-id
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Requests from other platforms to feature your posts are now handled correctly instead of being ignored.
--- a/includes/class-signature.php
+++ b/includes/class-signature.php
@@ -74,6 +74,37 @@ class Signature {
 	}
 
 	/**
+	 * Extract the signing keyId from a request's HTTP Signature headers.
+	 *
+	 * Supports both the draft HTTP Signatures `Signature` header and the RFC 9421
+	 * `Signature-Input` header.
+	 *
+	 * @since unreleased
+	 *
+	 * @param \WP_REST_Request $request The request object.
+	 *
+	 * @return string|null The keyId, or null when no signature is present.
+	 */
+	public static function get_key_id( $request ) {
+		$signature = $request->get_header( 'signature' );
+		if ( $signature && \preg_match( '/keyId="([^"]+)"/i', $signature, $matches ) ) {
+			return $matches[1];
+		}
+
+		/*
+		 * RFC 9421 Signature-Input: keyid is a `;`-delimited parameter whose value may be
+		 * quoted or unquoted. Anchoring on `;` (or string start) avoids matching a `keyid=`
+		 * substring inside another parameter's value.
+		 */
+		$signature_input = $request->get_header( 'signature-input' );
+		if ( $signature_input && \preg_match( '/(?:^|;)\s*keyid="?([^";,\s]+)/i', $signature_input, $matches ) ) {
+			return $matches[1];
+		}
+
+		return null;
+	}
+
+	/**
 	 * If a request with RFC-9421 signature fails, we try again with the Draft Cavage signature.
 	 *
 	 * @param array  $response HTTP response.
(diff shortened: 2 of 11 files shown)


jeherve (dismissed):
This makes sense to me, and is defensive enough that it shouldn't cause any issues or conflicts. 🚢

PR #3386, opened 2026-06-10
Harden the Blurhash encoder against decode bombs, transparency, and metadata races

kraftbj:
Ports four hardenings to the Blurhash encoder that previously shipped in FOSSE's copy of the same code (https://github.com/Automattic/fosse/pull/198). The encoder was originally upstreamed from FOSSE; its copy has since been hardened and this brings the upstream version up to parity, adapted to this plugin's layout (the `imagetypes()` codec gating was already present here and is untouched).

## Proposed changes:

* **Decode-bomb gate.** Before `imagecreatefromstring()` fully decodes the source bytes, the declared dimensions from `getimagesizefromstring()` are checked against a new `MAX_ENCODE_PIXELS` cap (50 megapixels). The existing 8 MiB byte cap doesn't bound the *decoded* allocation: a tiny flat-color PNG declaring 30000×30000 sails past it and forces a multi-gigabyte GD allocation — an uncatchable OOM that kills the cron worker or aborts a CLI backfill mid-run. Over-cap sources are a policy skip, not a failure.
* **Transparency flattening.** The decoded (post-downscale) image is composited onto an opaque white truecolor canvas before per-pixel sampling, falling back gracefully to the un-flattened image if any GD call fails. Without this, `imagecolorsforindex()` reports the raw RGB (usually black) of transparent pixels, so transparent PNG/GIF logos and stickers encode to near-black blurhashes.
* **Cron deferral.** The encode event is scheduled at `time() + MINUTE_IN_SECONDS` instead of `time()`. The `wp_generate_attachment_metadata` filter runs *before* the caller persists the metadata, so an immediate cron tick could race the metadata write, find no thumbnail intermediate, and fall back to encoding the full-size original.
* **Three-state return.** `encode_from_attachment()` now returns `string|false|null` (hash / policy skip / unexpected failure) so the decode-bomb skip routes to the silent-skip path in both the cron callback and the WP-CLI backfill (skip counter, zero exit) instead of `error_log` + diagnostic action + warning + nonzero exit on every run forever. Both callers updated; persistence is gated on `is_string()` so `false` can never reach a string-typed parameter. The native union return type from FOSSE's version was dropped in favor of a docblock because union types require PHP 8.0 and this plugin supports 7.4.

### Other information:

New tests cover the decode-bomb gate (crafted PNG header declaring 30000×30000 must return the `false` sentinel, proving rejection happens *before* decode), the silent-skip routing in `run_encode()` (no `activitypub_blurhash_encode_failed` action, no meta), the transparency flattening (a fully transparent PNG must hash identically to a solid white one), and the cron deferral (scheduled timestamp at least one minute out).

## Testing instructions:

* Run the Blurhash test suite: `vendor/bin/phpunit --filter Test_Blurhash` — 17 tests should pass.
* Decode bomb: craft a PNG whose IHDR declares huge dimensions (see `generate_png_with_declared_dimensions()` in the test file), attach it to an image attachment, and confirm `Blurhash::encode_from_attachment()` returns `false` quickly instead of exhausting memory, and that `wp activitypub blurhash backfill` counts it as skipped with a zero exit code.
* Transparency: upload a transparent PNG logo, let cron run (or use the backfill command), and confirm the stored hash decodes to a light placeholder rather than a near-black one.
* Cron deferral: upload an image and check `wp cron event list` — the `activitypub_blurhash_compute` event should be scheduled about a minute out rather than immediately.

### Changelog entry

A changelog entry (`.github/changelog/add-blurhash-encoder-hardening`) is included in the branch, created with `composer changelog:add`.

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Harden the Blurhash encoder: skip decompression-bomb images before decoding, flatten transparency onto white so transparent logos no longer produce near-black placeholders, and defer the cron encode until attachment metadata is saved.

</details>


includes/class-blurhash.php
```diff
@@ -231,17 +259,47 @@ public static function encode_from_attachment( int $attachment_id ): ?string {
...
 
+		// Decode-bomb guard. `imagecreatefromstring()` fully decodes
+		// the compressed bytes into an uncompressed bitmap before we
+		// get a chance to downscale, so a small but highly
+		// compressible source declaring huge dimensions (e.g. a
+		// flat-color 30000×30000 PNG) would force a multi-gigabyte
+		// allocation and OOM the worker — uncatchable, so we can't
+		// recover with the try/catch below. Read the declared
+		// dimensions cheaply first and skip (silent, not an error)
+		// anything past the megapixel cap.
```
pfefferle: Project convention is `/* */` for multi-line comments (`//` for single-line). This file is currently the `//` outlier, but `/* */` is the dominant style across `includes/`.

```suggestion
		/*
		 * Decode-bomb guard. `imagecreatefromstring()` fully decodes
		 * the compressed bytes into an uncompressed bitmap before we
		 * get a chance to downscale, so a small but highly
		 * compressible source declaring huge dimensions (e.g. a
		 * flat-color 30000×30000 PNG) would force a multi-gigabyte
		 * allocation and OOM the worker — uncatchable, so we can't
		 * recover with the try/catch below. Read the declared
		 * dimensions cheaply first and skip (silent, not an error)
		 * anything past the megapixel cap.
		 */
```

includes/class-blurhash.php
```diff
@@ -231,17 +259,47 @@ public static function encode_from_attachment( int $attachment_id ): ?string {
...
+		$declared_height = (int) $dimensions[1];
+		if ( $declared_width < 1 || $declared_height < 1 ) {
+			return null;
+		}
+		if ( $declared_width * $declared_height > self::MAX_ENCODE_PIXELS ) {
+			// Policy skip, not a failure: `false` routes the caller
+			// to the same silent bucket as a non-raster mime, so a
+			// permanently over-cap source doesn't error_log on every
+			// cron run or hold the CLI backfill exit code nonzero
+			// forever.
```
pfefferle: Project convention is `/* */` for multi-line comments (`//` for single-line). This file is currently the `//` outlier, but `/* */` is the dominant style across `includes/`.

```suggestion
			/*
			 * Policy skip, not a failure: `false` routes the caller
			 * to the same silent bucket as a non-raster mime, so a
			 * permanently over-cap source doesn't error_log on every
			 * cron run or hold the CLI backfill exit code nonzero
			 * forever.
			 */
```

includes/class-blurhash.php
```diff
@@ -231,17 +259,47 @@ public static function encode_from_attachment( int $attachment_id ): ?string {
...
 		}
 
 		// $scaled holds a second GD resource created by imagescale when
-		// the image exceeds MAX_ENCODE_EDGE. It is kept separate from
-		// $original so both can be destroyed in finally regardless of
-		// which code path ran.
+		// the image exceeds MAX_ENCODE_EDGE, $canvas a third one
+		// created for the transparency flattening composite. Both are
+		// kept separate from $original so all of them can be destroyed
+		// in finally regardless of which code path ran.
```
pfefferle: Project convention is `/* */` for multi-line comments (`//` for single-line). This file is currently the `//` outlier, but `/* */` is the dominant style across `includes/`.

```suggestion
		/*
		 * $scaled holds a second GD resource created by imagescale when
		 * the image exceeds MAX_ENCODE_EDGE, $canvas a third one
		 * created for the transparency flattening composite. Both are
		 * kept separate from $original so all of them can be destroyed
		 * in finally regardless of which code path ran.
		 */
```

includes/class-blurhash.php
```diff
@@ -283,6 +341,27 @@ public static function encode_from_attachment( int $attachment_id ): ?string {
...
 
+			// Flatten transparency against a white background before
+			// sampling. `imagecolorsforindex()` reports the raw RGB of
+			// a transparent pixel (usually 0,0,0 → black) while
+			// discarding alpha, so a transparent PNG/GIF/WebP logo or
+			// sticker would otherwise encode to a near-black blurhash.
+			// Compositing onto an opaque white canvas yields the color
+			// a viewer actually sees over a typical light surface.
+			// Best-effort: if any GD call fails we keep sampling the
+			// un-flattened image rather than bail.
```
pfefferle: Project convention is `/* */` for multi-line comments (`//` for single-line). This file is currently the `//` outlier, but `/* */` is the dominant style across `includes/`.

```suggestion
			/*
			 * Flatten transparency against a white background before
			 * sampling. `imagecolorsforindex()` reports the raw RGB of
			 * a transparent pixel (usually 0,0,0 → black) while
			 * discarding alpha, so a transparent PNG/GIF/WebP logo or
			 * sticker would otherwise encode to a near-black blurhash.
			 * Compositing onto an opaque white canvas yields the color
			 * a viewer actually sees over a typical light surface.
			 * Best-effort: if any GD call fails we keep sampling the
			 * un-flattened image rather than bail.
			 */
```

includes/class-blurhash.php
```diff
@@ -549,11 +640,20 @@ public static function run_encode( int $attachment_id ): void {
...
+		if ( \is_string( $hash ) ) {
 			self::set( $attachment_id, $hash );
 			return;
 		}
 
+		// `false` is a policy skip (source over the decode-bomb
+		// dimension cap): deliberate, deterministic, and global to
+		// the source bytes — logging it would re-introduce the
+		// per-run noise this bucket exists to avoid. Only `null`
+		// (unexpected failure) falls through to diagnostics.
```
pfefferle: Project convention is `/* */` for multi-line comments (`//` for single-line). This file is currently the `//` outlier, but `/* */` is the dominant style across `includes/`.

```suggestion
		/*
		 * `false` is a policy skip (source over the decode-bomb
		 * dimension cap): deliberate, deterministic, and global to
		 * the source bytes — logging it would re-introduce the
		 * per-run noise this bucket exists to avoid. Only `null`
		 * (unexpected failure) falls through to diagnostics.
		 */
```

includes/cli/class-blurhash-command.php
```diff
@@ -201,6 +201,18 @@ public function backfill( $args, $assoc_args ) {
 				}
 
 				$hash = Blurhash::encode_from_attachment( $attachment_id );
+
+				// `false` is a policy skip (e.g. declared dimensions
+				// over the decode-bomb cap) — same bucket as a
+				// non-raster mime, not a failure. Warning on it every
+				// run (and exiting nonzero) would make a permanently
+				// over-cap attachment poison automation forever.
```
pfefferle: Project convention is `/* */` for multi-line comments (`//` for single-line). This file is currently the `//` outlier, but `/* */` is the dominant style across `includes/`.

```suggestion
				/*
				 * `false` is a policy skip (e.g. declared dimensions
				 * over the decode-bomb cap) — same bucket as a
				 * non-raster mime, not a failure. Warning on it every
				 * run (and exiting nonzero) would make a permanently
				 * over-cap attachment poison automation forever.
				 */
```

tests/phpunit/tests/includes/class-test-blurhash.php
```diff
@@ -132,4 +232,99 @@ public function test_encode_from_attachment_produces_hash() {
...
+		}
+
+		$transparent_id = $this->create_image_attachment_at( $this->generate_alpha_fixture_png( true ) );
+		$white_id       = $this->create_image_attachment_at( $this->generate_alpha_fixture_png( false ) );
+
+		$transparent_hash = Blurhash::encode_from_attachment( $transparent_id );
+		$white_hash       = Blurhash::encode_from_attachment( $white_id );
+
+		$this->assertIsString( $transparent_hash );
+		$this->assertSame( $white_hash, $transparent_hash );
```
pfefferle, line 392: `fully-transparent === solid-white` is a slightly weak lock: an implementation that flattened *every* image to white (or dropped color) would also pass, and nothing else asserts visible color survives the composite (`test_encode_from_attachment_produces_hash` only checks the hash *format*). Consider adding a fixture with an opaque colored shape on a transparent background and asserting the hash is neither the all-white nor all-black hash — that pins down both halves: transparent regions flatten to white *and* real color is preserved.

pfefferle (changes requested):
Solid hardening overall — the decode-bomb guard, the three-state `string|false|null` routing, and the +1 minute cron deferral all look correct and are well covered.

Two requests: (1) the new multi-line comments use `//`; please switch them to `/* */` to match the project convention (suggestions inline; the same applies to the deferral comment block in `schedule()` which I didn't inline since it's glued to a pre-existing `//` block). (2) a minor test-coverage note on the transparency test.

tests/phpunit/tests/includes/class-test-blurhash.php
```diff
@@ -17,13 +17,113 @@
...
+		$ihdr      = \pack( 'N', \strlen( $ihdr_data ) )
+			. 'IHDR' . $ihdr_data
+			. \pack( 'N', \crc32( 'IHDR' . $ihdr_data ) );
+
+		// Empty IEND chunk so the stream is well-formed enough for
+		// header parsers.
+		$iend = \pack( 'N', 0 ) . 'IEND' . \pack( 'N', \crc32( 'IEND' ) );
+
+		$path                  = \tempnam( \sys_get_temp_dir(), 'ap-blurhash-dim-' ) . '.png';
+		$this->fixture_files[] = $path;
```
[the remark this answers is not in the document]
pfefferle: @kraftbj Valid — `tempnam()` creates `/tmp/ap-blurhash-dim-XXXX` (empty), then `. '.png'` makes the write target a *different* path, so the original temp file is orphaned (only the suffixed path is in `$fixture_files`). Since the encoder reads bytes (`getimagesizefromstring`/`imagecreatefromstring` don't care about the extension) and the mime is set on the attachment, the cleanest fix is to drop the `.png` suffix and use the `tempnam()` path directly.

tests/phpunit/tests/includes/class-test-blurhash.php
```diff
@@ -17,13 +17,113 @@
...
+	 * black RGB — the worst case for an alpha-unaware sampler) or solid
+	 * opaque white. The pair drives the transparency flattening test.
+	 *
+	 * @param bool $transparent True for fully transparent, false for solid white.
+	 * @return string Absolute file path.
+	 */
+	private function generate_alpha_fixture_png( $transparent ) {
+		$path                  = \tempnam( \sys_get_temp_dir(), 'ap-blurhash-alpha-' ) . '.png';
+		$this->fixture_files[] = $path;
+
```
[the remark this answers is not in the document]
pfefferle: @kraftbj Same leak here as the dimension fixture — `tempnam() . '.png'` orphans the un-suffixed temp file. `imagepng( $image, $path )` writes PNG regardless of extension, so dropping the suffix (and tracking the real `tempnam()` path) fixes it.

tests/phpunit/tests/includes/class-test-blurhash.php
```diff
@@ -17,13 +17,113 @@
...
+		\add_filter(
+			'get_attached_file',
+			function ( $file, $id ) use ( $attachment_id, $path ) {
+				return ( (int) $id === $attachment_id ) ? $path : $file;
+			},
+			10,
+			2
+		);
+
```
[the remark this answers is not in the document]
pfefferle: @kraftbj Reasonable. The `get_attached_file` closure is never removed, and because each test rolls back in a transaction post IDs can repeat across tests — a stale closure matching a later test's reused attachment ID would return the wrong path. Capturing the callback and `remove_filter`-ing it in `tear_down` (or asserting via a single shared filter) avoids the cross-test risk.

tests/phpunit/tests/includes/class-test-blurhash.php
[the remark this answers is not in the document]
kraftbj: Fixed in 7f2cdb5b — dropped the `.png` suffix and use the `tempnam()` path directly, so the tracked path is the real temp file (the encoder reads bytes, not the extension, and the mime is set on the attachment).

tests/phpunit/tests/includes/class-test-blurhash.php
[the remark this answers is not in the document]
kraftbj: Fixed in 7f2cdb5b — same change. `imagepng()` writes a PNG stream regardless of extension, so the suffix is gone and the real `tempnam()` path is tracked in `$fixture_files`.

tests/phpunit/tests/includes/class-test-blurhash.php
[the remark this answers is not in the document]
kraftbj: Fixed in 7f2cdb5b — the `get_attached_file` callback is now stored in a `$attached_file_filters` property and `remove_filter`-ed in `tear_down()`, so no closure leaks across tests.

includes/class-blurhash.php
kraftbj: Applied in 7f2cdb5b. Also converted the deferral comment block in `schedule()` (the one glued to the pre-existing `//` block) to `/* */` in the same commit, merging both paragraphs into one block comment.

includes/class-blurhash.php
kraftbj: Applied in 7f2cdb5b.

includes/class-blurhash.php
kraftbj: Applied in 7f2cdb5b.

includes/class-blurhash.php
kraftbj: Applied in 7f2cdb5b.

includes/class-blurhash.php
kraftbj: Applied in 7f2cdb5b.

includes/cli/class-blurhash-command.php
kraftbj: Applied in 7f2cdb5b.

tests/phpunit/tests/includes/class-test-blurhash.php
kraftbj: Added in 7f2cdb5b — new `test_encode_preserves_color_over_transparency` builds an opaque red square on a transparent background and asserts its hash equals neither the all-white nor the all-black reference, pinning down both that transparent regions flatten to white and that real color survives the composite.

tests/phpunit/tests/includes/class-test-blurhash.php
```diff
@@ -132,4 +300,130 @@ public function test_encode_from_attachment_produces_hash() {
...
+			static function ( $failed_id ) use ( &$captured ) {
+				$captured[] = (int) $failed_id;
+			}
+		);
+
+		Blurhash::run_encode( $attachment_id );
+
+		$this->assertNull( Blurhash::get( $attachment_id ) );
+		$this->assertSame( array(), $captured, 'Policy skip must not fire the encode-failed action.' );
+	}
```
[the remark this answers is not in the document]
kraftbj: Fixed in b6a7e3c8 — the `activitypub_blurhash_encode_failed` callback is now captured in a variable and `remove_action`-ed in a `finally` around `run_encode()`, so it can't leak into later tests in the shared process.

includes/class-blurhash.php
```diff
@@ -195,14 +215,22 @@ public static function delete( int $attachment_id ): void {
...
 	 *
 	 * @param int $attachment_id Attachment post ID.
-	 * @return string|null
+	 * @return string|false|null Hash on success, false on policy skip, null on failure.
 	 */
-	public static function encode_from_attachment( int $attachment_id ): ?string {
+	public static function encode_from_attachment( int $attachment_id ) {
 		if ( ! self::is_encodable_attachment( $attachment_id ) ) {
 			return null;
 		}
```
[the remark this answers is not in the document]
kraftbj: Fixed in b6a7e3c8 — `encode_from_attachment()` now returns the `false` policy-skip sentinel (not `null`) for both predictable-unencodability guards (non-encodable attachment and unavailable encoder), matching the documented three-state contract and `run_encode()`'s own classification of those as silent skips. Updated the docblock and added `test_encode_from_attachment_returns_false_for_non_encodable` to pin it. Both production callers already pre-filter via `is_encodable_attachment()`, so this only corrects the contract for direct callers.

tests/phpunit/tests/includes/class-test-blurhash.php
```diff
@@ -132,4 +300,150 @@ public function test_encode_from_attachment_produces_hash() {
...
+	 * `time()`. The filter runs before the caller persists the metadata, so
+	 * an immediate cron spawn could race the metadata write and encode the
+	 * full-size original instead of the thumbnail.
+	 *
+	 * @covers ::schedule_encode
+	 */
+	public function test_schedule_encode_defers_cron_event() {
+		if ( ! Blurhash::is_encoder_runnable() ) {
+			$this->markTestSkipped( 'GD is not available.' );
+		}
```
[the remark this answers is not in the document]
kraftbj: Fixed in 5cc970ab — the test now also skips when `imagepng()` is unavailable, matching the other PNG-writing tests, so a GD-without-PNG host skips rather than fatals.

includes/class-blurhash.php
```diff
@@ -195,20 +215,35 @@ public static function delete( int $attachment_id ): void {
...
-	 * — never throws, never warns. Used by both the cron handler
-	 * and the WP-CLI backfill.
+	 * pixel array to the encoder. Never throws, never warns. Used by
+	 * both the cron handler and the WP-CLI backfill.
+	 *
+	 * Three-state return so callers can route outcomes to the right
+	 * bucket: a string is success; `false` means the source is
+	 * deliberately outside encode policy and must be skipped silently —
+	 * a non-encodable attachment (non-raster mime, missing/deleted, or a
+	 * format the host GD build can't decode), an unavailable encoder, or
```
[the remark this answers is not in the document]
kraftbj: Clarified in 5cc970ab — the docblock now says `false` covers a deleted/nonexistent attachment ID, and `null` covers a missing/unreadable/corrupt file behind an otherwise-encodable attachment, so the two aren't conflated.

pfefferle (approved):
Nice! Thanks a lot :)

PR #3468, opened 2026-06-26
Fix: Prevent duplicate Like comments from repeated deliveries

### LINKED ISSUE #3215: Getting repeated comments from a single Mastodon "like"
### Quick summary

A user from mas.to (a Mastodon server) "liked" a post on my WordPress ActivityPub site running 8.1.1 of the plugin and WordPress 6.9.4). 

I've gotten now 10 "comments" in WordPress for what I take to be a single "like" - different timestamps for "submitted" on the comment approval page. 

Any ideas where to look / what to check? It feels almost like the Mastodon server doesn't know the like has been registered and keeps trying? 

### Steps to reproduce

Approved the first "like" - expected there to be only one. 

Not entirely sure how reproducable it is - you'd have to make a post, like it on a subscribed Mastodon server, and have it fail in this way 

I've had other "likes" just fine - so not sure if it is mas.to that has an issue?

### Site owner impact

### Severity

Minor

No revenue impact

I guess the workaround is just to approve one and to call the others spam, or approve them and then delete them (so as not to train the system they are spam) 

I have only had it happen once

pfefferle:
Fixes #3215

## Proposed changes:

A single "Like" from a remote account could pile up as many duplicate "Like" comments (the reporter saw 10–16 from one user on one post). The trigger correlated with accounts whose display name contains emoji.

Root cause: the Like handler's dedupe guard looked up an existing comment by the **liked object's** URL, but `Interactions::add_reaction()` stores the comment's `source_id` as the **Like activity's** id. Those never matched, so the plugin's own idempotency guard was effectively dead code for Likes. Repeat deliveries then fell through to WordPress's native author/content duplicate check — which, on databases that are not `utf8mb4`, fails to match because the emoji author name is mangled/truncated on store. Each redelivery therefore created another comment.

* Dedupe incoming Likes on the Like activity id (`object_to_uri( $like )`, the value stored as `source_id`) instead of the liked object. This makes redelivery idempotent regardless of the author name or the site's database charset.
* This brings the Like handler into parity with the Announce handler, which already dedupes on the activity id.

The underlying `utf8` vs `utf8mb4` table issue (which also makes the name render as `??` in wp-admin) is site-side and out of scope here; the corrected guard stops the duplication either way.

### Other information:

## Testing instructions:

* Add a regression test `test_handle_like_duplicate_short_circuits_before_processing` that delivers the same Like twice and asserts the `activitypub_handled_like` hook fires exactly once. It fails on the previous code (the dead guard let the second delivery through) and passes with the fix.
* Run `npm run env-test -- --filter='Test_Like|Test_Announce'` — all pass.
* Manually: like a post from a remote (Mastodon) account whose display name contains emoji; confirm exactly one "Like" comment is recorded even if the activity is delivered more than once.

### Changelog entry

A changelog entry is included in the branch (`.github/changelog/fix-duplicate-like-comments`).


pfefferle:
@jeckman can you somehow check if this PR solves your issue?

jeckman:
> @jeckman can you somehow check if this PR solves your issue?

Well, I can install it and see if I stop getting multiple likes but I won't really know for a while whether it has had the desired impact. (I can't, or don't know how to, trigger the multiple likes from an external source with an emoji in their name to validate for sure they fail - but I can see whether i keep getting them)

jeckman:
ok, I've applied the patch / made the relevant changes to the like class - will let you know if I get any further duplicate likes or side effects

pfefferle:
thanks @jeckman !

jeherve (approved):

PR #3567, opened 2026-07-22
Use `esc_url_raw()` for URLs emitted into ActivityPub JSON

### LINKED ISSUE #3566: Actor JSON icon/image URLs are HTML-escaped with esc_url() — should be esc_url_raw()
## The bug

Someone decided to run URLs destined for **JSON output** through `esc_url()`. That's the *display* variant. It HTML-entity-encodes `&` into `&#038;` (or `&amp;`, depending on which layer touches it last). The result is then `json_encode`d, so the entity survives verbatim into the actor document.

JSON is not HTML. Nobody is rendering this string in a browser. There is no reason on earth to be HTML-escaping it. All you've accomplished is corrupting every URL that happens to carry more than one query parameter.

## Where

All in the actor JSON serialization path:

- `includes/functions.php:140` — site icon (`Blog::get_icon()` returns this)
- `includes/model/class-blog.php:242` — blog header image
- `includes/model/class-user.php:194` — user icon (attachment)
- `includes/model/class-user.php:198` — user icon (`get_avatar_url()`)
- `includes/model/class-user.php:231` — user header image

## Why it matters

A perfectly ordinary Photon/Gravatar avatar URL looks like this:

```
https://i0.wp.com/…?fit=215,215&ssl=1
```

After `esc_url()` it ships in the actor doc as:

```
https://i0.wp.com/…?fit=215,215&#038;ssl=1
```

Now think about what a receiving server does with that. The `&#038;` isn't a parameter separator anymore — the `#` opens a fragment, so `ssl=1` silently vanishes on the way to the origin. Tolerant clients behind a forgiving CDN won't notice. Stricter AP consumers, or any CDN that isn't Photon and actually cares about its query string, get handed a mangled URL and can fail the fetch. This is machine-readable output being fed corrupted data and *hoping* the other end is sloppy enough to cope.

## The fix

`esc_url_raw()`. That's it. It's the variant that exists precisely for non-display contexts and does **not** HTML-encode the ampersand.

The part that should be embarrassing: the transformers two directories over already know this. `includes/transformer/class-base.php` and `includes/transformer/class-post.php` use `esc_url_raw()` for exactly this kind of URL. So the codebase already contains the right answer — the actor models just didn't get the memo. Make it consistent.

## Note

This is a **correctness bug in the emitted JSON**, independent of any caching or federation-delivery symptom. Please don't close it as "couldn't reproduce a broken avatar" — the point is that the plugin is emitting an HTML-escaped URL into a JSON document, which is wrong on its face regardless of whether a given downstream server currently tolerates it.

pfefferle:
Fixes #3566

## Proposed changes:

URLs destined for **ActivityPub JSON** were escaped with `esc_url()`, whose *display* context HTML-encodes ampersands (`&` → `&#038;`). That is right for an HTML attribute but wrong for a machine-readable document: a multi-parameter URL like a Photon/Gravatar avatar (`…?fit=215,215&ssl=1`) ships with the ampersand entity-encoded, and a strict receiving server can treat `&#038;` as a fragment and silently drop the trailing parameter.

This switches every URL that is **serialized into JSON** to `esc_url_raw()` — the non-display variant that keeps the ampersand intact while still rejecting unsafe schemes (`javascript:`, etc.):

- **Actor models:** site icon, user/blog/application icon, avatar, header image, and profile URLs.
- **Transformers:** post permalinks, image and attachment URLs, and `Mention`/`Hashtag` hrefs.
- **Collections & REST:** extra-field `Link` hrefs, the Hashtag collection, the interaction redirect id (which is `rawurlencode`d), and the reactions REST response.

Deliberately **unchanged** (where `esc_url()` is correct): mention/hashtag anchor tags and media `` tags embedded in the AP `content` HTML, all `wp-admin/` UI, and the `esc_url( $guid )` SQL `WHERE guid = %s` lookups (which normalize against stored guids rather than emitting output).

The reporter noted the transformers "already use `esc_url_raw()`"; in fact they had the same bug, which is why this sweep is broader than the five spots in the issue.

### Other information:

## Testing instructions:

* Set a site icon or user avatar to a URL with more than one query parameter (a Jetpack/Photon or Gravatar avatar is a natural example, e.g. `…?resize=215&ssl=1`).
* Fetch the actor document, e.g. `curl -H "Accept: application/activity+json" https://example.com/author/<user>/`.
* **Before:** the `icon` / `image` URL contains `&#038;` and the trailing parameter is broken.
* **After:** the URL contains a plain `&` and is intact.
* The same applies to `url`, image attachments, and `Mention`/`Hashtag` hrefs in a transformed post, and to the reactions REST response.

### Changelog entry

A changelog entry is included in the branch (`.github/changelog/fix-actor-json-url-encoding`).

### DIFF
--- a/.github/changelog/fix-actor-json-url-encoding
+++ b/.github/changelog/fix-actor-json-url-encoding
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fixed
 
-Fix URLs with multiple query parameters (such as avatars, images, and profile links) being corrupted in content sent to the Fediverse.
+Fix URLs with multiple query parameters (such as avatars, images, profile links, and podcast media) being corrupted in content sent to the Fediverse.
--- a/.github/changelog/fix-actor-json-url-encoding
+++ b/.github/changelog/fix-actor-json-url-encoding
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix URLs with multiple query parameters (such as avatars, images, and profile links) being corrupted in content sent to the Fediverse.
--- a/includes/collection/class-extra-fields.php
+++ b/includes/collection/class-extra-fields.php
@@ -132,7 +132,7 @@ class Extra_Fields {
 					$attachment = array(
 						'type' => 'Link',
 						'name' => $title,
-						'href' => \esc_url( $tags->get_attribute( 'href' ) ),
+						'href' => \esc_url_raw( $tags->get_attribute( 'href' ) ),
 					);
 
 					$rel = $tags->get_attribute( 'rel' );
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -137,7 +137,7 @@ function site_icon() {
 
 	return array(
 		'type' => 'Image',
-		'url'  => \esc_url( $icon_url ),
+		'url'  => \esc_url_raw( $icon_url ),
 	);
 }
 
--- a/includes/model/class-application.php
+++ b/includes/model/class-application.php
@@ -156,7 +156,7 @@ class Application extends Actor {
 		if ( \has_header_image() ) {
 			return array(
 				'type' => 'Image',
-				'url'  => \esc_url( \get_header_image() ),
+				'url'  => \esc_url_raw( \get_header_image() ),
 			);
 		}
 
--- a/includes/model/class-blog.php
+++ b/includes/model/class-blog.php
@@ -94,7 +94,7 @@ class Blog extends Actor {
 		$permalink = \get_option( 'activitypub_use_permalink_as_id_for_blog', false );
 
 		if ( $permalink ) {
-			return \esc_url( \home_url( '/@' . $this->get_preferred_username() ) );
+			return \esc_url_raw( \home_url( '/@' . $this->get_preferred_username() ) );
 		}
 
 		return \add_query_arg( 'author', $this->_id, \home_url( '/' ) );
@@ -170,7 +170,7 @@ class Blog extends Actor {
 	 * @return string The User-Url.
 	 */
 	public function get_alternate_url() {
-		return \esc_url( \trailingslashit( \get_home_url() ) );
+		return \esc_url_raw( \trailingslashit( \get_home_url() ) );
 	}
 
 	/**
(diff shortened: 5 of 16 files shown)


donnchawp:
The `includes/` sweep looks correct and safe — every changed site serializes into AP JSON, security posture is unchanged (`esc_url_raw()` shares the same scheme whitelist as `esc_url()`), and `interaction-controller.php` even stops a double-encode since that id gets `rawurlencode()`d afterwards.

One gap before this is complete though: two active **integration** transformers override `get_attachment()` and emit the same JSON attachment structure this PR fixed in `class-base.php` / `class-post.php`, but still use `esc_url()`:

- `integration/class-seriously-simple-podcasting.php` — lines 34 (`url`) and 44 (`icon`)
- `integration/class-podlove-podcast-publisher.php` — lines 117 (`url`) and 137 (`icon`)

Podlove is the sharper one: when analytics tracking is enabled it deliberately swaps in a **tracking URL** (lines 108–113), which is canonically a multi-parameter URL — the exact `#3566` bug class. These four call sites should be `esc_url_raw()` too, otherwise podcast enclosure/tracking URLs stay corrupted.

Minor / probably out of scope: `includes/class-emoji.php:49` puts a URL into block attrs that get `wp_json_encode()`d into an `wp:activitypub/emoji` comment, so `&` is entity-encoded there as well — but that's incoming content rendered locally, not outbound federation, so arguably fine to leave.

donnchawp (dismissed):
Looks great, thanks!

pfefferle:
Good catch, thanks. Fixed in 2e93ce196 — all four podcast attachment sites now use `esc_url_raw()`:

- `class-seriously-simple-podcasting.php`: audio `url` and cover `icon`
- `class-podlove-podcast-publisher.php`: enclosure `url` and episode `icon`

You're right that Podlove is the sharp one: with analytics tracking enabled the enclosure becomes a multi-parameter tracking URL, which is exactly the `#3566` class.

I checked the rest of the `esc_url()` calls in `integration/` too, and left these as-is since they're genuine HTML, not AP JSON:

- `class-jetpack.php` (reader `<a href>`), `class-classic-editor.php` (`[audio src]` shortcode), `class-buddypress.php` (profile `<a href>`), `class-yoast-seo.php` (admin-notice `admin_url`).

And I agree on `class-emoji.php:49` being out of scope — that URL is entity-encoded into a local `wp:activitypub/emoji` block comment for incoming content, not outbound federation.

Also broadened the changelog to mention podcast media.

PR #3575, opened 2026-07-23
Federate podcast episodes from both Jetpack podcast features

pfefferle:
## Proposed changes:

Jetpack has two podcast features, built at the same time by different people, and an episode can come from either one:

* **Posts to Podcast** generates audio for a post and writes a `jetpack/podcast-episode` block carrying the audio URL, mime type and cover art.
* **Jetpack Podcast** treats every post in the configured podcast category as an episode, with the audio arriving as an ordinary WordPress enclosure.

The first version of this PR only handled the block, and I had it labelled as Jetpack Podcast, which is the other one. Thanks to Rob for catching that. It now covers both.

Neither surface reaches ActivityPub on its own. Episode audio is usually hosted off-site, so it has no attachment ID, and `Transformer\Base::filter_unique_attachments()` drops every media entry without one, so that audio never survives to the `activitypub_attachments` filter. The integration resolves the episode itself and either adds the audio, or replaces the artwork when the media library already contributed the file.

* Hooks `activitypub_attachments` rather than adding a transformer subclass. A subclass is winner-take-all, so a site running Podlove or Seriously Simple Podcasting alongside a Jetpack podcast would get one behavior instead of both.
* Cover art follows the podcast feed's own order: the episode's own artwork, then the post's featured image, then the show image, so a federated episode carries the artwork subscribers already see.
* Deduplicates against the existing attachment with `normalize_url()`, so a site that moved to HTTPS after publishing does not federate the same file twice.
* Respects the site's attachment maximum, which the transformer applies just before this filter runs.
* Omits `mediaType` when the block has no mime type, rather than sending an empty string that stops receivers classifying the attachment at all.

I also pulled the max-attachments resolution out of `Transformer\Post` into `Activitypub\get_max_attachments()`, since the integration needs the same value and I did not want a second copy of the meta, option and filter precedence.

### Known gap

`Podlove_Podcast_Publisher` and `Seriously_Simple_Podcasting` override `get_attachment()` and never apply `activitypub_attachments`, so on a site running either of those this callback does not fire. I left that alone here, but it means the three podcast integrations sit at three different levels for one problem.

### Worth a follow-up

`filter_unique_attachments()` drops media without an `id`, but `Post::get_enclosures()` deliberately builds complete entries for enclosures that are not in the media library, and `Base::transform_attachment()` opens with a passthrough for id-less entries that is unreachable today. Fixing that one place would let every site with an external enclosure federate it, and would remove most of the workaround here. It changes shared behavior for every post that merely links an audio file though, and there is no enclosure test coverage yet, so I think it should be its own PR.

### Other information:

## Testing instructions:

Posts to Podcast:

* Generate an episode for a post so it gets a `jetpack/podcast-episode` block.
* Fetch the post's ActivityPub representation and confirm `attachment` holds the audio with `type: Audio`, the right `mediaType`, and the episode cover art as `icon`.
* Confirm the audio is not listed twice when WordPress' core enclosure meta points at the same file, including when one of the two uses `http` and the other `https`.

Jetpack Podcast:

* Configure a podcast category and a show image, then publish a post in that category with an audio enclosure hosted off-site.
* Confirm the episode federates with that audio attached and the show image as `icon`.
* Repeat with the audio uploaded to the media library, and confirm the show image wins over the site icon.
* Confirm a post outside the podcast category that happens to have an enclosure is not treated as an episode.

### Changelog entry

Included in the branch (`.github/changelog/add-podcast-episode-attachments`).


integration/class-jetpack.php
```diff
@@ -172,4 +183,222 @@ public static function adapt_post_share() {
...
+				$attachments[ $index ]['icon'] = $icon;
+			}
+
+			return $attachments;
+		}
+
+		$audio = array(
+			'type' => $episode['type'],
+			'url'  => $episode['url'],
+			'name' => \esc_attr( \get_the_title( $post ) ),
```
jeherve: I noticed a small issue in the output. The attachment `name` comes through with raw HTML entities. Here's what my test episode actually federated:

```json
"name": "Rockwell &#8211; Somebody&#8217;s Watching Me"
```

That `&#8211;` should be an en-dash and `&#8217;` an apostrophe.

`get_the_title()` already returns the texturized, entity-encoded title, and then `esc_attr()` re-encodes it on top. But `name` here is a JSON string, not an HTML attribute, so a receiver like Mastodon shows the entities literally in the media description.

Maybe we could do like what we do here?
https://github.com/Automattic/wordpress-activitypub/blob/d0f254a71a1b9423ab6a30124055d47f220bf6ca/includes/transformer/class-post.php#L329


```suggestion
			'name' => \html_entity_decode( \wp_strip_all_tags( \get_the_title( $post ) ), ENT_QUOTES, 'UTF-8' ),
```

jeherve (commented):
Nice work on this. I tested it on a live site and the audio attachment federates cleanly.

I only have a small, non-blocking comment.

PR #3596, opened 2026-07-31
Only serve ActivityPub to JSON-only requests

pfefferle:
## Proposed changes:

Content negotiation now treats a request as ActivityPub only when *every* media type in the `Accept` header is JSON (`application/json`, `application/ld+json`, `application/activity+json`, or any other `*+json`). Before, we matched "contains JSON anywhere", so a request that also accepted HTML (e.g. `text/html, application/activity+json`) still got JSON.

The loose rule made our JSON responses fragile against any shared page cache that varies only on "JSON-only" requests. Tightening it keeps content-negotiated JSON out of those caches in general, not just one patched cache. I ran the JSON-only rule against Mastodon, Misskey, GoToSocial, Akkoma, Lemmy and Friendica `Accept` headers, and they all still resolve to JSON, because every one is JSON-only.

The decision lives in one dependency-free helper, `is_json_only_accept()`. `is_activitypub_request()` delegates to it, and the Surge cache drop-in (`integration/surge-cache-config.php`) reuses the same helper to pick its `representation` cache variant, so the bucket the cache keys on can never disagree with what the plugin serves. The drop-in runs from Surge's `advanced-cache.php` before the plugin loads, so it can only call the dependency-free helper, not `is_activitypub_request()` (which needs `Query`, `$wp_query` and the plugin loaded). Both sides also classify the header in the same raw form: the drop-in runs before WordPress' sanitizers are even loaded, so the plugin no longer sanitizes it either, or the two could classify the same request differently. The docblocks spell all of this out so no one reaches for the wrong one.

One behavior change worth calling out: a client sending a mixed `text/html, application/activity+json` now gets HTML instead of JSON. No major platform does that, and the `activitypub_is_activitypub_request` filter is still there as an escape hatch.

Not in this PR: the LiteSpeed `.htaccess` rule has the same coarse boundary and is harder to express as JSON-only in mod_rewrite, so I left it for a follow-up.

### Other information:

## Testing instructions:

* Publish a post.
* `curl -H "Accept: application/activity+json" <post-url>` returns `Content-Type: application/activity+json`.
* `curl -H "Accept: application/ld+json" <post-url>` still returns JSON.
* `curl -H "Accept: text/html, application/activity+json" <post-url>` now returns `text/html` (this is the change).
* A normal browser request still returns `text/html`.
* With Surge active, confirm the JSON and HTML variants land in separate cache buckets (a mixed or browser request never replays a pure-JSON response, and the other way round).

### DIFF
--- a/.github/changelog/tighten-content-negotiation-json-only
+++ b/.github/changelog/tighten-content-negotiation-json-only
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+ActivityPub responses are now served only to clients that ask for ActivityPub data and nothing else, which keeps that data out of page caches meant for regular web pages.
--- a/docs/how-to/caching.md
+++ b/docs/how-to/caching.md
@@ -55,11 +55,13 @@ Varnish does not pass through `Vary: Accept` by default. You need to configure i
 
 ```vcl
 sub vcl_hash {
-    if (req.http.Accept ~ "application/(ld\+json|activity\+json|json)") {
+    if (req.http.Accept ~ "(?i)^[\s,]*[^\s,;]*(/json|\+json)\s*(;[^,]*)?([\s,]+[^\s,;]*(/json|\+json)\s*(;[^,]*)?)*[\s,]*$") {
         hash_data("activitypub");
     }
 }
 ```
+
+The pattern matches only requests whose `Accept` header is *entirely* JSON, so a browser (which also accepts HTML) shares the HTML cache, not the JSON one. It mirrors how the plugin itself decides.
 
 Alternatively, add `Vary: Accept` in your Apache or Nginx config so Varnish sees it from the backend.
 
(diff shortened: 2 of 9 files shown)


jeherve (commented):
I looked at this with Claude, and I think there's a real correctness issue hiding in the shared helper.

`is_json_only_accept()` unslashes with `stripslashes()` unconditionally, but the two callers hand it the header at different stages.

- The plugin path in `class-query.php` runs after `wp_magic_quotes()`, which slashes all of `$_SERVER` (including `HTTP_ACCEPT`), so `stripslashes()` there gets you back to the real bytes.
- The Surge drop-in runs before WordPress, so the header is already raw, and `stripslashes()` strips backslashes the client actually sent. 

So the two paths classify different bytes the moment a `\` shows up in the header, which is exactly the drift the docblock says can never happen.
Concretely, with Surge active, an anonymous `Accept: application/json\` on a normal post URL:

- Drop-in: `stripslashes('application/json\')` → `application/json` → **json** bucket.
- Plugin: magic-quoted then stripslashed back to `application/json\` → suffix is `json\`, not `/json` → **not** an AP request → serves HTML.
- Surge caches that HTML under the `json` variant, and the next real `application/activity+json` fetch gets HTML back. That breaks federation for the URL until the entry expires.

I'm wondering if the cleaner fix is to not unslash inside the helper at all, and have each caller normalize first: the drop-in passes `$_SERVER['HTTP_ACCEPT']` as-is (already raw), the plugin passes `wp_unslash( $_SERVER['HTTP_ACCEPT'] )`, and the helper just classifies whatever it's given. Then both sides really are looking at the same client bytes.

What do you think?

pfefferle:
Good catch @jeherve, that's exactly the drift the docblock claimed couldn't happen.

Went with your suggestion in 1bfe73169: the helper no longer unslashes at all. The plugin passes `wp_unslash( $_SERVER['HTTP_ACCEPT'] )`, the Surge drop-in passes the raw value, and the helper just classifies whatever it gets. So both paths really are looking at the same client bytes, and your `application/json\` case now serves HTML on both.

While I was in there I also tightened the LiteSpeed rule and the Varnish/Nginx docs in f84826ed1. They had the same too-loose `Accept` boundary and would have poisoned the same way once the plugin got stricter.

Re-requested your review, thanks for digging into this one.

jeherve (approved):
This is looking good now! 🚢

PR #2887, opened 2026-02-06
Add block-based runtime caching for remote media

pfefferle:
Fixes #2742, #2743, #2744

## Proposed changes:

Introduces a dedicated `Cache` namespace for remote media caching with improved architecture:

* **New Cache classes**: `Cache\Avatar`, `Cache\Media`, `Cache\Emoji` - each handles type-specific caching with lazy loading via `activitypub_remote_media_url` filter
* **Block-based runtime caching**: Emoji and media are wrapped in WordPress blocks at insert time, cached lazily at render time
  - `activitypub/emoji` block: Wraps emoji shortcodes, renders with cached URLs
  - `activitypub/image`, `activitypub/audio`, `activitypub/video` blocks: Wrap remote media in posts (comments strip remote images for security)
* **Audio/video support**: Incoming posts with audio or video attachments now render as native `<audio>` and `<video>` elements via dedicated blocks
* **CLI commands**: `wp activitypub cache status` and `wp activitypub cache clear [--type=<type>]` for cache management
* **Filter-first architecture**: All remote media URLs pass through `activitypub_remote_media_url` filter, allowing CDN plugins (Jetpack Photon, Cloudflare) to intercept
* **Improved security**: Uses `finfo` for reliable MIME validation, escapes glob metacharacters, strips remote images from comments
* **Lazy avatar resolution**: Avatars are no longer eagerly downloaded and stored in post meta (`_activitypub_avatar_url`). Instead, they resolve lazily at render time through the `activitypub_remote_media_url` filter
* **Simplified Attachments class**: Incoming post attachments are no longer imported into the Media Library. Instead, remote media is cached as files in the uploads directory via the new Cache classes. The Attachments class now only handles outgoing post attachments
* **Granular control**: Can disable globally (`ACTIVITYPUB_DISABLE_MEDIA_CACHE`) or per-type via filters
* **New media helper functions**: `is_remote_url()`, `process_remote_images()`, `process_remote_media()`, and block generator functions in `functions-media.php`

### Breaking changes:

* **`activitypub_store_attachments_locally` filter removed** — This filter controlled whether incoming post attachments were imported into the Media Library. It is replaced by the `activitypub_should_cache_url` and `activitypub_cache_{type}_enabled` filters, and the `ACTIVITYPUB_DISABLE_MEDIA_CACHE` constant
* **`_activitypub_avatar_url` post meta no longer used** — Avatar URLs are now resolved lazily at render time instead of being stored in post meta

## Implementation Advantages

[...]

PR #3635, opened 2026-08-10
Clean up stale cached avatars automatically

### LINKED ISSUE #3583: actors folder clogs up directory
### Quick summary

It appears that the actors folder in the Wordpress directory is constantly filling up, making for bloated directory.
(wp-content->uploads->activitypub->actors)

Looks like profile pics are stored in here. But they are never cleared. 1000s of profile pics are stored, even if the WP AP account only follows 10 profiles. Whatsmore, some profile pics have new versions uploaded daily with the old versions still stored, leaving multiple copies of the same image. Looking at three sites using the AP plugin two have 35mb worth of data of profile pictures stored here, while the other has over 750mb! 

Could there be a way to flush this information so it is not stored eternally?

### Steps to reproduce

Go to wp-content->uploads->activitypub->actors and see how much data is in there.

### Site owner impact

### Severity

Moderate

_No response_

_No response_

faisalahammad:
Fixes #3583

## Proposed changes:

* Stop duplicate avatar copies from being created in the first place: when a remote actor changes their profile picture, the previous cached version is now removed at the moment the new one is downloaded. Remote actors rotate their icon URL often, and each change previously left an orphaned copy behind, which is how the actors folder kept filling up (the root cause of #3583).
* Add a daily cron job that scans the avatar cache directory and drains any backlog already on disk, removing orphaned actor directories that no longer match an actor post plus stale avatar files that no longer match an actor's current icon.
* Batch the cron cleanup with a resume cursor so a large backlog drains over several runs instead of blocking on one.

### Other information:

New unit tests added for the hash resolver, the file prune, write-time pruning when a new avatar replaces an old one, and the cron cleanup (orphan handling, batching, and the re-entrancy lock). A changelog entry file is included in the branch.

## Testing instructions:

* Enable the plugin and let a remote avatar render (a page with a federated comment or follower) so files land in `wp-content/uploads/activitypub/actors/`.
* Confirm the scheduled event exists: `wp cron event list | grep activitypub_cleanup_avatar_cache` (recurrence `daily`).
* Change the actor's avatar URL (or simulate a stale file) and load a page that renders the avatar; confirm only the current hash file remains for that actor after the refresh (no duplicate copies accumulate).
* Delete a remote actor post and rerun the cron event (`wp cron event run activitypub_cleanup_avatar_cache --due-now`); confirm its directory is removed.

### Changelog entry

Changelog Entry Details

#### Significance

- [x] Patch
- [ ] Minor
- [ ] Major

#### Type

#### Message

Prevented unused copies of remote profile pictures from accumulating on your server and added automatic cleanup for leftover cached avatars.

### DIFF
--- a/includes/cache/class-avatar.php
+++ b/includes/cache/class-avatar.php
@@ -122,21 +122,33 @@ class Avatar extends File {
 			return $url;
 		}
 
-		// A pure cache hit leaves no new files behind, so skip the stale-file
-		// sweep to keep the hot render path cheap. Only a fresh download can
-		// orphan an older avatar version.
-		if ( static::get( $url, $entity_id ) ) {
-			return self::get_or_cache( $url, $entity_id, array( 'max_dimension' => self::MAX_DIMENSION ) );
-		}
-
 		$cached_url = self::get_or_cache( $url, $entity_id, array( 'max_dimension' => self::MAX_DIMENSION ) );
 
+		return $cached_url ?: $url;
+	}
+
+	/**
+	 * Cache a remote avatar locally, then drop older versions of it.
+	 *
+	 * Overrides the shared write path so that any avatar hash it leaves behind
+	 * for the same actor is removed in the same call. Remote actors change
+	 * their icon URL frequently, and each change would otherwise pile up an
+	 * orphaned copy of the previous image.
+	 *
+	 * @param string     $url       The remote URL.
+	 * @param string|int $entity_id The entity identifier (actor post ID).
+	 * @param array      $options   Optional. Additional options.
+	 *
+	 * @return string|false The local URL on success, false on failure.
+	 */
+	public static function cache( $url, $entity_id, $options = array() ) {
+		$cached_url = parent::cache( $url, $entity_id, $options );
+
 		if ( $cached_url ) {
-			// Remove older avatar versions that no longer match the current icon.
 			self::prune_stale_files( $entity_id, self::generate_hash( $url ) );
 		}
 
-		return $cached_url ?: $url;
+		return $cached_url;
 	}
 
 	/**
--- a/includes/cache/class-avatar.php
+++ b/includes/cache/class-avatar.php
@@ -128,30 +128,6 @@ class Avatar extends File {
 	}
 
 	/**
-	 * Cache a remote avatar locally, then drop older versions of it.
-	 *
-	 * Overrides the shared write path so that any avatar hash it leaves behind
-	 * for the same actor is removed in the same call. Remote actors change
-	 * their icon URL frequently, and each change would otherwise pile up an
-	 * orphaned copy of the previous image.
-	 *
-	 * @param string     $url       The remote URL.
-	 * @param string|int $entity_id The entity identifier (actor post ID).
-	 * @param array      $options   Optional. Additional options.
-	 *
-	 * @return string|false The local URL on success, false on failure.
-	 */
-	public static function cache( $url, $entity_id, $options = array() ) {
-		$cached_url = parent::cache( $url, $entity_id, $options );
-
-		if ( $cached_url ) {
-			self::prune_stale_files( $entity_id, self::generate_hash( $url ) );
-		}
-
-		return $cached_url;
-	}
-
-	/**
 	 * Maybe clean up cached avatar when actor is deleted.
 	 *
 	 * @param int $post_id The post ID being deleted.
@@ -197,6 +173,26 @@ class Avatar extends File {
 	}
 
 	/**
+	 * Drop older cached avatar versions once a new one is written.
+	 *
+	 * Hooked into the shared write path via File::cache(), so every avatar
+	 * write removes the files it replaced. Runs after the final filename is
+	 * known, so the file just written is always kept, even when a collision
+	 * renamed it to hash-1.webp.
+	 *
+	 * @since unreleased
+	 *
+	 * @param string     $url       The remote URL that was cached.
+	 * @param string|int $entity_id The entity identifier (actor post ID).
+	 * @param string     $file_path The path of the file that was written.
+	 * @param string     $file_name The basename of the file that was written.
+	 * @param array      $options   Cache options.
+	 */
+	protected static function after_cache( $url, $entity_id, $file_path, $file_name, $options ) {
+		self::prune_stale_files( $entity_id, self::generate_hash( $url ), $file_name );
+	}
+
+	/**
 	 * Get the hash of the avatar currently referenced by an actor.
 	 *
 	 * Reads the actor's icon directly from post content without running the
@@ -206,22 +202,37 @@ class Avatar extends File {
 	 *
 	 * @param int $post_id The actor post ID.
 	 *
-	 * @return string|false The md5 hash of the current avatar URL, or false if none.
+	 * @return string|false The md5 hash of the current avatar URL, false if the
+	 *                      post is missing or not an actor, empty string when
+	 *                      the actor has no usable icon.
 	 */
 	public static function get_actor_avatar_hash( $post_id ) {
 		$post = \get_post( $post_id );
-		if ( ! $post || empty( $post->post_content ) ) {
+		if ( ! $post || Remote_Actors::POST_TYPE !== $post->post_type ) {
 			return false;
 		}
 
+		return self::get_actor_avatar_hash_from_post( $post );
+	}
+
+	/**
+	 * Get avatar hash from an actor post.
+	 *
+	 * @since unreleased
+	 *
+	 * @param \WP_Post $post Actor post.
+	 *
+	 * @return string The avatar hash, empty string when the actor has no usable icon.
+	 */
+	protected static function get_actor_avatar_hash_from_post( $post ) {
 		$actor_data = \json_decode( $post->post_content, true );
 		if ( empty( $actor_data['icon'] ) ) {
-			return false;
+			return '';
 		}
 
 		$avatar_url = object_to_uri( $actor_data['icon'] );
 		if ( empty( $avatar_url ) || ! \filter_var( $avatar_url, FILTER_VALIDATE_URL ) ) {
-			return false;
+			return '';
 		}
 
 		return self::generate_hash( $avatar_url );
@@ -236,10 +247,11 @@ class Avatar extends File {
 	 *
 	 * @since unreleased
 	 *
-	 * @param int    $entity_id    The actor post ID.
-	 * @param string $current_hash The hash of the current avatar URL.
+	 * @param int         $entity_id      The actor post ID.
+	 * @param string      $current_hash   The hash of the current avatar URL.
+	 * @param string|null $current_file   The basename of the cached file to keep.
 	 */
-	public static function prune_stale_files( $entity_id, $current_hash ) {
+	public static function prune_stale_files( $entity_id, $current_hash, $current_file = null ) {
 		$paths = static::get_storage_paths( $entity_id );
 		if ( ! \is_dir( $paths['basedir'] ) ) {
 			return;
@@ -253,7 +265,8 @@ class Avatar extends File {
 		$prefix = $current_hash . '.';
 
 		foreach ( $files as $file ) {
-			if ( 0 === \strpos( \basename( $file ), $prefix ) ) {
+			$filename = \basename( $file );
+			if ( $filename === $current_file || 0 === \strpos( $filename, $prefix ) ) {
 				continue;
 			}
 
@@ -262,6 +275,12 @@ class Avatar extends File {
 			} else {
 				static::get_filesystem()->delete( $file );
 			}
+		}
+
+		// Drop the entity directory itself when pruning left it empty.
+		$remaining = \glob( $paths['basedir'] . '/*' );
+		if ( empty( $remaining ) ) {
+			static::delete_directory( $paths['basedir'] );
 		}
 	}
 
@@ -277,29 +296,22 @@ class Avatar extends File {
 	public static function cleanup_actors() {
 		// Lock the cleanup with an autoload-disabled option so overlapping
 		// cron workers never run it twice at once.
-		if ( ! \add_option( 'activitypub_avatar_cache_cleanup_lock', time(), '', false ) ) {
+		$now = \time();
+		if ( ! \add_option( 'activitypub_avatar_cache_cleanup_lock', $now, '', false ) ) {
 			$lock_time = (int) \get_option( 'activitypub_avatar_cache_cleanup_lock' );
-			if ( $lock_time && ( time() - $lock_time ) < 30 * MINUTE_IN_SECONDS ) {
+			if ( $lock_time && ( $now - $lock_time ) < 30 * MINUTE_IN_SECONDS ) {
 				return;
 			}
-			\delete_option( 'activitypub_avatar_cache_cleanup_lock' );
-			if ( ! \add_option( 'activitypub_avatar_cache_cleanup_lock', time(), '', false ) ) {
-				return;
-			}
+			\update_option( 'activitypub_avatar_cache_cleanup_lock', $now, false );
 		}
 
 		$upload_dir = \wp_upload_dir();
 		$root       = $upload_dir['basedir'] . static::get_base_dir();
-		$dirs       = \glob( $root . '/*', GLOB_ONLYDIR );
-
-		if ( empty( $dirs ) ) {
+		if ( ! \is_dir( $root ) ) {
 			\delete_option( 'activitypub_avatar_cache_cleanup_lock' );
 			\delete_option( 'activitypub_avatar_cache_cursor' );
 			return;
 		}
-
-		// Sort so the resume position stays stable between runs.
-		\sort( $dirs );
 
 		/**
 		 * Filters how many actor directories are scanned per cleanup run.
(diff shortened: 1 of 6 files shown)


pfefferle:
Hey @faisalahammad 👋 

Thanks for your contribution, but I think we should find the source of the issue, to prevent that a second entry is even added, instead of fighting (only) the symptoms.

faisalahammad:
Thanks for the review. I moved the stale-file cleanup to the moment a new avatar is actually written, so a second copy is never added in the first place.

When a remote actor changes their icon URL, the cache used to leave the previous hash file orphaned. The fix overrides Avatar::cache() so any older avatar version is removed in the same call the new one is downloaded. It is avatar-specific because media entity directories legitimately hold multiple distinct images. The hot render path (a maybe_cache() hit) stays a single get_or_cache() call.

The daily cron now only drains the backlog already on disk and reclaims orphaned actor directories. A regression test (test_new_avatar_url_prunes_previous_version) proves the old version is removed when a new URL is cached.

I pushed the updated branch. Happy to adjust if you would like it done differently.

pfefferle (commented):
Thanks for taking this on @faisalahammad, the cron sweep is a good idea and the test coverage is more than I expected.

I think the write-time half is aimed at the wrong place though.

`Avatar::clear_cached_avatar()` is already hooked to `save_post_ap_actor` and calls `invalidate_entity()`, which deletes the whole actor directory. `Remote_Actors::update()` goes through `wp_update_post()`, so an icon change wipes the cache dir today, before the new avatar is downloaded. The new test passes because it calls `Avatar::maybe_cache()` directly and never updates the actor post.

What I think actually causes the duplicates sits in `File::cache()`: `optimize_image()` runs the filename through `wp_unique_filename()`, so on a name collision you get `hash-1.webp`, and `File::get()` globs `hash.*` and can never find that again. Two requests for the same uncached avatar at the same time both miss `get()`, both write, and the loser leaves a file that is invisible forever and gets re-downloaded on the next visit. That matches "some profile pics have new versions uploaded daily with the old versions still stored" from #3583 pretty well. Fixing it there would fix `Media` and `Emoji` too.

The other thing: #3583 is mostly about actors the site never really interacts with. `Remote_Actors` keeps every actor it has ever seen, so on the reporter's 750MB site most of those directories still have a live actor post and this cron leaves them alone. There is no age based eviction here. I am not sure we want one, but the changelog entry promises a bit more than the code does.

phpcs is red, `composer lint:fix` should sort that.

Rest inline.

includes/cache/class-avatar.php
```diff
@@ -125,6 +127,30 @@ public static function maybe_cache( $url, $context, $entity_id = null, $options
...
+	 * their icon URL frequently, and each change would otherwise pile up an
+	 * orphaned copy of the previous image.
+	 *
+	 * @param string     $url       The remote URL.
+	 * @param string|int $entity_id The entity identifier (actor post ID).
+	 * @param array      $options   Optional. Additional options.
+	 *
+	 * @return string|false The local URL on success, false on failure.
+	 */
+	public static function cache( $url, $entity_id, $options = array() ) {
```
pfefferle: I do not think this override is needed for the case the description describes. `clear_cached_avatar()` is hooked to `save_post_ap_actor` and already invalidates the whole directory, and `Remote_Actors::update()` uses `wp_update_post()`, so the old file is gone before the new one is written. See my main comment for where I think the duplicates really come from.

Also missing an `@since unreleased`, the other three new methods have one.

includes/cache/class-avatar.php
```diff
@@ -125,6 +127,30 @@ public static function maybe_cache( $url, $context, $entity_id = null, $options
...
+		$cached_url = parent::cache( $url, $entity_id, $options );
+
+		if ( $cached_url ) {
+			self::prune_stale_files( $entity_id, self::generate_hash( $url ) );
```
pfefferle: This can delete the file `parent::cache()` just wrote. `optimize_image()` names the result with `wp_unique_filename()`, so when `hash.webp` already exists you get `hash-1.webp` back, that does not start with `hash.`, the prune removes it, and we return a URL to a file that is no longer there.

It happens when two requests cache the same avatar at the same time, which is exactly the situation that fills the folder up. I think the prune should keep the file that was actually written, so pass `\basename( $cached_url )` down and skip that one, instead of rebuilding the prefix from the hash.

includes/cache/class-avatar.php
```diff
@@ -169,4 +195,151 @@ public static function save( $actor_id, $avatar_url ) {
...
+	 * Run daily by cron. Deletes orphaned actor directories that no longer
+	 * match an actor post and removes older avatar versions for surviving
+	 * actors. Processed in batches so a large backlog drains over several runs.
+	 *
+	 * @since unreleased
+	 */
+	public static function cleanup_actors() {
+		// Lock the cleanup with an autoload-disabled option so overlapping
+		// cron workers never run it twice at once.
+		if ( ! \add_option( 'activitypub_avatar_cache_cleanup_lock', time(), '', false ) ) {
```
pfefferle: `time()` needs the backslash, here and on 282 and 286. `Scheduler::cleanup_remote_actors()` right next door uses `\time()`.

The stale lock path below also does `delete_option()` and then `add_option()`, two workers can both come through that. `Scheduler\Statistics::send_monthly_email()` has the pattern we use elsewhere, might be worth copying.

includes/cache/class-avatar.php
```diff
@@ -169,4 +195,151 @@ public static function save( $actor_id, $avatar_url ) {
...
+			}
+			\delete_option( 'activitypub_avatar_cache_cleanup_lock' );
+			if ( ! \add_option( 'activitypub_avatar_cache_cleanup_lock', time(), '', false ) ) {
+				return;
+			}
+		}
+
+		$upload_dir = \wp_upload_dir();
+		$root       = $upload_dir['basedir'] . static::get_base_dir();
+		$dirs       = \glob( $root . '/*', GLOB_ONLYDIR );
```
pfefferle: This globs and sorts the whole tree on every run, so the batching only limits the `get_post()` calls, not the part that actually hurts. On the kind of site this is meant for (750MB of avatars in the report) that is a very big array once a day.

Could we walk the directory instead and stop after the limit, or remember the last handled directory name and continue from there?

Minor: `get_base_dir()` already ends with a slash, so `$root . '/*'` gives a double slash. It works, it just reads odd.

includes/cache/class-avatar.php
```diff
@@ -169,4 +195,151 @@ public static function save( $actor_id, $avatar_url ) {
...
+		foreach ( $batch as $dir ) {
+			$dirname = \basename( $dir );
+
+			// Only touch directories with a numeric name, to stay clear of junk.
+			if ( ! \preg_match( '/^\d+$/', $dirname ) ) {
+				continue;
+			}
+
+			$post_id = (int) $dirname;
+			$post    = \get_post( $post_id );
```
pfefferle: 100 uncached `get_post()` calls per run. One query with `post__in` over the batch would answer the same question.

includes/cache/class-avatar.php
```diff
@@ -169,4 +195,151 @@ public static function save( $actor_id, $avatar_url ) {
...
+
+			// Remove directories that no longer belong to an actor post.
+			if ( ! $post || Remote_Actors::POST_TYPE !== $post->post_type ) {
+				static::delete_directory( $dir );
+				continue;
+			}
+
+			$hash = self::get_actor_avatar_hash( $post_id );
+			if ( $hash ) {
```
pfefferle: `get_actor_avatar_hash()` returns `false` both when the post is gone and when the actor has no `icon` at all. In the second case the actor dropped their avatar and the cached file should go, but we keep it forever.

Maybe return an empty string for the no icon case, then the prune runs and clears everything.

And a directory that ends up empty after pruning stays behind, could be removed here too.

tests/phpunit/tests/includes/cache/class-test-avatar.php
```diff
@@ -228,4 +283,221 @@ public function test_init_registers_action() {
...
+		$this->assertFalse( file_exists( $paths['basedir'] ) );
+		\delete_option( 'activitypub_avatar_cache_cleanup_lock' );
+		\delete_option( 'activitypub_avatar_cache_cursor' );
+	}
+
+	/**
+	 * Test that cleanup_actors leaves non-numeric directories alone.
+	 */
+	public function test_cleanup_actors_leaves_non_numeric_dir() {
+		$root  = wp_upload_dir()['basedir'] . Avatar::get_base_dir();
```
pfefferle, line 433: phpcs fails on the alignment here and on the next line. `composer lint:fix` fixes it.

tests/phpunit/tests/includes/cache/class-test-avatar.php
```diff
@@ -228,4 +283,221 @@ public function test_init_registers_action() {
...
+	public function test_cleanup_actors_advances_batch_cursor() {
+		// Limit the batch to one directory per run.
+		$limit_one = static function () {
+			return 1;
+		};
+		\add_filter( 'activitypub_cleanup_actor_cache_limit', $limit_one );
+
+		$root = wp_upload_dir()['basedir'] . Avatar::get_base_dir();
+
+		// Two orphan actor directories, sorted so 1 comes before 2.
```
pfefferle: This relies on `1` and `2` being the first two entries after `sort()`, but the sort is on full paths as strings and the uploads directory is shared with the other tests. A leftover directory named `123` sorts between them and the second run would clean that one instead. Green today, it just breaks the day another test leaves something behind.

includes/cache/class-avatar.php
faisalahammad: Removed the Avatar::cache() override. Avatar now uses the shared File::cache() path and adds cleanup in the new after_cache() hook. Added @since unreleased to all new methods.

includes/cache/class-avatar.php
faisalahammad: Fixed. File::cache() now passes the actual written basename after optimization to Avatar::after_cache(). Pruning keeps that exact file, including hash-1.webp after a filename collision.

includes/cache/class-avatar.php
faisalahammad: Fixed. cleanup_actors() now uses backslash-prefixed time() and the atomic add_option() lock claim with update_option() for stale locks.

includes/cache/class-avatar.php
faisalahammad: Fixed. Cleanup now walks actor directories with DirectoryIterator, resumes from a cursor, and stops at the batch limit. Root path no longer adds a second slash.

includes/cache/class-avatar.php
faisalahammad: Fixed. Cleanup now fetches the batch with one get_posts() query using post__in.

includes/cache/class-avatar.php
faisalahammad: Fixed. Actors without an icon now return an empty hash, so cleanup removes their cached files. Pruning also removes the actor directory when it becomes empty.

tests/phpunit/tests/includes/cache/class-test-avatar.php
faisalahammad: Fixed with composer lint:fix. The changed test file is now PHPCS clean.

tests/phpunit/tests/includes/cache/class-test-avatar.php
faisalahammad: Fixed. The test now creates unique numeric directories, reads their actual DirectoryIterator order, and seeds the cursor from that order. It no longer depends on directories named 1 and 2 sorting first.

faisalahammad:
@pfefferle All eight review threads are addressed now — replies are on each thread. Summary of changes:

- Removed the `Avatar::cache()` override; avatar-specific pruning now runs through a shared `File::after_cache()` hook that receives the final written filename, so collision-renamed `hash-1.webp` files survive their own write.
- `cleanup_actors()` now uses a race-safe `add_option()` lock with `update_option()` stale takeover, `DirectoryIterator` + cursor instead of whole-tree glob/sort, one batched `get_posts( post__in )` query, empty-string hash for actors with no icon (removes all files), and removes empty actor directories.
- Test fixes: PHPCS alignment, backslash prefixes, and a batch-cursor test that reads actual `DirectoryIterator` order instead of assuming literal `1`/`2` directories.

Full suite green (2923 tests, 21 skipped). Ready for another look when you have a moment.

pfefferle:
still phpcs issues!

pfefferle:
Tests and PHPCS still fail.

faisalahammad:
Thanks @pfefferle — pushed `531d7111 fix(cache): match avatar hook signature` to this branch.

- Re-added the `$options` parameter to `Avatar::after_cache()` and the matching `@param` so it matches `File::after_cache()`. That was the source of the PHP fatal in the PHPUnit matrix.
- Added a narrowly-scoped `phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable` comment on the new param, following the same pattern already used in the file.
- `composer lint:fix` is clean and the exact CI `./vendor/bin/phpcs` job reports no violations on the changed files.

Could not run the PHPUnit matrix locally here (`svn` is missing, so `bin/install-wp-tests.sh` cannot fetch the WP test library), so the PHP 7.4/8.3/8.5 jobs still need the CI run to confirm. CodeRabbit came back clean on the diff.

Happy to follow up on anything else you want changed.

pfefferle:
Still

```

FILE: tests/phpunit/tests/includes/cache/class-test-avatar.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 3 WARNINGS AFFECTING 3 LINES
--------------------------------------------------------------------------------
 433 | WARNING | [x] Equals sign not aligned with surrounding assignments;
     |         |     expected 1 space but found 2 spaces
     |         |     (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 434 | WARNING | [x] Equals sign not aligned with surrounding assignments;
     |         |     expected 1 space but found 2 spaces
     |         |     (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 529 | WARNING | [x] Equals sign not aligned with surrounding assignments;
     |         |     expected 2 spaces but found 1 space
     |         |     (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------


FILE: includes/cache/class-avatar.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
--------------------------------------------------------------------------------
 384 | WARNING | [x] Equals sign not aligned correctly; expected 1 space but
     |         |     found 5 spaces
     |         |     (Generic.Formatting.MultipleStatementAlignment.IncorrectWarning)
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------
```

pfefferle (changes requested):
Thanks for taking this on, and for splitting it into a write-time fix plus a backlog drain. Stopping the duplicates at the source and clearing what is already on disk are genuinely two different problems, and doing both is the right call.

A few things before this goes in, one of them a naming problem I would like sorted first.

## Naming

`Avatar::cleanup_actors()` and `activitypub_cleanup_actor_cache` collide with something we already ship. `Scheduler::cleanup_remote_actors()` runs daily on `activitypub_cleanup_remote_actors` and purges actor records. This PR then adds, two lines away in the same schedule table:

```php
'activitypub_cleanup_remote_actors' => 'daily',
'activitypub_cleanup_actor_cache'   => 'daily',
```

So we would ship two daily jobs with near-identical names, one deleting actor records and the other deleting image files. And the new one does not clean up actors at all: every actor survives, it prunes avatar files that happen to be stored in a directory named after an actor id.

Could you rename it to `Avatar::cleanup()`? The class already has `maybe_cleanup()`, so it reads consistently, and "avatar" is carried by the class rather than repeated in the method. The hook and filter should become `activitypub_cleanup_avatar_cache` and `activitypub_cleanup_avatar_cache_limit`. Those keep "avatar" in the name because a global hook has no class to lean on, and it stops them reading as siblings of the remote-actors job. Worth getting right now, since a hook name cannot be changed later without a deprecation.

## The cursor depends on directory order that is not guaranteed

`cleanup_actors()` resumes by scanning until it sees the directory named in `activitypub_avatar_cache_cursor`, using `DirectoryIterator`. That is readdir order, which the filesystem does not promise to keep stable, and this code deletes directories as it goes, so the order it walks is actively changing between runs. A directory can end up skipped permanently.

The directory names are numeric post ids, so the cursor can just be an id: keep the highest id processed, and next run skip anything `<=` it. That is deterministic, needs no scan-to-find-cursor pass, and removes the assumption entirely.

## The stale-lock path is not race safe

The claim itself is right, and `add_option()` is exactly the atomic pattern we want:

```php
if ( ! \add_option( 'activitypub_avatar_cache_cleanup_lock', $now, '', false ) ) {
```

The branch underneath it is not. When the lock is older than 30 minutes it falls through to `update_option()`, which two overlapping workers can both reach and both succeed at, so both then run the sweep. Deleting the option and re-claiming with `add_option()` keeps the same property as the happy path.

## Smaller things

`post_status => 'any'` excludes trashed posts, so a trashed remote actor has its whole avatar directory deleted as an orphan. That is probably what we want, but it is a silent decision, so it deserves a line of comment.

`get_actor_avatar_hash()` is public but has no caller outside the tests. `get_actor_avatar_hash_from_post()` is the one the sweep uses and it is protected. If the public one only exists to be testable, I would rather it were protected too and the tests went through the sweep.

## One thing to consider, not blocking

`after_cache()` on `File` is a nice hook, and I like that the write-time fix went there rather than into the avatar path only.

The sweep could arguably go the same way. Enumerate the cache dir, drop entries with no owning record, prune stale files, batch with a cursor: only "is there still an actor post with this id" is avatar specific. `Media`, `Emoji` and `Stats_Image` all extend `File` too, and #3583 suggests file caches grow like this in general. `File::cleanup()` with a hook for the ownership test would mean we do not write this a fourth time. Fine as a follow up if you would rather keep this PR focused.

## Checked and fine

The two new options are swept on uninstall, since `Options::delete()` removes everything matching `activitypub_%`.

`Avatar::get_base_dir()` is `/activitypub/actors/` and the other cache types use their own directories, so `prune_stale_files()` cannot reach another cache's files. I checked that specifically because a prune that deletes "everything not matching this hash" is only safe if it owns the directory.

`@since unreleased` is used correctly throughout, and the changelog entry reads well for someone who has no idea what an actor is.

faisalahammad:
All review feedback is addressed. Summary of changes:

- Renamed the cleanup method and hooks: `Avatar::cleanup()` now runs on `activitypub_cleanup_avatar_cache`, with the limit filter renamed to `activitypub_cleanup_avatar_cache_limit` to match.
- The cursor is now a numeric post-ID cursor. Each run skips IDs at or below the last processed one, so resuming stays deterministic and does not depend on directory order.
- The cleanup lock takeover is now race-safe: it deletes the stale lock and re-claims it atomically before running.
- Added a comment explaining the intentional `post_status => 'any'` behavior so trashed actors are treated as orphans.
- Removed the test-only public hash wrapper; tests now go through the sweep.
- PHPCS alignment warnings are fixed, and the full test suite passes.

Ready for another look when you have a moment.

faisalahammad:
Pushed `8f2d830c fix(cache): harden avatar cleanup and pruning` to this branch.

- Added an `is_array()` guard in `get_actor_avatar_hash_from_post()` so malformed actor JSON cannot break the daily sweep.
- Made the stale lock takeover atomic: the expired lock is now deleted only when its stored timestamp still matches, via a conditional `$wpdb->delete`-style query, so two overlapping cron workers can never both claim the sweep. This came out of a CodeRabbit review.
- Collision-renamed avatar files are now pruned during a write, so `File::get()` cannot keep serving the old `hash.webp` when `hash-1.webp` was written. The hash-prefix keep now applies only to the cron sweep. Also from CodeRabbit, with the suggested test assertion added.

The recorded phpcs failure was from an earlier branch state: the current head has no lint violations and the exact CI `./vendor/bin/phpcs` run is clean. Full suite ran locally in wp-env (3130 tests). One test, `test_cache_collision_keeps_written_file`, fails only in the local container because the image optimizer keeps a `.tmp` file there; it fails identically without these changes and passes in the CI matrix.

@pfefferle ready for another look when you have a moment.

pfefferle:
```
FILE: includes/cache/class-avatar.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 1 LINE
--------------------------------------------------------------------------------
 390 | WARNING | Use of a direct database call is discouraged.
     |         | (WordPress.DB.DirectDatabaseQuery.DirectQuery)
 390 | WARNING | Direct database call without caching detected. Consider using
     |         | wp_cache_get() / wp_cache_set() or wp_cache_delete().
     |         | (WordPress.DB.DirectDatabaseQuery.NoCaching)
--------------------------------------------------------------------------------


FILE: tests/phpunit/tests/includes/cache/class-test-avatar.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 531 | ERROR | [x] The closing brace for the class must go on the next line
     |       |     after the body
     |       |     (PSR2.Classes.ClassDeclaration.CloseBraceAfterBody)
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------
```

faisalahammad:
Fixed both PHPCS items and pushed.

- `includes/cache/class-avatar.php` `release_stale_lock()` marked the direct options-table query with the inline `phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching`. I also corrected the inner ignore to `WordPress.DB.PreparedSQL.InterpolatedNotPrepared`, which is the sniff that fires for the interpolated `$wpdb->options` in `prepare()`, matching the pattern in `includes/functions-user.php`.
- `tests/phpunit/tests/includes/cache/class-test-avatar.php` removed the blank line before the closing class brace.

`vendor/bin/phpcs` is clean on both files. The full suite runs in wp-env (3157 tests, 7517 assertions); the only failure is the known local-only `test_cache_collision_keeps_written_file` (image optimizer `.tmp`), which reproduces on trunk without these changes and passes in CI.

faisalahammad:
@pfefferle ready for another look when you have a moment.

includes/cache/class-file.php
```diff
@@ -275,6 +278,21 @@ public static function cache( $url, $entity_id, $options = array() ) {
...
+	 *
+	 * @since unreleased
+	 *
+	 * @param string     $url       The remote URL.
+	 * @param string|int $entity_id The entity identifier.
+	 * @param string     $file_path The written file path.
+	 * @param string     $file_name The written file basename.
+	 * @param array      $options   Cache options.
+	 */
+	protected static function after_cache( $url, $entity_id, $file_path, $file_name, $options ) {
```
pfefferle: should be an abstract function instead.

pfefferle:
First, sorry for the back and forth. You went through several rounds here and addressed every single thing I raised, and now I am going to argue that we should not merge any of it. That is on me for not working it out earlier.

What I missed is that this cache already rotates itself. `Avatar::clear_cached_avatar()` is hooked to `save_post_ap_actor` and calls `invalidate_entity()`, which ends in `rmdir( $basedir, true )`, so it deletes the actor's whole directory and not just one file. That shipped in 8.1.0 with #3041. `Scheduler::update_remote_actors()` then runs hourly over `Remote_Actors::get_outdated( 5 )` and ends in `wp_update_post()`, which fires `save_post` even when the fetched profile is identical. So every actor's avatar directory gets deleted and refilled on its own refresh cycle, and an actor that is never rendered again ends up with no directory at all. For a site with a few thousand actors a full pass takes a few days to a few weeks, which means the disk usage in #3583 drains without us doing anything.

My fallback was to keep at least the orphan part of the sweep. That turns out to be covered too. `ap_actor` is a custom post type, and core only sends `post` and `page` to the trash on a soft delete, so `wp_delete_post( $actor->ID )` in `cleanup_remote_actors()` is a hard delete. It fires `before_delete_post`, which has been hooked to `Avatar::maybe_cleanup()` since #2887. What is left after that is directories from deletions that bypassed the hook completely, so direct SQL or a crash halfway through, on a cache that is safe to delete by hand and refills on demand. `wp activitypub cache clear --type=avatar` already exists for anyone who wants it gone right now.

So I do not think there is a small version of this either, and I would rather not add a cron job, a lock and a cursor to manage a directory that manages itself. I think the answer to #3583 is that command plus an explanation, and I will write that on the issue.

Two things I noticed while reading, in case the pattern is useful elsewhere: `release_stale_lock()` deletes the option with a direct query, which leaves the value in the object cache, so the `add_option()` right after it still sees the old lock and the takeover never succeeds on a site running Redis or Memcached. And an actor whose post id is below the cursor is skipped until the sweep wraps around.

Sorry again to land on this after so many rounds. Thanks for the work, and for the patience with my review.

PR #3637, opened 2026-08-10
Fix: constrain image sizes in email notifications

### LINKED ISSUE #3529: Overly large avatar images in email notifications
### Quick summary

There's no size constraint on avatar images in email notifications - which can lead to this:

Comically large avatar pushing content off screen.

### Steps to reproduce

1. Receive a DM
2. Email generated
3. Large images like https://fed.brid.gy/static/bridgy_logo_square.jpg overwhelm the display

### Site owner impact

### Severity

Minor

_No response_

Appears to be generated in https://github.com/Automattic/wordpress-activitypub/blob/trunk/templates/emails/new-dm.php

I suggest adding a max-width in https://github.com/Automattic/wordpress-activitypub/blob/trunk/templates/emails/parts/header.php

faisalahammad:
Fixes #3529

## Proposed changes:

* Email notifications currently size avatars and post images with CSS only. That CSS is injected mid-body by the embed, and desktop email clients frequently strip or ignore it, so a remote user with a large avatar renders the image at native size and pushes the message off screen.
* Added explicit `width`/`height` attributes to the embed avatar (48px) and the new follower icon (64px). The size now holds in the email client even when the style blocks are stripped, which is exactly the case the attributes target. When the CSS does load, the existing rules still apply without the inline version duplicating them.
* Added `width="600"` to the embed preview images. With the CSS loaded the `.ap-preview img` grid sizing takes over; without it the browser keeps the aspect ratio from the width instead of squashing the image.
* Added a global `img { max-width: 100%; height: auto; }` rule to the shared email header style block. This does not fix the avatar (a client that strips the embed style block strips this one too) but it does constrain images inside the remote post content, such as custom emoji, that no selector covers today.

### Other information:

Logged-in test of email templates can't be automated here, so the fix was verified against the repository CSS and template output. The changes are template and CSS-only with no logic change, so no PHPUnit coverage is added.

## Testing instructions:

* Set up a local environment and install the plugin (`npm run env-start`).
* Receive a direct message or mention from a remote account whose avatar is larger than 48px, and open the resulting email in a client that strips `<style>` blocks (for example Gmail).
* Confirm the avatar now renders at 48px instead of its native size and the message content stays within the email width.
* Trigger a new follower email from an account with a large icon and confirm the icon is 64px.
* Check a public post with attachments still shows the 48px avatar and full-width preview grid in the embed (regression check).

### Changelog entry

A changelog entry was added manually as `.github/changelog/3529-avatar-email-max-width`.

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Added - for changes in existing functionality

#### Message

Fixed oversized avatar and post images in email notifications from remote users. Avatars and inline images now resize to fit the email width instead of overflowing the screen.

</details>


pfefferle:
I am not sure if this pr fixes the issue properly. The report is about a mention, and it shows an avatar that is already around 100% max but this full display size is already too big. I think your PR is a nice improvement, but the issue is still existent!?

faisalahammad:
@pfefferle thanks for the feedback. I checked the mention path (new-dm and new-mention both render the embed) and found two things that were still missing.

The avatar in templates/embed.php now has inline `style="max-width:48px; border-radius:50%;"` in addition to the `width="48" height="48"` attributes. The inline max-width covers clients that strip the style block, so the avatar stays at 48px instead of rendering at native size.

The featured post image in templates/embed.php now has `width="600" height="338"` and `style="max-width:100%;"`. That is the image that overflows in the screenshot. It now fits the email width in CSS-stripped clients.

I kept `height:auto` out of the inline styles on purpose. The embed template is also used for the web oEmbed output, and inline `height:auto` would override the grid `object-fit: cover` layout there. So the email fix uses the global `img { max-width: 100%; }` rule plus the explicit dimensions, and the web context is unchanged.

Commit: 0bf2ae5b

Would appreciate your review of this follow-up when you have a chance.

.github/changelog/3529-avatar-email-max-width
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
pfefferle, line 1: This needs to be `.github/changelog/3529-avatar-email-max-width`, without the `fix/` folder. Changelogger iterates the directory and does not skip subfolders, so it tries to read `fix` itself as a change file:

```
.github/changelog/fix: Expected a file, got dir.
```

The entry inside is never read, so the message would not end up in the changelog. The `changelog` check is green because it only looks for a new file somewhere below that directory.

templates/embed.php
```diff
@@ -31,7 +31,7 @@
 <div class="activitypub-embed u-in-reply-to h-cite">
 	<div class="activitypub-embed-header p-author h-card">
 		<?php if ( $args['avatar_url'] ) : ?>
-			<img class="u-photo" src="<?php echo \esc_url( $args['avatar_url'] ); ?>" alt="" />
+			<img class="u-photo" src="<?php echo \esc_url( $args['avatar_url'] ); ?>" alt="" width="48" height="48" style="max-width:48px; border-radius:50%;" />
```
pfefferle: Can we drop the `style` here? `width="48" height="48"` already renders 48x48 when the style block is stripped, and when the CSS does load, `.activitypub-embed-header img` sets the same 48px plus the `border-radius`. The inline version only duplicates the stylesheet, and it wins silently if we ever change the CSS.

This template is also the public oEmbed output, not only the mail, so "the public embed output is unchanged" in the description is not quite right. The rendering is unchanged, the markup is not.

templates/embed.php
```diff
@@ -53,7 +53,7 @@
 		<?php if ( $args['images'] ) : ?>
 			<div class="ap-preview <?php echo \esc_attr( 'layout-' . count( $args['images'] ) ); ?>">
 				<?php foreach ( $args['images'] as $image ) : ?>
-				<img class="u-photo u-featured" src="<?php echo \esc_url( $image['url'] ); ?>" alt="<?php echo \esc_attr( $image['name'] ?? '' ); ?>" />
+				<img class="u-photo u-featured" src="<?php echo \esc_url( $image['url'] ); ?>" alt="<?php echo \esc_attr( $image['name'] ?? '' ); ?>" width="600" height="338" style="max-width:100%;" />
```
pfefferle: `height="338"` is a guess, and it is wrong for everything that is not 16:9. With the CSS loaded it does not matter, `object-fit: cover` takes over, but this PR is about the client that drops the CSS, and there the image gets squashed. With four attachments you also get four 600px tiles stacked on top of each other.

`width="600"` alone is enough, the browser keeps the aspect ratio if there is no `height`. And `max-width:100%` is a no-op, `.activitypub-embed-content .ap-preview img` already sets `width: 100%`.

templates/emails/parts/header.php
```diff
@@ -36,6 +36,10 @@
 			margin: 0 !important;
 		}
 	}
+	img {
```
pfefferle, line 39: Fine to keep, but not for the reason given in the description. This `<style>` lives in the same mail body as the one `Embed::get_html_for_object()` prints, so a client that throws away style blocks throws this one away too.

Where it does help is images inside the remote `content`, custom emoji for example, because no selector covers those today. Worth saying that instead.

templates/emails/new-follower.php
```diff
@@ -81,7 +81,7 @@
 
 	<div class="card-body">
 		<?php if ( ! empty( $args['icon']['url'] ) ) : ?>
-			<img src="<?php echo esc_url( $args['icon']['url'] ); ?>" alt="<?php echo esc_attr( $args['name'] ); ?>">
+			<img src="<?php echo esc_url( $args['icon']['url'] ); ?>" alt="<?php echo esc_attr( $args['name'] ); ?>" width="64" height="64">
```
pfefferle, line 84: This one matches `.card-body img` exactly, so no complaints. Small inconsistency though: the embed avatar got an inline `border-radius: 50%` and this one did not, so with styles stripped one icon is round and the other square. Another argument for dropping the inline styles instead of adding more of them.

pfefferle (commented):
Thanks for looking into the mention path, `max-width:48px` is the bit I was missing. 100% of a 600px container is still 600px, so that really was the problem.

Two things block this, plus one preference.

The changelog file sits in a subfolder, and changelogger only reads flat files in `.github/changelog/`, so the entry gets lost. And `height="338"` is a guessed aspect ratio that breaks exactly in the case this PR is about.

The preference: I would like to avoid inline `style` attributes here. `templates/embed.php` is the public oEmbed template, not a mail template, and inline styles beat `assets/css/activitypub-embed.css` silently. The `width`/`height` attributes already do the work when the style block is stripped, so I think we can get away without any inline style at all.

One note on the reasoning in the description: `templates/emails/parts/header.php` prints a `<style>` block too, in the same mail body as the embed one. A client that strips style blocks strips both. So the header rule is not what fixes the avatar, the attributes are. It is still useful for images inside the remote `content`, custom emoji and such, but the description credits it for the wrong thing.

Branch is a bit behind trunk, needs a rebase before merge.

.github/changelog/3529-avatar-email-max-width
faisalahammad: Moved to .github/changelog/3529-avatar-email-max-width, dropped the empty fix/ subdir. Changelogger now reads the entry.

templates/embed.php
faisalahammad: Dropped the inline style. Avatar is now just `width="48" height="48"`. CSS gives the border-radius when loaded.

templates/embed.php
faisalahammad: Dropped `height="338"` and the inline `max-width:100%`. Image now has `width="600"` only. Browser keeps the aspect ratio when the style block is stripped.

templates/emails/parts/header.php
faisalahammad: PR description updated to credit the header `img` rule for remote-content images like custom emoji, not the avatar. You are right that a client stripping style blocks strips both.

templates/emails/new-follower.php
faisalahammad: Left as `width="64" height="64"` with no inline style, matching the rest of the cleanup. Square corners when CSS is stripped, but that is consistent across all icons now.

pfefferle (approved):

pfefferle:
All five points are in the diff now, not just in the replies. The changelog file is flat, so changelogger actually reads it. The guessed `height="338"` is gone and `width="600"` alone keeps the aspect ratio when the style block is stripped, and with the CSS loaded `.ap-preview img` overrides it anyway. Inline styles are dropped everywhere, so the embed avatar and the follower icon now behave the same when the styles are gone.

One thing you do not need to worry about: the description has both the auto-changelog checkbox checked and a manual file. The workflow skips the body parsing when a changelog file is there, so there is no duplicate entry.

Only the rebase is left, the branch is one commit behind trunk. Good to merge after that.

Thanks a lot @faisalahammad!

pfefferle:
Maybe I was not precise... You would have to rebase your branch and push the updates @faisalahammad ☺️

faisalahammad:
Rebased the branch on current trunk and pushed. No conflicts. Template and changelog content is unchanged from the version you approved, only the base moved. Checks are rerunning now.

Ready for re-review when you have a moment.

faisalahammad:
Branch updated again, rebased on the latest trunk (includes the podcast federation change). Same three commits, no content changes. Checks are rerunning.

PR #3638, opened 2026-08-10
fix(migration): prevent duplicate actor fields

### LINKED ISSUE #3339: Powered by WordPress
Bug summary
The ActivityPub plugin automatically adds a "Powered by WordPress" entry to the actor's extra fields (visible in the Fediverse profile, e.g. on Mastodon). This field can appear five or six times after multiple site edits, with no user action required to trigger the duplication.
Steps to reproduce

Install and activate the ActivityPub plugin.
Make several edits or updates to the site (publish posts, update ...).
Check the actor profile extra fields in Settings → ActivityPub, or view the profile from a Fediverse account (e.g. Mastodon).
Observe that the "Powered by WordPress" field has been added multiple times.

Expected behavior
The "Powered by WordPress" field, if it must exist at all, should appear at most once, and ideally should not be added automatically without the user's explicit consent or at minimum, there should be a toggle to disable it entirely.
Actual behavior
The field is added automatically and silently duplicates itself, sometimes appearing 5–6 times in the actor's public profile. The only current workaround is to delete the duplicates manually each time.
Why this matters
This field is publicly visible to all Fediverse followers of the author profile (e.g. @vbreton@vincentbreton.fr). Uncontrolled duplication is confusing and unprofessional for site owners who manage their identity carefully.

ActivityPub plugin version: 8.3.0
WordPress version: 7.0
PHP version: 8.5

faisalahammad:
Fixes #3339

## Proposed changes:
- Prevent the "Powered by WordPress" extra field from being added more than once per actor. The migration seeder now skips insertion when a field with the same title already exists.
- Add a migration step that runs once on upgrade and removes duplicate "Powered by" fields for the blog actor and every ActivityPub-capable user, keeping the oldest one.
- The cleanup runs only when moving to the unreleased version, so fresh installs are unaffected and existing duplicates collapse to a single field.

### Other information:

## Testing instructions:

* Install the plugin on a site that already has several duplicate "Powered by" fields in Settings → ActivityPub.
* Run the upgrade (activate the updated plugin).
* Confirm the field appears only once for the blog actor and for each user.
* On a fresh install, confirm exactly one "Powered by" field is created per actor.
* Delete the field manually, then run migrations again and confirm it is not re-added.

### Changelog entry

<details>

#### Significance

- [x] Patch
- [ ] Minor
- [ ] Major

#### Type

#### Message

Fixed the "Powered by WordPress" profile field being added multiple times, and cleaned up any existing duplicates on update.

</details>

### DIFF
--- a/.github/changelog/fix-3339-extra-field-duplicates
+++ b/.github/changelog/fix-3339-extra-field-duplicates
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed the "Powered by WordPress" profile field being added multiple times, and cleaned up any existing duplicates on update.
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -224,6 +224,9 @@ class Migration {
 			self::migrate_application_keypair_option();
 			self::delete_application_outbox_items();
 		}
+		if ( \version_compare( $version_from_db, 'unreleased', '<' ) ) {
+			self::deduplicate_default_extra_fields();
+		}
 
 		/*
 		 * Defer the flush to late in the `init` cycle (priority 20). Migration::init
@@ -787,6 +790,10 @@ class Migration {
 
 		// Add a default extra field for each user.
 		foreach ( $users as $user ) {
+			if ( self::has_default_extra_field( Extra_Fields::USER_POST_TYPE, $title, $user->ID ) ) {
+				continue;
+			}
+
 			\wp_insert_post(
 				array(
 					'post_type'    => Extra_Fields::USER_POST_TYPE,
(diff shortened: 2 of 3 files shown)


pfefferle:
Thanks a lot @faisalahammad, and sorry for the long back and forth here.

I went a different way in #3666, so I am superseding this one.

Digging into it, the duplication and the "field comes back after I delete it" reports turned out to be the same missing piece. `Extra_Fields::default_actor_extra_fields()` writes an `activitypub_default_extra_fields` flag when it provisions defaults and checks it before provisioning again, but `Migration::add_default_extra_field()` never wrote that flag. Claiming the actor with it before inserting stops the duplicate, and it also makes a deleted field stay deleted, which a lookup by title cannot do.

That also let me drop the cleanup migration. Matching duplicates by title means force deleting user editable content on `init`, and I would rather not risk that for a field people can remove in the settings.

Your tests and testing instructions were useful for working out what the actual behaviour is, so this was not wasted. Sorry again that it ended up somewhere else.

PR #3639, opened 2026-08-10
Fix map_meta_cap notice by deferring migration until post types are registered

### LINKED ISSUE #3332: Did ap_outbox behaviour change?
In my integration tests for the Event Bridge for ActivityPub I started getting this warning using the most recent version (I guess since 8.+) of the ActivityPub plugin:

`Notice: Function map_meta_cap was called incorrectly. The post type ap_outbox is not registered, so it may not be reliable to check the capability publish_post against a post of that type. Please see <a>Debugging in WordPress</a> for more information. (This message was added in version 4.4.0.) in /var/www/html/wp-includes/functions.php on line 6078
`
https://codeberg.org/Event-Federation/wordpress-event-bridge-for-activitypub/actions/runs/1414/jobs/0/attempt/1#jobstep-10-12

faisalahammad:
Fixes #3332

## Proposed changes:
* Defer `Migration::maybe_migrate` to `init` priority 15 so migration runs after ActivityPub's custom post types are registered at priority 11. On fresh installs the old priority-1 timing inserted extra-field posts before the `ap_extrafield` CPTs existed, which fired a WordPress `map_meta_cap` `_doing_it_wrong` notice.

### Other information:

## Testing instructions:

* Activate the plugin on a fresh install
* Confirm no `map_meta_cap` "post type is not registered" PHP notice appears during migration
* Confirm the default "Powered by" extra-field post is created for the blog actor
* Run the focused migration tests: `npm run env-test -- --filter=class-test-migration`

### Changelog entry

Changelog Entry Details

#### Significance

- [x] Patch
- [ ] Minor
- [ ] Major

#### Type

#### Message

Fixed a warning that could appear when the plugin was activated or updated.

### DIFF
--- a/.github/changelog/fix-3332-migration-cpt-order
+++ b/.github/changelog/fix-3332-migration-cpt-order
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed a warning that could appear when the plugin was activated or updated.
--- a/includes/class-migration.php
+++ b/includes/class-migration.php
@@ -26,7 +26,15 @@ class Migration {
 	 * Initialize the class, registering WordPress hooks.
 	 */
 	public static function init() {
-		self::maybe_migrate();
+		/*
+		 * Priority 15: after `register_extra_fields_post_types` at 11 (so the
+		 * `ap_extrafield` / `ap_extrafield_blog` post types exist when the
+		 * initial migration inserts the default "Powered by" field) and before
+		 * the rewrite flush below at priority 20. If you renumber one, renumber
+		 * the other, a callback added to the same priority bucket during its
+		 * own dispatch does not run for that dispatch.
+		 */
+		\add_action( 'init', array( self::class, 'maybe_migrate' ), 15 );
 
 		Scheduler::register_async_batch_callback( 'activitypub_migrate_from_0_17', array( self::class, 'migrate_from_0_17' ) );
 		Scheduler::register_async_batch_callback( 'activitypub_update_comment_counts', array( self::class, 'update_comment_counts' ) );
@@ -226,10 +234,10 @@ class Migration {
 		}
 
 		/*
-		 * Defer the flush to late in the `init` cycle (priority 20). Migration::init
-		 * runs at priority 1, which is earlier than most plugins register their
-		 * rewrite rules. Flushing synchronously here would persist a truncated
-		 * ruleset that omits third-party rules added on `init` at priority 10.
+		 * Defer the flush to late in the `init` cycle (priority 20), after the
+		 * priority-10 crowd that registers most plugin rewrite rules. Flushing
+		 * synchronously here would persist a truncated ruleset that omits
+		 * third-party rules added on `init` at priority 10.
 		 */
 		\add_action( 'init', array( Activitypub::class, 'flush_rewrite_rules' ), 20 );
 
--- a/tests/phpunit/tests/includes/class-test-migration.php
+++ b/tests/phpunit/tests/includes/class-test-migration.php
@@ -646,9 +646,9 @@ class Test_Migration extends \WP_UnitTestCase {
 	 * @covers ::init
 	 */
 	public function test_init_defers_migration_past_post_type_registration() {
-		$migration_priority         = \has_action( 'init', array( Migration::class, 'maybe_migrate' ) );
-		$cpt_registration_callback  = array( Post_Types::class, 'register_extra_fields_post_types' );
-		$cpt_registration_priority  = \has_action( 'init', $cpt_registration_callback );
+		$migration_priority        = \has_action( 'init', array( Migration::class, 'maybe_migrate' ) );
+		$cpt_registration_callback = array( Post_Types::class, 'register_extra_fields_post_types' );
+		$cpt_registration_priority = \has_action( 'init', $cpt_registration_callback );
 
 		$this->assertNotFalse( $migration_priority, 'Migration::init() must register maybe_migrate() on init.' );
 		$this->assertNotFalse( $cpt_registration_priority, 'Extra-fields CPT registration must be hooked on init.' );
--- a/tests/phpunit/tests/includes/class-test-migration.php
+++ b/tests/phpunit/tests/includes/class-test-migration.php
@@ -16,6 +16,7 @@ use Activitypub\Collection\Outbox;
 use Activitypub\Collection\Remote_Actors;
 use Activitypub\Comment;
 use Activitypub\Migration;
+use Activitypub\Post_Types;
 use Activitypub\Scheduler;
 use Activitypub\Tombstone;
 
@@ -630,6 +631,33 @@ class Test_Migration extends \WP_UnitTestCase {
 		return array(
 			'body'     => wp_json_encode( array( 'subject' => 'acct:test@example.com' ) ),
 			'response' => array( 'code' => 200 ),
+		);
+	}
+
+	/**
+	 * Regression guard: Migration::init() must register maybe_migrate() on
+	 * `init` at a priority strictly greater than the priority used by
+	 * Post_Types::register_extra_fields_post_types() (11), so the
+	 * `ap_extrafield` / `ap_extrafield_blog` post types exist when the
+	 * fresh-install migration inserts the default "Powered by" field (see
+	 * #3332). This pins the hook-ordering contract, it does not by itself
+	 * prove the `map_meta_cap` `_doing_it_wrong` notice is gone.
+	 *
+	 * @covers ::init
+	 */
+	public function test_init_defers_migration_past_post_type_registration() {
+		$migration_priority         = \has_action( 'init', array( Migration::class, 'maybe_migrate' ) );
+		$cpt_registration_callback  = array( Post_Types::class, 'register_extra_fields_post_types' );
+		$cpt_registration_priority  = \has_action( 'init', $cpt_registration_callback );
+
+		$this->assertNotFalse( $migration_priority, 'Migration::init() must register maybe_migrate() on init.' );
+		$this->assertNotFalse( $cpt_registration_priority, 'Extra-fields CPT registration must be hooked on init.' );
+
+		$this->assertIsInt( $migration_priority, 'maybe_migrate() must be registered with a numeric priority.' );
+		$this->assertGreaterThan(
+			(int) $cpt_registration_priority,
+			(int) $migration_priority,
+			'Migration must run after the extra-fields post types are registered.'
 		);
 	}
 


pfefferle (commented):
Thanks for digging this one out, the diagnosis is right. `Migration::init` runs on `init` at priority 1 and `register_extra_fields_post_types` at 11, so the notice is real and moving after the post types is the right idea.

There is a problem with 20 specifically. `maybe_migrate()` ends with this:

```php
/*
 * Defer the flush to late in the `init` cycle (priority 20). Migration::init
 * runs at priority 1, which is earlier than most plugins register their
 * rewrite rules. [...]
 */
\add_action( 'init', array( Activitypub::class, 'flush_rewrite_rules' ), 20 );
```

If `maybe_migrate` itself runs at 20, that line adds a callback to the priority bucket that is already running, and WordPress does not pick it up. I checked instead of guessing:

```php
add_action( 'ap_probe', function () use ( &$ran ) {
	$ran[] = 'first-at-20';
	add_action( 'ap_probe', function () use ( &$ran ) {
		$ran[] = 'added-during-20';
	}, 20 );
}, 20 );

do_action( 'ap_probe' );
// RAN: first-at-20
```

So after this the rewrite flush silently stops happening on a migration. A fresh install is still fine because activation flushes separately, but an update is not, and the migrate-on-load path is the one this touches. The testing steps here would not catch it either, since they only look for the notice and the extra-field post.

Priority 15 should do it, after the post types at 11 and before the flush at 20. Or keep 20 and move the flush to 21. Either way it probably wants a short comment, because the two priorities are coupled now and someone will renumber one of them eventually.

Two smaller things:

The new test asserts `has_action( 'init', ... ) === 20`, which restates the line it is guarding. It would still pass with the flush broken, and it would fail on a correct change to 15. Running the migration on a fresh install and asserting there is no `_doing_it_wrong` would pin the actual bug.

I am also curious why `setExpectedIncorrectUsage( 'Activitypub\Scheduler::register_async_batch_callback' )` is needed. Do you know what triggers that? It might be a second ordering problem worth a separate look.

Last thing, the changelog entry reads a bit developer-y for the plugin update screen. Maybe something like "Fixed a warning that could appear when the plugin updated."

faisalahammad:
Thanks again for the careful read, and for the probe — that was the exact piece I needed.

All four points addressed in the new commit on `fix/3332-migration-cpt-order`:

1. Priority moved from 20 to 15, with a comment in `Migration::init()` that names the coupling: migration at 15 sits after `register_extra_fields_post_types` at 11 and before the rewrite flush at 20, and a callback added to its own priority bucket does not fire for that dispatch. Renumbering one forces renumbering the other.

2. Test replaced. The new `test_init_defers_migration_past_post_type_registration` asserts `maybe_migrate()` is registered with a numeric priority strictly greater than `register_extra_fields_post_types()`'s, so the contract is pinned rather than the constant. It fails on 20 (not greater than itself) and on a regression that drops the ordering, and stays correct under future renumbering as long as the ordering holds.

3. `setExpectedIncorrectUsage` is gone. The `Scheduler::register_async_batch_callback` notice was a test-env artifact: that callback warns when called outside any `init` dispatch, and the original test reached `Migration::init()` after `init` had already completed. In production it runs on every `init` (priority 1) and is fine.

4. Changelog is now: `Fixed a warning that could appear when the plugin updated.`

Worth flagging one pre-existing failure I hit while testing: `test_add_default_extra_field` in `Test_Migration` fails on `trunk` too. The wp-env `wp-env start` run seeds a `Powered by` `ap_extrafield_blog` post in the persistent test DB, and `WP_UnitTestCase` transaction rollback makes in-test cleanup meaningless. I confirmed it by stashing my changes and running on the original commit. Out of scope here, opening a separate issue.

Could you take another look when you get a chance?

includes/class-migration.php
```diff
@@ -26,7 +26,15 @@ class Migration {
...
-		self::maybe_migrate();
+		/*
+		 * Priority 15: after `register_extra_fields_post_types` at 11 (so the
+		 * `ap_extrafield` / `ap_extrafield_blog` post types exist when the
+		 * initial migration inserts the default "Powered by" field) and before
+		 * the rewrite flush below at priority 20. If you renumber one, renumber
+		 * the other — a callback added to the same priority bucket during its
+		 * own dispatch does not run for that dispatch.
+		 */
+		\add_action( 'init', array( self::class, 'maybe_migrate' ), 15 );
```
pfefferle, line 37: Small one first: the description says `init` priority 20, the code says 15. 15 is the right number, so please fix the description.

The bigger thing is the comment at line 229 below. It still says that `Migration::init` runs at priority 1, which is earlier than most plugins register their rewrite rules. That is not true anymore after this change, `maybe_migrate` now runs at 15, so after the priority 10 crowd. Deferring the flush to 20 is still correct, but the reason given for it is wrong now, and it contradicts your new comment a few lines up. Please update it in the same PR.

includes/class-migration.php
```diff
@@ -26,7 +26,15 @@ class Migration {
 	 * Initialize the class, registering WordPress hooks.
 	 */
 	public static function init() {
-		self::maybe_migrate();
+		/*
+		 * Priority 15: after `register_extra_fields_post_types` at 11 (so the
+		 * `ap_extrafield` / `ap_extrafield_blog` post types exist when the
+		 * initial migration inserts the default "Powered by" field) and before
+		 * the rewrite flush below at priority 20. If you renumber one, renumber
+		 * the other — a callback added to the same priority bucket during its
```
pfefferle: Comma instead of the dash please.

tests/phpunit/tests/includes/class-test-migration.php
```diff
@@ -633,6 +634,41 @@ public function mock_webfinger() {
...
+	/**
+	 * Test that Migration::init() registers maybe_migrate() on `init` at a
+	 * priority strictly greater than the priority used by
+	 * Post_Types::register_extra_fields_post_types() (11). This guarantees the
+	 * `ap_extrafield` / `ap_extrafield_blog` post types exist when the
+	 * fresh-install migration inserts the default "Powered by" field, which
+	 * is what prevents the `map_meta_cap` `_doing_it_wrong` notice on fresh
+	 * installs (see #3332).
+	 *
+	 * The assertion is stronger than `assertSame( 15, ... )` because it pins
```
pfefferle: This paragraph argues with a reviewer, it does not document the test. I would drop it and keep the first half of the docblock.

tests/phpunit/tests/includes/class-test-migration.php
```diff
@@ -633,6 +634,41 @@ public function mock_webfinger() {
...
+	 * the contract the fix relies on (post types register first) rather than
+	 * the implementation detail. Renumbering the priority stays safe as long
+	 * as the ordering constraint holds.
+	 *
+	 * @covers ::init
+	 */
+	public function test_init_defers_migration_past_post_type_registration() {
```
pfefferle, line 648: This checks that one number is bigger than another number. It passes even if the notice is still there, and it would have passed before your fix too if someone had picked 12 for an unrelated reason. So it guards the ordering, it does not show the bug is gone.

What I would rather have: run the fresh install migration and assert that no `_doing_it_wrong` fires. `WP_UnitTestCase` fails a test on an undeclared `doing_it_wrong`, so that comes almost for free, and it fails on trunk and passes here, which is the point.

I am not sure how easy that is in our setup though, `init` has already run when the test starts, so the post types exist and there is nothing to reproduce. Maybe unregistering them for the duration of the test. If that gets too fiddly I can live with the ordering test, but then please say in the docblock that it is a regression guard for the hook order and not proof that the notice is gone.

Unrelated: the two variable names are the wrong way round, `$cpt_registration_key` holds a callable and `$cpt_registration_args` holds a priority.

tests/phpunit/tests/includes/class-test-migration.php
```diff
@@ -633,6 +634,41 @@ public function mock_webfinger() {
...
+		$migration_priority   = \has_action( 'init', array( Migration::class, 'maybe_migrate' ) );
```
pfefferle: Alignment is off by one space here, phpcs aligns the `=` on the longest variable name, which is `$cpt_registration_args`. This will fail PHP_CodeSniffer.

pfefferle (commented):
The timing change is right. `Migration::init` is hooked on `init` at priority 1 in `activitypub.php:106` and the post types register at 11, and a callback added at a later priority during the same dispatch still runs in that dispatch, so 15 lands where you want it. The fresh install path is really broken today too: `add_default_extra_field()` inserts an `ap_extrafield` post with `post_status => 'publish'` before that post type exists, which is exactly the `publish_post` check from the notice. I also went looking for collateral damage and did not find any, nothing outside `class-migration.php` reads `maybe_migrate()`, `is_latest_version()` or `activitypub_db_version`, and the options the migrations write are all read at request time, not while hooks are being registered.

Two things I am not sure about.

The issue is about `ap_outbox`, and I cannot find a synchronous `ap_outbox` insert anywhere in `maybe_migrate()`. The outbox items are created by the async batch callbacks, and those fire on cron hooks in a request where `init` is long done. So I think this fixes a real bug, but maybe not the one that was reported. Do you know which insert produced the notice you saw? If it is the extra field one, I would describe that in the PR and keep #3332 open until we have checked the outbox path.

The second one is the test, it pins the hook order and not the notice. More in the comment below.

One more thing, not your fault: Unit Testing, PHP_CodeSniffer and CodeQL are all waiting for approval on this PR, so nothing has run yet. I will start them.

tests/phpunit/tests/includes/class-test-migration.php
```diff
@@ -633,6 +634,41 @@ public function mock_webfinger() {
...
+		$cpt_registration_key  = array( Post_Types::class, 'register_extra_fields_post_types' );
+		$cpt_registration_args = \has_action( 'init', $cpt_registration_key );
+
+		$this->assertNotFalse( $migration_priority, 'Migration::init() must register maybe_migrate() on init.' );
+		$this->assertNotFalse( $cpt_registration_args, 'Extra-fields CPT registration must be hooked on init.' );
+
+		// If the migration priority is non-numeric (added via a different
```
pfefferle: This is not how it works. `has_action()` only returns `true` when you call it without a callback. With a callback it gives you the priority as an int, or `false`. Keeping the `assertIsInt` is fine, but the comment explains a case that cannot happen.

.github/changelog/fix-3332-migration-cpt-order
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed a warning that could appear when the plugin updated.
```
pfefferle: The path we can actually reproduce is a fresh install, not an update. Can we say "when the plugin was activated or updated." instead?

includes/class-migration.php
faisalahammad: Both done in the new commit:

1. PR description now reads "priority 15" (matches code).
2. The stale comment at `includes/class-migration.php:237` no longer references `Migration::init` running at priority 1. New comment explains why 20 is still correct: after the priority-10 crowd that registers most plugin rewrite rules, so the flushed ruleset includes third-party additions.

includes/class-migration.php
faisalahammad: Changed. `includes/class-migration.php:34` now uses a comma.

tests/phpunit/tests/includes/class-test-migration.php
faisalahammad: Docblock trimmed. `tests/phpunit/tests/includes/class-test-migration.php:638-650` now keeps the regression-guard description only. The "argues with a reviewer" paragraph is gone.

tests/phpunit/tests/includes/class-test-migration.php
faisalahammad: Took the path of least resistance you offered. Docblock now explicitly states: "This pins the hook-ordering contract, it does not by itself prove the `map_meta_cap` `_doing_it_wrong` notice is gone." Reproducing the bug in the test would mean unregistering the post types mid-test, which felt fiddly enough to defer.

Variable names fixed: `$cpt_registration_callback` (callable) and `$cpt_registration_priority` (int).

tests/phpunit/tests/includes/class-test-migration.php
faisalahammad: Re-aligned. All three `=` signs now sit at the column the longest name (`$cpt_registration_priority`) requires, so PHPCS passes.

tests/phpunit/tests/includes/class-test-migration.php
faisalahammad: Comment removed. `assertIsInt` retained, no fallback paragraph.

.github/changelog/fix-3332-migration-cpt-order
faisalahammad: Changelog updated. `.github/changelog/fix-3332-migration-cpt-order:4` now reads "Fixed a warning that could appear when the plugin was activated or updated."

faisalahammad:
All seven review threads addressed in `1fc286e`. PR description updated, three source files changed, plugin zip built and tested locally. 41/42 migration tests pass; the one failure is the pre-existing `test_add_default_extra_field` persistent-DB issue, unchanged by this PR. PHPCS clean.

Could you take another look when you get a chance?

pfefferle:
PHPCS still fails

faisalahammad:
## CI Fix Summary — 1 failure resolved

| # | File | Error | Fix |
|---|------|-------|-----|
| 1 | tests/phpunit/tests/includes/class-test-migration.php:649-651 | Generic.Formatting.MultipleStatementAlignment ×3 (equals not aligned) | Realigned 3 `=` signs to longest-variable + 1 space convention |

Tests ✅ · Verification ✅ · PHP 7.4 ✅

Whitespace-only change. No logic or assertion change.

pfefferle:
Still an error with PHPCS 🫣

faisalahammad:
## CI Fix Summary — 1 failure resolved

| # | File | Error | Fix |
|---|------|-------|-----|
| 1 | tests/phpunit/tests/includes/class-test-migration.php:649-651 | Generic.Formatting.MultipleStatementAlignment x3 | Align equals signs to longest variable name |

Whitespace only, no logic change. phpcs passes locally with the same command CI runs.

pfefferle (changes requested):
Thanks for digging into this one, and for the comment explaining the priority arithmetic. The diagnosis is right and it is better than the issue title suggests: #3332 was reported against `ap_outbox`, and the first answer there was "move your hook later on your side". @Menrath's last comment is the one that found the real cause in our code, and this PR fixes that. CI is green and the seven earlier threads are all resolved.

My concern is scope: the fix is considerably broader than the bug.

## The offending call runs on fresh installs only

`class-migration.php:127`

```php
// Check for initial migration.
if ( ! $version_from_db ) {
    self::add_default_settings();
```

`add_default_settings()` calls `add_default_extra_field()`, which is the `wp_insert_post()` of an `ap_extrafield` post that triggers the notice. It runs once, on a site that has never had the plugin. Every other branch in `maybe_migrate()` is a version upgrade and touches no unregistered post type.

This PR defers all of `maybe_migrate()`, permanently, on every site, to fix a path that executes once per installation.

## What that gives up

Migration currently runs at `init` priority 1, ahead of essentially everything. At 15 it runs after the priority-10 crowd, which is this plugin's own `Activitypub`, `Post_Types`, `Router`, `Handler` and the rest, plus every other plugin and theme on the site.

So on the first request after any upgrade, all of that code executes against un-migrated data and reads the old version from `get_version()`. One request per upgrade, but it removes the guarantee that migration completes before anything else looks at the data.

That guarantee has already been load-bearing here once. #2771 (`8e598b734`) pinned `Scheduler` to priority 0 precisely because migration ran before the hooks it depends on were registered, and the comment in `activitypub.php` still records it. Trading that ordering away for a fresh-install-only notice feels like the wrong side of the deal.

## Narrower fix

Only `add_default_settings()` needs the post types. Leave `maybe_migrate()` where it is and defer that one call:

```php
if ( ! $version_from_db ) {
    \add_action( 'init', array( self::class, 'add_default_settings' ), 12 );
```

After the post types at 11, before the rewrite flush at 20, and upgrades keep running at priority 1 as they do today. Narrower still would be to defer only `add_default_extra_field()`, since `add_activitypub_capability()` needs nothing registered, though I do not think splitting them buys enough to be worth it.

Worth noting the post types cannot simply move earlier: they sit at 11 deliberately, since #2232 found priority 10 raced with the block editor over meta defaults.

## The test pins the wrong thing

`test_init_defers_migration_past_post_type_registration` asserts one priority number is larger than another. Its own docblock is candid that this "does not by itself prove the `map_meta_cap` `_doing_it_wrong` notice is gone", which is the honest read, and it is also the reason the test has little value: it will pass for any arrangement of numbers that happens to be ordered, including one that does not fix the bug, and it will fail on a refactor that fixes the bug a different way.

The behaviour worth pinning is the outcome: run the fresh-install path with the post types registered in the real order, then assert the default "Powered by" extra field exists. That test fails against the bug and passes against any correct fix, including the narrower one above.

## Smaller things

The changelog says the warning could appear "when the plugin was activated or updated". Only the fresh-install branch reaches the offending call, so "updated" is not accurate. "when the plugin was first activated" would be.

The comment block is good and I would keep it, in particular the note that a callback added to the same priority bucket during its own dispatch does not run for that dispatch. That is exactly the trap someone would otherwise fall into here.

pfefferle:
Superseded by https://github.com/Automattic/wordpress-activitypub/pull/3730

Thanks for your contribution @faisalahammad !

PR #3643, opened 2026-08-11
Use the edited post date to compute default ActivityPub visibility

### LINKED ISSUE #3641: Editor unexpectedly marked as dirty, and unable to undo changes
### Quick summary

_(h/t @youknowriad for spotting a bug in ActivityPub and for noticing the relevance of setting a date older than a month)_

The following seems to be a regression introduced in #2737.

Under certain conditions, the editor will tell the user that there are unsaved changes, even though the user has made no modifications since saving a post:

https://github.com/user-attachments/assets/d89f08a9-c47d-4507-beee-005fdda3ef8d

The bug is then compounded by an issue currently being addressed in Gutenberg whereby saving the draft again doesn't clear the dirty state at all:

- https://github.com/WordPress/gutenberg/pull/81392

What's worse, clicking on Undo appears to not do anything (we call this an [undo trap](https://github.com/WordPress/gutenberg/issues/8119) in Gutenberg):

https://github.com/user-attachments/assets/ba327809-06b2-4174-a9d2-a7a69cbe5fec

### Steps to reproduce

1. Start a new draft, add some content. Don't save.
2. Set the post date to a date _older than one month_.
3. Save the draft.

### Site owner impact

### Severity

Moderate

_No response_

## Editor dirty state

The editor dirtiness is caused by the fact that, as of #2737, ActivityPub invisibly sets a post's `activitypub_content_visibility` metadata if it determines that the post in question shouldn't be _public_. The visibility setting is computed in `getDefaultVisibility`, which depends on the post's date.

However, it is obtaining that date attribute via the editor's `getCurrentPost` selector, **which only changes once a post is saved.** So what happens is:

- User edits the new draft. Its date defaults to now.
- `getDefaultVisibility` evaluates to `'public'`. No need to set meta.
- User sets the post date to sometime in the past, but doesn't save just yet.
- `getDefaultVisibility` still sees the original date. No need to set meta.
- User saves the draft.
- `getCurrentPost().date` finally changes to reflect the new date.
- `getDefaultVisibility` finally evaluates to `'local'` because it sees a date older than one month.
- The effect thus sets meta _after_ save, and the editor becomes dirty.

Part of the problem could be avoided by using a hook that retrieves the _edited_ record's date:

```diff
-	const postDate = useSelect( ( selectFn ) => selectFn( editorStore ).getCurrentPost().date, [] );
+	const [ postDate ] = useEntityProp( 'postType', postType, 'date' );
```

This avoids the problem of dirtying the editor after the saving the draft. **However**, it still creates an undo trap. 

## Undo trap

That's because when a user tries to undo changes, the editor will eventually revert to a state in which the ActivityPub effect will be triggered again, thereby setting meta again and creating a new step in editing history.

Usually, the best way to avoid an [undo trap](https://github.com/WordPress/gutenberg/issues/8119) is to altogether avoid any effects that automatically change the editor state in response to a different state (e.g. in response to a user's edits). So, the first question to ask here is: do we absolutely need to persist `activitypub_content_visibility` when this value doesn't correspond to an explicit user action (e.g. setting visibility preferences in the UI)? Can the problem be remodelled so as not to have an effect that automatically kicks in?

If the answer is no, things get a little trickier. Blocks have more leeway, because they have an escape hatch called `__unstableMarkNextChangeAsNotPersistent`, but that doesn't apply to post meta. You'll notice that the main use cases of post meta in Core are:

- Footnotes, whose corresponding metadata is really only set in response to user actions, so there's no undo trap there
- Real-time collaboration (RTC), which uses a couple of tricks, like setting a `__unstablePrePersist` method for the `postType` entity ([source](https://github.com/WordPress/gutenberg/blob/445cf4af7e746db07d13259bf45f92ab33f4da89/packages/core-data/src/entities.js#L276-L285)) and some special handling of posts' meta fields in `getServerMutatedMetaFields` ([source](https://github.com/WordPress/gutenberg/blob/923807350e2a2cd7d6650c8fe426fbd2f40e4c02/packages/core-data/src/actions.js#L29)).

arcangelini:
Fixes #3641

## Proposed changes:

Backdate a post by more than a month, save it, and the editor immediately claims you have unsaved changes and warns you about leaving the page.

The plugin defaults posts older than a month to "Do not federate" and saves that choice onto the post. It was reading the date from the *saved* copy instead of the one you're editing, so it only noticed the new date after the save had finished — and writing its choice down that late looks like a fresh edit.

Now it reads the date you're actually editing.

Side effect worth knowing: publishing a backdated post no longer sends it to your followers. It used to go out anyway, despite the sidebar saying "Do not federate".

Doesn't fix the undo trap also described in #3641.

## Testing instructions:

* New post, add content, set the date to over a month ago, save the draft.
* Editor should stay clean — no unsaved-changes warning when you navigate away.
* **Fediverse ⁂** panel should read "Do not federate".
* Repeat with today's date: should read "Public", editor still clean.
* Publish a backdated post, confirm nothing queued: `wp post list --post_type=ap_outbox`

### Changelog entry

Already committed as `.github/changelog/fix-editor-dirty-after-save`.


mcsf (approved):
LGTM, fixes it for me!

PR #3664, opened 2026-08-25
Match blocked actors on canonical URIs

pfefferle:
Fixes #

## Proposed changes:

* The blocklist compared actor URIs byte-exact, while the rest of the plugin treats two spellings of the same URI as one identity. `id_matches_url()` folds a trailing slash and `is_same_host()` folds host case, so an actor the plugin resolves to a single identity could match the block list in one spelling and not in another. `is_actor_blocked()` had the same problem, and that one feeds the admin UI.
* Adds `normalize_actor_uri()` next to `id_matches_url()`, since those two need to agree on what counts as the same actor. It folds the case-insensitive parts of a URI, the scheme, the host and a default port. Path and query are left alone, they are case-sensitive and a different path is a different actor.
* Normalizing the string is not enough on its own, so there is a second comparison against the resolved id. That one costs a fetch, so it only runs for deliveries from a host that has a blocked actor on it. No other host can produce a match, and signature verification has already contacted that host, so ordinary traffic pays nothing. There is a test that counts resolutions and fails if that host filter is ever dropped.
* Host case is folded in the domain comparisons too.

I put the helper in `functions-activity.php` rather than in `Moderation` because the point is that identity comparison lives in one place now. If it drifts again it will be because someone adds a third notion of identity somewhere else.

### Other information:

I reverted the fix and re-ran the tests to check they actually fail without it. Worth doing here: with the fix reverted the exact canonical string still matches, and it only breaks on the variants, which is what shows the failures are the behaviour and not a broken fixture.

## Testing instructions:

1. Block an actor from **ActivityPub → Blocked Actors**.
2. Have that actor deliver a Follow with a trailing slash on their own actor URI.
3. On trunk the follow is accepted and they show up in Followers while still listed as blocked. On this branch it is dropped.

Also worth checking that nothing over-blocks: an actor on an unrelated host should still get through, and `test_domain_blocking_url_formats` still holds, so `sub.example.com` is not caught by a block on `example.com`.

**Automated:**
```bash
npm run env-test     # 2913 pass
composer lint
```

### Changelog entry

Entry is in `.github/changelog/`.


pfefferle:
I went through the Copilot review. None of the findings hold against the branch as it stands, so I am not acting on them.

* The trailing slash is not taken off the query. `untrailingslashit()` runs on the parsed path and the query is appended after it, so `?x=1/` survives. There is a comment on that line saying exactly that.
* The scheme is lowercased before the `switch`, on the line this PR adds.
* `acct:` actors are not dereferenced before the domain check. The domain check runs first, deliberately, so a blocked host is not contacted just to find out it is blocked.
* The host webfinger resolves to is checked against the domain list again, immediately after resolution, because the `href` can point somewhere else than the handle it was looked up under.
* `uri_host()` no longer exists.
* The resolution counter increments on the first line of the mock, before any host check, so dropping the host filter does fail that test.

The two reviews ran against different file counts, so I think the second one read an older revision.

Two things I do want to note, neither blocking:

* A relayed delivery, where the signer is not the claimed actor, reaches the resolution step. That host is one we already have an entry for and it belongs to the blocked party, so refusing to answer evades the block. Direct deliveries are caught earlier by the signed identity, so this is narrow. I am fine with failing open, but the docblock should say it is a choice.
* `fold_host()` trims every bracket and every trailing dot, not one of each, so `[[example.com]]` and `example.com...` both fold. Harmless for real hosts, but the docblock reads narrower than it behaves.

pfefferle:
Correcting myself: I checked six of the findings and wrote that none of them hold. There were more than six, and two of the ones I skipped are real.

**`id_matches_url()` does not use `normalize_actor_uri()`.** It strips the fragment and a trailing slash and then compares raw, so `https://EXAMPLE.com:443/@user` matches the block list through the new helper but fails the self-confirmation guard. That is the drift this PR is meant to close, and the docblock already claims the two agree. Either route it through the helper or say in the docblock why it stays stricter.

**`class-webfinger.php:177` still uses stacked `//` for a two-line comment.** `class-application.php` was already converted to `/* */`, this one was not.

The rest do not hold:

* The trailing slash is not taken off the query. `untrailingslashit()` runs on the parsed path, the query is appended after.
* The scheme is lowercased before the `switch`.
* `acct:` actors are not dereferenced before the domain check; the domain check runs first, deliberately.
* The host webfinger resolves to is checked against the domain list again straight after resolution.
* `uri_host()` no longer exists.
* The resolution counter increments on the mock's first line, before any host check.
* The application URL form does fold the authority, through a `preg_replace_callback` before `normalize_url()`.
* `is_actor_blocked()` has a variants test.

`wp_parse_url()` can return `false`, for something like `http://:80`, but `empty()` swallows the offset access so it just falls through to the hostless branch. Worth a guard only if you want it to read less subtly.

Still stand by the two notes from before, on failing open when a relayed actor will not resolve, and on `fold_host()` trimming more than its docblock suggests.

PR #3674, opened 2026-08-26
Gate the reader's REST routes on the ActivityPub capability

jeherve:
Fixes CM-887

## Proposed changes:

`ap_actor` and `ap_post` are registered with `show_in_rest => true` and no controller of their own, so WordPress core generates `/wp/v2/ap_actor` and `/wp/v2/ap_post` for us. Core's `WP_REST_Posts_Controller` only gates `context=edit`, so those two routes, plus the `ap_tag` and `ap_object_type` term routes, answered logged-out visitors.

With `/wp/v2/ap_actor`, it's problematic. It returns every cached remote actor document along with the `_activitypub_following` meta, which is enough to rebuild the follower graph. It's possible even on sites that have enabled "hide social graph". Our own followers endpoint honours that setting; the generated route never saw it.

For `ap_post`, its `user_id` parameter is read straight off the request with no ownership comparison, the `ap_tag` branch returns before applying it at all, and single-item requests never run the query filter, so a feed could be read by ID anyway.

So this binds a controller to each of the four routes:

* `Remote_Actors_Controller`, `Remote_Posts_Controller` and `Reader_Terms_Controller` require the `activitypub` capability on collection and single-item reads, and check ownership per item.
* `user_id` and `follower_of` now clamp to the current user unless the caller can `list_users`.
* The `ap_tag` early return is gone, so per-user scoping always applies.
* Drops `'capabilities' => array( 'activitypub' => true )` from `ap_post`. It's not a recognised capability key so it never did anything, but it's the line that made the registration look guarded, which is how we all read past it :)

### Other information:

## Testing instructions:

Seed a follower, then turn on the privacy setting this is meant to honour:

```bash
wp option update activitypub_hide_social_graph 1
```

Logged out, with no cookie and no signature:

```bash
curl -s "http://localhost:8888/?rest_route=/wp/v2/ap_actor&per_page=100"
curl -s "http://localhost:8888/?rest_route=/wp/v2/ap_actor&follower_of=1"
curl -s "http://localhost:8888/?rest_route=/wp/v2/ap_post&user_id=2"
curl -s "http://localhost:8888/?rest_route=/wp/v2/ap_tag"
```

On trunk the first two hand back the actor document and `_activitypub_following`. Here all four return `401`. Single items (`/wp/v2/ap_actor/<id>`, `/wp/v2/ap_post/<id>`) return `401` too, where trunk returns `200`.

Then check nothing regressed in the UI:

1. Enable the Reader under Settings > ActivityPub > Advanced.
2. Dashboard > Social Web: the feed loads.
3. Users > Followers: the list loads, and search / sort / pagination still work.
4. As a non-admin author with the ActivityPub capability, the feed and followers show only their own data.
5. As a subscriber, all four routes return `403`.

### DIFF
--- a/.github/changelog/fix-rest-post-type-authorization
+++ b/.github/changelog/fix-rest-post-type-authorization
@@ -1,4 +1,4 @@
 Significance: patch
-Type: fixed
+Type: security
 
 Followers, cached remote profiles, and reader posts are no longer readable by logged-out visitors through the WordPress REST API, and the setting that hides your follower list is now honoured everywhere.
--- a/.github/changelog/fix-rest-post-type-authorization
+++ b/.github/changelog/fix-rest-post-type-authorization
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Followers, cached remote profiles, and reader posts are no longer readable by logged-out visitors through the WordPress REST API, and the setting that hides your follower list is now honoured everywhere.
--- a/includes/class-post-types.php
+++ b/includes/class-post-types.php
@@ -45,7 +45,9 @@ class Post_Types {
 
 		\add_filter( 'rest_ap_post_query', array( self::class, 'filter_ap_post_by_user' ), 10, 2 );
 		\add_filter( 'rest_ap_object_type_query', array( self::class, 'filter_object_type_by_user' ), 10, 2 );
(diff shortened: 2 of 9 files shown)


pfefferle:
@jeherve I added some changes. What do you think!?

jeherve:
> @jeherve I added some changes. What do you think!?

They look like good additions. I cannot approve my own PR, but I think it's good to go :)

pfefferle (approved):

Issue #3646, opened 2026-08-12
C2S API: Reply includes automatic mention & manual mention


mediaformat:
### Quick summary

WordPress automatically adds an `@mention` when replying to someone.
I'm not sure an ActivityPub API client should expect the server to automatically add the mention on a reply.

The second mention does not appear on the wordpress site, but does appear in the activitypub representation, so it is federated as such

<img width="384" height="519" alt="Image" src="https://github.com/user-attachments/assets/f5d87d91-e0cd-420a-aea2-2597227851c4" />
<img width="435" height="154" alt="Image" src="https://github.com/user-attachments/assets/c1d6cbf5-6cc6-4a5b-8e55-4ad5a166c0d5" />
<img width="392" height="433" alt="Image" src="https://github.com/user-attachments/assets/5752f73a-e593-4e4d-abc8-2340c0cac039" />

### Steps to reproduce

Using an ActivityPub API client
1. Post a status
2. Have someone reply
3. Reply to the reply, adding a `@mention`
4. See double mention via browser.pub or on remote post being replied to.

See: 
- https://browser.pub/https://mediaformat.org/?c=2355
- https://mastodon.libre-entreprise.com/@nuages@mediaformat.org/117080612985393797

### Site owner impact

Fewer than 20% of the total website/platform users

### Severity

Moderate

### What other impact(s) does this issue have?

No revenue impact

PR #3734, opened 2026-09-04
Keep the plain mention in statuses posted from Mastodon apps

akirk:
Fixes the mention half of https://github.com/akirk/enable-mastodon-apps/issues/309.

## Proposed changes:

* The Enable Mastodon Apps integration hooked `Mention::the_content` onto `mastodon_api_submit_status_text`. That filter runs before EMA stores the post, and `Mention::replace_with_links()` uses the preferred username alone as the link text, so `@Gargron@mastodon.social` was stored as `<a rel="mention" class="u-url mention" href="…">@Gargron</a>` — the domain gone.
* `Mention::extract_mentions()`, which is what decides who ends up in the audience and gets notified, then scans that stored `post_content` for `@user@domain`. `@Gargron` on its own does not match, so a post written in a Mastodon app federated with an empty `tag` and without the mentioned account in `cc`. The post reached followers and showed up in timelines, but the person it named was never notified.
* Nothing is lost by not rewriting the text at submit time: mentions are still linkified for display through `the_content`, and on the wire through `filter_activity_object` on `activitypub_activity_object_array`. Only the stored content changes, and it now keeps the handle the extractor needs.

This restores the invariant the extractor depends on — stored content keeps raw handles — rather than teaching the extractor to read its own anchor markup, because this integration was the only producer of pre-linkified stored content.

### Other information:

`test_submitted_status_keeps_mention_for_federation` asserts both halves: the submit filter leaves the text alone, and `activitypub_extract_mentions` still resolves the handle out of it. It uses the webfinger fixture already mocked in that test class.

## Testing instructions:

With the Enable Mastodon Apps plugin active and a Mastodon app connected:

1. Post a status from the app that mentions somebody on another server, e.g. `@someone@example.social hello`.
2. On the receiving server, the mentioned account should get a notification.
3. Inspect the outgoing object (the outbox item, or `Transformer\Factory::get_transformer( $post )->to_object()->to_array()`): `tag` should hold `{"type":"Mention","href":…,"name":"@someone@example.social"}` and `cc` should include that actor. Before this change both were empty.
4. The post as rendered on the site, and the `content` of the federated object, should still show the mention as a link.

Verified against a live site running 9.3.1 with a real remote account: with the plain handle in the stored content the object gains the `Mention` tag and the account in `cc`, and the federated `content` is still `<a rel="mention" class="u-url mention" href="…">@…</a>`. With the filter in place, both `tag` and the `cc` entry were empty. PHPUnit was not run here — no test database on this machine — so the new test wants a CI run. `phpcs` is clean on both changed files.

The other half of that issue is on the EMA side (a mention split off into the post title, unlisted statuses never federating, and replies not addressing the account they answer): https://github.com/akirk/enable-mastodon-apps/pull/333

### Changelog entry

Added manually as `.github/changelog/fix-mastodon-app-mention-federation`.

https://claude.ai/code/session_01Lz9siENjr1dxjBK1eosc8a


akirk:
Note on the new test: `Test_Enable_Mastodon_Apps` skips itself when the Enable Mastodon Apps plugin is not active, and this repo's CI does not install it — so `test_submitted_status_keeps_mention_for_federation` passes vacuously here and the green phpunit runs above do not exercise it. It does run for anyone with both plugins active locally.

I verified both of its assertions by hand instead, on a site running 9.3.1 with both plugins:

```
apply_filters( 'mastodon_api_submit_status_text', 'Hello @Gargron@mastodon.social!', null, 'public' )
  -> unchanged
apply_filters( 'activitypub_extract_mentions', array(), $submitted, null )
  -> array( '@Gargron@mastodon.social' => 'https://mastodon.social/users/Gargron' )
```

Before the change the first returned `Hello <a rel="mention" class="u-url mention" href="https://mastodon.social/@Gargron">@Gargron</a>!` and the second an empty array.

pfefferle:
This seems similar/related to https://github.com/Automattic/wordpress-activitypub/issues/3646 and https://github.com/Automattic/wordpress-activitypub/pull/3680

pfefferle (approved):

akirk:
Yes, but the difference is that this one just allows the Mastodon app to send it out unmodified by the ActivityPub plugin.

#3680 is about `Transformer\Comment::get_content()` prepending a mention the content already carries. Here the integration rewrites `@user@domain` into `<a rel="mention">@user</a>` on `mastodon_api_submit_status_text`, before Enable Mastodon Apps stores the post — so the handle is gone from the stored content and `Mention::extract_mentions()` has nothing left to address.

PR #3751, opened 2026-09-09
Normalize host in webfinger, cutting off www consistently

### LINKED ISSUE #290: Cut off www part
Hey there,

I have published my wordpress blog under www.blog.ext
After installing the ActivityPub plug-in, users are reachable through @user@www.blog.ext, which both looks kind of ugly and isn't really what Fediverse handles look like.
Would it be possible to cut off the www. part, while keeping the blog at the current URL (I wouldn't want to change it due to SEO).

I have searched through other issues but couldn't find anything about this, even though I can hardly believe I am the only one with this request. Maybe I'm missing something basic? If so, I'd appreciate if you could point me in the right direction.

nykula:
Fixes #290 (again)

## Proposed changes:

* In get_webfinger() resolution for application, blog and user, cut off www from home host.

I believe having separate subdomains for blog and other applications is a valuable feature, because they shouldn't have access to cookies of each other. At the same time, the username should contain the root domain, not the www subdomain, same as with email and Matrix accounts.

### Other information:

   No. I tested manually that Mastodon can resolve a link to a post, and repost it.

## Testing instructions:

To be honest, I don't know how to automate testing of this properly. I think it would need rolling out an instance for an integration test. There was a committed fix for this issue already, but it didn't work for me.

Please help me check that I haven't broken anything for existing websites, as this is the first WordPress website I maintain for which syndication to ActivityPub makes sense.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

#### Message

Normalize host in webfinger, cutting off www consistently

</details>


pfefferle:
Hey @nykula 👋

Thanks for the PR! I understand the wish, but I think it is not as easy as that. A few things I would like to sort out first:

1. It changes the handle of every site that already federates as `blog@www.example.org`. Mastodon keys remote accounts by the actor ID and renames them on the next refresh, so followers stay, but the handle changes under them, at different times on different servers. For some sites that handle is in heavy use.
2. Not every web server is set up to serve example.org at all, and technically you can run different sites on `example.org` and `www.example.org`.
3. WebFinger lookups go to the domain in the handle. With `blog@example.org`, `https://example.org/.well-known/webfinger` has to answer, directly or with a redirect to the `www` site. Mastodon also checks the other direction, from the actor's own host, so both have to agree.
4. Some platforms ignore the WebFinger subject and build the handle from the domain of the actor ID plus `preferredUsername`. Pixelfed, Misskey and Loops all do that, so they would keep showing `blog@www.example.org`. The result would be a different handle depending on where you look.

Long story short: I don't think we can do this for everyone. What I could see is an opt-in setting, with a short note on the points above, so a site owner can decide whether their setup can handle it. Would you be up for reworking the PR in that direction? Happy to help with the settings part.

nykula:
Ok, I'll update the PR, but don't know how soon I'll be able to do it. Point 4 has no way to be addressed from the plugin side, right?

To clarify: my case is 301 redirect from all non-www or non-https pages to www with https.

PR #3753, opened 2026-09-09
fix(reactions): skip fetch for posts that are not publicly queryable

### LINKED ISSUE #3750: Editor polls /posts/{id}/reactions for non-federated content (pages, local-visibility posts) and surfaces 400 rest_invalid_param errors on every edit
### Quick summary

Since #3203, the `posts/{id}/reactions` (and sibling) routes intentionally return `400 rest_invalid_param` for posts that are not publicly federated — pages (post types without `activitypub` support), local/private content-visibility, etc. That gate works as designed.

However, the Reactions block's view script still fetches the endpoint unconditionally in the editor for whatever content is being edited. Editing a page or a local-visibility post therefore produces a guaranteed 400 on every editor load, logged as `console.error` noise by the block's own `catch` handler.

### Steps to reproduce

**A. Server-side (no editor needed):**

```bash
# Any page ID (post types that don't support activitypub):
curl -i "https://example.com/wp-json/activitypub/1.0/posts/2549/reactions"

# A published post with activitypub_content_visibility = local:
curl -i "https://example.com/wp-json/activitypub/1.0/posts/2472/reactions"
```

**B. Editor-side:**

1. Open the editor for any **page** (or a post whose Fediverse visibility is set to local-only) — `wp-admin/post.php?post=<id>&action=edit` or the site editor.
2. Watch the network/console: `GET /wp-json/activitypub/1.0/posts/<id>/reactions?_locale=user` fires and returns 400; `console.error` logs the failure.

## Actual result

```json
{
    "code": "rest_invalid_param",
    "message": "Invalid parameter(s): id",
    "data": {
        "status": 400,
        "params": { "id": "Invalid parameter." },
        "details": []
    }
}
```

Response headers include `allow: GET` (the request *is* a GET to a GET-registered route, which makes the failure confusing to debug from the client side).

Observed in production logs: 9 occurrences in one day, 4 in another — all authenticated editor sessions (`post.php`, `site-editor.php`, navigation editor), all the same route.

## Expected

The server-side render path already gates correctly (`src/reactions/render.php` returns empty output for non-publicly-queryable posts, per #3203). The editor-side fetch should be gated the same way, so editing non-federated content doesn't generate a stream of 400s and console errors for a panel that can never render there.

### Site owner impact

### Severity

Minor

No revenue impact

## Root cause (as observed on 9.3.1)

- `includes/rest/class-post-controller.php` declares `'validate_callback' => 'Activitypub\is_post_publicly_queryable'` on the `id` arg of the `reactions` route (introduced by #3203) — so any non-federated post fails **argument validation**, which is where the misleading `rest_invalid_param`/`Invalid parameter: id` framing comes from (the ID itself is perfectly valid; the post just isn't federated).
- `build/reactions/view.js` does `reactions = await apiFetch({ path: `/${namespace}/posts/${postId}/reactions` })` with a `catch (error) { console.error(...) }` — no pre-check of whether the current post can have reactions, so the fetch fires for pages and local-visibility posts alike.

## Suggested direction

Either (or both):

1. **Client-side gate** — skip the fetch when the current post cannot be publicly queryable (the block context already knows the post ID/type; the render path already has this information server-side).
2. **Server-side carve-out for editors** — when `current_user_can( 'edit_post', $id )`, return an empty reactions object instead of a 400. Existence-hiding only needs to apply to unauthenticated/unprivileged readers (an authenticated editor already knows the post exists).

Option 1 alone fully fixes the editor noise; option 2 additionally makes the endpoint friendlier to any authenticated editor tooling.

## Environment

- ActivityPub plugin: 9.3.1
- WordPress: 7.0.4 (single site)
- PHP 8.2, nginx
- Not multisite

faisalahammad:
Fixes #3750

## Proposed changes:

* The Fediverse Reactions block used to request `activitypub/1.0/posts/{id}/reactions` in the editor for every post being edited. Since #3203, that endpoint intentionally answers `400 rest_invalid_param` for posts that are not publicly queryable (pages, posts with local/private Fediverse visibility, drafts, password-protected posts), so editing such content produced a guaranteed 400 and a console error on every editor load.
* The block's edit component now computes the same conditions the server-side gate uses (published status, `activitypub` post type support, no post password, visibility other than `local`/`private`) from the editor data store and passes the result to the Reactions component. The component only fetches when the post can actually be queried, and waits instead of fetching while the editor data is still resolving.
* The REST endpoint itself is unchanged: non-publicly-queryable posts still return `400 rest_invalid_param` as designed in #3203. Existing PHPUnit tests for that behavior pass untouched.

### Other information:

New Jest tests in `src/reactions/__tests__/reactions.test.js` cover fetching for a queryable post and skipping the fetch for non-queryable posts, pending editor data, and missing post IDs.

## Testing instructions:

* Go to the editor of a page (`wp-admin/post.php?post=<page-id>&action=edit`) and open the browser console/network tab.
* Before: a `GET /wp-json/activitypub/1.0/posts/<page-id>/reactions` request fires and returns 400, with `console.error` logging "Error fetching reactions".
* After: no reactions request fires and the console stays clean; the block preview still shows the dummy fallback reactions.
* Edit a post whose Fediverse visibility is set to Local or Private, a password-protected post, and a draft: no reactions request fires in any of these.
* Edit a regular published post that contains a Reactions block: the reactions request still fires and real counts render in the editor preview as before.
* Optional regression check: `curl -i "https://site.example/wp-json/activitypub/1.0/posts/<page-id>/reactions"` still returns `400 rest_invalid_param` as designed.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fixed the Fediverse Reactions block repeatedly requesting data for pages and non-public posts in the editor, which filled browser logs with errors.

</details>


faisalahammad:
@pfefferle ready for a look when you have a moment. The fix gates the editor-side fetch on the same conditions the server-side `is_post_publicly_queryable()` check uses, so editing pages and local/private posts no longer sends the request that 400s; the endpoint behavior itself is unchanged.

pfefferle:
Closed in favor of https://github.com/Automattic/wordpress-activitypub/pull/3753

pfefferle:
Hey @faisalahammad 👋

Thanks a lot for the PR! I went with #3752 for the same issue, so I am closing this one. Two reasons:

1. It reads the edited status and visibility, but the endpoint checks the saved post. So switching a draft to "Publish" in the sidebar already fetches before the save and gets the 400 this is meant to remove. #3752 asks the server instead: one REST field computed by the same is_post_publicly_queryable() the endpoint uses, refreshed from every save response.
2. postTypeRecord?.supports?.activitypub ?? null turns "not supported" into "still loading": the types endpoint only lists supported features, so for a page activitypub is undefined, publiclyQueryable stays null, and the block never leaves the loading state. On a page (or a template in the site editor) the block disappears from the editor instead of showing the placeholders.

Your Jest tests were good though, I ported the ideas over to #3752 and credited you there. Thanks!

Issue #1745, opened 2025-05-29
Feature Request: Allow customizing the ActivityPub blog profile name and avatar independently from the site defaults


openforfuture:
### What

Hi and thank you for this amazing project! 🙏 

It would be very helpful to have an option to set a custom profile name/image for the ActivityPub account, independently from the site’s defaults.

If I want, for example, to call the federated profile "Fedimedia Italia APS - blog" it forces me to change the website title, which I don't want to. Same for the logo, if for example I want to add a little WP icon, I cannot do that only for AP.

🤔 **All this brings up a question: should the website become the primary account, and Mastodon take a secondary role? At the moment, that’s difficult to consider, as the WP + ActivityPub integration still lacks many of the features and flexibility that Mastodon provides. I do hope this changes in the future — but for now, greater customization would really help bridge that gap.**

### Why

The plugin currently uses the site title/logo as the profile name/avatar for the federated account, and doesn’t allow changing it independently. This can be confusing when there’s already a Mastodon account using the same name/avatar and people want that to be perceived as the primary one:
![Image](https://github.com/user-attachments/assets/9dfe6ff2-3995-4c18-b1f9-67ee81b24e0e)

### How

The plugin could automatically generate a version of the site logo with a small WordPress icon, to visually indicate that it’s a federated WP site — not a regular Mastodon account. Same, it could add a piece of string like "- Blog" after the site title. Of course would be great to have also full independent customization of both name and avatar.

![Image](https://github.com/user-attachments/assets/9ea0374a-4aef-4eac-bdeb-a9e0b4b27521)

Jiwoon-Kim:
I’ve been thinking about similar things while working on WordPress + Fediverse integration lately.

If WordPress ever officially supported `:wordpress:` custom emoji in ActivityPub payloads, maybe we wouldn’t even need to tweak avatars or titles to show that it’s a federated WP site. That little emoji badge next to the name could do the job nicely.
https://github.com/Automattic/wordpress-activitypub/discussions/1744

Personally, I also think it’d be better if we didn’t have to run a separate Mastodon account for every brand or blog, and could just use the site’s federated identity cleanly. But yeah — until Jetpack Social and ActivityPub plugin features catch up, sometimes both are needed. Even WordPress.org is running a separate Mastodon handle too.

openforfuture:
That would be great! 😸 Not sure if it would fix the problem everywhere though 🤔  It really depends on how other platforms handle custom emojis.

[...]

PR #3755, opened 2026-09-09
Add custom display name and avatar for the blog profile

### LINKED ISSUE #3685: Feature Request: Custom display name & avatar for the profile blog
### What

Hi! Currently, the ActivityPub blog profile gets its display name (the public name of the federated account as seen in the Fediverse) from the Site title. It would be great to have a specific field for that. Same applies for the site icon / avatar. 
Thanks!

### Why

The site title may not always match the desired name for the federated account feed, i.e. "Lorem ipsum - Travel Agency" vs "Free city guides" or "Jane Doe · Photographer in New York" vs. "NY images by JD". Sometimes the blog content needs its own name or brand & icon.

### How

Just like the description field works: by default it uses the site tagline but accepts a custom description.

faisalahammad:
Fixes #3685

## Proposed changes:
The blog profile currently takes its display name from the site title and its avatar from the site icon. Site owners had no way to give their Fediverse blog account a different name or picture, even though the profile description already supports this through a dedicated field.

This PR adds two new options, `activitypub_blog_name` and `activitypub_blog_icon`, to the blog profile:

* The blog actor's `name` and `icon` prefer the new options and fall back to the site title and site icon when they are not set, so nothing changes for existing sites.
* The blog profile settings get a "Change Name" text field (with the site title as placeholder, matching the description field) and the avatar field becomes a media picker with Choose/Change/Remove buttons.
* The avatar picker crops the image to a 512x512 square before saving, using the same cropper as the header image field, so the profile picture stays square instead of federating a full-size photo. It reuses one shared picker script driven by the button's data attributes.
* Changes to the new options schedule a profile update so followers receive an Update activity.
* Profile edits coming from Mastodon clients (Enable Mastodon Apps) now write the new options instead of overwriting the site title and site icon in Settings > General. Before this change, editing the blog profile from a Mastodon client changed the actual WordPress site title.

Related: #1745 (closed as a duplicate of this request).

### Other information:

New tests in `tests/phpunit/tests/includes/model/class-test-blog.php` cover custom name and icon precedence, fallbacks, and that the setters no longer write the global site options. The scheduler tests cover that changing the new options schedules a blog profile update. The options tests cover that both `activitypub_blog_name` and `activitypub_blog_icon` are exposed to the REST settings API.

## Testing instructions:
* Go to 'Settings > ActivityPub > Blog Profile'.
* Enter a custom name in the new "Change Name" field and pick an avatar with the media picker, then save.
* Expected: picking a non-square image opens the cropper. After cropping, the preview shows the square crop. The saved avatar is the cropped attachment, not the original full-size image.
* After saving, check `https://yoursite.test/?author=0` (or your fediverse preview of the blog profile): the `name` and `icon` show the custom values, and the avatar is square.
* Press "Remove Custom Avatar": the preview should show the site icon again (or the bundled default icon when no site icon is set), not go blank. Clear the name field, save again: the profile falls back to the site title and the site icon.
* Dismiss the cropper without cropping, then reopen it: the picker should still work (no JS errors in the console).
* Follow the blog account from a Mastodon account, then change the name or avatar in the blog profile settings: the follower receives an Update and the profile refreshes.
* With Enable Mastodon Apps active, edit the blog profile from a Mastodon client: the Fediverse profile changes, while the site title and site icon in Settings > General stay untouched.
* Open your own user profile (Users > Profile) and use the header image field: it should keep working exactly as before (the header image picker shares the same script).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

Added a custom name and avatar for the blog profile. Your Fediverse blog account can now use its own display name and profile picture, independent of the WordPress site title and site icon.

</details>

### DIFF
--- a/assets/js/activitypub-blog-avatar.js
+++ b/assets/js/activitypub-blog-avatar.js
@@ -1,0 +1,69 @@
+/**
+ * Handle the blog avatar setting on the blog profile settings page.
+ */
+
+/* global jQuery, wp */
+
+( function ( $ ) {
+	var $chooseButton = $( '#activitypub-choose-blog-avatar-button' ),
+		$previewWrapper = $( '#activitypub-blog-avatar-preview-wrapper' ),
+		$preview = $( '#activitypub-blog-avatar-preview' ),
+		$hiddenDataField = $( '#activitypub_blog_icon' ),
+		$removeButton = $( '#activitypub-remove-blog-avatar' ),
+		frame;
+
+	/**
+	 * Initializes the media frame for selecting an avatar.
+	 */
+	$chooseButton.on( 'click', function () {
+		// Create the media frame.
+		frame = wp.media( {
+			title: $chooseButton.attr( 'data-choose-text' ),
+			button: {
+				// Set the text of the button.
+				text: $chooseButton.attr( 'data-update-text' ),
+				close: true,
+			},
+			library: { type: 'image' },
+			multiple: false,
+		} );
+
+		// When an image is selected, run a callback.
+		frame.on( 'select', function () {
+			var attachment = frame.state().get( 'selection' ).first();
+
+			$hiddenDataField.val( attachment.id );
+			$preview.attr( 'src', attachment.attributes.url );
+			$previewWrapper.removeClass( 'hidden' );
+			$removeButton.removeClass( 'hidden' );
+
+			$chooseButton
+				.attr( {
+					class: 'button',
+					'data-state': '1',
+				} )
+				.text( $chooseButton.attr( 'data-update-text' ) );
+
+			frame.close();
+		} );
+
+		frame.open();
+	} );
+
+	/**
+	 * Handles the click event of the remove button.
+	 */
+	$removeButton.on( 'click', function () {
+		$hiddenDataField.val( '' );
+		$preview.attr( 'src', $previewWrapper.attr( 'data-fallback-url' ) );
+		$( this ).addClass( 'hidden' );
+
+		$chooseButton
+			.attr( {
+				class: 'button upload-button button-add-media button-add-blog-avatar',
+				'data-state': '',
+			} )
+			.text( $chooseButton.attr( 'data-choose-text' ) )
+			.trigger( 'focus' );
+	} );
+} )( jQuery );
--- a/assets/js/activitypub-blog-avatar.js
+++ b/assets/js/activitypub-blog-avatar.js
@@ -5,8 +5,8 @@
 /* global jQuery, wp */
 
 ( function ( $ ) {
-	var $chooseButton = $( '#activitypub-choose-blog-avatar-button' ),
-		$previewWrapper = $( '#activitypub-blog-avatar-preview-wrapper' ),
+	var fallbackUrl = window.activitypubBlogAvatar && window.activitypubBlogAvatar.fallbackUrl,
+		$chooseButton = $( '#activitypub-choose-blog-avatar-button' ),
 		$preview = $( '#activitypub-blog-avatar-preview' ),
 		$hiddenDataField = $( '#activitypub_blog_icon' ),
 		$removeButton = $( '#activitypub-remove-blog-avatar' ),
@@ -34,7 +34,6 @@
 
 			$hiddenDataField.val( attachment.id );
 			$preview.attr( 'src', attachment.attributes.url );
-			$previewWrapper.removeClass( 'hidden' );
 			$removeButton.removeClass( 'hidden' );
 
 			$chooseButton
@@ -55,7 +54,7 @@
 	 */
 	$removeButton.on( 'click', function () {
 		$hiddenDataField.val( '' );
-		$preview.attr( 'src', $previewWrapper.attr( 'data-fallback-url' ) );
+		$preview.attr( 'src', fallbackUrl );
 		$( this ).addClass( 'hidden' );
 
 		$chooseButton
--- a/includes/class-options.php
+++ b/includes/class-options.php
@@ -437,6 +437,18 @@ class Options {
 		 */
 		\register_setting(
 			'activitypub_blog',
+			'activitypub_blog_name',
+			array(
+				'type'              => 'string',
+				'description'       => 'The Name of the Blog-User',
+				'show_in_rest'      => true,
+				'default'           => '',
+				'sanitize_callback' => 'sanitize_text_field',
+			)
+		);
+
+		\register_setting(
+			'activitypub_blog',
 			'activitypub_blog_description',
 			array(
 				'type'         => 'string',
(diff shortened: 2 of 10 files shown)


faisalahammad:
@Automattic/fediverse Could you take a look when you get a chance? Reviewer request from a fork PR is not possible, so tagging here.

assets/js/activitypub-blog-avatar.js
```diff
@@ -0,0 +1,68 @@
...
+		$hiddenDataField = $( '#activitypub_blog_icon' ),
+		$removeButton = $( '#activitypub-remove-blog-avatar' ),
+		frame;
+
+	/**
+	 * Initializes the media frame for selecting an avatar.
+	 */
+	$chooseButton.on( 'click', function () {
+		// Create the media frame.
+		frame = wp.media( {
```
pfefferle: This stores whatever image gets picked and we federate it in full size, so a 4000x3000 photo ends up as the avatar. The site icon flow we replace here always crops to a square. We already have a cropper in `activitypub-header-image.js` (`ImageCropperNoCustomizer`, based on core's site-icon.js), including the "already the right ratio, skip the crop" shortcut. Can you build this frame the same way, a Library state plus that cropper with width and height 512, and store the id of the cropped attachment? Maybe even one script for both buttons, driven by the data attributes the button already has.

includes/wp-admin/class-admin.php
```diff
@@ -287,6 +287,24 @@ public static function enqueue_scripts( $hook_suffix ) {
...
+			),
+			array( 'jquery', 'media-editor' ),
+			ACTIVITYPUB_PLUGIN_VERSION,
+			false
+		);
+		\wp_localize_script(
+			'activitypub-blog-avatar',
+			'activitypubBlogAvatar',
+			array(
+				'fallbackUrl' => \esc_url_raw( \get_site_icon_url() ),
```
pfefferle: `get_site_icon_url()` is empty when there is no site icon, so "Remove" leaves the preview blank. The PHP side uses `site_icon()`, which falls back to our default icon. I would use the same here.

includes/class-options.php
```diff
@@ -458,6 +470,17 @@ public static function register_settings() {
 			)
 		);
 
+		\register_setting(
+			'activitypub_blog',
+			'activitypub_blog_icon',
```
pfefferle, line 475: `activitypub_blog_name` has `show_in_rest`, this one does not. Either both or neither?

pfefferle (changes requested):
Thanks, the option handling looks good and the Mastodon-client path writing our own options instead of the site title was a real bug. One thing blocks it for me: the avatar picker takes any image as is, see the inline comments.

assets/js/activitypub-blog-avatar.js
faisalahammad: Done. The header image script is now a shared script, activitypub-media-picker.js. One script for both buttons, driven by the data attributes the buttons already have. The avatar button crops to 512x512 with the same cropper as the header image and stores the cropped attachment ID.

includes/wp-admin/class-admin.php
faisalahammad: Done. Remove now shows the default icon, same as the PHP side. The fallback URL comes from site_icon() and is passed with wp_localize_script, so the preview no longer depends on get_site_icon_url().

includes/class-options.php
faisalahammad: Done. activitypub_blog_icon has show_in_rest now, same as activitypub_blog_name. I added a test that checks both.

faisalahammad (commented):
Thanks for the review. All three points are fixed in 408bec3a.

1. The avatar picker now uses the same cropper as the header image. I turned the header image script into a shared script, activitypub-media-picker.js, and both buttons use it. Each button controls its picker from its data attributes, so the avatar crops to 512x512 and we store the cropped attachment ID.

2. Remove no longer blanks the preview. The fallback URL comes from site_icon(), the same function the PHP side uses, and is passed with wp_localize_script.

3. activitypub_blog_icon has show_in_rest now, same as activitypub_blog_name. Added a test for both.

Tests: PHPUnit suite is green, phpcs and JS lint are clean.

There is one thing I want to mention. I did not put the fallback URL in a data attribute. An earlier commit on this branch removed that pattern because of a CodeQL DOM XSS alert, so I pass it through wp_localize_script and key it by the picker context. Let me know if you prefer it another way.

faisalahammad:
@pfefferle All three review points are fixed in 408bec3a and the threads are resolved. Could you take another look when you have time? The one thing to check is the fallback URL for the avatar preview. I did not use a data attribute for it because an earlier commit on this branch removed that pattern after a CodeQL DOM XSS alert, so it goes through wp_localize_script instead.

assets/js/activitypub-media-picker.js
```diff
@@ -0,0 +1,290 @@
...
+
+	/**
+	 * Initializes a single media picker, driven by the data attributes of its button.
+	 *
+	 * Every picker keeps its own state, so multiple pickers can run on the same page.
+	 *
+	 * @param {jQuery} $chooseButton The button that opens the media frame.
+	 */
+	function initPicker( $chooseButton ) {
+		var $preview = $( $chooseButton.attr( 'data-preview' ) ),
```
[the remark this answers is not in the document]
faisalahammad: Fixed in 0e7e5ae1. The picker read its element selectors from data attributes and passed them to jQuery. jQuery parses a string that starts with "<" as HTML, which is what CodeQL flagged. The selectors now go through document.querySelector, so they are never parsed as HTML.

assets/js/activitypub-media-picker.js
```diff
@@ -0,0 +1,290 @@
...
+			$previewWrapper = $( $chooseButton.attr( 'data-preview-wrapper' ) ),
```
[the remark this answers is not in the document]
faisalahammad: Fixed in 0e7e5ae1. The selectors now come from document.querySelector instead of jQuery, so they are never parsed as HTML.

assets/js/activitypub-media-picker.js
```diff
@@ -0,0 +1,290 @@
...
+			$hiddenDataField = $( $chooseButton.attr( 'data-input' ) ),
```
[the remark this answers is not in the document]
faisalahammad: Fixed in 0e7e5ae1. The selectors now come from document.querySelector instead of jQuery, so they are never parsed as HTML.

assets/js/activitypub-media-picker.js
```diff
@@ -0,0 +1,290 @@
...
+			$removeButton = $( $chooseButton.attr( 'data-remove' ) ),
```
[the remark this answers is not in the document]
faisalahammad: Fixed in 0e7e5ae1. The selectors now come from document.querySelector instead of jQuery, so they are never parsed as HTML.

pfefferle (approved):
Thanks, this looks good now. The cropper is exactly what I had in mind, and moving `update_name()`/`update_icon()` off `blogname`/`site_icon` fixes a real bug: a Mastodon client could rename the whole site so far.

One thing I noticed while reading, nothing to do for you: the settings form only `absint`s `activitypub_blog_icon`, while `update_icon()` checks `wp_attachment_is_image()`. The header image option has the same gap (no sanitize callback at all), so I will align the attachment-id options in a follow-up.

Issue #383, opened 2023-07-31
Dealing with translations on a site with mulitple languages


ghost:
Dealing with translations might become necessary, and it might be done via different approaches. As there are not too many widely used translation plugins there might be solutions that work for most users.

Easy first step solution: Select/choose languages that should be published via ActivityPub in the settings.

More advanced solution: Create some sort of group actors that publish certain languages.

#multilang

pfefferle:
FYI: https://wordpress.org/support/topic/post-language-seems-to-default-to-blog-default/

Menrath:
Just as a reference: I made a dirty hack, so that on our site with PolyLang installed only english posts get send as activities.
https://github.com/Menrath/wordpress-activitypub/commit/8fe1a1ac7a19a499260c9c01bca970d0a973a45b

bruvellu:
For me, the core basic functionality for multilingual support would be to set the language of the ActivityPub's content to match that of the blog's post. In this way, the post would be federated with the correct language tag. This allows followers to filter out posts in specific languages if they want to (like it's possible on Mastodon).

Since each WP multilingual plugin handles language differently, one solution is creating compatibility classes like `class-multilingual` or `class-polylang` (or other plugins) as described by @pfefferle, where the post's language can be discovered using plugin-specific functions like `pll_get_post_language` from Polylang. And then the language can be passed on to the `"@context": ["https://www.w3.org/ns/activitystreams", {"@language": "en"}]` in the ActivityPub object.

I might be completely off track here (sorry, only rudimentary php and activitypub knowledge), but what do you think?

Menrath:
I would not pass those via @language but use the `contentMap` instead. As many multilingual plugins save an extra post for each translation this would require only creating a single activity for multiple posts.

As you see there are many solutions for this problem. People should try to build a solution for their needs, and then we can see whether there are missing hooks/filters to fullfill their needs.

Of course one can also make new settings pages and add a proper management of this functionality, but I would do this in a second step. At the same time for the project I do some infrastructure work I am currently fine with my hot-fix.

bruvellu:
Something I just thought. Would it be possible to set the ActivityPub object's language using a filter placed in the `functions.php`? That way, the logic of getting the post's language would be up to the site's owner, which makes sense since each site and multilingual plugin is different.

Menrath:
It totally is possible, but I did not fully understand, what you would prefer.

bruvellu:
[...]

PR #3756, opened 2026-09-11
Add Polylang integration

pfefferle:
## Proposed changes:

Adds a Polylang integration, the same thing the WPML integration does: the `activitypub_locale` filter returns the language Polylang assigned to the post, so `contentMap`, `nameMap` and `summaryMap` are keyed by the post's language instead of the site locale. A comment takes the language of the post it belongs to, which the WPML integration does not do yet. A post without a Polylang language keeps the site locale.

That is all there is to it. I checked whether Polylang needs more handling and did not find anything: its `home_url` filter only applies to calls from theme files and a short whitelist, so our actor and object IDs are not touched; its post type settings only list public post types, so our internal ones stay out of it; and the `term_id` it puts on every request is handled generically since #3744.

What this does not do: build a multi-language `contentMap` from the linked translations (`pll_get_post_translations()`). Each translation is its own post and federates as its own object, so merging them would either post twice or need one canonical object per translation group. That is a product decision, not an integration detail.

### Other information:

Three tests with a small mock of `pll_get_post_language()`: a post gets its Polylang language, a comment gets the language of its post, a post without a language keeps the site locale.

## Testing instructions:

* Install Polylang, add two languages, and publish a post in the non-default one.
* Fetch the post with `Accept: application/activity+json`. `contentMap` is keyed by the post's language, e.g. `de`, not the site locale.
* Reply to it from the Fediverse and fetch the comment the same way. Same key.
* A post Polylang has no language for still uses the site locale.

### Changelog entry

Added manually in `.github/changelog/add-polylang-integration`.

### DIFF
--- a/.github/changelog/add-polylang-integration
+++ b/.github/changelog/add-polylang-integration
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Posts and comments on sites using Polylang are now federated with the language Polylang assigned to them.
--- a/integration/README.md
+++ b/integration/README.md
@@ -58,6 +58,7 @@ A few integrations are always initialized (`Nodeinfo`, `Webfinger`, `Surge`, `Li
 | Integration | What it does |
 |---|---|
 | **WPML** | Supplies the correct per-object locale for federated content under WPML. |
+| **Polylang** | Supplies the correct per-object locale for federated posts and their comments under Polylang. |
 | **Multisite Language Switcher** | Keeps ActivityPub data consistent across MSLS post translations. |
 
 ### Other
--- a/integration/class-polylang.php
+++ b/integration/class-polylang.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * Polylang integration.
+ *
+ * @package Activitypub
+ */
+
+namespace Activitypub\Integration;
+
+/**
+ * Compatibility with the Polylang plugin.
+ *
+ * @see https://polylang.pro/
+ */
+class Polylang {
+	/**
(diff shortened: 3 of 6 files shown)


pfefferle:
Can you maybe have a look @bruvellu? Is this working?

bruvellu:
Hi @pfefferle, that would be wonderful. I'll test it out. If I understand it correctly, this addresses the [use case](https://github.com/Automattic/wordpress-activitypub/issues/383#issuecomment-1913662391) described in #383?

louischance:
Hi there thanks for this patch!
I just tested it and it seems to work perfectly. Here's what I did exactly:

1. Started a staging WordPress website.
2. Installed Polylang (Free edition), added French, English and Spanish as languages. Set up French to be the primary language.
3. Installed Activitypub from the repo and quickly set it up to use Blog profile only.
4. I followed the website through my Mastodon account (using Tusky on mobile, to check the post language).
5. Created and published a post in French -> appears in Mastodon in FR
6. Created a translation in English -> appears in Mastodon in FR (normal, it's the repo version of the plugin).
7. Then uploaded the current version to update the plugin
8. Created a Spanish trasnlation of the post -> appears in Mastodon in ES
9. Opened the English post, hit the Save button -> now appears in Mastodon in EN and not FR

So from my experience it seems to work as expected :)

I also tried to change the Site language to Deutsch, and then created a new post, and it it published in Mastodon as using French (as expected, as this is the Polylang language of the post).

pfefferle:
@louischance thanks a lot for testing!!

PR #3769, opened 2026-09-14
Fix <title> tag on the ActivityPub settings page

pilarmera:
Fixes the admin title of the plugin's settings page: currently it simply shows  "Welcome" (followed by the standard WP admin title suffix: " < {site_name} -- WordPress" ), without any mention to ActivityPub, and it doesn't change when navigating through the settings tabs. 

## Proposed changes:

* Changes the default title of the settings page from the un-meaningful "Welcome" to "ActivityPub" (changing the title string in the `add_options_page` args from `class-menu.php`).
* Updates the title of the settings page with the label of the settings tab currently viewed ("Welcome", "Settings", "Advanced", "Blog Profile", etc.) by using the `admin_title` filter.

### Other information:
I don't know if `class-admin.php` is the right place to filter the admin title; I tried to do it within class-menu.php or class-settings.php but I couldn't make the `admin_title` filter to work there. 

## Testing instructions:
*Check admin <title> tag on the ActivityPub Settins screen

Before:
admin-title-before

After:
admin-title-after (Copy)

### Changelog entry

Added on commit

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


pfefferle:
Thanks for the PR! I think we can make this simpler and reuse more of what is already there.

Core has one tabbed screen, Site Health, and it does not filter `admin_title`. `wp-admin/site-health.php` just sets the global `$title` to `sprintf( __( 'Site Health - %s' ), $tabs[ $current_tab ] )` before the header is printed, and core builds the rest ("‹ Site name — WordPress") from that. We can do the same: the `load-settings_page_activitypub` action runs before core includes `admin-header.php`, and `get_admin_page_title()` returns `$title` as is when it is already set. So the `admin_title` filter and the hand-built suffix (which bypasses core's translatable string and is different in the network admin) are not needed.

The tab labels already exist in `Settings::settings_page()`, they are just built too late, in the page callback after the header. My suggestion:

1. Move the tab array (up to and including the `activitypub_admin_settings_tabs` filter and the empty fallback) out of `settings_page()` into `Settings::get_settings_tabs()`, and the "which tab is active" part into `Settings::get_current_tab( $settings_tabs )`. `settings_page()` then just calls both.
2. Add a third `load-settings_page_activitypub` callback in `Settings::init()`, next to the two that are already there:

```php
/**
 * Set the page title to the tab that is being viewed.
 *
 * Runs before core prints the admin header, so `get_admin_page_title()` picks it up.
 */
public static function set_page_title() {
	$settings_tabs = self::get_settings_tabs();
	$tab           = self::get_current_tab( $settings_tabs );

	/* translators: %s: The label of the settings tab. */
	$GLOBALS['title'] = \sprintf( \__( 'ActivityPub - %s', 'activitypub' ), $settings_tabs[ $tab ]['label'] ); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
}
```

3. Remove `Admin::set_tab_title()`, the `admin_title` filter and the switch with the duplicated labels.

This way tabs that other plugins add through the `activitypub_admin_settings_tabs` filter get the right title too, and the labels live in one place. Your "ActivityPub - Settings" format stays, it is the same shape core uses for Site Health.

phpcs complains about writing to `$GLOBALS['title']`, but we have the same `phpcs:ignore` for `$plugin_page` in `templates/bulk-actor-delete-confirmation.php`, so I think that is fine here. I only read through core's `admin.php`, `admin-header.php` and `site-health.php`, I have not tested it in the browser, so please double check.

The "ActivityPub" fallback in `add_options_page()` is good, please keep that.

Small thing on the changelog entry: users see it on the plugin update screen, so I would write something like "The browser tab title of the settings page now shows the name of the current tab." (with a period at the end).

pfefferle:
Sorry for the complicated review! I thought it might be easier to commit my suggestion directly, so I pushed 0b0de1a3 to your branch, I hope that is fine.

It builds on your PR: the title now reuses the labels from the settings tabs, and core adds the "‹ Site name — WordPress" part by itself, so the `admin_title` filter and the switch are gone. Would be great if you could check that it still works for you and tell me what you think.

pfefferle (approved):

pilarmera:
Hi! Thank you very much for the corrections and explanations. They make total sense :)

I suspected it had to be a better way to do it without having to "redeclare" the tab layers nor the core `admin_title` part, but I didn't dare / didn't know how to mess with the global $title or the actions order. 

I learned a lot with this PR -one of my fews-, I'm still becoming familiar with git. Thank you for you time, I'm loving ActivityPub!

Issue #3792, opened 2026-09-17
One entry point for remote objects, with a pluggable cache


pfefferle:
We talked about the `proxyUrl` endpoint in the ActivityPub API meeting today, and I think it points at a structure we should have on the server side anyway.

Right now remote data is cached in three places that don't know about each other: `Remote_Actors` (a CPT that is cache and storage at the same time), the transients in `Http::get()`, and `Remote_Posts` (a CPT filled by the inbox, read by the Reader). The proxy endpoint reads through the first two, the Reader reads the third directly. The database grows with every remote actor and post we ever saw, and there is no way for a host to put a Redis in front of any of it.

## Idea

One `Proxy` class that is the only way to get a remote object:

- `Proxy::get( $id )` resolves an acct via WebFinger, looks in the cache, fetches on a miss (signed, with the id/host checks we already have in `Http::get_remote_object()`), stores, returns. Plus `get_many()`, `delete()` and `refresh()`.
- The cache behind it is a PSR-16 `CacheInterface` (version 1, works on PHP 7.4). A host can hook in any existing Redis or Memcached adapter with one filter. We ship a transient provider (default, works everywhere), an object cache provider (used automatically when a persistent object cache is active), and a post type provider (what the CPTs do today).
- The REST proxy endpoint becomes a thin wrapper around it, so clients and the server go through the same path.
- Invalidation follows federation: `Update`, `Delete` and `Move` call `Proxy::delete()`, TTL is the safety net, misses on dead servers are cached for a short time.

`Http::get_remote_object()`, `Remote_Actors::fetch_by_uri()`, `get_remote_metadata_by_actor()` and friends forward to `Proxy` first and get deprecated over time. `Http` keeps only the transport.

## What is not a cache

`Remote_Actors` and `Remote_Posts` each do two jobs, and only one of them is caching bodies. The other parts need their own home before the post types can go:

- Relations (followers, following, blocks, inbox lists) are ours and permanent. Today they are meta on the actor post.
- The Reader needs an index ("the next 20 posts for this user"), which a cache can't answer. Replies and quotes on remote posts also hang off the `ap_post` row via `comment_post_ID`.

So I would stage it: bodies first (the Proxy owns them from day one), relations second, the post index third. Each of the two classes shrinks with every step and ends up as a facade.

## Open questions

[...]

PR #3793, opened 2026-09-17
Proxy remote objects through one cached entry point

pfefferle:
First step of #3792.

## Proposed changes:

* `Proxy::get( $id )` is the one way to fetch a remote object now, and every fetch in the plugin goes through it. It resolves an acct via WebFinger, answers from the cache, and on a miss fetches with a signed request, checks that the object is served under its own id (the checks that lived in `Http::get_remote_object()`), and stores it. Failures are remembered for one to fifteen minutes. `delete()` and `refresh()` for invalidation.
* An object is stored once, under its declared id. Any other URL it was requested by gets an alias pointing at that id, so dropping the id drops every spelling. Anything served from another host after a redirect gets no alias, it is cached under its own id only. Fragments are ignored, so a key id like `…#main-key` shares the actor's entry.
* The cache is the WordPress object cache, with transients as the fallback where no persistent object cache exists. Entries are addressed by kind plus id, the key is a hash. A host swaps the backend with its object cache drop-in and nothing else; the cache code is private to the Proxy.
* The Update and Delete handlers drop the cached object first, so a changed or removed object never outlives its cache entry. `Http::get_remote_object()` stays as a forwarding alias for anyone outside the plugin.

Nothing about `Remote_Actors` or `Remote_Posts` changes in this step. Moving those onto the Proxy, followers holding only ids, one cache group per kind, and taking the caching out of `Http::get()` come after this.

### Other information:

The existing `activitypub_pre_http_get_remote_object` filter keeps working, it now fires inside the Proxy. The transport no longer caches raw responses for this path; the Proxy caches the decoded object instead. `cached => false` bypasses the read but still stores the fresh copy.

## Testing instructions:

* `npm run env-test`.
* On a site with a persistent object cache, fetch a remote profile through the proxy endpoint twice and check that the second request does not hit the remote server.
* Edit a post on a remote server that your site has fetched; after the Update arrives, the next fetch returns the new content.

### Changelog entry

Added on the branch.

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

</details>


pfefferle:
@jeherve @akirk I would love to get your feedback on this general idea. I thought about using the cache for remote data like stream objects and actors with a lazy loading mechanism, mainly to keep the CPT (and the WordPress db structure in general) a bit cleaner and to be able to hook in more effective storing mechanisms like redis or memcachedb. I am also curious if this is also possible on WordPress.com or if this might spam the cache there!?

PR #3796, opened 2026-09-18
Stop signature reads from creating key pairs

### LINKED ISSUE #3766: Signature::get_key_pair() persists during a read and has no in-process memo
### Description

`Signature::get_key_pair()` (`includes/class-signature.php`) performs an `update_option()` as part of what is otherwise a read:

```php
$key_pair = \get_option( $option_key );

if ( $key_pair ) {
    return $key_pair;
}

$key_pair = $legacy_callback ? $legacy_callback() : false;

if ( ! $key_pair ) {
    $key_pair = self::generate_key_pair();
    // ...
}

\update_option( $option_key, $key_pair );

return $key_pair;
```

The stored option row is the only cache. There is no static memo, object-cache entry or transient in `Signature`, `Application` or `Actors` — `Application::get_keypair()` and `Actors::get_keypair()` are thin pass-throughs, and `Application::get_public_key()` / `get_private_key()` each call `get_keypair()` separately, so a caller that needs both pays twice.

Two consequences:

**1. A read can create state, on whichever site happens to be current.** `Http::get()` builds `key_id` / `private_key` from `Application::get_private_key()` eagerly on every uncached request, so an ordinary outbound fetch can mint a keypair. On a multisite host, plugin code may run before the request has been switched to the site it is serving — in that window the row is written to the wrong site. Because generation is non-deterministic, that site ends up with a *different* keypair rather than a stray copy of the right one.

**2. If the write does not stick, generation repeats forever.** Since `get_option()` is the only memo, a persist that is filtered, blocked or simply fails means `generate_key_pair()` runs again on the very next call. That is an unconditional 2048-bit RSA keygen (`openssl_pkey_new`, `sha512`) — I measured ~280 ms per keypair, with the long tail RSA keygen normally has. Inbound signature verification resolves remote actors through several `Http::get()` calls per request, so this compounds quickly.

### Suggested change

- Add a static memo keyed on `$option_key`, so repeated calls within a request cost nothing regardless of whether the row exists.
- Separate "read the keypair" from "generate and persist one", so a caller that only wants to read cannot trigger a write as a side effect.

The memo alone is a clear win independent of the multisite angle.

faisalahammad:
Fixes #3766

## Proposed changes:

* Reading a key pair no longer writes to the database. `Signature::get_stored_key_pair()` returns only a pair the site already has, and a new read-only `Application::get_stored_keypair()` delegates to it.
* `Http::get()` uses that read-only pair. When the site has no key pair yet, the request goes out unsigned instead of generating one as a side effect, which is also what `sign_request()` checks before it signs.
* Both lookups are memoized in a non-persistent object cache group keyed per site, so a key pair is read from the options table at most once per request and private keys never reach a persistent cache.
* Paths that deliberately need a key still create it: `Http::post()` deliveries and the starter kit import, which now passes the signing keys explicitly.
* `Migration::migrate_application_keypair_option()` flushes the memo, since the rows it moves would otherwise be served stale for the rest of the request.

Previously a plain outbound fetch on a site that had never published an actor could mint a 2048-bit RSA key pair (~280ms, sha512) and write it to the options table, on multisite possibly on the wrong site. If the write did not stick, the keygen repeated on every read.

### Other information:

## Testing instructions:

* On a fresh site, run `wp option get activitypub_application_keypair`. It should report the option does not exist.
* Fetch a remote actor, for example `wp activitypub fetch https://mastodon.social/@Gargron`.
* Check `wp option get activitypub_application_keypair` again. It should still be missing. Before this change, the fetch would have created it.
* On a site that already has a key pair, confirm signed requests still work: the option is unchanged and federation (follow, deliver) continues to work.
* Repeat the fetch and confirm the key pair is looked up once per request rather than re-read on every call.
* On multisite, switch to a second site that has no key pair and confirm a fetch there does not create one on the first site.
* Import a starter kit from the admin. It should still work and be allowed to create a key pair.
* Run the PHPUnit suite.

### Changelog entry

<details>

#### Significance

-   [x] Patch

#### Type

-   [x] Fixed

#### Message

Fetching remote content no longer creates a signing key pair on your site. Sites that already have one keep signing their requests, and the key pair is now looked up at most once per request, so pages that contact other servers load faster.

</details>


pfefferle:
Closing this one, sorry for the late turn. Two things made me change my mind on the whole approach:

`Http::get()` has to keep signing and creating the key when there is none. The Application key is only ever created lazily, and this was the place that did it. Inbound signature verification resolves the remote key through `Http::get()`; unsigned, GoToSocial and Mastodon in secure mode answer 401, we cache that for 15 minutes, and the first Follow to a fresh site is lost.

And the cache buys nothing: `get_option()` already memoizes per request, misses included. The only thing worth remembering is a failed generation, and that is a ten line change on its own if we ever want it.

I am closing the issue as well, the setup that triggered it turned out to be an edge case of one hosting environment and is handled there.

PR #3799, opened 2026-09-21
feat(oauth): switch accounts on authorize screen

### LINKED ISSUE #3632: Feature Request: ActivityPub API login different user
### What

When logging into wordpress via the ActivityPub Authorize, the authorize page will pick the logged-in user. 
The page offers to either Authorize or Cancel (which returns to the app url)
There isn't a great way to log in as a different user.

### Why

I use different author/actors for different purposes, and it currently requires opening another window to log out, before trying to authorize again.

### How

Perhaps an additional link/button to logout the current user, while keeping all the authorization parameters

faisalahammad:
Fixes #3632

## Proposed changes:

* Add a "Log in as a different user" link to the ActivityPub OAuth consent screen.
* Build the link with WordPress's `wp_logout_url()` and the validated authorization form URL, preserving the OAuth request through logout and the next login.
* Add a PHPUnit test covering the rendered link and preserved authorization parameters.

### Other information:

## Testing instructions:

Automated checks completed:

* Focused OAuth test: passed.
* OAuth PHPUnit group: 200 tests, 480 assertions, passed.
* Full PHPUnit suite: 3,220 tests, 7,647 assertions, 25 pre-existing skips, passed.
* PHPCS: passed for the changed files and the full configured ruleset.
* Build: passed.

Manual test steps:

1. Enable the ActivityPub API and register an OAuth client with a callback URL.
2. Log in as WordPress user 1.
3. Start the client's authorization request.
4. On the consent screen, confirm that **Log in as a different user** is shown.
5. Click the link, then log in as WordPress user 2.
6. Confirm that the same client, requested permissions, callback URL, and state are shown.
7. Click Authorize and confirm that the access is granted to user 2.
8. Repeat the flow and click Cancel. Confirm that the client receives `access_denied` with the original state value.

### Changelog entry

A changelog entry is already included in `.github/changelog/add-oauth-authorize-switch-user`.

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message

You can now log in as a different WordPress user while authorizing an ActivityPub app without losing the authorization request.

</details>


pfefferle (changes requested):
I had a look at this, since I was about to open the same thing (#3800, I will close that one in favour of this PR). Same approach, `wp_logout_url( $form_url )`, so we are on the same page. Two things I ran into while testing the round trip:

1. `add_query_arg()` does not encode values, so `$form_url` breaks for a `state` containing `&`, `+` or `?` (or a `redirect_uri` with a query string). It did not matter so far because the form posts hidden fields and ignores the action URL, but the logout `redirect_to` now round-trips through that URL, and core splits it at the first raw `&`. The client's state check then fails. Fix is small, in `render_authorize_form()`:

   ```php
   \array_merge( array( 'action' => 'activitypub_authorize' ), \array_map( 'rawurlencode', $authorize_params ) ),
   ```

   I verified with curl that `state=abc%26def%2Bghi` comes back byte-identical after logout → login as another user with that change. It would be good to add such a state to the test, the current `test-state` cannot catch it. Note that the test's expected URL is built the same lossy way, so it needs the same encoding once you change it.

2. phpcs fails on the new stubs file (5 "unused function parameter" warnings), the `phpcs:ignore` codes do not match what CI reports. See https://github.com/Automattic/wordpress-activitypub/actions/runs/35582756816/job/106279125266

The stubs for `login_header()`/`login_footer()` are a good idea, I had skipped a test for exactly that reason. Thanks!

templates/oauth-authorize.php
```diff
@@ -79,6 +80,11 @@
 		);
 		?>
 		</p>
+		<p>
+			<a href="<?php echo esc_url( $logout_url ); ?>">
+				<?php esc_html_e( 'Log in as a different user', 'activitypub' ); ?>
```
pfefferle:
```suggestion
				<?php esc_html_e( 'Not you? Log in as a different user.', 'activitypub' ); ?>
```

faisalahammad: Applied the suggested copy: "Not you? Log in as a different user."

faisalahammad:
Addressed the review feedback. The authorize URL now rawurlencodes OAuth parameters before add_query_arg(), and the test covers reserved characters and checks the full round trip. The login page stubs now use the CI-reported VariableAnalysis suppression code. The suggested copy is applied. OAuth tests and CodeRabbit review pass.

faisalahammad:
Thanks for the review, all three points are addressed in dd10b901.

- The OAuth parameters are now encoded where the form URL is built, so a state or redirect with reserved characters survives the logout round trip. The test now uses `abc&def+ghi?jkl/xyz` as the state and checks every parameter after the round trip, not only two.
- The phpcs ignore codes on the login page stubs are corrected.
- The link copy is now "Not you? Log in as a different user."

@Automattic/wordpress-activitypub team, @pfefferle could you take another look when you have time?

faisalahammad:
## CI Fix Summary

| Check | Result |
|---|---|
| PHPCS warning fix | Passed. Added the missing unused-parameter sniff ignores to the WordPress login stubs. |
| OAuth server tests | Passed. 66 tests, 121 assertions in wp-env. |
| Focused PHPCS | Passed. The committed stub file exits 0 with `phpcs.xml`. |
| Manual zip test | Confirmed by tester. |
| CodeRabbit | Not run because the review rate limit was reached. |

Commit: `ca34ce72`

faisalahammad:
Update on the CodeRabbit line above.

The review rate limit reset, so I ran it after all. It found no issues and reviewed 5 files, including the login stub file that this fix touches.

So all checks are clear on my side.

includes/oauth/class-server.php
```diff
@@ -432,12 +432,17 @@ private static function render_authorize_form() {
 		$scopes       = Scope::validate( Scope::parse( $authorize_params['scope'] ) ); // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
 
 		// Build form action URL.
+		// The values must be encoded: add_query_arg() does not encode them, so raw
+		// reserved characters in the OAuth params would break the query string.
 		// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
```
pfefferle, line 437:
```suggestion
		/*
		 * Build the form action URL. add_query_arg() does not encode values, so the
		 * OAuth parameters are encoded first; raw reserved characters would break the query string.
		 */
		// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
```

pfefferle (approved):
Thanks, all points addressed and the round-trip test is a nice one. One nit for the multi-line comment (project style is `/* */` for those), suggestion below.

includes/oauth/class-server.php
faisalahammad: Applied the suggestion. The comment now uses the project block comment style. Commit 4ab32caf.

PR #3800, opened 2026-09-21
Add switch-user link to the OAuth authorize screen

### LINKED ISSUE #3632: Feature Request: ActivityPub API login different user

pfefferle:
Fixes #3632

## Proposed changes:
* Adds a "Not you? Log in as a different user." link to the OAuth consent screen. It points to `wp_logout_url()` with the current authorize URL as `redirect_to`, so core logs the user out, sends them to the login form and back to the consent screen with all OAuth parameters (client, redirect URI, scope, state, PKCE) intact.
* Encodes the parameters of the consent form URL before passing them to `add_query_arg()`, which does not encode values. Before, a `state` containing `&` or `+` was mangled in that URL. It was harmless as the form action (the POST uses hidden fields), but the new link round-trips through it.

### Other information:

No PHPUnit test: the template calls `login_header()`, which only exists inside `wp-login.php` and is not loaded in the test bootstrap. Verified end to end with curl, see below.

## Testing instructions:

* Enable the ActivityPub API and register a client (e.g. via `Client::register()` or any Social Web app).
* Log in as user A and open `wp-login.php?action=activitypub_authorize&client_id=…&redirect_uri=…&response_type=code&scope=read&state=abc%26def%2Bghi&code_challenge=…&code_challenge_method=S256`.
* Click "Not you? Log in as a different user." You land on the login form.
* Log in as user B. You should be back on the consent screen, now "Logged in as B", and the hidden `state` field should still be `abc&def+ghi`.
* Authorize or Cancel as usual.

### Changelog entry

Added in `.github/changelog/add-oauth-authorize-switch-user`.


pfefferle:
Closing in favour of #3799, which was first. I left the encoding fix and the phpcs note as a review over there.

